PR #19, opened 2020-09-10
Fix imagem alt in readme

leonardoelias:
The documents have a beautiful image with some texts explaining more about the library, but we must not forget the screen readers, so I added an alternative text for it.

More: https://www.a11yproject.com/posts/2013-01-14-alt-text/


dai-shi (approved):
Makes sense. Thanks!

eptaccio:
Wow, this is amazing! Thanks @leonardoelias 💖

PR #20, opened 2020-09-10
chore: delete item in dependentsMap if is empty

leonardoelias:
was it to do that?

all tests passed

😥


eptaccio:
Good Luck. You are amazing! 💖

dai-shi (approved):
Yeah, this is correct. It's probably something I forgot while changing implementations.

(Note: we don't have test specs for deleting dependents, yet..)

leonardoelias:
you are awesome @dai-shi 🤗
thank you very much for your work in the react community

PR #21, opened 2020-09-10
chore: remove disable selection in demo

leonardoelias:
The page selection is disabled for all elements, so the user does not select the code.

'user managing to select the code'

Is there any reason for that?


dai-shi (approved):
I'm not sure where this originates from, but it should be fine to remove it.
(Or, do we want to enable `user-select` only on code snippet?)

PR #22, opened 2020-09-10
improve atom initialization (lazy read)

dai-shi:
This is something I gave up as an initial implementation.
This will be annoying soon or later, so I'd put some effort to investigate and possibly fix it. It can be a big change.

~For now, there's only a failing test.~

This is a big change in the end.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -24,9 +24,9 @@
     "gzipped": 1922
   },
   "index.js": {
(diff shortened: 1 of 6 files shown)


dai-shi (commented):
This is almost rewriting almost a half part of the implementation. There might be some issues in edge cases. I see some room for improvement. We'd need more tests. Having said that, let me merge this to move forward.

PR #33, opened 2020-09-12
Fix adding atom and deleting atom

dai-shi:
There's some bugs when we create and delete atoms in React lifecycles.
This fixes with various techniques.

### DIFF
--- a/src/useAtom.ts
+++ b/src/useAtom.ts
@@ -52,6 +52,14 @@ export function useAtom<Value, Update>(
     )
   )
 
+  useIsoLayoutEffect(() => {
+    const id = Symbol()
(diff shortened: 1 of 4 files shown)


dai-shi (commented):
should be fine.

src/useAtom.ts
```diff
@@ -40,14 +43,23 @@ export function useAtom<Value, Update>(
         if (initialPromise) {
           throw initialPromise
         }
-        if (!pendingPartialStateRef.current) {
+        if (pendingPartialState.size) {
           pendingPartialStateRef.current = pendingPartialState
```
dai-shi, line 47: Hm, how can we avoid mutating ref in render.

dai-shi: #34

Issue #26, opened 2020-09-10
How to avoid extra rerender?


xiaoyu-tamu:
with Jotai, the `<FakeCard />` will re-render whenever editId changed, but Recoil.js does not because i can select only setter function. How to achieve the same behavior with Jotai? 

## Recoil
```
import React from 'react';
import { atom, useRecoilState, useSetRecoilState } from 'recoil';

const editIdAtom = atom<number | null>({
  key: `editIdAtom`,
  default: null,
});

function FakeCard() {
  const setEditId = useSetRecoilState(editIdAtom);
  const handleClick = React.useCallback(
    (e: React.MouseEvent<HTMLAnchorElement>) => {
      setEditId(1);
    },
    [setEditId]
  );
  return (
    <div>
      <a onClick={handleClick}>toggle</a>
      <FakeModal />
    </div>
  );
}

function FakeModal() {
  const [editId, setEditId] = useRecoilState(editIdAtom);
  const handleClose = React.useCallback(() => {
    setEditId(null);
  }, [setEditId]);

  return editId ? <span>editing {editId}</span> : null;
}
```

## Jotai
```
import React from 'react';
import { atom, useAtom } from 'jotai';

const editIdAtom = atom<number | null>(null);

function FakeCard() {
  const [,setEditId] = useAtom(editIdAtom);
  const handleClick = React.useCallback(
    (e: React.MouseEvent<HTMLAnchorElement>) => {
      setEditId(1);
    },
    [setEditId]
  );
  return (
    <div>
      <a onClick={handleClick}>toggle</a>
      <FakeModal />
    </div>
  );
}

function FakeModal() {
  const [editId, setEditId] = useAtom(editIdAtom);
  const handleClose = React.useCallback(() => {
    setEditId(null);
  }, [setEditId]);

  return editId ? <span>editing {editId}</span> : null;
}
```

dai-shi:
Hi, thanks for opening up the issue.
Jotai core doesn't provide `useSetRecoilState` equivalent. Instead, we use write-only atoms.

```ts
const editIdAtom = atom<number | null>(null);
const setEditIdAtom = atom(null, (_get, set, x: number) => set(editIdAtom, x));

  const [,setEditId] = useAtom(setEditIdAtom);
```

----
Now, I wonder if we can build a generalized hook.
```js
const useSetAtom = (anAtom) => {
[...]

PR #35, opened 2020-09-13
utils bundle and use update atom

dai-shi:
ref: #26 

This adds new `utils` bundle, that is separated from core.

```js
import { useUpdateAtom } from 'jotai/utils'

const countAtom = atom(0)

const Counter = () => {
  const [count] = useAtom(countAtom)
  return <div>count: {count}</div>
}

const Controls = () => {
  const setCount = useUpdateAtom(countAtom)
  const inc = () => setCount(c => c + 1)
  return <button onClick={inc}>+1</button>
}
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -57,7 +57,7 @@
       },
       "alias": {
         "extensions": [".js", ".jsx", ".ts", ".tsx", ".json"],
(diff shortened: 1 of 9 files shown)


dai-shi:
https://codesandbox.io/s/react-typescript-forked-3q11k
Working as expected.

PR #45, opened 2020-09-15
atom family util

### LINKED ISSUE #23: Is atomFamily behavior possible?
Recoil has atomFamily which allows atoms to be created and referenced with something akin to a groupId.  It seems like that kind of behavior can not be replicated without some kind string key identifiers.

dai-shi:
Close #23 

- [x] example snippet
- [x] example codesandbox
- [x] docs/utils.md (how to use areEqual for recoil-like-behavior)

```js
import { atomFamily } from 'jotai/utils'

const todoFamily = atomFamily((name) => name)

  todoFamily('foo')
  // this will create a new atom('foo'), or return the one if already created
```

```js
import { atomFamily } from 'jotai/utils'

const todoFamily = atomFamily(
  (name) => (get) => get(todosAtom)[name],
  (name) => (get, set, arg) => {
    const prev = get(todosAtom)
    return { ...prev, [name]: { ...prev[name], ...arg } }
  }
)
```

```js
import { atomFamily } from 'jotai/utils'

const todoFamily = atomFamily(
  ({ id, name }) => ({ name }),
  null,
  (a, b) => a.id === b.id
)
```

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -48,9 +48,9 @@
     "gzipped": 1866
   },
   "utils.js": {
-    "bundled": 1251,
-    "minified": 629,
-    "gzipped": 311,
+    "bundled": 1827,
+    "minified": 865,
+    "gzipped": 434,
     "treeshaked": {
       "rollup": {
         "code": 28,
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -48,9 +48,9 @@
     "gzipped": 1927
   },
   "utils.js": {
-    "bundled": 1827,
-    "minified": 865,
-    "gzipped": 434,
+    "bundled": 1823,
+    "minified": 870,
+    "gzipped": 436,
     "treeshaked": {
       "rollup": {
         "code": 28,
--- a/docs/utils.md
+++ b/docs/utils.md
@@ -110,3 +110,67 @@ const countReducerAtom = atomWithReducer(0, countReducer)
 ```
 
 https://codesandbox.io/s/react-typescript-forked-g3tsx
+
+## atomFamily
+
+Ref: https://github.com/pmndrs/jotai/issues/23
+
+### Usage
+
+```js
+atomFamily(initializeRead, initializeWrite, areEqual): (param) => Atom
+```
+
+This will create a function that takes `param` and returns an atom.
+If it's already created, it will return it from the cache.
+`initializeRead` and `initializeWrite` are functions and return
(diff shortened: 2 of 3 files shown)


dai-shi:
I'm not really happy with replicating types from `atom()`. Are there any better ways, including fixing types of `atom()`?

brookslybrand:
So sorry @dai-shi, this took me longer than I expected to test. I kept having issues linking the package to my demo, and basically had to end up hacking it together.

That being said, it was actually really easy to migrate! I ended up removing a net of 49 lines with Jotai from Recoil. I want to get the diff up on GitHub to be able to share at some point.

The only thing I'm really not sure about is the `remove` method. I think it could be a bit confusing, because if you call `myAtomFamily.remove(params)`, it will remove the reference inside of the map, but it's not deleting the atom itself. So if you just call that but you still have a hook like `useAtom(myAtomFamily(params)` setup, you'll immediately add the atom back. Maybe this is an expected result, but I found it a bit confusing and not really useful since I had to coordinate it's removal along with the actual unmounting of the component using the atom.

I can try and whip up a simple example to show you what I mean if that was at all confusing.

Otherwise great work! I'm really excited for when this gets officially pushed in, definitely makes me want to try jotai more!

dai-shi:
Sounds great! For `remove()`, I'm not sure either. I just feel bad without it, because a user can't avoid memory leak. We could try something like `gc()`, if that makes more sense.

dai-shi:
> remove()

Now, I feel like `remove(param)` is better than `remove(atom)`.

dai-shi:
Hmm,
```ts
type T = typeof atom extends (read: number) => infer R ? R : never  
```
This can't infer type properly.

dai-shi:
Finally made a working and fully typed demo.
https://codesandbox.io/s/react-typescript-forked-8zfrn?file=/src/App.tsx

brookslybrand:
Glad you got a demo up, it's looking good!

I definitely get the desire to have a `remove` and the worry about a memory leak that the a user will face. I still think people aren't necessarily going to realize or get used to the fact that they'll need to _both_ remove the atom from the atomFamily, as well as from some other atom that likely holds it's params (or however they're creating and storing the params).

> Now, I feel like `remove(param)` is better than `remove(atom)`.
I totally agree, and I didn't even realize it was the other way 😅 I was trying to do it with the param initially, so it wasn't even working. It's a bit difficult though to know whether or not you've removed the atom, because by checking you'll basically just recreate the atom.

PR #49, opened 2020-09-17
use iso layout effect

dai-shi:
- `lastState` in Provider should be updated in layout effect. too late with passive effect
- if `isSSR`, invoke the effect in render otherwise effect never fires.


dai-shi:
(CodeSandbox CI, where are you?

PR #55, opened 2020-09-19
support optional action/update

### LINKED ISSUE #54: Allow reducer updates to be optional
The `dispatch` function of reducer atoms should support optional updates if the reducer defines the update as optional.

To detect this, Jotai could use the same pattern that React's `useReducer` does:

```ts
type ReducerWithoutAction<V> = (value: V) => V;
const reducerWithOptionalAction: ReducerWithoutAction<number> = (value, action?: boolean) => v;
```

Use case:

```ts
const sidebarAtom = atomWithReducer(true, (prev, update?: boolean) => {
	return update == null ? !prev : update
});

	const [isSidebarVisible, toggleSidebar] = useAtom(sidebarAtom);

	return <button onClick={() => toggleSidebar()}>toggle</button>;
}
```

dai-shi:
Close #54

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -48,7 +48,7 @@
     "gzipped": 2013
   },
   "utils.js": {
-    "bundled": 1823,
+    "bundled": 2126,
     "minified": 870,
     "gzipped": 436,
     "treeshaked": {
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -48,9 +48,9 @@
     "gzipped": 2013
   },
   "utils.js": {
-    "bundled": 2126,
-    "minified": 870,
-    "gzipped": 436,
+    "bundled": 1966,
+    "minified": 880,
+    "gzipped": 441,
     "treeshaked": {
       "rollup": {
         "code": 28,
--- a/src/types.ts
+++ b/src/types.ts
@@ -13,8 +13,14 @@ export type Atom<Value> = {
   read: (get: Getter) => Value | Promise<Value>
 }
 
+type Write<Update> = [Update] extends [never]
+  ? (get: Getter, set: Setter) => void | Promise<void>
+  : undefined extends Update
+  ? (get: Getter, set: Setter, update?: Update) => void | Promise<void>
+  : (get: Getter, set: Setter, update: Update) => void | Promise<void>
+
 export type WritableAtom<Value, Update> = Atom<Value> & {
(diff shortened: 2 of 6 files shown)


dwelle:
This was actually the approach I first suggested in the issue, but it has two problems:

1. you need to specify `update` as `| undefined` ([sandbox](https://codesandbox.io/s/react-typescript-forked-03fee?file=/src/App.tsx))
2. `| undefined` doesn't technically mean optional then I found out how React does it and felt it was better.

That's why I updated my OP when I realized how React does it, which solves both counts.

dai-shi:
I'm not sure if I understand your argument.
I thought you would like to do this: https://codesandbox.io/s/react-typescript-forked-r65o5

> `| undefined` doesn't technically mean optional.

Yeah, but seems like it's the only way to know it.

dwelle:
> I thought you would like to do this: codesandbox.io/s/react-typescript-forked-r65o5

Exactly, but in a way that's transparent to the user so that you can simply mark the prop as optional: [sandbox](https://codesandbox.io/s/react-typescript-forked-1r8p5?file=/src/App.tsx)

> Yeah, but seems like it's the only way to know it.

No, you just showed how to do it in your sandbox: codesandbox.io/s/react-typescript-forked-r65o5 -- this is how React's `useReducer` does it, too. Question is, whether you can tie it into your existing types or not --- I'm not sure if that's what you meant by "only way to know it".

dai-shi:
> https://codesandbox.io/s/react-typescript-forked-1r8p5?file=/src/App.tsx

I see your point. Yeah, this would be nice. As far as I tried with TypeScript playground, I wasn't able to find a solution. Would be nice if you or somebody has an idea.

dai-shi:
Finally fixed: https://codesandbox.io/s/react-typescript-forked-b8n06?file=/src/App.tsx

----

> `| undefined` doesn't technically mean optional.

This could be solved with additional generics param, but it's going to be too messy. I guess it's fine with this. (Also, `Setter` could handle optional update, but didn't come up with a good solution. Leaving it for now.)

dwelle:
The `prev` argument is now set to `any` if the `update?` is made non-optional: [sandbox](https://codesandbox.io/s/react-typescript-forked-57011) :thinking:

dai-shi:
So, I gave up `never` handling.

dai-shi:
Merging this. Let me know if there's still some issues, then it will be a follow up PR.

PR #57, opened 2020-09-21
Bridge for multiple rerenders

### LINKED ISSUE #50: Limitation: Does not work accross multiple roots
Unlike zustand, jotai doesn't work across rerenderers.

This is because jotai is React state oriented.
Also partly intentional to distinguish jotai from zustand.

We might be able to provide an escape hatch for this if there's certain demand and use cases.
Refs:
- https://github.com/pmndrs/react-three-fiber/blob/master/api.md#consuming-context-from-a-foreign-provider
- https://recoiljs.org/docs/api-reference/core/useRecoilBridgeAcrossReactRoots

dai-shi:
Close #50 

```jsx
import { useBridge, Bridge } from 'jotai'

const Component = () => {
  return (
    <Canvas>
      <Bridge value={useBridge()}>
        {/* ... */}
```


drcmda:
How does it work? Will usebridge find all context objects?

Also could the API be trimmed? useBridge could perhaps return a primed Bridge component.

dai-shi:
No magic, but I needed to modify the upstream use-context-selector. We use two **special** contexts internally.

> useBridge could perhaps return a primed Bridge component.

That's what I originally tried. But returning a component function from a hook is considered a bad pattern. Here's non official document: https://www.reddit.com/r/reactjs/comments/9yq1l8/how_do_you_feel_about_a_hook_returning_components/ea3o8ue/

stephencorwin:
> returning a component function from a hook is considered a bad pattern

This still returns a component via the CreateElement call.

stephencorwin:
This isn't really composable if it doesn't take an array of contexts -- which the hook usage doesn't really lend itself to. That's why the version that I offered up composes an array.
```jsx
import React, {FC, useMemo} from 'react';
import {IContextBridgeProps} from './IContextBridge';

const ContextBridge: FC<IContextBridgeProps> = ({contexts = [], children}) => {
  const reversed = useMemo(() => contexts.reverse(), [contexts]);
  return reversed.reduce(
    (acc, {context: Context, value}) => (
      <Context.Provider value={value}>{acc}</Context.Provider>
    ),
    children
  );
};

export default ContextBridge;
```

The typical use case is for folks to forward multiple components that are nested within each other.
```jsx
<Context1.Provider>
  <Context2.Provider>
    <Context3.Provider>
      // ....
    <Context3.Provider>
  <Context2.Provider>
<Context1.Provider>
```
This is why I wrote mine to use the reduce function the way it does. If the intent is to switch the api to use a mutable source in the future and break compatibility. I can submit this more general use bridge to the `drei` library for folks to use in the meantime.

dai-shi:
> This still returns a component via the CreateElement call.

Which one are you referring?

stephencorwin:
Your Bridge.ts

dai-shi:
> This isn't really composable if it doesn't take an array of contexts

The way how jotai is using contexts is internal implementation details and we can't expose it: a) because it's not a normal context, and b) we will be changing the implementation with useMutableSource in the future and the way we use contexts will be changed. We just shouldn't assume it's using contexts.

dai-shi:
> Your Bridge.ts

The `Bridge` function may confuse you without JSX, but it's just a normal function component: a function that returns a react element.

stephencorwin:
I wasn't confused about it, I was just pointed out that in your comment here:
https://github.com/pmndrs/jotai/pull/57#issuecomment-695942652

You mention that it was a bad pattern to return a Component from a hook, but that you were still doing that in your implementation.

dai-shi:
Here's the bad pattern, returning a react component from a hook. (I assume @drcmda means this as trimmed api.)

```ts
export const useBridge = () => {
  const actions = useContext(ActionsContext)
  const state = useContext(StateContext)
  const Bridge: React.FC = ({ children }) => createElement(
    BridgeProvider,
    { context: ActionsContext, value: actions },
    createElement(
      BridgeProvider,
      { context: StateContext, value: state },
      children
    )
  )
  return Bridge
}
```

dai-shi:
I tried it with react-three-fiber and it worked. Merging this.

PR #69, opened 2020-09-26
Fix async stale value

### LINKED ISSUE #65: Race conditions in rapidly updating async atoms
@dai-shi's example sandbox: https://codesandbox.io/s/nextjs-with-jotai-async-pm0l8?file=/pages/_app.js

### Current Behavior
Clicking the button rapidly immediately opts-out of the fallback UI and shows a stale success state, then switches to the current state.

### Expected Behavior
Suspense should not opt-out of fallback until the current Promise state has been resolved.

dai-shi:
Close #65.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -24,9 +24,9 @@
     "gzipped": 1922
   },
   "index.js": {
-    "bundled": 15164,
-    "minified": 6081,
-    "gzipped": 2212,
+    "bundled": 15236,
+    "minified": 6092,
+    "gzipped": 2225,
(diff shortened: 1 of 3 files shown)


dai-shi:
https://codesandbox.io/s/nextjs-with-jotai-async-sns34 is working good!

dai-shi:
There can still be a race condition in some edge cases.
```
atomA ------------------------> atomB ------> atomC
        atomA ------> atomB ----------------------------> atomC
```
We should write value anyway with keeping promise.

PR #72, opened 2020-09-28
Fix async atom init

### LINKED ISSUE #71: Imperatively refresh atomFamily
What would be the best way to imperatively refresh an `atomFamily`?

Let's say I have following case, where I use `get(updateAtomFamily)` in order to track it as dependency in the `atomFamily` so that I trigger the fetch imperatively:

```javascript
export const fetchAtomFamily = atomFamily(
  (param) => async (get) => {
    get(updateAtomFamily);
    const { data } = await getData({
      param1: param.param1,
      param2: param.param2,
    });
  },
  null,
  deepEqual
);

export const updateAtomFamily = atom(0);
```

Then my idea was to imperatively `useUpdateAtom` in my component as a refresh function like this:

```javascript
  const refresh= useUpdateAtom(updateAtomFamily);
```

and then have the imperative call:

```html
<button onClick={() => refresh(prev => prev + 1)}>refresh</button> 
```

However, when I trigger the refresh function, I get following error:

```
Uncaught Error: atom state not found. possibly a bug.
```

In my opinion however, even if this approach worked, it feels sluggish and cumbersome to me. It would be extremely helpful if the API had a native approach to re-fetch async atoms :)

dai-shi:
Fix #71 `atom state not found. possibly a bug.` error.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -24,9 +24,9 @@
     "gzipped": 1922
   },
   "index.js": {
-    "bundled": 16284,
-    "minified": 6490,
-    "gzipped": 2362,
+    "bundled": 16314,
+    "minified": 6493,
+    "gzipped": 2366,
     "treeshaked": {
       "rollup": {
         "code": 300,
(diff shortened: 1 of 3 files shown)


dai-shi:
It doesn't seem to fix the original case: https://codesandbox.io/s/brave-greider-cp8vu?file=/package.json

dai-shi:
https://codesandbox.io/s/smoosh-mountain-lxixr?file=/src/atoms.js
It's fixed now. I'm not 100% sure this is the right fix. There could be a better one.
(It's not only this case, but Provider's `readAtom` in general though.)

goncy:
I was having this exact problem and it works perfectly on that sandbox

dai-shi:
Published: https://www.npmjs.com/package/jotai/v/0.7.3

PR #75, opened 2020-10-01
avoid running write before initialization

### LINKED ISSUE #74: Atom initialization bug with two atoms and useEffect
Reported in Discord channel.

Reproduction: https://codesandbox.io/s/competent-rubin-zf1nd?file=/src/App.js

Simply by opening it, it shows an error. 😱

dai-shi:
close #74 

This fixes this certain case, but it's not 100% certain if there's another case failing similarly.
So far, I wasn't able to create another failing test.

### DIFF
--- a/src/Provider.ts
+++ b/src/Provider.ts
@@ -510,7 +510,7 @@ const runWriteThunk = (
   writeThunkQueue: WriteThunk[]
 ) => {
   while (true) {
-    if (lastStateRef.current === null) {
+    if (!lastStateRef.current) {
       return
     }
     if (writeThunkQueue.length === 0) {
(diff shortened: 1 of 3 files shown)


src/Provider.ts
```diff
@@ -546,16 +540,24 @@ export const Provider: React.FC = ({ children }) => {
...
 
+  const [state, setStateOrig] = useState(initialState)
+  const lastStateRef = useRef<State | null>(null)
+  const setState = (setStateAction: SetStateAction<State>) => {
+    lastStateRef.current = null
+    setStateOrig(setStateAction)
+  }
+
+  useIsoLayoutEffect(() => {
+    if (state !== initialState) {
```
dai-shi, line 551: It feels weird, because this only works for initial Provider mount.
Anyway, let's move forward. We have some tests so it won't break.

PR #79, opened 2020-10-02
Reuse read pending atom state

### LINKED ISSUE #76: Concurrent mode async data fetching with atomFamily
Hey let's say I have a `Parent` component that fetches data and renders a list of that data. So far so good. Then for each element of the parent data fetched a `Child` component is rendered that uses `atomFamily`. Now each `Child` component initialized the `atomFamily` with the same ID, so I thought the caching mechanism should kick in and after it sees that data was fetched once from the `atomFamily`, subsequent calls with `useAtom` in other children will not make HTTP requests anymore but will read data from cache.

codesandobox: https://codesandbox.io/s/delicate-rain-f89kt You can open the console in the network tab to see that `Child` is making as many HTTP requests as list elements in `Parent`

dai-shi:
close #76

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 16348,
-    "minified": 6507,
-    "gzipped": 2375,
+    "bundled": 17032,
+    "minified": 6774,
+    "gzipped": 2438,
     "treeshaked": {
       "rollup": {
         "code": 300,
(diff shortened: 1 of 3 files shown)


dai-shi:
code becomes messy. size gets bigger. we'd need to do some refactoring eventually.

dai-shi:
https://codesandbox.io/s/elastic-northcutt-ilhcu
Seems like working as expected? cc @fkocovski

fkocovski:
> https://codesandbox.io/s/elastic-northcutt-ilhcu
> Seems like working as expected? cc @fkocovski

Yes indeed, now only one fetch request is being made by the `Child` component! This is a game changer, terrific work, thank you!

PR #80, opened 2020-10-03
useContextUpdate in use-context-selector v1.2

dai-shi:
The whole point of v0.8.0 was updating to use-context-selector v1.2, but I forgot to apply this change to use useContextUpdate. I also found a bug in use-context-selector v1.2 which is fixed.

### DIFF
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -1,4 +1,4 @@
-import React, { StrictMode, useEffect } from 'react'
+import React, { StrictMode, useEffect, useRef } from 'react'
 import { fireEvent, render } from '@testing-library/react'
 import { Provider, atom, useAtom, WritableAtom } from '../src/index'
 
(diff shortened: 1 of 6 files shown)


tests/basic.test.tsx
```diff
@@ -177,11 +177,14 @@ it('only re-renders if value has changed', async () => {
...
     const [count, setCount] = useAtom(countAtom)
-    const renderCount = React.useRef(0)
+    const commits = useRef(1)
+    useEffect(() => {
+      ++commits.current
+    })
     return (
       <>
         <div>
-          renderCount: {++renderCount.current}, {name}: {count}
```
dai-shi, line 184: Because use-context-selector v1.2 useContextUpdate forces double renders for better CM support, we can't rely on `renderCount` for testing. So, we use commits count instead.

tests/basic.test.tsx
```diff
@@ -177,11 +177,14 @@ it('only re-renders if value has changed', async () => {
   type Props = { countAtom: typeof count1Atom; name: string }
   const Counter: React.FC<Props> = ({ countAtom, name }) => {
     const [count, setCount] = useAtom(countAtom)
-    const renderCount = React.useRef(0)
+    const commits = useRef(1)
```
dai-shi, line 180: However, I'm not sure if I understand how RTL and `findBy*` work. It looks like the initial effect is ignored, and I made `1` initially, which seems totally wrong...

BTW, testing with `findBy*` is handy but not like the typical usage. There would be a better way. (Hope someone chimes in eventually to help on it.)

PR #91, opened 2020-10-04
add tests for useUpdateAtom

merisbahti:
I saw that you needed some tests, so I added this one.

I'm not sure on the format that we want to add tests for utils, I suggest the convention `utils.${function}.test.tsx` or similar, how do you feel about this? Or should we make tests for _all_ functions in one file?

If that's alright, I'll make PR's similar to this one for the remaining functions.

### DIFF
--- a/tests/utils/useUpdateAtom.test.tsx
+++ b/tests/utils/useUpdateAtom.test.tsx
@@ -1,0 +1,69 @@
+import React, { StrictMode, useEffect, useRef } from 'react'
+import { fireEvent, render } from '@testing-library/react'
+import { Provider, atom, useAtom, WritableAtom } from '../../src/index'
+import { useUpdateAtom } from '../../src/utils'
+
+const useRerenderCount = () => {
+  const rerenderCountRef = useRef(0)
+  useEffect(() => {
+    rerenderCountRef.current += 1
+  })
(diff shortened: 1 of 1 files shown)


merisbahti:
Alright, I've moved the files now, please take a look again :+1:

tests/utils/useUpdateAtom.test.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { StrictMode, useEffect, useRef } from 'react'
+import { fireEvent, render } from '@testing-library/react'
+import { Provider, atom, useAtom, WritableAtom } from '../../src/index'
+import { useUpdateAtom } from '../../src/utils'
+
+const useRerenderCount = () => {
+  const rerenderCountRef = useRef(0)
+  useEffect(() => {
+    rerenderCountRef.current += 1
```
dai-shi, line 9: I called this commit counts or `commits` in other files, but let's revisit it later because I don't know which is better.

dai-shi (approved):
👍

tests/utils/useUpdateAtom.test.tsx
merisbahti: I see, I agree, it would be better to be completely consistent.

Issue #84, opened 2020-10-04
[Question] Where do atoms and derived atoms values 'live'?


fjcalzado:
Hi @dai-shi, first of all thank you for this library. It is great to have different options, approaches and improvements around the React state topic. Good job!

I'm playing a bit with it and trying to better understand what's happening under the hood, which use cases can be easily covered with jotai and which doesn't (if any). And several questions arised. I didn't check code yet, sorry, so my questions are based on guessing.

My first assumption when I saw the `Provider` wrapping the tree:
```
<Provider>
  <App />
</Provider>
```
and also based [on this quote](https://github.com/pmndrs/jotai/issues/13#issuecomment-698961443):
>  If you need React Context alternatives, Jotai comes with enough features

was to think that:

- [ ] Provider is kind of a contextual store where all the atom values live. Is that correct?

Also, for learning/debug purposes:

- [ ] Is there a way we can check atom values directly from its "store" without consuming them? Similar to what Redux DevTools offer.

fjcalzado:
I'll continue in this post

dai-shi:
> Provider is kind of a contextual store where all the atom values live. Is that correct?

Yes, totally. Provider holds a big `Map` of atom and atom values.

> Is there a way we can check atom values directly from its "store" without consuming them? 

I had thought about adding `useDebugValue` in Provider, which we already have in useAtom, but it wouldn't be very useful, because an atom is not printable. Maybe we could create a special serializer to show the big `Map`, but I'm not sure how.

fjcalzado:
Can I ask you why is not printable?

By the way, I edited my post 2 above with a new sample and question, maybe you missed it, sorry.

dai-shi:
> the user data is persisted in the Provider, so the second time PageA is mounted it has user data available.

Yes, it's by design and `read` must be pure (in react terms), so it's technically correct.
Notice if you unmount PageA, the value in Provider is garbage collected, and after that if you mount PageA again, it will re-run the `read`.

> My guess is this behaviour has something to do with referential equality for the atom, being different on each mount maybe?

No, it has nothing to do with ref equality. I would do the same if we used string keys.
This behavior is because we store atom values in Provider, and only initialize it once when we add it in Provider.

dai-shi:
> Can I ask you why is not printable?

I simply mean this:
```js
> const atomA = {}, atomB = {}
undefined
> const map = new Map()
undefined
> map.set(atomA, 1)
Map { {} => 1 }
[...]

PR #93, opened 2020-10-04
useDebugValue in Provider

dai-shi:
ref #84 

> I had thought about adding useDebugValue in Provider

> `countAtom.debugLabel = "count"`

Trying this approach.


dai-shi:
@fjcalzado Would you try this if it works for you?

fjcalzado:
Also, another possible one-liner approach:

```js
const countAtom = atom(0).label("count");
```
> Would you try this if it works for you?

I should see the label in React DevTools > Provider > State Map, right? I didn't see it. Checking the PR, `useDebugValue` should be placed whithin a hook maybe...

dai-shi:
> React DevTools > Provider > State Map

It should be in `React DevTools > Provider > DebugValue`, if I remember correctly.

dai-shi:
> ```js
> const countAtom = atom(0).label("count");
> ```

This would require a tiny bit of code. Not sure if it's worth it.

fjcalzado:
@dai-shi I am afraid I can't see any debug label in dev tools.

image

This is the [sandbox](https://codesandbox.io/s/jotai-debug-value-provider-40w8v), forked from your build [#60001](https://ci.codesandbox.io/status/pmndrs/jotai/pr/93).

src/Provider.ts
```diff
@@ -646,6 +647,16 @@ export const Provider: React.FC = ({ children }) => {
     }),
     []
   )
+  useDebugValue(
```
fjcalzado: As far as I know, `useDebugValue` sets a label for custom hooks, so it must be located whithin a hook. However, this is not under any hook but a component.

dai-shi: I see. I wasn't aware that. Thanks for confirming. So, should we create a custom hook for this purpose? Sounds a bit weird, but let's try and see how it goes.

dai-shi:
It's now in a custom hook.

dai-shi:
Trying with todo example: https://codesandbox.io/s/react-typescript-forked-cvfeh?file=/src/App.tsx

Looks like it's working.
image

dai-shi:
Should string `undefined` be something else?

dai-shi:
BTW, I'm not sure why those 4 first items are there. We need to investigate it.
So, it proves this addition is worth it. Nice input! @fjcalzado

src/core/Provider.ts
```diff
@@ -623,6 +625,10 @@ export const Provider: React.FC = ({ children }) => {
     }),
     []
   )
+  if (process.env.NODE_ENV !== 'production') {
```
dai-shi, line 628: Notice it doesn't increase the bundle size at all in production, I suppose.

dai-shi:
```js
const countAtom = atom(0, "count")
const countAtom = atom(0).label("count")
addAtomDebugLabel(countAtom, "count")
```

If we want any of new syntax, we could try to have it in utils. 

> Should string undefined be something else?

Maybe `"<no debugLabel>"`, which tells what it is. I'll make this change and merge it.

fjcalzado:
Sorry @dai-shi , long day at work. It looks really good!

> BTW, I'm not sure why those 4 first items are there. We need to investigate it.

You mean the 4 first items of map? I also saw something weird, like entries repeated in the map.

dai-shi:
No problem! Yes, the 4 first items of map.

PR #100, opened 2020-10-05
add atomWithReducer.test

merisbahti:
I was just thinking, is there any reason there is no reducer-without-argument function signature for this one.

The specific one I'm thinking about is:

```typescript
export function atomWithReducer<Value, Action>(
  initialValue: Value,
  reducer: (v: Value) => Value
): WritableAtom<Value, undefined>
``` 
(Or similar)

### DIFF
--- a/tests/utils/atomWithReducer.test.tsx
+++ b/tests/utils/atomWithReducer.test.tsx
@@ -1,0 +1,84 @@
+import React from 'react'
+import { fireEvent, render } from '@testing-library/react'
+import { Provider, useAtom } from '../../src/index'
+import { atomWithReducer } from '../../src/utils'
+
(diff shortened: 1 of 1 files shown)


dai-shi:
> I was just thinking, is there any reason there is no reducer-without-argument function signature for this one.

I originally thought there's no use case, but actually this reducer doesn't run in render, which means we can make side effects. So, it's totally valid to add the signature.

But, I'm not sure if we should call it "reducer", what do you think?
On the other hand the reducer in `useReducerAtom` runs in render, so it must be "pure".

dai-shi:
Merging this.

@merisbahti I would like to discuss about my previous comment, either in this closed PR or maybe a new issue.

PR #101, opened 2020-10-05
add atomFamily.test

merisbahti:


tests/utils/atomFamily.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <button onClick={() => setCount((oldValue) => oldValue + 1)}>
+          increment #{index}
+        </button>
+      </div>
+    )
+  }
+
+  const Parent: React.FC = () => {
+    const [indices] = useAtom(
+      atom((get) => [...new Array(get(arrayAtom).length)])
```
dai-shi, line 81: Is it intentional to create an atom in each render, instead of creating it outside component?

dai-shi (approved):
wow, it's very interesting to see how atomFamily is used.

another non-blocking comment: would you like to add a test using a custom `areEqual`?

tests/utils/atomFamily.test.tsx
merisbahti: Well, yeah... I think of it "it's the parents responsibility to create it, no one else is interested in it".

Should have I created it outside?

Is this a possible memory leak?

merisbahti:
> wow, it's very interesting to see how atomFamily is used.
Thank you. I will treat it both as a compliment and as something else :)

Did you have any other intention for it? How would you have used it? I can add your use case too in the test.

> another non-blocking comment: would you like to add a test using a custom `areEqual`?

That is a good idea, I was blinded by that it said 100% test coverage so I didn't think about it. Thank you.

tests/utils/atomFamily.test.tsx
dai-shi: > Should have I created it outside?

Yes, please. It's more idiomatic.

> Is this a possible memory leak?

It doesn't leak memory, but every time the component renders, it will run initialization and garbage collection. So, it's a performance issue.

dai-shi:
> Did you have any other intention for it?

Honestly, I built it upon request, and I didn't have any use case in mind at that point.
The only my trial is this example: https://codesandbox.io/s/react-typescript-forked-8zfrn

tests/utils/atomFamily.test.tsx
merisbahti: > It doesn't leak memory, but every time the component renders, it will run initialization and garbage collection. So, it's a performance issue.

I understand. Fortunately, in this example there's no way to make Parent re-render ;)

But yes, I think this answers another question I've had about jotai, the way I've been using it there's this "commit" in the performance tab of react tools, this commit always comes before the real component update, that might be the initialization/garbage-collection commit!

merisbahti:
Hey, it looks like this test started breaking after me rebasing to the newest jotai with this error:
``` 
FAIL tests/utils/atomFamily.test.tsx
  ● atomFamily functionality as usual

    Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.

      24 |   atom: Atom<Value> | WritableAtom<Value, Update>
      25 | ) {
    > 26 |   const actions = useContext(ActionsContext)
         |                   ^
      27 | 
      28 |   const value = useContextSelector(
      29 |     StateContext,

      at resolveDispatcher (node_modules/react/cjs/react.development.js:1465:13)
      at Object.useContext (node_modules/react/cjs/react.development.js:1473:20)
      at useContextSelector (node_modules/use-context-selector/src/index.js:93:30)
      at useContext (node_modules/use-context-selector/src/index.js:157:40)
      at useAtom (src/core/useAtom.ts:26:19)
      at Object.<anonymous> (tests/utils/atomFamily.test.tsx:79:21)
``` 

Do you have any idea what this might be because?

dai-shi:
Here's the fix:

```diff
diff --git a/tests/utils/atomFamily.test.tsx b/tests/utils/atomFamily.test.tsx
index 162c3bc..432f078 100644
--- a/tests/utils/atomFamily.test.tsx
+++ b/tests/utils/atomFamily.test.tsx
@@ -76,11 +76,10 @@ it('atomFamily functionality as usual', async () => {
     )
   }
 
-  const [indices] = useAtom(
-    atom((get) => [...new Array(get(arrayAtom).length)])
-  )
+  const indicesAtom = atom((get) => [...new Array(get(arrayAtom).length)])
 
   const Parent: React.FC = () => {
+    const [indices] = useAtom(indicesAtom)
     return (
       <div>
         {indices.map((_, index) => (
```

merisbahti:
> Here's the fix:
> 
> ```diff
> diff --git a/tests/utils/atomFamily.test.tsx b/tests/utils/atomFamily.test.tsx
> index 162c3bc..432f078 100644
> --- a/tests/utils/atomFamily.test.tsx
> +++ b/tests/utils/atomFamily.test.tsx
> @@ -76,11 +76,10 @@ it('atomFamily functionality as usual', async () => {
>      )
>    }
>  
> -  const [indices] = useAtom(
> -    atom((get) => [...new Array(get(arrayAtom).length)])
> -  )
> +  const indicesAtom = atom((get) => [...new Array(get(arrayAtom).length)])
>  
>    const Parent: React.FC = () => {
> +    const [indices] = useAtom(indicesAtom)
>      return (
>        <div>
>          {indices.map((_, index) => (
> ```

I must have had a weak moment, thanks.

dai-shi (approved):
Looks good. Thanks!

PR #103, opened 2020-10-06
Refactor step 1 with immutable map

dai-shi:
We plan code refactoring and optimization. This is step 1.

- Introduce new ImmutableMap. This abstracts some logic and hopefully improve readability.
- Simplify readAtom. I think I did something unnecessary...

There some points that we need to investigate further, but I think it's working as is and safe to merge.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 18209,
-    "minified": 7331,
-    "gzipped": 2733,
+    "bundled": 18330,
+    "minified": 7401,
+    "gzipped": 2776,
     "treeshaked": {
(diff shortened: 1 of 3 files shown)


.size-snapshot.json
```diff
@@ -1,27 +1,27 @@
 {
   "index.js": {
-    "bundled": 17771,
-    "minified": 7164,
-    "gzipped": 2549,
+    "bundled": 18330,
+    "minified": 7401,
+    "gzipped": 2776,
```
dai-shi, line 5: 😭

dai-shi:
Todo example with atomFamily is working fine.
https://codesandbox.io/s/react-typescript-forked-5d62g?file=/src/App.tsx

PR #106, opened 2020-10-06
Batched writes

dai-shi:
This will improve write performance. Test are passing, but there could be some edge cases that aren't taken care of.


dai-shi:
While testing this, I found the issue #110. As it's not a regression of this PR change, I'd merge this and work on it after that.

Issue #122, opened 2020-10-10
Add docs and test for jotai/immer


dai-shi:
This is a follow up task for #120.

Aslemammad:
Which part should I edit? And how it works?

dai-shi:
For docs: Create a new file `./docs/immer.md`

For tests: Create three new files `./tests/immer/atomWithImmer.d.ts` `./tests/immer/useImmerAtom.d.ts` `./tests/immer/withImmer.d.ts`

`yarn test` should run all tests.

Aslemammad:
@dai-shi I did the same thing for the test, yesterday. I'll work on them. thanks

Aslemammad:
People, if you have any idea about immer docs in jotai, share with us here.

dai-shi:
(Should we keep this issue open for a while?

Aslemammad:
@dai-shi Choose the best. I think we shall close it because if someone wanted to edit the docs, he/she would create a new issue, so do the best.

dai-shi:
Closing this as the initial work is done with #125. Please file a new issue/PR for any improvements.

PR #125, opened 2020-10-11
immer docs

Aslemammad:
#122 @dai-shi I wrote the docs for immer support.


dai-shi (approved):
Looks good. We can keep improving it. Merging this.

PR #126, opened 2020-10-11
clear lastStateRef in write

dai-shi:
This may or may not solve write batch issues in some edge cases.


dai-shi:
This doesn't seem to change the behavior. Closing.

PR #128, opened 2020-10-12
Fix updating async atoms with depndencies

### LINKED ISSUE #127: Batched write dependency
Batched write dependencies can lead to queueing of state updates:

https://codesandbox.io/s/white-smoke-hzw12?file=/src/atoms.js

If selector `atom` is marked as dependency in another `atom`, batched write does not propagate writes anymore.

Tested both with `atom` and `atomFamily`

dai-shi:
close #127

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,27 +1,27 @@
 {
   "index.js": {
-    "bundled": 16393,
-    "minified": 7275,
-    "gzipped": 2794,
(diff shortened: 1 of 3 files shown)


dai-shi:
The sandbox in #127 is working. Let's merge this in.

dai-shi:
Published: https://www.npmjs.com/package/jotai/v/0.9.2

PR #133, opened 2020-10-16
new util useSelector

### LINKED ISSUE #36: [Tips] useSelector equivalent
While the atom model fits with bottom-up approach (composing small atoms), we could create a big atom and select pieces out of it.

```js
const bigAtom = atom({
  num: 0,
  str: 'hi',
  bool: true,
  obj: { a: 1, b: 2, arr: [3, 4, 5] },
})

const numAtom = atom(
  get => get(bigAtom).num,
  (get, set, update) => {
    const prev = get(bigAtom)
    const num = typeof update === 'function' ? update(prev.num) : update
    set(bigAtom, { ...prev, num })
  }
)

  const [num, setNum] = useAtom(numAtom)
```

We can make this a hook. Let's try for a read-only atom.

```js
const useSelector = (anAtom, selector) => {
  const selectedAtom = useMemo(() => atom(get => selector(get(anAtom))), [anAtom, selector])
  return useAtom(selectedAtom)[0]
}
```

dai-shi:
close #36 

- [x] implementation
- [x] example
- [x] docs

----

### Example

```tsx
import * as React from "react";
import { Provider } from "jotai";
import { useSelector, atomWithReducer, useUpdateAtom } from "jotai/utils";
import "./styles.css";

type State = {
  count: number;
  text: string;
};

const initialState: State = {
  count: 0,
  text: "hello"
};

type Action =
  | {
      type: "INC";
    }
  | {
      type: "SET_TEXT";
      text: string;
    };

const reducer = (state: State, action: Action): State => {
  if (action.type === "INC") {
    return { ...state, count: state.count + 1 };
  } else if (action.type === "SET_TEXT") {
    return { ...state, text: action.text };
  } else {
    throw Error("no such action");
  }
};

const stateAtom = atomWithReducer(initialState, reducer);

const selectCount = (state: State) => state.count;

const Counter: React.FC = () => {
  const dispatch = useUpdateAtom(stateAtom);
  const count = useSelector(stateAtom, selectCount);
  return (
    <div>
      {count} <button onClick={() => dispatch({ type: "INC" })}>+1</button>
    </div>
  );
};

const selectText = (state: State) => state.text;

const TextBox: React.FC = () => {
  const dispatch = useUpdateAtom(stateAtom);
  const text = useSelector(stateAtom, selectText);
  return (
    <div>
      {text}{" "}
      <input
        value={text}
        onChange={(e) => dispatch({ type: "SET_TEXT", text: e.target.value })}
      />
    </div>
  );
};

const App: React.FC = () => (
  <Provider>
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <h2>Start editing to see some magic happen!</h2>
      <Counter />
      <TextBox />
    </div>
  </Provider>
);

export default App;
```

https://codesandbox.io/s/react-typescript-forked-i4880


Aslemammad:
@dai-shi Could you talk more about the problem and ....?

PR #138, opened 2020-10-16
fix: esm build with new use-context-selector

### LINKED ISSUE #136: Problem with ESM/CJS configuration and Rollup
Seems your config is sending mixed messages about your package state to Rollup. I don't claim to understand the issue, but the problem arises when using [Snowpack](https://www.snowpack.dev/). @FredKSchott describes here: https://github.com/pikapkg/snowpack/discussions/1321

dai-shi:
close #136

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     ]
   },
   "dependencies": {
-    "use-context-selector": "1.2.8"
+    "use-context-selector": "1.2.9"
   },
   "devDependencies": {
     "@babel/core": "^7.11.6",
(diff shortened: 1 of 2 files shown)


dai-shi:
@DynamicArray 
Could you try if this fixes your use case?
You can install the package built by codesandbox ci: https://ci.codesandbox.io/status/pmndrs/jotai/pr/138/builds/63576

DynamicArray:
Works a treat! Thank you for fixing that so quickly.

FredKSchott:
LGTM! Didn't run myself but checked the new use-context-selector and it looks good. Thanks for the quick fix!

dai-shi:
Thanks both for confirmation! Merging this.

PR #139, opened 2020-10-16
installation notes for yarn 2

### LINKED ISSUE #134: scheduler not mentioned as peerDependency in documentation
When using this package with yarn v2 (yarn berry), it throws while building:

```
Module not found: Error: jotai tried to access scheduler (a peer dependency) but it isn't provided by your application; this makes the require call ambiguous and unsound.
```

I can resolve it by adding [scheduler](https://www.npmjs.com/package/scheduler) to my dependencies.

To solve the problem from your side, I see 2 possibilities:
- You could add an installation instruction, that mentions the need to install react and scheduler or 
- you could define scheduler as a dependency
- (or remove the scheduler-dependency if not needed ;-) )

### LINKED ISSUE #135: add react-dom as a peer dependency (yarn 2 compatibility)
When using this package with yarn v2 (yarn berry), it throws while building:

```
Module not found: Error: use-context-selector tried to access react-dom (a peer dependency) but it isn't provided by its ancestors; this makes the require call ambiguous and unsound.
```

> this error comes after resolving #134 by adding scheduler as a dependency ;-) (resolving this error is a bit easier than #134 - so I created another issue, hope this is fine for you)

You depend on [use-context-selector](https://www.npmjs.com/package/use-context-selector), and they declared `react-dom` as a peerDependency. So you should also add `react-dom` to your peerDependencies to make sure, it is available.

## workaround at the moment

Currently workaround is adding this to `.yarnrc.yml`

```yaml
packageExtensions:
  jotai@*:
    peerDependencies:
      "react-dom": "*"
```

info about packageExtensions:

> Some packages may have been specified incorrectly with regard to their dependencies - for example with one dependency being missing, causing Yarn to refuse it the access. The packageExtensions fields offer a way to extend the existing package definitions with additional information.

https://yarnpkg.com/configuration/yarnrc#packageExtensions

dai-shi:
close #134 
close #135

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -173,9 +173,12 @@ yarn add jotai react scheduler react-dom
 ```
 
 For React Native users:
+
 ```bash
 yarn add jotai react scheduler react-native
 ```
+
+For users who use workspaces, check out [this issue](https://github.com/pmndrs/jotai/issues/135) for some workarounds.
 
 ## More information
 
--- a/readme.md
+++ b/readme.md
@@ -164,6 +164,19 @@ function Controls() {
 
 ----
 
+## Installation notes for Yarn 2 users
+
+If you are using yarn v2 (yarn berry), you need to install all peer dependencies by yourself.
+
+```bash
+yarn add jotai react scheduler react-dom
+```
+
+For React Native users:
+```bash
+yarn add jotai react scheduler react-native
+```
+
 ## More information
 
 We will be organizing some more information later. Meanwhile, please see WIP materials.


dai-shi:
@PutziSan 
Could you review this and possibly suggest some improvements?

PutziSan:
This will close #134 but not #135 - yarn 2 will search for the `react-dom` package in the jotai-space and does not find it.

You can see the error here: https://github.com/PutziSan/jotai-yarn-2-workspaces-error-demo
(in fact when I tried to reproduce it seems, that the error only appears for yarn 2 with workspaces-feature enabled. which is... interesting ;-) )

Maybe than also adding peerDependencyMeta optional for react-dom could do the trick.

dai-shi:
Thanks so much for the reproduction repo! Now, I get clearer understanding.

> the error only appears for yarn 2 with workspaces-feature enabled

If this is the case, it seems like an issue with yarn 2. For the error message `tried to access react-dom (a peer dependency) but it isn't provided by its ancestors; this makes the require call ambiguous and unsound.`, we do have react-dom provided in package.json which is the one of the ancestors.

Here's my workaround for your repo:
```diff
diff --git a/package.json b/package.json
index 08b386d..5137e63 100644
--- a/package.json
+++ b/package.json
@@ -4,6 +4,10 @@
   "workspaces": [
     "packages/*"
   ],
+  "dependencies": {
+    "react": "^16.14.0",
+    "react-dom": "^16.14.0"
+  },
   "devDependencies": {
     "prettier": "^2.1.2"
   }
```

PutziSan:
Yep, something seems not correct here, but I thought that the error is the expected behaviour (from yarn 2 perspective ;-) ), because their "mantra" is, that everything needs to be strictly defined.

I opened https://github.com/yarnpkg/berry/issues/1992 for this. So I think for the moment this PR is fine to solve most of the issues.

dai-shi:
Sounds good. I will add my workaround in #135 and add a link to the issue in readme. This will close both issues for now but we keep eyes on it.

readme.md
```diff
@@ -164,6 +164,22 @@ function Controls() {
 
 ----
 
+## Installation notes for Yarn 2 users
+
+If you are using yarn v2 (yarn berry), you need to install all peer dependencies by yourself.
```
merceyz, line 169: These instructions shouldn't be Yarn 2 specific, anyone using this needs to have these dependencies present

dai-shi: As you said in the other issue, yarn 2 (pnp) is especially strict. For other environments, one wouldn't need these instructions. They are often installed implicitly. Personally, I never needed them.

The difference between yarn 2 and others like yarn 1 is not about whether you need those dependencies (because you always need them), but about how dependencies are specified in package.json. These instruction would ease the use case with yarn 2, but for yarn 1 users and npm users, they are not necessary.

merceyz: `jotai` has the same peer dependencies no matter what package manager you're using. I don't mind the instruction, my point is just that the same is true when using npm or pnpm so it shouldn't be Yarn 2 specific as it makes it seem like it's more work to use `jotai` with Yarn 2, which isn't the case.

> They are often installed implicitly. Personally, I never needed them.

I don't think users *usually* install `scheduler`, it just happens to work because hoisting brings it up to an accesible location from `react-dom`. You can read more about hoisting here https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

dai-shi: Hm, so we shouldn't depend on the hoisting behavior. I don't want to make it seem more work with yarn 2.

Issue #92, opened 2020-10-04
atoms and error boundaries


JLarky:
I'm trying out jotai in one of my projects and it was crashing with code like this
```
 React.useEffect(() => {
    setState(s => {
      const x = {};
      delete x[123][123];
    });
  });
 ```
and I was confused why this error wasn't caught by the react error boundary (of `react-error-boundary` fame). Then I realized that my component three was like this `JotaiProvider -> ErrorBoundary -> MyComponent` so doing something like `ErrorBoundary -> JotaiProvider -> ErrorBoundary -> MyComponent` helped, but it makes me realize interesting consideration that with smaller context providers I can have them closer to my component and thus making errors more local, but errors in jotai are basically globally scoped :)

I don't know if that's an issue per se, but maybe a good point to consider when deciding between custom context vs global context

dai-shi:
ErrorBoundary should be inside JotaiProvider.
Provider should catch errors in the render phase and useAtom will throw it.
So, if what you observe is different, it's likely to be a bug.
Would you be willing to create a repro in codesandbox?

JLarky:
OK :) that wasn't easy and I'm still not 100% there :) I managed to reproduce it basically by accident :)

So here are the steps:
1. go to https://codesandbox.io/s/jotai-issue-92-forked-uih92?file=/src/App.tsx
2. edit App.tsx source code and you will see
![image](https://user-images.githubusercontent.com/7026/95647231-a67a6f00-0a82-11eb-9861-f9e2cb298cb8.png)
and under overlay:
![image](https://user-images.githubusercontent.com/7026/95647236-b09c6d80-0a82-11eb-989f-533ac0fbc347.png)
3. expected behavior:
![image](https://user-images.githubusercontent.com/7026/95647251-d6297700-0a82-11eb-902a-fbcbaffcb0bc.png)
and under overlay:
![image](https://user-images.githubusercontent.com/7026/95647265-e7728380-0a82-11eb-9406-7088134ed51e.png)

I was trying to reproduce the issue with `unmount crash` button, which is closer to the code that I have in the real app, but I'm really struggling :) for me clicking on the button works as expected :) so the only way that I could manage to reproduce it was unmounting that is triggered by hot reload, but in my real app it has nothing to do with hot reload, and I see the same issue in production mode. So I will try to improve this code when I would have more time (I'm thinking maybe unmounting is async in the real app or just has more layers of components?), but I hope that the steps that I provided are enough to understand the issue :)

[...]

PR #141, opened 2020-10-17
fix batched writes behavior

dai-shi:
ref: #92 

- [x] failing test (test case for throwing error in useEffect cleanup)
- [x] fix batched writes behavior

We use `Promise.resolve()` to fix this. Not sure this is the ideal solution. We'd need to collect more use cases and feedbacks.


dai-shi:
Let's merge this in and see how it goes.

dai-shi:
> We use `Promise.resolve()`

Looks like preact uses `setTimeout(1)` ? https://preactjs.com/guide/v8/differences-to-react/

PR #142, opened 2020-10-17
no implicit peer dependency

dai-shi:
ref: https://github.com/pmndrs/jotai/issues/135#issuecomment-710782084 https://github.com/pmndrs/jotai/pull/139#discussion_r506916713

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,12 @@
   },
   "peerDependencies": {
     "react": ">=16.8",
-    "scheduler": ">=0.19"
+    "scheduler": ">=0.19",
+    "react-dom": "*"
+  },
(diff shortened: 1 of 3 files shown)


dai-shi:
@merceyz I guess this fixes for react-dom users, but for react-native users we should add `react-native` as an optional dependency, shouldn't we?

merceyz:
Correct, I would like it if you could update the install instructions to not be Yarn 2 specific as well though

dai-shi:
@merceyz If you are still around, would you check the changes?

merceyz (approved):
Perfect, thank you :)

PR #145, opened 2020-10-19
fix: return write error in promise

dai-shi:
This might be a bit controversial. 
Currently, we throw read errors in useAtom, but for write errors we throw them (or reject promises) in write function (more precisely SetAtom function).
This PR fixes some edge cases where it fails to return errors in a promise.

If a write function is async, we always return a promise, however, if a write function is sync, we may throw an error or return a promise which can contain an error. It's not very predictable for users.

Unless there's any better idea, it can't be helped much...

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 18149,
-    "minified": 8200,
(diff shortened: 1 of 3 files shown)


dai-shi:
Hmm, but what would be the behavior if we used bare context and state without jotai?

Issue #147, opened 2020-10-20
State not found v0.10.0 for async atoms


fkocovski:
Async `atoms` that have primitive `atoms` as dependents can't find the state if multiple ones are used inside a component.

Repro with `atomFamily`: https://codesandbox.io/s/zen-hellman-plh9w

Repro with `atom`: https://codesandbox.io/s/suspicious-germain-zr0wu

garrettmk:
I had a similar issue, but it doesn't look like it was solved by #148. Here's a [sandbox](https://codesandbox.io/s/crimson-silence-xvg3o) to illustrate. The atom dependencies look like this:

Primitive atom -> async atom (simulated fetch) -> derived atom

Changing the primitive atom causes the async atom to re-fetch, as expected. However the derived atom never resolves.

dai-shi:
@garrettmk Thanks for reporting. The sandbox link has the code yet unchanged? I'll try to repro in test.

dai-shi:
@garrettmk I'll merge #148, and work on the issue reported as a new PR. Would you create a new issue with the correct sandbox link please?

PR #149, opened 2020-10-20
fix: async atom in the middle of dependency chain

dai-shi:
ref: https://github.com/pmndrs/jotai/issues/147#issuecomment-712997774


dai-shi:
didn't get a response, but merging this.

PR #152, opened 2020-10-23
docs: add typescript version requirement note

### LINKED ISSUE #151: Declaration or statement expected "export type"
https://github.com/pmndrs/jotai/blob/9bd3ea476d1a69f6f967a42a587c9bf8b0d16e41/src/index.ts#L4

`import type`/`export type` construct started being supported on TypeScript 3.8 and I just wanted to point that out because CRA still ships with `3.7.2` and this syntax won't work if you don't update TypeScript. It might be nice to add that on the documentation or use a different syntax for exporting types.

italodeandra:
Close #151

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -35,3 +35,7 @@ const [num, setNum] = useAtom(primitiveNumAtom)
 const [num] = useAtom(readOnlyNumAtom)
 const [, setNum] = useAtom(writeOnlyNumAtom)
 ```
+
+#### Version requirement
+
+jotai uses TypeScript 3.8+ syntax. Upgrade your TypeScript version if you're on 3.7.5 or lower.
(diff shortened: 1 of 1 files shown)


docs/typescript.md
```diff
@@ -35,3 +35,7 @@ const [num, setNum] = useAtom(primitiveNumAtom)
 const [num] = useAtom(readOnlyNumAtom)
 const [, setNum] = useAtom(writeOnlyNumAtom)
 ```
+
+#### Version requirement
```
dai-shi, line 39:
```suggestion
## Version requirement
```

This would be better.

I don't mind putting this section before `## Notes` but it's up to you.

italodeandra: I updated it on commit https://github.com/pmndrs/jotai/pull/152/commits/388994a3727656ca0191498fecfe8850c7a5f433.

dai-shi:
Thanks for your suggestion!

PR #105, opened 2020-10-06
Update package.json to use shx

LuckyPigeon:
refer to #104 
Update package.json copy command using shx package for windows user

PR #154, opened 2020-10-26
Update rollup plugins and fix their ordering #105

jcoussard:
Only dev related changes. No need for a new version release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,8 @@
     "@babel/plugin-transform-react-jsx": "^7.12.1",
     "@babel/plugin-transform-typescript": "^7.12.1",
     "@babel/preset-env": "^7.12.1",
+    "@rollup/plugin-node-resolve": "^9.0.0",
+    "@rollup/plugin-typescript": "^6.0.0",
     "@testing-library/react": "^11.1.0",
     "@types/jest": "^26.0.15",
     "@types/react": "^16.9.53",
@@ -117,9 +119,7 @@
     "react-dom": "^17.0.0",
     "rollup": "^2.32.0",
     "rollup-plugin-babel": "^4.4.0",
-    "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-size-snapshot": "^0.12.0",
-    "rollup-plugin-typescript2": "^0.28.0",
     "scheduler": "^0.19.1",
     "shx": "^0.3.2",
     "typescript": "^4.0.3"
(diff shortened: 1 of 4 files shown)


jcoussard:
This pull request fixes the bundles to inline the local code dependencies instead of importing it. This was noticed after trying to build the project and noticing the .size-snapshot.json was way different.

The first attempts at fixing it were suggested by @dai-shi. I switched the following rollup packages:
- rollup-plugin-node-resolve => @rollup/plugin-node-resolve
- rollup-plugin-typescript2 => @rollup/plugin-typescript

This didn't solve the root issue, but instead resulted in new issues. The @rollup/plugin-typescript didn't like the fact that declarations were outputted without providing an "outDir" value in tsconfig. But if specifying one, then new problems arose in the existing rollup bundles.

After looking for ideas from other projects, I noticed that the resolve() plugin was always put first in all rollup configurations, I tried it and it worked! Then all was left was to re-emit the typescript declaration files. Since I thought it was a good idea to keep the updated dependencies, I created a new rollup entry specifically to emit them.

jcoussard:
Also I guess zustand could profit from the same fix: https://github.com/pmndrs/zustand/blob/master/rollup.config.js

rollup.config.js
```diff
@@ -14,16 +14,27 @@ const getBabelOptions = (targets) => ({
...
 })
 
+function createDeclarationConfig(input, output) {
+  return {
+    input,
+    output: {
+      dir: output,
+    },
+    external,
+    plugins: [typescript({ declaration: true, outDir: output })],
```
dai-shi, line 24: I tried almost all what you tried, and agree this would be the best so far. 😄

package.json
```diff
@@ -117,9 +119,7 @@
     "react-dom": "^17.0.0",
     "rollup": "^2.32.0",
     "rollup-plugin-babel": "^4.4.0",
```
dai-shi, line 121: Would you please try https://www.npmjs.com/package/@rollup/plugin-babel too?

dai-shi:
Thanks for your investigation!

> Also I guess zustand could profit from the same fix

I agree. Would you be interested to work on it too? Otherwise, I can ask someone for help in the discord channel.

package.json
jcoussard: I thought the same while going to bed last night. I'll try it later today.

dai-shi (approved):
LGTM. Thanks for working on this.

PR #158, opened 2020-10-29
add shouldRemove in atomFamily

### LINKED ISSUE #155: atomFamily caches too aggressively async requests
When using `atomFamily` starting from v0.10.0 the `get` part is cached thus preventing new refetch of data even if initialising parameters change. This behaviour does not happen with 0.9.2.

https://codesandbox.io/s/sharp-oskar-2kmtx?file=/src/atoms.js

In order to check, open the console and try to navigate between components `A` and `B`.

dai-shi:
close #155 

This adds a new functionality to atomFamily.
Typical use case would be to remove atoms if it's created a long time ago, and refresh them.

This use case is probably for experts. Naive shouldRemove can lead to infinite loop.

----

### Usage

```js
const family = atomFamily(...)
family.setShouldRemove((createdAt, param) => createdAt + 60 * 1000 < Date.now())
```

### DIFF
--- a/src/utils/atomFamily.ts
+++ b/src/utils/atomFamily.ts
@@ -7,7 +7,7 @@ type ShouldRemove<Param> = (createdAt: number, param: Param) => boolean
 type AtomFamily<Param, AtomType> = {
   (param: Param): AtomType
   remove(param: Param): void
-  gc(shoudRemove: ShouldRemove<Param> | null): void
+  setShouldRemove(shouldRemove: ShouldRemove<Param> | null): void
 }
 
 // writable derived atom
@@ -90,7 +90,7 @@ export function atomFamily<Param, Value, Update>(
       atoms.splice(index, 1)
     }
   }
-  createAtom.gc = (fn: ShouldRemove<Param> | null) => {
+  createAtom.setShouldRemove = (fn: ShouldRemove<Param> | null) => {
     shouldRemove = fn
     if (!shouldRemove) return
     let index = 0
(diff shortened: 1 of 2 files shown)


src/utils/atomFamily.ts
```diff
@@ -2,9 +2,12 @@ import { atom, Atom, WritableAtom, PrimitiveAtom } from 'jotai'
 
 import type { Getter, Setter } from '../core/types'
 
+type ShouldRemove<Param> = (createdAt: number, param: Param) => boolean
+
 type AtomFamily<Param, AtomType> = {
   (param: Param): AtomType
   remove(param: Param): void
+  gc(shoudRemove: ShouldRemove<Param> | null): void
```
dai-shi, line 10: I wonder if there's better name. Would one prefer `setShouldRemove()`?

src/utils/atomFamily.ts
```diff
@@ -80,5 +90,18 @@ export function atomFamily<Param, Value, Update>(
...
+    if (!shouldRemove) return
+    let index = 0
+    while (index < atom.length) {
+      const item = atoms[index]
+      if (shouldRemove(item[2], item[0])) {
+        atoms.splice(index, 1)
+      } else {
+        ++index
+      }
+    }
```
dai-shi, line 104: So, it sets `shoudRemove` and check all atoms just once.
Is it clearer if we expose two functions? set and run?

src/utils/atomFamily.ts
```diff
@@ -80,5 +90,18 @@ export function atomFamily<Param, Value, Update>(
       atoms.splice(index, 1)
     }
   }
+  createAtom.gc = (fn: ShouldRemove<Param> | null) => {
+    shouldRemove = fn
```
dai-shi, line 94: passing `null` will clear `shouldRemove`.

dai-shi:
https://codesandbox.io/s/nifty-silence-2f5y5
@fkocovski Please check this.

PR #160, opened 2020-10-29
Detect react experimental and fix some bugs

### LINKED ISSUE #156: Rerendering issue starting with 0.8.1
It seems that Jotai rerenders all connected components, even ones that are not subscribed to changing atoms.

https://codesandbox.io/s/jotai-demo-forked-4fzwi

This issue starts with 0.8.1. It's okay with 0.8.0. If you type into the input and look at the console, you will see `Static` is rerendering.

dai-shi:
close #156 

The behavior introduced with use-context-selector v1.2 is confusing developers.
As it would only be required for experimental concurrent mode support,
this adds a detection and changes the behavior in legacy mode.

It revealed some bugs around pending state, and they are fixed.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,27 +1,27 @@
 {
   "index.js": {
-    "bundled": 19093,
-    "minified": 8544,
-    "gzipped": 3130,
+    "bundled": 19439,
+    "minified": 8732,
+    "gzipped": 3243,
     "treeshaked": {
       "rollup": {
-        "code": 318,
-        "import_statements": 85
+        "code": 390,
+        "import_statements": 92
       },
       "webpack": {
-        "code": 1441
+        "code": 1519
       }
     }
   },
   "index.cjs.js": {
-    "bundled": 22735,
-    "minified": 10562,
-    "gzipped": 3638
+    "bundled": 23291,
+    "minified": 10914,
+    "gzipped": 3785
   },
   "index.iife.js": {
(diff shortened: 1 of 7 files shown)


dai-shi:
@paulshen
It seems good: https://codesandbox.io/s/jotai-demo-forked-vozqs
If you have a chance, please try your app with this build, and see if it's ok.

paulshen:
@dai-shi it works great when i use this build. thank you!

fkocovski:
@dai-shi it also seems solid from my side, if I detect anything suspicious I'll report back, thank you!

dai-shi:
Thanks guys. Merging this and release it later.

merisbahti:
Is it possible to get some info regarding these bugs? Any test cases? I found that one of my use cases (regarding atomfamily) started working after this upgrade... (from 0.10.1)

dai-shi:
@merisbahti 
I tried to reproduce the bug as failing test cases while working on this, but all my trials were not successful. Only, with the feature flag, I could repro the bug.

I also found the bug while investigating #155.
The failing pattern is something like this:
```js
const family = atomFamily(...);

// in a component
  const [data, refresh] = useAtom(family(param));
  React.useEffect(() => {
    refresh();
  }, [refresh]);
```
I mean I tried to repro this bug behavior in v0.10.1 as a test case, which I couldn't.
Was your use case like this? Would be great if you can create a test case that fails in v0.10.1.

Issue #159, opened 2020-10-29
Consider React Native for persistence support


likern:
Hello! I now it might not be the issue for right now. 
But if you are going to add persistence support, please, _consider React Native also_. So persistence layer would be optional / configurable.

There is no HTML and DOM in React Native and components (`View`, `Text`, `TextInput`) map to Native Android / iOS components. And **there is no Web API** like https://www.w3.org/TR/IndexedDB or https://dev.w3.org/html5/webdatabase
Also there is several solutions for persistence like SQLite database https://github.com/andpor/react-native-sqlite-storage or AsyncStorage https://github.com/react-native-async-storage/async-storage

dai-shi:
Hi,
Please check `docs/persistence.md` and feel free to suggest improvements (and possibly a PR).

We don't add persistence support in core to keep it premitive, but utils can.
So far, because there are many patterns, we haven't yet come up with a single recommended solution.

btw, have you tried jotai with RN? Would be great to hear your experience.

likern:
> Hi,
> Please check `docs/persistence.md` and feel free to suggest improvements (and possibly a PR).
> 
> We don't add persistence support in core to keep it premitive, but utils can.
> So far, because there are many patterns, we haven't yet come up with a single recommended solution.
> 
> btw, have you tried jotai with RN? Would be great to hear your experience.

I'm going to try and test jotai for my React Native project.
I will share my experience.

likern:
I'm using React Native version `0.62.2` with Hermes JS Engine https://hermesengine.dev `0.4.1` version.
When I wrapped my application into `<Provider>` component from jotai I see white screen / nothing.

There is not error, error logs and indicator that something doesn't work.
But one time error message appeared, as far as I remember it stated `No proxy support`.

For more modern Hermes JS there is Proxy support. But for stated versions jotai doesn't work out-of-the-box.

While doing the same with `RecoilJS` `0.1.2` at least I see application screens.

dai-shi:
@likern Thanks for your trial. That's totally unexpected for me. Let me create a PR to remove Proxy usage. I will ask you for testing it.

dai-shi:
Let's close this. If we have a real issue, please open a new one.

Albert-Gao:
Actually, Proxy support for Hermes just landed with 0.64, But I am perfectly fine. Jotai is my go-to statement for React/React Native now.

Issue #48, opened 2020-09-15
Browser support (IE11 supported?)


bricejar:
On which browsers will jotai work ?

Is IE11 supported ?

dai-shi:
Good question. Jotai intensively uses `Map`, `WeakMap` and `Symbol`. IE11 seems to support those, so it's expected to work. Can you confirm?

To use the CJS build:
```js
import { Provider, atom, useAtom } from 'jotai/index.cjs.js'
```

edgesoft:
@bricejar https://github.com/pmndrs/jotai/blob/master/src/Provider.ts#L35

Here Proxy is used and that is not supported by IE11. However there are polyfills but they are mostly not very stable.

dai-shi:
@edgesoft Oh, you are right. If it's the only blocking issue, I can probably change it to object getters. Polyfills should work fine in this case of simple proxy getters.

edgesoft:
@dai-shi Thanks! I don't think it will be an issue since is not used extensively.

dai-shi:
Let me close this. Please reopen if you need further discussion.

The current understanding of IE11 support is: It should work with proxy-polyfill.

dai-shi:
Now, #162 removed proxies, it no longer requires proxy-polyfill.

dai-shi:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
According to this, Map(iterable), Map.keys() and Map.entries() are not supported in IE11. We use those now.

Does anyone have access to IE11?

sandren:
Would supporting IE11 compromise the Jotai bundle size for modern browsers? If so, it's not worth it at this point. [Even Microsoft doesn't support IE11 anymore](https://www.theverge.com/2020/8/17/21372487/microsoft-internet-explorer-11-support-end-365-legacy-edge).

Thisen:
I don't think we should stress about supporting IE11, we're also using symbols which isn't supported in IE11.

dai-shi:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
I also noticed Set(iterator) and similar, which are used a lot.

> compromise the Jotai bundle size for modern browsers

So, seems like so.

> also using symbols

Yeah, but this one is easy. `Symbol = () => ({})`

---
So, concluding this:
- We don't officially support IE11.
- Map/Set/Symbol are blocking
  - (If one could polyfill them well, it might work.)

PR #162, opened 2020-11-02
fix: avoid using proxy

dai-shi:
https://github.com/pmndrs/jotai/issues/159#issuecomment-720697657

It seems like some RN versions doesn't support Proxies.

https://github.com/pmndrs/jotai/issues/48#issuecomment-701025932

It's also reported before for IE11 compatibility.

This PR removes the usage of Proxies. It does reduce bundle size too.

@likern Would you please if this solves your use case? (Check codesandbox bot message how to install the bundle.)

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,27 +1,27 @@
 {
   "index.js": {
-    "bundled": 19439,
-    "minified": 8732,
-    "gzipped": 3243,
+    "bundled": 19414,
+    "minified": 8714,
+    "gzipped": 3232,
     "treeshaked": {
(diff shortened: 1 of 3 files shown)


likern:
@dai-shi I've tested and can confirm that this PR works on my codebase.
At least I see application running, if it's wrapped into `<Provider />`. I haven't yet tested more features.

dai-shi:
@likern Thanks for testing! Good to know.
I'll merge this and publish a patched version as it's basically good not only with RN.
Let me know if you find any issues when you have a chance to try further.

PR #165, opened 2020-11-05
add package.json exports

dai-shi:
I'm not super confident, but does this help importing `jotai/utils` in nodejs environment?


dai-shi:
@wilcoschoneveld 
Would you mind trying this?
My expectation is this allows `import { ... } from 'jotai/utils'` instead of `jotai/utils.cjs` in nodejs env.

dai-shi:
I tried this bundle with nextjs on codesandbox, and it didn't work. Feels like I should give this up for the time being.

dai-shi:
This doesn't help at all except for latest node, but my hope is TS eventually supports it.

dai-shi:
Probably, this doesn't do anything, but hope it won't hurt anyway. Let's merge this in and see how it goes.

PR #169, opened 2020-11-07
Add ReduxDevTools Integration to Jotai

c0d3t3k:
Enable [Redux Devtools](https://github.com/reduxjs/redux-devtools) browser extension integration for Jotai.


dai-shi (approved):
Should be good. Let's work on docs in a separate PR.

Issue #167, opened 2020-11-06
Discussion: atom values should be freezed/immutable


likern:
Based on discussion in https://github.com/pmndrs/jotai/issues/41

dai-shi:
oops.

dai-shi:
Do you agree the issue exists even with `React.useState`?

likern:
> Do you agree the issue exists even with `React.useState`?

I agree that if some object is shared between components, it's change changes all components. But because callbacks with updated value are used - UI is still valid.

Let's say I have that array of tags. If tag value is changed (isSelected = true) it doesn't cause a rerender.
But I get `onPress(tagId: string, isSelected: boolean)` callback fired. And it will rerender valid components.

likern:
If atom is changed inherently, there is no notification of that, we don't know that it's happened.

And if `useAtom(atom)` will be called in some other components we accidentally get wrong value. Without even knowing about it.

likern:
If I'm correct Recoil has immutable atom values.
https://github.com/facebookexperimental/Recoil/issues/112#issuecomment-631068581

dai-shi:
> But because callbacks with updated value are used - UI is still valid.

I'm not sure if I understand this.

> If atom is changed inherently, there is no notification of that, we don't know that it's happened.

> And if useAtom(atom) will be called in some other components we accidentally get wrong value. Without even knowing about it.

But, I get these two.

> If I'm correct Recoil has immutable atom values.

Hm, wonder if they deep freeze it.

dai-shi:
I would like to revisit this: https://github.com/pmndrs/jotai/issues/41#issuecomment-722705742

```js
const​ ​updatedTags​ ​=​ ​produce​(​Array​.​from​(​taskTags​)​,​ ​(​draft​)​ ​=>​ ​{​
  ​const​ ​tag​ ​=​ ​draft​.​find​(​(​it​)​ ​=>​ ​it​.​id​ ​===​ ​action​.​id​)​;​
  ​if​ ​(​tag​ !== ​undefined​)​ ​{​
    ​tag​.​isSelected​ ​=​ ​action​.​isSelected​;​
  ​}​
}​)​;​
​console​.​log​(​
  ​`taskTagsAtom: updatedTags are ​${​JSON​.​stringify​(​updatedTags​,​ ​null​,​ ​2​)​}​`​
​)​;​
```

Why on earth immer goes wrong?

likern:
[...]

PR #171, opened 2020-11-08
feat: freeze prop in Provider

dai-shi:
Implementation of this: https://github.com/pmndrs/jotai/issues/167#issuecomment-723515366

### DIFF
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -697,7 +697,9 @@ export const Provider: React.FC<{
       // XXX this is not very efficient
       mKeys(state).forEach((a) => {
         const aState = mGet(state, a) as AtomState<unknown>
-        freeze(aState.value)
+        if (aState.value) {
+          freeze(aState.value)
+        }
       })
     }
   }
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -552,7 +552,8 @@ const InnerProvider: React.FC<{
 
 export const Provider: React.FC<{
   initialValues?: Iterable<readonly [AnyAtom, unknown]>
-}> = ({ initialValues, children }) => {
+  freeze?: (obj: unknown) => void
+}> = ({ initialValues, freeze, children }) => {
   const contextUpdateRef = useRef<ContextUpdate>()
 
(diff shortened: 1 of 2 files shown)


src/core/Provider.ts
```diff
@@ -697,7 +697,9 @@ export const Provider: React.FC<{
       // XXX this is not very efficient
       mKeys(state).forEach((a) => {
         const aState = mGet(state, a) as AtomState<unknown>
-        freeze(aState.value)
```
likern: If CodeSandbox CI can be package for latest PR commit that would be awesome.

dai-shi: I think it did?

likern: Sorry, my mistake.  CodeSandbox edited original comment. Smarter than I fought.

likern:
I expected `freeze` will prevent my original bug saying I'm trying to modify freezed object.
```ts
const taskTags = get(taskTagsAtom);
const updatedTags = produce(Array.from(taskTags), (draft) => {
  const tagIndex = draft.findIndex((it) => it.id === action.id);
  if (tagIndex >= 0) {
    // Found tag with this id
    draft[tagIndex].isSelected = action.isSelected;
  }
});
```
But it doesn't. I don't see any errors. Still can silently update original `taskTags`

src/core/Provider.ts
dai-shi: You can also click the green ✔️ mark left to the commit.

dai-shi:
Hm, that was my expectation too. 🤔

likern:
@dai-shi I created a simple reproduction for React Native. Though I think it's not specific to it.

https://github.com/breeffy/jotai-react-native
https://expo.io/@malvikus/projects/jotai-react-native

It can be run locally using 
```bash
yarn start && yarn android
```
or 
```bash
yarn start && yarn ios
```

if `expo-cli` in installed https://docs.expo.io/.

dai-shi:
@likern Thanks for the repro.
I tried the same in codesandbox: https://codesandbox.io/s/react-forked-hk6ei
And it doesn't work as expected.

I noticed freezing in Provider is too late in certain scenarios (derived atoms).
Let's drop this idea.

Issue #108, opened 2020-10-06
Benchmark immutableMap (create a benchmark tool)


dai-shi:
After #107, or in parallel with it, we would want to create a benchmark tool for immutableMap.
It would allow us to evaluate the current implementation, compare possible alternative implementations, and improve the entire performance of jotai.

Anybody keen about performance?

dai-shi:
btw, benchmarking of jotai itself would also be interesting and important.

Aslemammad:
@dai-shi  How we write it? and what to learn? suggestions? ...

dai-shi:
First of all, this is not a trivial issue. It requires some general knowledge for benchmarking and understanding how it would be used in jotai (we need to design benchmark scenario.) I don't have a full answer.

That said, if I were to do this, I'd use [benchmark](https://www.npmjs.com/package/benchmark) and start with a simple scenario.

You might want to start with #107 before working on benchmarking.

Aslemammad:
Got it, I'm just learning and I love challenges. Thanks.

dai-shi:
New implementation might be based on different abstractions. Closing this for now.

Issue #107, opened 2020-10-06
Write tests for immutableMap


dai-shi:
immutableMap is an internal module. It's an abstract module.
I wonder if anyone would be interested in writing tests for it.
Note that we want to test the api, not the implementation (the type `ImmutableMap` is representing implementation), because the goal is to come up with better implementation with same functionality.

Aslemammad:
@dai-shi What we should do? Could you explain more? or what we need to learn?

dai-shi:
> What we should do?

It is going to be something like this:

- Read ./src/core/immutableMap.ts
- Ask questions if something is not clear
- Create ./tests/core/immutableMap.test.ts
- Send a PR and discuss

> Could you explain more?

Jotai uses a single provider component to store all atom states (values) in its component tree.
It is important that the provider works with good performance, because it can be a bottleneck.
One of the challenges is immutable map. It's something like `Map` but has to be updated immutably for React Concurrent Mode.
So, I abstract out that part in the file and created a naive (but should work fairly well) implementation.
We'd like to know if we can handle thousands or millions of atoms.

> or what we need to learn?

If you are not familiar with the standard `Map`. Check some docs:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

Learning about testing & benchmarking in general should also help.

Aslemammad:
@dai-shi I appreciate this, thanks.

dominictwlee:
@Aslemammad Are you still working on this?

Aslemammad:
@dominictwlee Hey, yes I'm working on it.

dominictwlee:
Cool, I'll have a look at it.

Aslemammad:
if u need any help, tell me, I'd love to work on it.

dominictwlee:
@dai-shi So given that the api for immutableMap methods are dependent on the implementation detail of receiving an immutable map type, I'm not entirely sure we can avoid writing tests that avoid its implementation details.

Could you clarify what you had in mind for this? Thanks!

dai-shi:
@dominictwlee Glad you start investigating it.
So, the background of this work is to improve performance.
If we have many atoms, our immutableMap becomes big which can cause bad performance of the entire library.

Originally, the code is embedded in Provider.ts, but as we move it to immutableMap.ts,
we can compare different implementations.
[...]

PR #174, opened 2020-11-08
refactor: implement mForEach

### LINKED ISSUE #172: [Refactor] implement mForEach instead of mKeys in immutableMap.ts
It should be more efficient and reduce bundle size.

dominictwlee:
closes #172 

@dai-shi Is this what you had in mind for this refactor?
Basically removes the need for the extra loop for the unique keys.

### DIFF
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -132,8 +132,7 @@ const replaceDependencies = (
 ) => {
   const dependencies = new Set(dependenciesToReplace)
   let nextState = prevState
-  mForEach(nextState, (a) => {
-    const aState = mGet(nextState, a) as AtomState<unknown>
+  mForEach(nextState, (aState, a) => {
     if (aState.deps.has(atom)) {
       if (dependencies.has(a)) {
         // not changed
@@ -609,8 +608,7 @@ export const Provider: React.FC<{
     let deleted: boolean
     do {
       deleted = false
-      mForEach(nextState, (a) => {
-        const aState = mGet(nextState, a) as AtomState<unknown>
+      mForEach(nextState, (aState, a) => {
         // do not delete while promises are not resolved
         if (aState.writeP || aState.readP) return
         const depsSize = aState.deps.size
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -32,9 +32,9 @@ import {
   mGet,
   mSet,
   mDel,
-  mKeys,
   mMerge,
   mToPrintable,
+  mForEach,
 } from './immutableMap'
 
 // guessing if it's react experimental channel
@@ -132,7 +132,7 @@ const replaceDependencies = (
 ) => {
   const dependencies = new Set(dependenciesToReplace)
   let nextState = prevState
-  mKeys(nextState).forEach((a) => {
+  mForEach(nextState, (a) => {
     const aState = mGet(nextState, a) as AtomState<unknown>
     if (aState.deps.has(atom)) {
       if (dependencies.has(a)) {
@@ -609,7 +609,7 @@ export const Provider: React.FC<{
     let deleted: boolean
     do {
       deleted = false
-      mKeys(nextState).forEach((a) => {
+      mForEach(nextState, (a) => {
         const aState = mGet(nextState, a) as AtomState<unknown>
         // do not delete while promises are not resolved
         if (aState.writeP || aState.readP) return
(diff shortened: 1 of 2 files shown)


dai-shi:
@dominictwlee Yes, exactly!!! So glad you could work on it with such a short description.
I'll add some comments for changes and discussions.

Now, you seem to be somewhat familiar with immutableMap.ts, you might be interested in #107 and #108. Feel free to start some discussions in those issues if you are interested.

src/core/immutableMap.ts
```diff
@@ -9,6 +9,7 @@ type MGet = <K, V>(m: ImmutableMap<K, V>, k: K) => V | undefined
 type MSet = <K, V>(m: ImmutableMap<K, V>, k: K, v: V) => ImmutableMap<K, V>
 type MDel = <K, V>(m: ImmutableMap<K, V>, k: K) => ImmutableMap<K, V>
 type MKeys = <K, V>(m: ImmutableMap<K, V>) => Set<K>
```
dai-shi, line 11: Let's remove mKeys as it's not used now, and basically it's not efficient.

src/core/immutableMap.ts
```diff
@@ -71,6 +72,21 @@ export const mKeys: MKeys = <K, V>(m: ImmutableMap<K, V>) => {
   return keys
 }
 
+export const mForEach: MForEach = <K, V>(
+  map: ImmutableMap<K, V>,
+  cb: (a: K) => void
+) => {
+  const cache = new Map()
```
dai-shi, line 79: I think we can use `Set` instead of `Map`.
(It can be even be `WeakSet`, which can be more performant, but not sure.)

I'd name it `seen` or something. `const seen = Set<K>()`

src/core/Provider.ts
```diff
@@ -132,7 +132,7 @@ const replaceDependencies = (
 ) => {
   const dependencies = new Set(dependenciesToReplace)
   let nextState = prevState
-  mKeys(nextState).forEach((a) => {
+  mForEach(nextState, (a) => {
     const aState = mGet(nextState, a) as AtomState<unknown>
```
dai-shi, line 136: Oh, I missed this. I wanted to avoid `mGet` too.

```suggestion
  mForEach(nextState, (aState, a) => {
```

We would like to this. callback is `(value, key) => ...`. cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach

dai-shi (changes requested):
It's the right direction. We need some more changes.

src/core/immutableMap.ts
```diff
@@ -8,7 +8,10 @@ type MCreate = <K, V>() => ImmutableMap<K, V>
 type MGet = <K, V>(m: ImmutableMap<K, V>, k: K) => V | undefined
 type MSet = <K, V>(m: ImmutableMap<K, V>, k: K, v: V) => ImmutableMap<K, V>
 type MDel = <K, V>(m: ImmutableMap<K, V>, k: K) => ImmutableMap<K, V>
-type MKeys = <K, V>(m: ImmutableMap<K, V>) => Set<K>
+type MForEach = <K extends object, V>(
+  m: ImmutableMap<K, V>,
+  cb: (value: V, key: K, map: Map<K, V>) => void
```
dai-shi, line 13: My bad, I didn't mean to add the third argument, because it's not something we expect.
```suggestion
  cb: (value: V, key: K) => void
```
Let's do this.

src/core/immutableMap.ts
```diff
@@ -61,14 +64,20 @@ export const mDel: MDel = <K, V>(m: ImmutableMap<K, V>, k: K) => {
...
-    for (const key of map.keys()) {
-      keys.add(key)
-    }
+export const mForEach: MForEach = <K extends object, V>(
+  map: ImmutableMap<K, V>,
+  cb: (value: V, key: K, map: Map<K, V>) => void
+) => {
+  const seen = new WeakSet()
+  map.forEach((m) => {
+    m.forEach((...args) => {
```
dai-shi, line 73:
```suggestion
    m.forEach((v, k) => {
```
Let's do this.

dai-shi (commented):
Looks good. Let's make a small fix, which I misdirected.

src/core/immutableMap.ts
```diff
@@ -61,14 +64,20 @@ export const mDel: MDel = <K, V>(m: ImmutableMap<K, V>, k: K) => {
...
-  const keys = new Set<K>()
-  m.forEach((map) => {
-    for (const key of map.keys()) {
-      keys.add(key)
-    }
+export const mForEach: MForEach = <K extends object, V>(
+  map: ImmutableMap<K, V>,
+  cb: (value: V, key: K, map: Map<K, V>) => void
+) => {
+  const seen = new WeakSet()
```
dai-shi, line 71:
```suggestion
  const seen = new WeakSet<K>()
```
Just to make it so with a little confidence.

dai-shi (approved):
👍 
Thanks for your contribution!

PR #176, opened 2020-11-09
rewrite atomFamily using Map

likern:
I expect this patch should improve performance when using `atomFamily`

### DIFF
--- a/src/utils/atomFamily.ts
+++ b/src/utils/atomFamily.ts
@@ -1,5 +1,4 @@
 import { atom, Atom, WritableAtom, PrimitiveAtom } from 'jotai'
-
 import type { Getter, Setter } from '../core/types'
 
 type ShouldRemove<Param> = (createdAt: number, param: Param) => boolean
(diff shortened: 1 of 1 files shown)


dai-shi:
This is nice! I don't remember why I didn't use `Map` (maybe confused with `WeakMap` which only works with objects.)

(BTW, are you in pmndrs discord server?)

likern:
> This is nice! I don't remember why I didn't use `Map` (maybe confused with `WeakMap` which only works with objects.)
> 
> (BTW, are you in pmndrs discord server?)

Yes, I'm there in **jotai** group as a likern user.

src/utils/atomFamily.ts
```diff
@@ -61,45 +60,52 @@ export function atomFamily<Param, Value, Update extends never = never>(
...
       initializeRead(param),
       initializeWrite && initializeWrite(param)
     ) as AtomType
-    atoms.unshift([param, newAtom, Date.now()])
+    atoms.set(param, [newAtom, Date.now()])
     return newAtom
   }
+
   createAtom.remove = (p: Param) => {
-    const index = atoms.findIndex((x) => x[0] === p)
```
dai-shi, line 88: Just noticed my original implementation doesn't take care of `areEqual`.
So, this is not a regression in this PR, but we should use areEqual here too to be consistent.

dai-shi (approved):
Nice work! @likern 

Also a big thank to @merisbahti who created the test #101 !

src/utils/atomFamily.ts
dai-shi: #177 follow up issue.

PR #179, opened 2020-11-09
update use-context-selector

### LINKED ISSUE #178: Weird error without Provider: TypeError Cannot read property 'Symbol()' of null
In the latest version (0.10.4), if we forgot to add `<Provider>`, the error message is not understandable.

This is maybe because #162 changed the way we show the message which revealed the bug in use-context-selector.

Repro: https://codesandbox.io/s/lucid-wright-0wc7o?file=/src/App.js

image

dai-shi:
close #178

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 19645,
-    "minified": 8810,
-    "gzipped": 3270,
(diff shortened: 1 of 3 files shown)


dai-shi:
https://codesandbox.io/s/frosty-lovelace-4h5bm
Fixed!

image

PR #181, opened 2020-11-09
fixes #177: use areEqual in atom remove

### LINKED ISSUE #177: atomFamily.remove should follow areEqual
atomFamily's createAtom.remove should follow areEqual if it is specified.
ref: https://github.com/pmndrs/jotai/pull/176#discussion_r519700888

Preferably with tests.

likern:
fixes #177

### DIFF
--- a/src/utils/atomFamily.ts
+++ b/src/utils/atomFamily.ts
@@ -67,6 +67,10 @@ export function atomFamily<Param, Value, Update>(
   let shouldRemove: ShouldRemove<Param> | null = null
   const atoms: Map<Param, [AtomType, CreatedAt]> = new Map()
   const createAtom = (param: Param) => {
+    if (param === undefined) {
+      throw new Error("param can't be undefined")
+    }
+
     let item: [AtomType, CreatedAt] | undefined
     if (areEqual === undefined) {
       item = atoms.get(param)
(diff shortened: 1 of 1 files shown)


dai-shi:
Hi, thanks for the work.

Do we really need to reject `undefined` param?

```js
    if (areEqual === undefined) {
      atoms.delete(param)
    } else {
      for (let [key, _] of atoms) {
        if (areEqual(key, param)) {
          atoms.delete(key)
          break
        }
      }
    }
```

Does this work?

A minor note:  Does `[key]` work instead of `[key, _]`?

likern:
> Hi, thanks for the work.
> 
> Do we really need to reject `undefined` param?
> 
> ```js
>     if (areEqual === undefined) {
>       atoms.delete(param)
>     } else {
>       for (let [key, _] of atoms) {
>         if (areEqual(key, param)) {
>           atoms.delete(key)
>           break
>         }
>       }
>     }
> ```
> 
> Does this work?
> 
> A minor note: Does `[key]` work instead of `[key, _]`?

Yes, we can allow `undefined`. Initially I've written code as you, but then decided to rewrite. I can update PR.
`[key]` will also work. Can update too.

dai-shi:
If there's no drawbacks, please update. There wouldn't be any specific use case of `undefined`, but I don't prefer having the if-undefined-throw statement.

dai-shi (approved):
Looks good. Thanks for your quick work.

PR #183, opened 2020-11-10
chore: useAtomDevtools typing

dai-shi:
How about enabling devtools only in development mode? It would be a little bit better in production mode.

### DIFF
--- a/src/devtools/useAtomDevtools.ts
+++ b/src/devtools/useAtomDevtools.ts
@@ -29,21 +29,21 @@ type Extension = {
   connect: (options?: Config) => ConnectionResult
 }
 
-type UseAtomDevtools = <Value>(
+export function useAtomDevtools<Value>(
   anAtom: WritableAtom<Value, Value>,
   name?: string
-) => void
-
-const useAtomDevtoolsImpl: UseAtomDevtools = <Value>(
-  anAtom: WritableAtom<Value, Value>,
-  name?: string
-) => {
+) {
   let extension: Extension | undefined
   try {
     extension = (window as any).__REDUX_DEVTOOLS_EXTENSION__ as Extension
   } catch {}
   if (!extension) {
-    console.warn('Please install/enable Redux devtools extension')
+    if (
+      process.env.NODE_ENV === 'development' &&
+      typeof window !== 'undefined'
+    ) {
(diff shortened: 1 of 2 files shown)


src/devtools/useAtomDevtools.ts
```diff
@@ -11,45 +11,45 @@ interface Config {
...
 
-interface Message {
+type Message = {
   type: string
   payload?: any
   state?: any
 }
 
-interface IConnectionResult {
+type ConnectionResult = {
```
dai-shi, line 20: Sorry if this conflicts with your preference.
I'd use `interface` only when required: interface merging, module augmentation.

src/devtools/useAtomDevtools.ts
```diff
@@ -95,3 +95,8 @@ export function useAtomDevtools<Value>(
     }
   }, [anAtom, extension, atomName, value])
 }
+
+export const useAtomDevtools: UseAtomDevtools =
+  process.env.NODE_ENV === 'development' && typeof window !== 'undefined'
```
dai-shi, line 100: Any comments on the condition? I'm not sure what's good.

src/devtools/useAtomDevtools.ts
c0d3t3k: This looks good, switching to `type` is probably better for the reasons you mentioned.

src/devtools/useAtomDevtools.ts
c0d3t3k: Again, no problems with using `type` here for the function. Having the `type` and then the `useAtomDevtoolsImpl` implementation isn't a pattern I normally use in my own projects, but I can see where this would be helpful for a library like Jotai.

c0d3t3k: With regards to disabling in production, there should maybe be an additional parameter on the hook (i.e. `useAtomDevtools(..., enableInProduction = false)`) to override this behavior if for some reason one wishes to enable the devtools in a production build.

dai-shi: OK, that was my concern. Unfortunately, the additional parameter doesn't work because of the rules of hooks. I'll revert the change.

dai-shi:
Merging this just for changes to prefer types to interfaces.

PR #186, opened 2020-11-10
test: immutable map

### LINKED ISSUE #107: Write tests for immutableMap

dominictwlee:
closes #107 

Tests the api of immutable map.

### DIFF
--- a/tests/immutableMap.test.ts
+++ b/tests/immutableMap.test.ts
@@ -13,14 +13,16 @@ import {
 describe('immutable writes', () => {
   it('returns copy on set item', () => {
     const m = mCreate()
-    const m2 = mSet(m, 1, 'a')
+    const key = new Map()
+    const m2 = mSet(m, key, 'a')
     expect(m).not.toEqual(m2)
   })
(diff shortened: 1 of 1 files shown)


tests/immutableMap.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  it('returns copy on merge', () => {
+    const m = mCreate()
+    const m2 = mCreate()
+    const m3 = mMerge(m, m2)
+    expect(m).not.toEqual(m3)
+    expect(m2).not.toEqual(m3)
+  })
+
+  it('does not mutate on set', () => {
```
dai-shi, line 35: on get?

dai-shi (approved):

dai-shi:
Thanks for your work!

PR #188, opened 2020-11-11
refactor with atom revision number and dependents weak map

### LINKED ISSUE #184: Updating an atom dependency without having the dependent in the same component doesn't update the dependent
## Context

I have an `attackerSkillSetAtom` atom:

```jsx
import { atom } from 'jotai';

export const attackerSkillSetAtom = atom({
  someExampleSkill: 3,
  anotherSkill: 2
});
```

I have a component with a `skillSetAtom` prop that accepts an atom:

```jsx
import { useAtom } from 'jotai';

export const UpdateSkill = ({ skillSetAtom, skill }) => {
  const [skillSet, setSkills] = useAtom(skillSetAtom);

  const handleChange = change => {
    setSkills({
      ...skillSet,
      [skill]: skillSet[skill] + change
    });
  };

  return (
    <>
      <button onClick={() => handleChange(-1)}>-</button>
      <input type='text' value={skillSet[skill]} disabled />
      <button onClick={() => handleChange(1)}>+</button>
    </>
  );
};
```

I have a page in my app that displays skills (for a game) in a list with these `UpdateSkill` components on it. The route for this page is `/skills/:party` because I'd like to show a different skill set for the attacking player and the defending player. My routing component feeds the `party` prop, and the component selects the appropriate atom based on that. To keep it simple, here is a simplified example of that page component `SkillsScreen` that feeds one single `UpdateSkill` component:

```jsx
import { attackerSkillSetAtom, defenderSkillSetAtom } from '../atoms';

const skillSetAtom = {
  attacker: attackerSkillSetAtom,
  defender: defenderSkillSetAtom
};

export const SkillsScreen = ({ party }) => {
  return (
    <>
      <h2>Some Example Skill</h2>
      <UpdateSkill skillSetAtom={skillSetAtom[party]} skill='someExampleSkill' />
    </>
  );
};
```

However, users of the app aren't mainly interested in setting skills. They are interested in the effects their skills have. So I have another atom that keeps track of skill effects: `attackerSkillEffectsAtom`:

```jsx
const getSkillEffects = skillSetAtom => get => {
  const skillSet = get(skillSetAtom);

  // for simplicity of the example, I'm just adding up whatever their skill values are
  return Object.keys(skillSet).reduce((sum, skill) => sum + skillSet[skill], 0);
};

export const attackerSkillEffectsAtom = atom(getSkillEffects(attackerSkillSetAtom));
```

The reason I provide a curried function to `atom()` is that I do the same for `defenderSkillEffectsAtom` and I don't want to duplicate code.

I have a "homepage" component that lists a player's skill effects:

```jsx
import { useAtom } from 'jotai';
import { attackerSkillEffectsAtom } from '../atoms';

export const HomeScreen = () => {
  const [skillEffects] = useAtom(attackerSkillEffectsAtom);

  return (
    <pre>
      Attacker skill effects are: {JSON.stringify(skillEffects, null, 2)}
    </pre>
  );
};
```

## Problem

When navigating to `/skills/attacker` and changing the skill value, then navigating back to `/`, the derived atom doesn't appear to have updated.

However, if I use my derived atom in `SkillsScreen`, the same place where the dependency atom is imported, it will update immediately as expected:

```diff
- import { attackerSkillSetAtom, defenderSkillSetAtom } from '../atoms';
+ import { attackerSkillSetAtom, defenderSkillSetAtom, attackerSkillEffectsAtom, defenderSkillEffectsAtom } from '../atoms';

  const skillSetAtom = {
    attacker: attackerSkillSetAtom,
    defender: defenderSkillSetAtom
  };

+ const skillEffectsAtom = {
+   attacker: attackerSkillEffectsAtom,
+   defender: defenderSkillEffectsAtom
+ };
+
  export const SkillsScreen = ({ party }) => {
+ const [skillEffects] = useAtom(skillEffectsAtom[atom]);
+
    return (
      <>
        <h2>Some Example Skill</h2>
        <UpdateSkill skillSetAtom={skillSetAtom[party]} skill='someExampleSkill' />
+       <pre>{skillEffects}</pre>
      </>
    );
  };
```

Now when navigating back to `/`, the value remains updated.

**Is this designed behaviour of jotai?**

dai-shi:
close #184 

This is rather a big refactoring:
- atom revision
- atom dependencies (not dependents)
- dependents map (weak map)

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,27 +1,27 @@
 {
   "index.js": {
-    "bundled": 19623,
-    "minified": 8787,
-    "gzipped": 3280,
+    "bundled": 20883,
+    "minified": 9189,
+    "gzipped": 3330,
     "treeshaked": {
       "rollup": {
         "code": 274,
         "import_statements": 92
       },
       "webpack": {
-        "code": 1407
+        "code": 1377
       }
     }
   },
   "index.cjs.js": {
-    "bundled": 23566,
-    "minified": 10980,
-    "gzipped": 3842
+    "bundled": 24827,
+    "minified": 11396,
+    "gzipped": 3924
   },
   "index.iife.js": {
-    "bundled": 25033,
-    "minified": 9007,
-    "gzipped": 3504
+    "bundled": 26332,
+    "minified": 9414,
+    "gzipped": 3582
   },
   "utils.js": {
     "bundled": 3413,
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -259,7 +259,7 @@ const readAtomState = <Value>(
         return aState.value
       }
       // a === atom
-      const aState = mGet(nextState.s, a)
+      const aState = mGet(nextState.s, a) || atomStateCache.get(a)
       if (aState) {
         if (aState.readP) {
           throw aState.readP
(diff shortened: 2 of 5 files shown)


dai-shi:
@mdingena 's reproduction fixed: https://codesandbox.io/s/gallant-dawn-mbi0l

dai-shi:
There should still be room for improvement, but for now let's merge this in.

PR #189, opened 2020-11-11
use object instead of map for debug state

dai-shi:
https://github.com/pmndrs/jotai/issues/184#issuecomment-725037679
Previously, the debug state is Map and not very easy to see in react devtools.
image

This fixes it by using objects.
image


dai-shi:
This should be good to go.

Issue #175, opened 2020-11-09
RFC: jotai/optics


dai-shi:
@merisbahti originally filed the issue #44.
As I wasn't familiar with optics then, a separate package `jotai-optics` made sense.
I'm not yet very familiar with it, but I start getting to know its power and suitability with jotai api.
So, it would be a good option to provide `jotai/optics` officially.

Just yesterday, @c0d3t3k announced [an article](https://dev.to/c0d3t3k/jotai-now-with-optics-29ld) with a good example using jotai and optics, thought it's a good chance to start a discussion.

merisbahti:
Really interesting and fun to hear, I can make a PR for this. 👍

PR #193, opened 2020-11-14
fix(optics): Add optics bundle.

merisbahti:
This fixes: https://github.com/pmndrs/jotai/issues/175

I did not include `useAtomArrayFamily` from https://github.com/merisbahti/jotai-optics/blob/master/src/index.ts#L61-L115 because it would make this too large of a PR, and I want to get more tests in before including it. But it is definately in the pipe.


dai-shi:
Thanks! I will take a look later.
@c0d3t3k I will assign you as a reviewer too.

c0d3t3k:
This looks good to me. Great work @merisbahti!

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
+import { atom } from 'jotai'
+import * as O from 'optics-ts'
+import { PrimitiveAtom, SetStateAction, WritableAtom } from '../core/types'
+
+export type FocusAtom = {
+  <S, A>(
+    atom: PrimitiveAtom<S>,
+    callback: (optic: O.OpticFor<S>) => O.Prism<S, any, A>
```
dai-shi, line 8:
```suggestion
    callback: (optic: O.OpticFor<S>) => O.Prism<S, unknown, A>
```
Can all `any`s be `unknown`s?

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
...
+    (_, set, update) => {
+      const newValueProducer =
+        update instanceof Function
+          ? O.modify(focus)(update)
+          : O.set(focus)(update)
+
+      set(baseAtom, (oldBaseValue) => {
+        const newBaseValue = newValueProducer(oldBaseValue)
+        return newBaseValue
+      })
```
dai-shi, line 50:
```suggestion
      set(baseAtom, newValueProducer)
```
Can this be simplified like this assuming it's not using `this`?

dai-shi (commented):
1st iteration done.

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
+import { atom } from 'jotai'
+import * as O from 'optics-ts'
+import { PrimitiveAtom, SetStateAction, WritableAtom } from '../core/types'
+
+export type FocusAtom = {
```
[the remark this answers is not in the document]
merisbahti: It's much easier to share the generic overload if we do it like this, I don't know how to share the overload between two functions otherwise

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
...
+  ) =>
+    | O.Lens<S, any, A>
+    | O.Equivalence<S, any, A>
+    | O.Iso<S, any, A>
+    | O.Prism<S, any, A>
+    | O.Traversal<S, any, A>
+): any => {
+  const focus = callback(O.optic<S>())
+  return atom<A, SetStateAction<A>>(
+    (atomGetter) => {
```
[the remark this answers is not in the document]
merisbahti: fixed

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
...
+  baseAtom: PrimitiveAtom<S>,
+  callback: (
+    optic: O.OpticFor<S>
+  ) =>
+    | O.Lens<S, any, A>
+    | O.Equivalence<S, any, A>
+    | O.Iso<S, any, A>
+    | O.Prism<S, any, A>
+    | O.Traversal<S, any, A>
+): any => {
```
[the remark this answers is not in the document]
merisbahti: fixed

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
+import { atom } from 'jotai'
+import * as O from 'optics-ts'
+import { PrimitiveAtom, SetStateAction, WritableAtom } from '../core/types'
```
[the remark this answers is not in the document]
merisbahti: fixed

src/optics/focusAtom.ts
merisbahti: nope, unfortunately not :(

src/optics.ts
```diff
@@ -0,0 +1,2 @@
+export * from './optics/useFocus'
+export * from './optics/focusAtom'
```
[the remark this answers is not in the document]
merisbahti: fixed

docs/optics.md
```diff
@@ -0,0 +1,103 @@
...
+
+const objectAtom = atom({a: 5, b: 10})
+
+const Counter = () => {
+  const [count] = useAtom(countAtom)
+  return <div>count: {count}</div>
+}
+
+const Controls = () => {
+  const aAtom = useFocus(objectAtom, optic => optic.prop('a'))
```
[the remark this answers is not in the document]
merisbahti: Uh, no that's working as expected... Think when "focusing" an atom in 3 component levels, then it makes sense. But that might not be a common use-case...

merisbahti:
ugh, not sure why codesandbox keeps failing?

src/optics/focusAtom.ts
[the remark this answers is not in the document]
dai-shi: OK, it's fine for now (nobody would require module augmentation)

src/optics/focusAtom.ts
dai-shi: How about changing as far as we can? Let's see.

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
+import { atom } from 'jotai'
+import * as O from 'optics-ts'
+import type { PrimitiveAtom, SetStateAction, WritableAtom } from '../core/types'
+
+export type FocusAtom = {
+  <S, A>(
+    atom: PrimitiveAtom<S>,
+    callback: (optic: O.OpticFor<S>) => O.Prism<S, any, A>
+  ): WritableAtom<A | undefined, SetStateAction<A>>
```
dai-shi, line 9: When does it become `undefined` ? It's preferable if this is avoidable as A can include undefined.

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
+import { atom } from 'jotai'
+import * as O from 'optics-ts'
+import type { PrimitiveAtom, SetStateAction, WritableAtom } from '../core/types'
+
+export type FocusAtom = {
```
dai-shi, line 5:
```suggestion
type FocusAtom = {
```

I would like to not export this as this can be changed something else in the future.
Please use `typeof focusAtom` instead.

Oh, it's just internal. So, it might be OK, if you prefer this over `typeof focusAtom`.

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
...
+    | O.Equivalence<S, any, A>
+    | O.Iso<S, any, A>
+    | O.Prism<S, any, A>
+    | O.Traversal<S, any, A>,
+  bigValue: S
+) => {
+  if (focus._tag === 'Traversal') {
+    const values = O.collect(focus)(bigValue)
+    return values
+  } else if (focus._tag === 'Prism') {
```
dai-shi, line 67:
```suggestion
  }
  if (focus._tag === 'Prism') {
```
I'm not sure which is more efficient, but just to be consistent with coding pattern in others.

src/optics/focusAtom.ts
```diff
@@ -0,0 +1,74 @@
...
+    const value = O.preview(focus)(bigValue)
+    return value
+  } else {
+    const value = O.get(focus)(bigValue)
+    return value
+  }
```
dai-shi, line 73:
```suggestion
  }
  const value = O.get(focus)(bigValue)
  return value
```

src/optics/useFocus.ts
```diff
@@ -0,0 +1,26 @@
...
+):
+  | WritableAtom<A | undefined, SetStateAction<A>>
+  | WritableAtom<Array<A>, SetStateAction<A>>
+  | PrimitiveAtom<A> => {
+  return React.useMemo(() => {
+    return focusAtom(
+      baseAtom,
+      callback as any /* we're just copying the type signature from focusAtom, so this should be fine. */
+    )
+  }, [baseAtom, callback])
```
dai-shi, line 25: If this is the implementation, I don't think we need to provide `useFocus`, and let users do it, unless there's a typing issue. (But, it there were a typing issue, I'd think that's something solvable in `focusAtom` types.)

dai-shi (changes requested):
second iteration. some of them may require discussion..

rollup.config.js
```diff
@@ -77,13 +77,10 @@ function createIIFEConfig(input, output, globalName) {
 
 export default [
   createDeclarationConfig('src/index.ts', 'dist'),
-  createESMConfig('src/index.ts', 'dist/index.js'),
-  createCommonJSConfig('src/index.ts', 'dist/index.cjs.js'),
+  createCommonJSConfig('src/index.ts', 'dist/index.js'),
```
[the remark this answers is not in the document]
merisbahti: Haha will not, I was experimenting to see why code sandbox is failing 👍

src/optics/useFocus.ts
merisbahti: Sounds reasonable!

src/optics/focusAtom.ts
merisbahti: So it will return a `WritableAtom<A | undefined, SetStateAction<A>>` whenever a `Prism` is returned.

A prism can for example be: When getting an index out of an array `['hello', 'there'][5]` you never know if the type is `string` or `undefined`.

Another example of using prism would be when using `guard`, https://github.com/akheron/optics-ts#guardu-extends-ag-a-a--a-is-u-prisms-_-u

src/optics/focusAtom.ts
merisbahti: Yeah `typeof focusAtom` didn't work since it didn't apply the generic type... hmm, but I might just remove `useFocus`

merisbahti: removed useFocus and used function overloading instead

src/optics/useFocus.ts
merisbahti: removed useFocus

src/optics/focusAtom.ts
dai-shi: Oh, I see. So, a prism is like an array, `A` only represents the array item type. Got it.

tests/optics/focus-lens.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+  await findByText('bigAtom: {"a":1}')
+})
+
+it('double-focus on an atom works', async () => {
+  const bigAtom = atom({ a: { b: 0 } })
+
+  const Counter: React.FC = () => {
+    const [bigAtomValue, setBigAtom] = useAtom(bigAtom)
+    const atomA = focusAtom(bigAtom, (optic) => optic.prop('a'))
+    const atomB = focusAtom(atomA, (optic) => optic.prop('b'))
```
dai-shi, line 83: Should we use useMemo here? for tests could be a good reference.

dai-shi: Or, can we move these out of Counter?

dai-shi (commented):
Code looks good. A comment in tests.

docs/optics.md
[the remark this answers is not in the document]
merisbahti: removed `useFocus`, instead do `focusAtom`

dai-shi (commented):
you want to update docs too.

tests/optics/focus-lens.test.tsx
merisbahti: fixed

merisbahti:
fixed docs too

tests/optics/focus-lens.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+  rtl.fireEvent.click(getByText('set zero'))
+  await findByText('count: 0')
+  await findByText('bigAtom: {"a":0}')
+})
+
+it('focus on an atom works', async () => {
+  const bigAtom = atom({ a: 0 })
+
+  const Counter: React.FC = () => {
+    const aAtom = focusAtom(bigAtom, (optic) => optic.prop('a'))
```
dai-shi, line 51: This one too. There are three other places to change.

merisbahti: oops, I have fixed this.

dai-shi (approved):
Looks great!

dai-shi:
Thanks for your work! Let's merge this.

PR #194, opened 2020-11-15
fix(useSelector): do not call equalityFn before initialization

dai-shi:
I noticed this while I was creating an example on codesandbox.

Would be nice if we can create a test to detect this. cc @Aslemammad

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -24,9 +24,9 @@
     "gzipped": 3582
   },
   "utils.js": {
-    "bundled": 3413,
-    "minified": 1477,
-    "gzipped": 653,
+    "bundled": 3483,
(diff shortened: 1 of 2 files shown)


Aslemammad:
@dai-shi Sure, I'll work on it.

Aslemammad:
@dai-shi Do you have the codesandbox?

dai-shi:
https://codesandbox.io/s/jotai-forked-lbdzd?file=/src/App.js
This is the example (discussed in discord) that I noticed this bug.

I will merge this PR soon, and you can work on a branch based on `master`.

Aslemammad:
@dai-shi Thanks. Ok, I'm gonna work but will not send PR until you merge this.

PR #195, opened 2020-11-15
feat: support atom scope

### LINKED ISSUE #192: Jotai for libraries
As a library author, I want to make sure the values of the atoms my library relies upon are only updated in the `jotai.Provider`s I control and not the ones, the users of my library might add to the tree for their own purposes.

I believe we need some API in jotai to support it. The initial discussion happened [here](https://discord.com/channels/740090768164651008/752920140483526677/776731682232729621)

dai-shi:
close #192 

Usage:
```jsx
import { atom, Provider } from 'jotai'

const myScope = Symbol()

const countAtom = atom(0)
countAtom.scope = myScope

<Provider scope={myScope}>
  ...
</Provider>
```


src/core/Bridge.ts
```diff
@@ -1,9 +1,11 @@
 import React, { createElement, useMemo } from 'react'
 import { BridgeProvider, useBridgeValue } from 'use-context-selector'
 
-import { StateContext, ActionsContext } from './Provider'
+import { getContexts } from './Contexts'
+import { Scope } from './types'
 
-export const useBridge = () => {
+export const useBridge = (scope?: Scope) => {
```
dai-shi, line 7: Another option is an option object.
```js
const useBridge = (opts?: { scope?: Scope }) => {
```

src/core/types.ts
```diff
@@ -7,9 +7,12 @@ export type Setter = <Value, Update>(
   update: Update
 ) => void
 
+export type Scope = symbol | string | number
```
dai-shi, line 10: We could technically support `object`, but for now let's limit it to primitives as I don't see use cases of `object`.

src/core/Bridge.ts
dai-shi: https://github.com/pmndrs/jotai/pull/195#discussion_r523709569
As we don't allow `object` for scope for now, this should be fine, even if we changed our mind to support option object.

src/core/types.ts
Aslemammad: @dai-shi I like the symbol solution more than other solutions, I don't think an object is necessary, What's your idea? and for strings or numbers, as you know, they are not unique, maybe libraries use them, and jotai would find the same scope for two or more libraries, and mess them up.

dai-shi: I agree that most use cases we have in our mind is just `symbol`s. `string`s and `number`s might be useful for certain cases if they know what they are doing. Like two separate libs share atoms? Don't know. I just think we don't need to limit too much.

Aslemammad: That's ok, but maybe two libs choose a scope like this:

Aslemammad: That's ok, but maybe two libs choose a scope like this:
```jsx
// in lib 1
countAtom.scope = 1
// in lib 2 
anotherAtom.scope = 1
```
And yes, I agree with you that we shouldn't limit too much.

dai-shi: Yes, I know. That use case might be valid. Thanks for your notice anyway. It's nice to discuss this new feature in deep.

Aslemammad: Your welcome, I think we can tell the folks in the docs that it's better to use symbols, it's another good idea.

dai-shi: Oh, I forgot to add docs, at least something minimal.

Aslemammad:
@dai-shi Shall I create tests for it after merging?

dai-shi:
@Aslemammad That would be super nice!

PR #198, opened 2020-11-16
tests for scope option in providers and atoms

Aslemammad:
@dai-shi I wrote them in a separate file, Any suggestion?


dai-shi (commented):
Hey, thanks!
They are good for basic specs, but would they pass even without the scope support? We'd want something like countAtom and scopedCountAtom having different values, a value is kept if an inner provider unmounts, and throw an error if scoped provider doesn't exist.

I can merge this as is, if you want.

dai-shi:
@Aslemammad I added two more specs. Merging this.

Aslemammad:
@dai-shi Thanks, That's awesome.

PR #202, opened 2020-11-19
improve useAtom error message

Aslemammad:
@dai-shi I was playing with codesandbox, then jotai showed me that error, but I had a provider, then I realized that I didn't put the scope. Maybe helpful for the folks.

### DIFF
--- a/src/core/useAtom.ts
+++ b/src/core/useAtom.ts
@@ -2,13 +2,14 @@ import { useEffect, useCallback, useDebugValue } from 'react'
 import { useContext, useContextSelector } from 'use-context-selector'
 
 import { getContexts } from './contexts'
-import { Atom, WritableAtom, AnyWritableAtom } from './types'
+import { Atom, WritableAtom, AnyWritableAtom, Scope } from './types'
 
-function assertContextValue<T extends object>(x: T | null, scope?: unknown): asserts x is T {
+function assertContextValue<T extends object>(
+  x: T | null,
+  scope?: Scope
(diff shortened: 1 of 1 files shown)


src/core/useAtom.ts
```diff
@@ -6,7 +6,9 @@ import { Atom, WritableAtom, AnyWritableAtom } from './types'
 
 function assertContextValue<T extends object>(x: T | null): asserts x is T {
```
dai-shi, line 7:
```suggestion
function assertContextValue<T extends object>(x: T | null, scope?: unknown): asserts x is T {
```

Let's show the note about the scope only if scope is specified. I suppose, the majority of users don't use scopes.

dai-shi (changes requested):
Thanks for the suggestion! can you change a bit?

src/core/useAtom.ts
```diff
@@ -6,7 +6,9 @@ import { Atom, WritableAtom, AnyWritableAtom } from './types'
...
   if (!x) {
-    throw new Error('Please use <Provider>')
+    throw new Error(
+      'Please use <Provider> or if you have scoped atoms, use <Provder scope={yourAtomScope}>'
```
dominictwlee, line 10:
```suggestion
      'Please use <Provider> or if you have scoped atoms, use <Provider scope={yourAtomScope}>'
```

Aslemammad:
I think it's ready now.

dai-shi (approved):
LGTM

PR #203, opened 2020-11-21
fix: async dependency behavior

### LINKED ISSUE #201: Error: Atom state not found. Weirdness with async derived atom loading depending on where atom is used.
Hi there, I am running into an issue using async atoms. Here is the repro code sandbox: https://codesandbox.io/s/suspicious-nobel-bgzce

Expected:
- `roomClientAtom` should properly load after `currentSpaceAtom` and `roomServiceAtom` are configured.

Actual:
- `roomClientAtom` errors out
image

Some interesting behavior — if I lift `const [room] = useAtom(roomClientAtom)` from `<Room>` into `<Loader>`, the `roomClientAtom` is now properly loaded. Curious why behavior of derived atom depends on where it is called with `useAtom`? 

Repro this by uncommenting line 44

image

dai-shi:
close #201

### DIFF
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -201,7 +201,8 @@ const readAtomState = <Value>(
     const promiseOrValue = atom.read(((a: AnyAtom) => {
       if (dependencies) {
         dependencies.add(a)
-      } else {
+      }
+      if (!isSync) {
         setState((prev) => addDependency(prev, atom, a))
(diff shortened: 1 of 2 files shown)


dai-shi:
Confirmed that the issue is resolved. Merging.

tarngerine:
thank you! will test this after holidays

Issue #86, opened 2020-10-04
[Umbrella] Improve showcase


dai-shi:
https://github.com/pmndrs/jotai/blob/master/docs/showcase.md has only a few examples with no description.

- Add descriptions
- Add examples
- Update examples
- Make examples pretty Look&Feel
- Organize examples with category or something

LuckyPigeon:
I can work on this, and would you prefer all the code in a PR or separate them into a few PR, for example, an example a PR?

dai-shi:
Good question. For the time being, I'd just put codesandbox/github links in showcase.md.
We also have examples folder in the repo, and we'd like to improve it eventually.

We do have a channel in the pmndrs discord server. please find the discord badge in readme.

Aslemammad:
@dai-shi in the future, I'll write articles about jotai and deep dive in it.

LuckyPigeon:
@dai-shi 
I guess I need to split this to several PRs, at least a PR for one enhancement. I will start from adding descriptions.

dai-shi:
Sure. That's expected.

LuckyPigeon:
@dai-shi 
Is there a way to modify and commit the code in the showcase(`codesandbox`)?

dai-shi:
@LuckyPigeon No. Just fork it and replace with the new URL.

LuckyPigeon:
@dai-shi 
`github` can easily link to `codesandbox` by change `https://github.com/<repo_name>` to `https://githubbox.com/<repo_name>`. 
Due to this feature, is it a good idea that we download all of examples in showcase.md and put them into `examples` folder?
Then we can modify and commit it easily, and we can also connect to codesandbox as easy as before.

dai-shi:
@LuckyPigeon Alright, I understand your suggestion. I was going to work on `examples` some time later, but as you express your motivation, let me work a bit on `examples` folder so that you can put multiple examples in different folders. Hold on a moment.

dai-shi:
@LuckyPigeon 
It's ready. We have one example `./examples/hello`.
This can work both locally or in codesandbox. If it runs locally it will use the local jotai src, if in codesandbox it will use the published jotai package. http://githubbox.com/pmndrs/jotai/tree/master/examples/hello

Would you try adding another example as a start?

LuckyPigeon:
@dai-shi 
Sure!

dai-shi:
Let me close this as the initial iteration is done.
We'd want more examples official/unofficial.
[...]

PR #206, opened 2020-11-23
Fix share name issue

LuckyPigeon:
refer to #86 
@dai-shi 
Fix the share name issue inside `examples/*/package.json`


dai-shi (approved):
Thanks!

Issue #212, opened 2020-11-27
useAtomValue util


dai-shi:
Analogous to recoil's `useRecoilValue`, it'd be nice to have an equivalent hook.

This should be in `jotai/utils` along with `useUpdateAtom`.

The implementation in JS is as simple as this.
```js
const useAtomValue = (anAtom) => useAtom(anAtom)[0]
```

So, all we need would be:
- [ ] implement in TS
- [ ] add docs
- [ ] write test

Aslemammad:
@dai-shi I'm gonna work on it. Thanks.

Aslemammad:
#213 Merged.

PR #213, opened 2020-11-28
add useAtomValue

Aslemammad:
I'll do useAtomValue( suggested by sophia) tests and ... in other PR. #212


dai-shi (commented):
Thanks for your work. Would you update docs/utils.md?
I think it should be colocated with useUpdateAtom

Aslemammad:
@dai-shi Can I do that in different PR, I think it would be cleaner.

merisbahti:
In this PR: https://github.com/pmndrs/jotai/pull/193

I both added util, updated docs, and updated the `.size-snapshot.json` in one. I think there's merit to make these kind of changes in a consistent package, where all related info is updated in one PR (but ofc, can be multiple commits within that PR).

Just giving my 2 eurocent :)

Aslemammad:
@merisbahti Thanks for your suggestion. done.

docs/utils.md
```diff
@@ -23,6 +23,30 @@ const Controls = () => {
...
+Ref: https://github.com/pmndrs/jotai/issues/212
+
+```js
+import { atom, Provider, useAtom } from "jotai"
+import { useAtomValue } from "jotai/utils"
+
+const countAtom = atom(0)
+
+const Counter = () => {
+  const [, setCount] = useAtom(countAtom)
```
merisbahti, line 37: Maybe this one isn't needed, and we can comment out the one below?

src/utils.ts
```diff
@@ -5,4 +5,5 @@ export { atomWithReducer } from './utils/atomWithReducer'
 export { atomFamily } from './utils/atomFamily'
 export { useSelector } from './utils/useSelector'
 export { useAtomCallback } from './utils/useAtomCallback'
+export { useAtomValue } from './utils/useAtomValue'
```
dai-shi, line 8: This is just a minor thing, and we never had a strict rule so far.
But, how about moving this in L2, to be consistent with the order in docs?

Aslemammad: @dai-shi done.

dai-shi (approved):

Aslemammad:
@dai-shi Thanks.

PR #215, opened 2020-11-29
fix: useEffect not called if another setState bails out

### LINKED ISSUE #208: useEffect not called with latest atom value
In a very specific case, a `useEffect` isn't triggered on state update (render) despite dependencies being changed.

(**EDIT:** actually, the two atoms aren't necessary as updating a single one, twice, is enough to reproduce. See sandbox in [comment below](https://github.com/pmndrs/jotai/issues/208#issuecomment-733280403)) After stripping down the whole app into the smallest reproducible example I was able to narrow it down to this: when you update two atoms where the second atom is updated right in the next frame of the previous atom, and you also update the state (in the given `useEffect`) of the component that consumes those atoms, the `useEffect` isn't triggered for the 2nd atom's value.

It's possible the conditions can be simplified even further but this is what I came up with.

Note: this isn't an academic exercise. It came up in a production code and resulted in an app-breaking bug because the state didn't resolve (i.e. the `useEffect` wasn't called with updated, final value).

Reproducible example: https://codesandbox.io/s/frosty-field-tgduk?file=/src/index.js&expanddevtools=1

You should see something like this (you'll probably need to refresh the page via the in-app browser refresh button):

image

Does not reproduce 100%, but often enough.

Note: plain Context [isn't affected by this bug](https://codesandbox.io/s/agitated-glitter-yvroi?file=/src/index.js&expanddevtools=1).

- **Jotai version:** `0.6.2`, `0.11.1`
- **React versions:** `16.14.0`, `17.0.0`
- **Reproduces in strict mode:** yes (both strict and non-strict)

dai-shi:
Close #208 
It's an edge case which had to dealt with upstream use-context-selector.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     ]
   },
   "dependencies": {
-    "use-context-selector": "1.2.11"
+    "use-context-selector": "1.2.12"
   },
   "devDependencies": {
     "@babel/core": "^7.12.7",
--- a/tests/async.test.tsx
+++ b/tests/async.test.tsx
@@ -213,7 +213,7 @@ it('updates an async atom in child useEffect on remount without setTimeout', asy
   const countAtom = atom(0)
   const asyncCountAtom = atom(
     async (get) => get(countAtom),
-    async (get, set) => set(countAtom, (get(countAtom) as number) + 1)
+    async (get, set) => set(countAtom, get(countAtom) + 1)
   )
 
   const Counter: React.FC = () => {
@@ -263,7 +263,7 @@ it('updates an async atom in child useEffect on remount', async () => {
     },
     async (get, set) => {
       await new Promise((r) => setTimeout(r, 10))
-      set(countAtom, (get(countAtom) as number) + 1)
+      set(countAtom, get(countAtom) + 1)
     }
   )
 
--- a/tests/basic.test.tsx
+++ b/tests/basic.test.tsx
@@ -751,7 +751,7 @@ it('set atom right after useEffect (#208)', async () => {
     }
     useEffect(() => {
       effectFn(count)
-      setState(null) // this is important to repro
+      setState(null) // this is important to repro (set something stable)
     }, [count, setState])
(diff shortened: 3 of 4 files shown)


dai-shi:
https://codesandbox.io/s/angry-wilbur-6rf9j
It works.

@dwelle Would you check if this fixes your real app?

dwelle:
As luck would have it, it doesn't :frowning_face:.

I can confirm the case from the reproducible example I posted is now fixed (even locally), but not the exact scenario which prompted my issue, so there must be some tiny difference in what is really happening.

I guess we can merge this one as it fixes at least some case. When I get time I'll try to do another round of reproducing the bug on a small example :).

Thanks for the effort, so far! :heart:

dwelle:
Ok, here's a new repro, with 2 atoms again, and some weird things like setting the target atom within the same useEffect (in my app I don't do that, but I couldn't easily repro those circumstances without doing this. Either way, even this workflow may arise in real-world apps, so...).

https://codesandbox.io/s/morning-night-fhwvh?file=/src/index.js&expanddevtools=1

dai-shi:
Wow, how could you come up with this? I mean thanks for the repro.
That's really something we should solve as it seems nothing wrong in the usage.
However, I can't think of any clue now.
I'm planning another (or two) refactoring shortly. Let's revisit this after that.

@dwelle Would you create a new dedicated issue for the new repro?

merisbahti:
Just a question, we're looking for results that look like this, right:

image

To be clear: render values and useEffect values should be exactly the same?

dwelle:
> To be clear: render values and useEffect values should be exactly the same?

It should eventually end up in the same value, as in if the last-ever render value is `[100]`, then at some point (in next tick or whenever react decides to call `useEffect`) it should end up in `[100]` as well.

dai-shi:
Thanks. I'm merging this, but won't release it immediately. I will try another one and see how it goes.

Issue #220, opened 2020-11-30
UseEffect not invoked on atom value change in special circumstances


dwelle:
Follow-up to https://github.com/pmndrs/jotai/issues/208 (via https://github.com/pmndrs/jotai/pull/215).

UseEffect sometimes doesn't invoke with updated atom value, despite it being listed in dependencies. There's some weird race condition that probably cancels it or something. Don't understand what's happening and repro itself is insane.

Repro: https://codesandbox.io/s/morning-night-fhwvh?file=/src/index.js&expanddevtools=1

- **Jotai version:** https://pkg.csb.dev/pmndrs/jotai/commit/b989cdbf/jotai (the one after `0.11.1`)
- **React versions:**  `17.0.0`
- **Reproduces in strict mode:** yes (both strict and non-strict)

PR #223, opened 2020-12-01
upgrade use-context-selector

dai-shi:
https://www.npmjs.com/package/use-context-selector/v/1.3.0

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -24,9 +24,9 @@
     "gzipped": 3683
   },
   "utils.js": {
-    "bundled": 4187,
-    "minified": 1909,
-    "gzipped": 835,
+    "bundled": 4265,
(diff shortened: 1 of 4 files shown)


dai-shi:
rollup warns something.
```bash

src/index.ts → dist...
(!) Plugin typescript: @rollup/plugin-typescript TS4023: Exported variable 'useBridge' has or is using name 'CONTEXT_VALUE' from external module ".../jotai/node_modules/use-context-selector/dist/src/index" but cannot be named.
src/core/Bridge.ts: (7:14)

7 export const useBridge = (scope?: Scope) => {
               ~~~~~~~~~
```

dai-shi:
This doesn't help #220, unfortunately.

dai-shi:
The new version fixes the rollup/ts error.

PR #221, opened 2020-11-30
export RESET symbol and add tests (#217)

JLarky:
To continue the discussion in https://github.com/pmndrs/jotai/issues/217 I'm adding PR that will allow you to use `useResetAtom` in more cases.

* this exports `RESET` symbol, so bundle becomes slightly bigger
* add tests to cover:
  * resetting read-write atom (but it does nothing special, I could have added `+1` or `* 2` to illustrate the point more)
  * writing custom atom that could use `RESET` value just as a regular value or action (thus `atomWithReducer`, but I don't know if that's ok to create a test that tests both `atomWithReducer` and `useResetAtom`, in some sense this particular tests could be in either `useResetAtom.test.tsx` or `atomWithReducer.test.tsx`)

PR #228, opened 2020-12-04
we can now split atomWithReset and useResetAtom

JLarky:
* it wasn't possible before since RESET wasn't exported
* there's probably some good reason why we can put `RESET` symbol in its own file or in `useResetAtom`, but I don't know it
* this continues #221


dai-shi (approved):
Yeah, it makes sense now we exported `RESET`. We'd need to provide docs or guides to show use cases.

PR #229, opened 2020-12-04
format md with prettier for more consistent code examples

JLarky:
* prettier is only running when files are committed, there probably more files that need to be formatted, I did `yarn prettier --write docs/*.md readme.md`
* it might be possible to set different code style for `md` files then for code. I tried for example with `semi: true` and `singleQuote: false`, but the amount of changes seemed about the same, as half of the code in docs is in one style and the other in other :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "printWidth": 80
   },
   "lint-staged": {
-    "*.{js,ts,tsx}": [
+    "*.{js,ts,tsx,md}": [
       "prettier --write"
     ]
   },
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "printWidth": 80
   },
   "lint-staged": {
-    "*.{js,ts,tsx}": [
+    "*.{js,ts,tsx,md}": [
```
dai-shi, line 40: 👀

dai-shi (approved):
Thanks for the chore! First time to use prettier for md.

PR #233, opened 2020-12-11
Refactor with vanilla

### LINKED ISSUE #220: UseEffect not invoked on atom value change in special circumstances

### LINKED ISSUE #235: "Primitive" vs. "primitive" atoms
I'm not sure if this is a bug. Even if it is, I'm not sure it's worth fixing---it may be the best you can manage with typescript. But it did confuse me for a bit. 

The documentation talks about "primitive" atoms, which are atoms created with an initial value:
```
const anAtom = atom('foo')
```
There is also a public type `PrimitiveAtom<T>`. All "primitive" atoms are of type `Primitive<T>`, but you can create a derived atom and give it type PrimitiveAtom<T> without it actually being  "primitive". E.g.,
```
const pAtom: PrimitiveAtom<string> = atom(
  (get) => get(someOtherAtom)
  (get, set, update: SetStateAction<string>) => {set(someOtherAtom, update)}
)
```
I got confused because I knew I needed a "primitive" atom for some purpose, and thought that any object with type `PrimitiveAtom` would work. But that wasn't the case. Everything type-checked okay, but it didn't actually work at run time.

dai-shi:
This is a complete refactoring of the core.

The primary goal is to make the core code more readable and maintainable.
Split the core module out of Provider.ts into vanilla.ts without React dependency.
Now, the code is much cleaner with single WeakMap, and unified state.
All tests are passing with some tiny exceptions.

There are some changes.
- We avoid Promise.resolve() batching and let React batch updates. (Fix a bug with `<input>` which has been existed since v0.10.0)
- Drop `atom.key` and add `atom.toString()` instead. (To be documented.)
- Fix #220 
- Better `PrimitiveAtom` typing, close #235


dai-shi:
#220 
https://codesandbox.io/s/eloquent-kirch-zwy7v?file=/src/index.js
Should be fixed.
cc @dwelle

dwelle:
Sadly I can't repro the original bug anymore (I've saved the original commit) :(. But the CSB repro is fixed, so that's great :tada:  :heart:

dai-shi:
https://codesandbox.io/s/jotai-r3f-fri9d
jotai with r3f works fine with useBridge/Bridge.

dai-shi:
https://codesandbox.io/s/jotai-todos-ijyxm
jotai todos with `key={String(anAtom)}` works fine.

dai-shi:
This is going to be v0.12.0.

tests/error.test.tsx
```diff
@@ -393,7 +373,6 @@ it('throws an error while updating in effect', async () => {
     </Provider>
   )
 
-  expect(console.error).toHaveBeenCalledTimes(0)
```
dai-shi, line 396: This is removed, maybe because it's now not async.

Issue #231, opened 2020-12-11
Focus Atoms with async atoms?


mhcoffin:
Focus atoms seem incredibly cool, but they require a primitive base atom. Unless I'm missing something, that means I lose all the nice asynchronous initialization, right? Is there any way to get both?
 
-mike

merisbahti:
Mey Mike,

Let me see if I understand what you mean, so if we have an atom that's initialized using the `async` constructor like this:

```ts
const asyncAtom  = atom(async () => {
  const response = await Promise.resolve('just an example')
  return response
})
```

Right?

Which results in basically an `Atom<string>`, which is read-only.

So the problem is that `focusAtom` is only able to work on `Primitive` atoms, i.e: `Atom<string, SetState<String>>`.

The reason I did not expand the `focusAtom` to work on read-only-atoms (or async for that matter), is that the value that `focusAtom` provides is that it _both_ sets a getter and a setter. 

But it would be pretty easy to expand the type definition to work on read-only (or async) atoms!

is that kinda what you mean?

mhcoffin:
No, it's not read-only atoms that I want (although for orthogonality that seems like it might be a good idea, too). 

I created a [sandbox](https://codesandbox.io/s/focusatom-with-non-primitive-atoms-dyxqv?file=/src/App.tsx) that shows what I think would be very nice to have. Suspense would take care of the delay in resolving the promise, but the atom could still be updated later. Obviously, for optics to make sense, the atom would have to be WritableAtom<T, T>, with the update type the same as the base type.

What actually happens is that attempting to increment the age sets the original atom's value undefined, I guess because there is a set() function defined. Naively, I would have expected the focus atom to create the updated value using optics and then call 'set(baseAtom, newValue)', but maybe that can't work?

merisbahti:
Oh, I think I understand what you mean, but the sandbox does (for me) not currently contain any async atoms:
![image](https://user-images.githubusercontent.com/2946728/101957256-64e07e80-3c01-11eb-8e16-9aa620e55bcb.png)

So maybe I'll take a look at the sandbox for more detail before I ponder more upon this, but it _sounds_ to me that you have a valid case here, it's just that I didn't consider Read/Write async atoms when I wrote the optics package, but very interesting case and I think we should support it if we can!

mhcoffin:
[...]

PR #236, opened 2020-12-12
fix(focus-atom): add focus of async atom test

merisbahti:
This discussion lead me to making a test-case that just asserts this behavior:
https://github.com/pmndrs/jotai/issues/231

I think it's good to have a test case so that we don't accidently break something in the future.


dai-shi (approved):
Would be good to add such tests before #233. Merging this.

PR #244, opened 2020-12-20
fix: use array.from instead of spread operator for cjs

### LINKED ISSUE #243: testing with create react app - warning "Atom State not found"
Hello,

Thanks again for the great library, and sorry for spamming with so many issues.

I am seeing a pretty weird thing in my tests. My app runs fine, and I don't see any warnings, but when I try to test some components that use jotai, I am seeing the "atom state not defined" warning.

I am using create-react-app, so the tests are run through react-scripts.

I made a repro by copying one test case from your test suite that is similar to my use case (setting an atom from useEffect), and I get the warning. I tried the same test in your original test suite (cloning your repo and running the tests) and I get no warnings.

It looks like this isn't directly an issue with your library, but maybe an incompatibility with something react-scripts does, but maybe you have some ideas of where to look at?

The repro is here: https://codesandbox.io/s/jotai-testing-with-react-scripts-ytd89
If you open a new terminal there and do "yarn test" you should see the warning.
(I made it in js instead of ts to get the most minimalist reproduction I could)

Thanks!

dai-shi:
close #243 

The current config for cjs build doesn't transpile `[...<iterator>]` well. There might be a way to change the babel config, but this PR just changes it in the code to `Array.from(<iterator>)` for now.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 18722,
-    "minified": 8242,
-    "gzipped": 2861,
+    "bundled": 18750,
+    "minified": 8270,
+    "gzipped": 2866,
     "treeshaked": {
       "rollup": {
         "code": 143,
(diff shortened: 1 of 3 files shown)


dai-shi:
@jcane86 @Aslemammad 
Could you check if this really fixes the problem?

jcane86:
amazing, that's working beautifully now... thanks a bunch @dai-shi and @Aslemammad !

Aslemammad:
@jcane86 That's awesome, thanks @dai-shi.

dai-shi:
Thanks, guys. Merging this.

dai-shi:
Published: https://www.npmjs.com/package/jotai/v/0.12.1

PR #246, opened 2020-12-21
Refactor Provider.ts and vanilla.ts some more

dai-shi:
This refactors the core code to improve readability/maintainability and reduce bundle size.

There's a small behavioral change in async atoms, which wouldn't be a problem. Hope someone to file an issue if it causes an unexpected behavior.


dai-shi:
Published: https://www.npmjs.com/package/jotai/v/0.12.2

PR #251, opened 2021-01-06
feat: preact support

dai-shi:
Issue reported in https://github.com/dai-shi/use-context-selector/issues/35

This will allow preact users to use jotai with aliases:
```json
  "alias": {
    "react": "preact/compat",
    "react-dom": "preact/compat",
    "scheduler": false
  }
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     ]
   },
   "dependencies": {
-    "use-context-selector": "1.3.4"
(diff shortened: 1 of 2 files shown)


dai-shi:
Published: https://www.npmjs.com/package/jotai/v/0.12.3

cc @mischnic

PR #260, opened 2021-01-14
Fix Provider crash when not having any children

marcelltoth:
## Problem

Currently `<Provider />` crashes as `InnerProvider` returns `undefined` when it has no children. React components may not return `undefined` so you get an error.

### Real world use-case

I know this is an edge case but imagine a component that conditionally renders something that uses `Provider`, like:
```tsx
return <Provider>{someCondition ? <ChildThatUsesJotai /> : undefined}</Provider>
```

I hit it in a unit test where I wanted to test jotai-based persistence of a component by unmounting the child (while keeping `Provider` alive) then remounting it and seeing if it retains the values.

## Alternative solutions

Instead of the cast (which hides the bug), wrapping `children` in a Fragment makes sure this doesn't happen. The alternative would be to make it `children ?? null`, I don't have a strong opinion, I just picked one.

### DIFF
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -1,5 +1,4 @@
 import React, {
-  Fragment,
   MutableRefObject,
   ReactElement,
   createElement,
@@ -41,7 +40,7 @@ const InnerProvider: React.FC<{
   if (isReactExperimental && r.current === defaultContextUpdate) {
     r.current = (f) => contextUpdate(f)
   }
-  return createElement(Fragment, {}, children)
+  return (children as ReactElement) ?? null
 }
 
 export const Provider: React.FC<{
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -40,7 +40,7 @@ const InnerProvider: React.FC<{
   if (isReactExperimental && r.current === defaultContextUpdate) {
     r.current = (f) => contextUpdate(f)
   }
-  return children as ReactElement
+  return <>{children}</>
 }
 
 export const Provider: React.FC<{
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -1,4 +1,5 @@
 import React, {
+  Fragment,
   MutableRefObject,
   ReactElement,
   createElement,
@@ -40,7 +41,7 @@ const InnerProvider: React.FC<{
   if (isReactExperimental && r.current === defaultContextUpdate) {
     r.current = (f) => contextUpdate(f)
   }
-  return <>{children}</>
+  return createElement(Fragment, {}, children)
 }
 
 export const Provider: React.FC<{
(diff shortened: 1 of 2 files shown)


dai-shi:
Nice catch!
I actually concerned a bit when I added the type assertion.
I prefer `children ?? null`. Would you make a fix?

src/core/Provider.ts
```diff
@@ -41,7 +40,7 @@ const InnerProvider: React.FC<{
   if (isReactExperimental && r.current === defaultContextUpdate) {
     r.current = (f) => contextUpdate(f)
   }
-  return createElement(Fragment, {}, children)
+  return (children as ReactElement) ?? null
```
dai-shi, line 43:
```suggestion
  return (children ?? null) as ReactNode
```
How about this? We need to import `ReactNode`.

marcelltoth: That's not good as `ReactNode` is `ReactChild | ReactFragment | ReactPortal | boolean | null | undefined`, including `undefined` and `boolean` which are not valid return values.

marcelltoth:
Thanks for the quick reply! Sure thing, I changed it, also added a test case

src/core/Provider.ts
dai-shi: Oh, I see.

tests/provider.test.tsx
```diff
@@ -61,3 +61,7 @@ it('only uses initial value from provider for specific atom', async () => {
     getByText('pet: dog')
   })
 })
+
+it('renders correctly without children', () => {
+  render(<Provider />)
+})
```
dai-shi, line 67: 👍

src/core/Provider.ts
dai-shi: Does that mean `<Provider>{true}</Provider>` is invalid?

dai-shi (approved):
I'm not sure how to type this correctly (I guess there was a related issue in TS),
but the fix should be good.

src/core/Provider.ts
marcelltoth: I'm sorry I was wrong, booleans are fine (although they render nothing), `undefined` is not.

marcelltoth: And even `undefined` is fine when being a child, like `<Provider>{undefined}</Provider>`. What is not fine is returning it from a component.

marcelltoth:
Cool! 🎉 Please merge this whenever you feel like it's good to go.

src/core/Provider.ts
dai-shi: Okay, thanks for the clarification. I guess InnerProvider is tricky.

dai-shi:
Good. Merging this. The new version will be released with other fixes later.
Thanks for your contribution.

PR #264, opened 2021-01-17
fix(utils): add splitAtom utility

merisbahti:
Heya, just adding the `useAtomSlice` utility to make jotai able to make 100% optics-based applications.

----

Renamed to `splitAtom`.

- [x] impl
- [x] tests
- [x] docs


dai-shi:
Hey, thanks for working on this.
Can it be in utils? It doesn't depend on optics-ts, right?

I would like to refactor without using other utils (no big reason). As there are tests, I think I can work on the refactoring.

merisbahti:
Fixed. Gonna try and see why the performance test isn't succeeding...

merisbahti:
For some reason, there are unneccesary renders being done in the react-experimental mode in the second jest run. Any clue why?

dai-shi:
Even with commit counts? That's weird.

merisbahti:
Hey... I still can't figure out what's the problem? Maybe I should skip that performance test?

Here's a CSB with an example:

https://codesandbox.io/s/react-typescript-forked-15zs2?file=/src/App.tsx

dai-shi:
I expect `key={countAtom}` to work, but no? Is it expected?
I'll try some refactors.

dai-shi:
So, I ended up with rewriting the whole function, which turns into `sliceAtom`.
I don't think we need to expose `useAtomSlice` hook.

merisbahti:
Hey, the perf seems to have improved in the example, but remove stops working after removing one item:

https://codesandbox.io/s/remove-not-working-36cde

dai-shi:
Can you create a failing test?

merisbahti:
I've added it now.

I just realized - remove only fails for last item. That might be a clue...

dai-shi:
Thanks for the test and nice catch. Fixed!

merisbahti:
From what I can see, I'm happy with this.

My original intention with this was to make my `mega-form` idea be performant in jotai. Which it is, proof here: https://github.com/pmndrs/jotai/pull/296

dai-shi:
@merisbahti Sounds great. Mind writing some docs?

merisbahti:
Will do that!

merisbahti:
added docs, please check it out

docs/api/utils.md
```diff
@@ -386,3 +386,81 @@ const countAtom = atom(0)
 `atomFrozenInDev` is another function to create a frozen atom.
 The atom is frozen only in the development mode.
 In production, it works as the normal `atom`.
+
+## splitAtom
+
+The `splitAtom` utility is useful for when you want to get an atom for each element in a list.
+It works for read/write atoms that contain a list.
+
```
dai-shi, line 394: It would be nice to explain how this util works. Like, splitAtom takes an atom representing a list of items and return an array of atoms, each representing item. Maybe, typescript definition would help?

merisbahti: I've updated the docs, tell me if there's anything I've missed!

docs/api/utils.md
```diff
@@ -386,3 +386,93 @@ const countAtom = atom(0)
...
+## splitAtom
+
+The `splitAtom` utility is useful for when you want to get an atom for each element in a list.
+It works for read/write atoms that contain a list. When used on such an atom, it returns an atom
+which itself contains a list of atoms, each corresponding to the respective item in the original list.
+
+A simplified type signature would be:
+
+```ts
+type SplitAtom = <Element>(arrayAtom: PrimitiveAtom<Array<Elmement>>): Atom<Array<PrimitiveAtom<Element>>>
```
dai-shi, line 399:
```suggestion
type SplitAtom = <Item>(arrayAtom: PrimitiveAtom<Array<Item>>): Atom<Array<PrimitiveAtom<Item>>>
```

Issue #263, opened 2021-01-16
[Question] How to read/set from outside of react tree


Noitidart:
I know this is specifically to use with react tree. However I'm using this in react-native and I have a background task that needs to update the latitude/longitude as my users travel around. So the app starts up in background mode, none of the react tree is mounted, the task is started in background.

For this use case any ideas on how to read/write atom's from outside the react tree?

dai-shi:
You should use `useEffect` for read/set from outside. It's the same idea with `useState`.

Now, in case you have the background task to update values before react is ready, you might need to have a store outside react anyway. But unless that values are used anywhere outside react, you could only subscribe to the source in useEffect. I'm not sure if I understand the full requirement.

Noitidart:
Thanks so much for your fast reply. I was creating persisted atoms - https://github.com/pmndrs/jotai/blob/master/docs/persistence.md - for lattitude/longitude.

My background job needs to update the latitude/longigude of these persisted atoms. If the app is launched in background, nothing from react fired up, so its ok to update the file on disk. But if app is launched by user, then I wanted to show the atom values in the UI. But the job continues to update the files on disk, which doesn't trigger re-renders in react components. 

If curious here is what I'm doing - https://docs.expo.io/versions/v40.0.0/sdk/task-manager/#example

dai-shi:
So, the background task keeps writing locations to disk, and
only if the react app is running, you want to show the location real-time.

Maybe, you want an external store and then subscribe from useEffect.

```js
const store = {
  latitude: 0,
  longitude: 0,
  listeners: new Set(),
  update: (nextLatitude, nextLongitude) => {
     store.latitude = nextLatitude
     store.longitude = nextLongitude
     listeners.forEach(listener => listener())
  },
}

const Component = () => {
  const [location, setLocation] = useAtom(locationAtom)
  useEffect(() => {
    const callback = () => {
      setLocation({
        latitude: store.latitude,
        longitude: store.longitude,
      })
    }
    store.listners.add(callback)
    callback()
    return () => store.listeners.delete(callback)
  }, [setLocation])
  return (
    <>{location.latitude} {location.longitude}</>
  )
}
```

Notice, there's nothing special for jotai. The same thing for useState, if you change the line:
```diff
-  const [location, setLocation] = useAtom(locationAtom)
+  const [location, setLocation] = useState({ latitude: 0, longitude: 0 })
```

Noitidart:
[...]

PR #248, opened 2020-12-31
feat: jotai/query (powered by react-query)

dai-shi:
This is to add new lib `jotai/query` with react-query.

- [x] `atomWithQuery` implementation
- [x] tests
- [x] docs
- [x] examples

Note this PR is for the basic use case. It doesn't cover advanced use cases including mutation.

PR #266, opened 2021-01-19
feat: atom onMount

### LINKED ISSUE #211: [RFC] atom effects
Continuing from discussions in the discord channel.
Recoil proposes [Atom Effects](https://recoiljs.org/docs/guides/atom-effects/).
We don't need to provide the same API, and probably it's impossible as there are some conceptual differences between jotai and recoil. (It is interesting to me that things get to be seen differently in jotai from recoil, because originally jotai is modeled after recoil.)

However, there are good use cases. It seems like there are three possibilities for each use case.
1. can be implemented as derived atoms
2. can't be implemented as derived atoms, but something can be added in jotai core
3. can never be implemented in jotai

Collecting use cases might be important. Please share your ideas.

For 1, we could create new functions in jotai/utils: `withEffect` and `atomWithEffect`.

Any thoughts? volunteers?

### LINKED ISSUE #234: Good pattern for updating atoms based on subscription
Lets say I have an API that I can subscribe to, and any time a value changes (can happen once a frame at 60FPS) I would want to update an atom to always reflect that

Currently I can just create a useEffect to do this at the App.tsx level, but wondering if there's an atoms-only way to do this.

some pseudocode:
```
useEffect(() => {
  async function load() {
    setAtom(client.get(someId))
  }
  client.subscribe(someId, (msg) => setAtom(msg));
  () => {
    client.unsubscribe(someId);
  }
}, [])
```

dai-shi:
close #211 
close #234 
related #263 
related #248 

- [x] first hypothetical implementation
- [x] second implementation `onMount`
- [x] third implementation `onMount(setAtom)`
- [x] add tests
- [x] fix implementation
  - mutual dependents (maybe not for this pr)
  - [x] `[Bug] a dependency is not mounted`
- [x] refactor code
- [x] docs


src/core/vanilla.ts
```diff
@@ -336,9 +336,7 @@ export const delAtom = (
   if (mounted) {
     const [dependents] = mounted
     dependents.delete(useId)
-    if (!dependents.size) {
-      unmountAtom(state, deletingAtom)
-    }
```
dai-shi: @Aslemammad The intuition behind this is when this deletingAtom has other dependents, we shouldn't unmount it. Could you create some other test cases (I can think at least two cases (1) two components use an atom directly; (2) one component use it, another component use it indirectly), and see this change fails.

tests/onmount.test.tsx
```diff
@@ -144,7 +144,7 @@ it('mount/unmount test', async () => {
   expect(onUnMountFn).toBeCalledTimes(1)
 })
 
-it('one derived atom, one onMount for the derived one, and one for the regular atom + onUnMount', async () => {
+it('derive chain', async () => {
```
dai-shi, line 147: oh, we can leave the descriptive version. my list in the comment is just dumb.

Aslemammad: Aha, Ok

dai-shi:
Made an example in which an atom is sync with redux store.
https://codesandbox.io/s/react-typescript-forked-pftk3?file=/src/App.tsx

We can even sync the whole store and use `useSelector` from `'jotai/utils'`. 😲 
(and create dispatchAtom for `useDispatch`. In essence, creating `react-redux` with `jotai`. )

.size-snapshot.json
```diff
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 17865,
-    "minified": 8067,
-    "gzipped": 2823,
+    "bundled": 19712,
+    "minified": 8864,
+    "gzipped": 3071,
```
dai-shi, line 5: I don't know why it increases so much.

Aslemammad: Sorry, my wrong, Let me fix it.

dai-shi: I don't think so, I run quite lately.

Aslemammad: Wow, that's weird, the same thing, I thought it's my old build problem.

dai-shi (commented):
Should be good to go.

PR #278, opened 2021-01-24
fix: extra commits with changing atoms

### LINKED ISSUE #275: Atom hook causes redundant re-renders with wrong atom value
@dai-shi @merisbahti , thanks for the quick fix. in (https://github.com/pmndrs/jotai/issues/273)

However, I seemed to find another problem on the re-rendering. 
In v0.12.6, using atom hook causes redundant re-renders and the first render produces wrong atom value.

**Example** ( https://codesandbox.io/s/romantic-forest-lixid?file=/src/App.tsx )
Switch from atomA to atomB will trigger two renders on ItemWithAtomHook component. (which should be one I believe)
The first render of ItemWithAtomHook produces a wrong atom value with correct atom label.

```ts
Item render:  atomA 
ItemWithAtomHook render:  atomA b 
ItemWithAtomHook render:  atomA a 
```

Edit: v0.11.3 works fine

_Originally posted by @kueychuj in https://github.com/pmndrs/jotai/issues/273#issuecomment-765599255_

dai-shi:
close #275

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 17908,
-    "minified": 8090,
-    "gzipped": 2827,
+    "bundled": 17865,
+    "minified": 8067,
+    "gzipped": 2823,
     "treeshaked": {
       "rollup": {
         "code": 139,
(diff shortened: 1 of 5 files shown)


dai-shi:
https://codesandbox.io/s/nifty-euler-3mf5q
This doesn't fix the original issue.

https://codesandbox.io/s/beautiful-jepsen-311de
But, this one is fixed. That's the goal of this PR.

Issue #85, opened 2020-10-04
Improve docs/*.md


dai-shi:
#83 added some drafts. we need to improve them.

There would be several levels of contributions.
- Fix typo, grammar, and/or expressions (please file a PR)
- Ask questions in a new issue to clarify, add,  and improve descriptions (a followup PR would be nice)
- Or, you can start a PR directly with a draft doc and discuss there

Feel free to ask general questions in this issue. Also we have a channel in the pmndrs Discord server.

swagatikapanda123:
I am excited to work on this , Could you please assign it to me?

dai-shi:
This is like an umbrella issue. So, please create a sub task and assign yourself.

eyuu24:
Hi, can I take the first sub task? _Fix typo, grammar, and/or expressions (please file a PR)_

Does this involve reviewing the readme.md and other md files?

dai-shi:
> Does this involve reviewing the readme.md and other md files?

Basically, all docs/*.md files.

> can I take the first sub task?

Sorry for confusion. I meant sub tasks are more granular ones. Unless many contributors work conflict each other, it's not so important. See #90 for example.

In other words, please go ahead.

JLarky:
I guess this is the best place to discuss this for now :) for context https://github.com/pmndrs/jotai/pull/221#issuecomment-738484829 https://github.com/pmndrs/jotai/pull/228#pullrequestreview-544707395

So first answer to the question for documentation of RESET symbol:

 - https://github.com/JLarky/jotai/blob/new-docs/docs/guides/resettable.md#derived-atom-with-reset-symbol
 - https://github.com/JLarky/jotai/blob/new-docs/docs/api/utils.md#RESET

Now the overall idea behind the `new-docs` branch:

 - I have split docs into three categories: API, guides, introduction
 - introduction will explain basic concepts, talk about async, and comparison with zustand/recoil
 - guides will contain help on how to use jotai in some particular scenarios: nextjs, typescript, devtools; and how to implement some of the patterns: persistence, resettable atoms
 - and API describes APIs each bundle provides which consists of type (or some other way to display function signature), what arguments it takes, etc. and some small examples, examples here I think should be as small as possible so it will serve as a reference instead of the whole working application, but I think it's OK for API references to duplicate some of the text or examples that could be already there in guides or introduction

So now back to resettable atoms :)

[...]

PR #280, opened 2021-01-26
Split documentation into categories #85

JLarky:
this continues work I started in https://github.com/pmndrs/jotai/issues/85#issuecomment-738643270 unfortunately I didn't do everything I planned :) but this step should be good enough for now.

Just to reiterate the main points:
* I have split docs into three categories: API, guides, introduction
* Unfortunately this means that all existing links to docs on the internet will be wrong like https://github.com/pmndrs/jotai/blob/master/docs/core.md will become https://github.com/pmndrs/jotai/blob/master/docs/api/core.md but I at least made sure that (hopefully) all links inside the repo are going to be pointing to the right place
* added docs for `RESET` symbol
* I described things that didn't make it into the PR [here](https://github.com/pmndrs/jotai/issues/85#issuecomment-767394946)


dai-shi:
Merging this and let us continue discussion in #85.
Thanks tons!

Issue #281, opened 2021-01-26
SyntaxError: Cannot use import statement outside a module


MarkLyck:
When using Create-react-app with Jest.

Any tests that ends up important a Component using `jotai/utils` fail with the following error:

```
Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/mark/ypm/x/c1-frontend-v4/node_modules/jotai/utils.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import { useMemo, useCallback } from 'react';
                                                                                             ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      1 | import React, { useState } from 'react'
      2 | import { useAtom } from 'jotai'
    > 3 | import { useUpdateAtom } from 'jotai/utils'
        | ^
      4 | import { ReactKeycloakProvider } from '@react-keycloak/web'
      5 | import keycloak from 'src/lib/keycloak'
      6 | import { IDP_HINT } from 'src/common/constants'

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
      at Object.<anonymous> (src/lib/keycloak/AuthProvider.tsx:3:1)
```

I am using "vanilla" create-react-app. There's no custom babel configuration or anything just the default.

If I switch to just use `useAtom` from 'jotai' it works fine. But I want to be able to use the jotai/utils.

dai-shi:
Thanks for reporting.

> When using Create-react-app with Jest.

This sounds like one of the most popular use cases...

I know what is probably happening, but don't know what's best to fix. If anyone has any ideas, I would like to hear.

Meanwhile, if you changed like this, it would work.
```diff
- import { useUpdateAtom } from 'jotai/utils'
[...]

Issue #265, opened 2021-01-19
jotai breaks ie11 on nextjs with arrow function


lanyusan:
Hi, I am testing this great lib and found out it breaks ie11 on next js with the arrow function.

ie 11 reports this error:

```
SCRIPT1002: Syntax error

_app.js (11916,31)
```

And it points to this line of code:

```
const hasInitialValue = atom => 'init' in atom;
```

I understand jotai supports ie11. It seems its standard dist output doesn't transpire arrow function.

Any idea how to transpire it to work with nextjs on ie11?  Sorry if it is a naive issue. Couldn't find an issue addressing this.

Aslemammad:
@dai-shi I think we just need to configure babel to compile things to ES5.

dai-shi:
currently, the cjs build is targeting ie11.
https://github.com/pmndrs/jotai/blob/4ef6cfb06d6400080cd1de6dcb68a5d51dbfb4c5/rollup.config.js#L50

Does it work with ie11 if it's not on nextjs? I'm not sure how nextjs build the bundle.

lanyusan:
My understanding is that Nextjs doesn't transpire use imported node modules.

I just use `import { atom ... } from 'jotai'` with no other nextjs configration.

I have tried `const { atom ... }  = require('jotai')` , still the same problem.

So do I need to specifically ask nextjs/webpack to use the `cjs` build? Any idea how to do so with nextjs or webpack? Did a search but couldn't find a straight answer.  Sorry if it is naive.

dai-shi:
@lanyusan 
Could you try this?
```js
import { atom } from 'jotai/index.cjs'
```
to explicitly use cjs build.

lanyusan:
Got this compile error:

`Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './index.cjs' is not defined by "exports" in /.../node_modules/jotai/package.json`

dai-shi:
Oh, I see. ie11 is traditional, but nextjs (node14?) is very new. Let me see what we can do.

dai-shi:
Unfortunately, I couldn't come up with a good solution.

if `const { atom } = require('jotai')` doesn't work, I would like you to try transpiling module at your end.

https://nextjs.org/docs/advanced-features/customizing-babel-config
Seems like you can customize the babel config.

https://github.com/martpie/next-transpile-modules
This package might help easing the work.

lanyusan:
[...]

Issue #61, opened 2020-09-21
[Next.js] Error using useUpdateAtom


lishine:
Using Next.js
Attached screenshot
The code is like this
```ts
import { atom, useAtom } from 'jotai'
import { useUpdateAtom } from 'jotai/utils'

const eventsAtom = atom<Events>([])

export const useCreateSSE = () => {
    let updateEvents = useUpdateAtom(eventsAtom)
}
```

Wrapping Component in _app with Provider
And then using this custom hook in some component

When I used useUpdateAtom code directly, just copied from the repo, it worked.

my tsconfig:
```json
{
    "compilerOptions": {
        "target": "es6",
        "lib": ["dom", "dom.iterable", "esnext"],
        "allowJs": true,
        "skipLibCheck": true,
        "strict": true,
        "forceConsistentCasingInFileNames": true,
        "noEmit": true,
        "noImplicitAny": true,
        "strictNullChecks": true,
        "strictFunctionTypes": true,
        "strictPropertyInitialization": true,
        "noImplicitThis": true,
        "alwaysStrict": true,
        "noImplicitReturns": true,
        "noFallthroughCasesInSwitch": true,
        "esModuleInterop": true,
        "module": "esnext",
        "moduleResolution": "node",
        "resolveJsonModule": true,
        "isolatedModules": true,
        "downlevelIteration": true,
        "baseUrl": ".",
        "allowSyntheticDefaultImports": true,
        "jsx": "preserve",
        "paths": {
            "~/*": ["./src/*"]
        }
    },
    "exclude": ["node_modules", "public", "not_used"],
    "include": ["next-env.d.ts", "src/**/*.ts", "src/**/*.tsx", "src/**/*.jsx", "src/**/*.js"]
}
```

![image](https://user-images.githubusercontent.com/6741645/93819110-faecb480-fc63-11ea-8d69-2dbf9b4498aa.png)

lishine:
![image](https://user-images.githubusercontent.com/6741645/93819607-a0078d00-fc64-11ea-8792-e1ed6875b64f.png)

lishine:
"react": "16.13.1"

dai-shi:
So, I assume that esm doesn't work in nextjs server, maybe? Let me create the cjs build.

lishine:
Ok, I added dev-dependency esm
And run NODE_OPTIONS='-r esm' next dev
And it works fine

Is it the way people import esm ?
I am using node 14.8. isn't es modules supported already?

[...]

PR #282, opened 2021-01-27
[BREAKING] change default js files to commonjs

dai-shi:
follow-up: #165 

It turns out that exports field is not yet supported, and there are so many envs still to depend on cjs.
- nextjs #61 
- ie11 #265 
- cra/jest #281 

This change the default `*.js` to be cjs.


dai-shi:
@MarkLyck Would you try this build, and see how it works?

PR #284, opened 2021-02-01
fix: dist mjs files in modules

dai-shi:
Looks like having cjs files and mjs files in the same directory is troublesome with webpack.
It's unfortunate, but put mjs files in `modules`.
Also drop iife.js assuming it's never used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,28 +3,29 @@
   "private": true,
   "version": "0.13.0",
   "description": "👻 Next gen state management that will spook you",
+  "type": "commonjs",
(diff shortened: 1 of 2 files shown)


dai-shi:
At least, the codesandbox is working fine with this.
https://codesandbox.io/s/react-typescript-forked-2n2fe?file=/src/App.tsx

dai-shi:
I don't get enough feedback, but I'm pretty sure v0.13.0 is troublesome.
Let me merge this, release it, and see how it goes.

PR #288, opened 2021-02-01
fix: module build again

### LINKED ISSUE #287: Can't import the named export 'BridgeProvider' from non EcmaScript module (only default export is available)
Upgraded to 0.13.1 and now getting this error:

`Can't import the named export 'BridgeProvider' from non EcmaScript module (only default export is available)`

dai-shi:
fix #287 

How's this?


dai-shi:
This seems good.

PR #293, opened 2021-02-05
chore: add a note about useBridge

dai-shi:
Ref: https://github.com/pmndrs/jotai/discussions/292

### DIFF
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -162,6 +162,11 @@ The `updateValue` takes just one argument, which will be passed to the third arg
 
 ## useBridge/Bridge
 
+Libraries such as react-three-fiber use custom renderer/reconciler a.k.a multiple roots.
+Contexts don't go beyond different renderer, so it requires a workaround.
+Bridge is a known technique for the escape hatch.
+See [this](https://github.com/pmndrs/react-three-fiber/issues/43) for more information.
+
 ```ts
 const useBridge = (scope?: Scope) => [Actions, State]
 ```


dai-shi:
@github0013 can review please?

github0013:
I am sorry, but I am still scratching my head hard trying to comprehend what they are all about.

> You get a bridge value with useBridge in the outer component and pass it to Bridge in the inner component.
> https://github.com/pmndrs/jotai/blob/master/docs/api/core.md#usebridgebridge

So in this example what is it that the inner is getting from the outside?
https://codesandbox.io/s/jotai-r3f-fri9d?file=/src/App.tsx:143-152

I only see `boxes` in Main component passed from above, but that is just a regular prop.., no?
It's maybe because I don't know what `Canvas` is really doing.

https://standard.ai/blog/introducing-standard-view-and-react-three-fiber-context-bridge/
I stumbled upon this article, and it is barely telling me that `Canvas` converts JSX to DOM so that React world's Context can't reach into the DOM?

Anyways, would you say I won't have a chance to use `useBridge/Bridge` unless I use this `Canvas` component?  (they are parts of the core, so I am thinking I should know about them before I start using this library)

github0013:
Oh, I guess the `boxes` are internally coming from a React's context, so any changes happen in the context will be bridged over to Canvas component?

dai-shi:
> Anyways, would you say I won't have a chance to use `useBridge/Bridge` unless I use this `Canvas` component?

`Canvas` is the component from react-three-fiber.
Yes, you won't have a chance to use this unless you use custom renderers.
Here I found some (many) of custom renderers.
https://github.com/chentsulin/awesome-react-renderer

Only when you use at least two of them together **and** you want to bridge them with a single `<Provider>`, you need `useBridge/Bridge` or context bridge in general.

dai-shi:
Merging this now. Suggestions to docs in general is highly appreciated. #85

PR #296, opened 2021-02-06
fix(examples): add example with mega form

merisbahti:
# Mega form example

Adds an example to demonstrate 3 points about the optics-based jotai pattern. Also suggests fix for the broken apps `examples` folder. 

## Developer ergonomics

Using the new `splitAtom` utility, in combination with `focusAtom`, we can finally make `Atom`-based components that let us ergonomically handle _huge complex_ state. This complexity is handled by focusing, and slicing down state, into smaller parts, and giving those smaller parts to smaller components, that are responsible for small parts of state.

Our state is contained in `src/App.tsx`, and it is a complex "form of forms", expressed in ts by: `Record<string, Record<string, string>>`. Yet, our `Field`-component simply takes a `PrimitiveAtom<{name: string, value: string}>` as a prop. This makes the `Field`-component, and other similar components, very easy to test. Since they don't depend on any global state, only on input atoms! This has accomplished by using the `splitAtom` and `focusAtom` utilities.

## Extremely performant, without any hassle.

This has extremely good performance, in two ways:

1. Only neccesary elements are updated, when changing the value of a field, only that `Field` atom is updated:
image
2. The updates are really cheap. This is svelte level performance ;)
image

This without doing any sorts of optimizations!

## State fully serializable

Ok good ergonomics, easy testing of components, good performance, but in addition: the state is 100% serializable, you might aswell have used a `localStorageAtom` for this! Or done it over websocket!

## Notes:
I couldn't get any of the apps in `examples` to work, so I built this example app using a very simple parcel setup. I think this is the way to go for our example apps, as `react-app-rewired` is bloated and messy.


dai-shi:
> as react-app-rewired is bloated and messy.

yeah, we can drop that.. the intention was to run the example with local jotai code.
but, it seems hard to maintain with react-app-rewired.

> so I built this example app using a very simple parcel setup.

I'm not sure if we need parcel. in valtio repo, we don't use react-app-rewind, it works fine. https://github.com/pmndrs/valtio/tree/master/examples/counter

There's nothing wrong with parcel, but if possible, we want to make them consistent.

---

on the other hand, the purpose of examples is to run directly in codesandbox with `githubbox`.

See the doc https://github.com/pmndrs/jotai/blob/master/docs/introduction/showcase.md
which has the link https://githubbox.com/pmndrs/jotai/tree/master/examples/text_length
which works without react-app-rewired

Would you try it with mega-form and make it work?
https://githubbox.com/pmndrs/jotai/tree/meris/mega-form-example/examples/mega-form

merisbahti:
> > as react-app-rewired is bloated and messy.
> 
> yeah, we can drop that.. the intention was to run the example with local jotai code.
> but, it seems hard to maintain with react-app-rewired.
> 
> > so I built this example app using a very simple parcel setup.
> 
> I'm not sure if we need parcel. in valtio repo, we don't use react-app-rewind, it works fine. https://github.com/pmndrs/valtio/tree/master/examples/counter
> 
> There's nothing wrong with parcel, but if possible, we want to make them consistent.
> 
> on the other hand, the purpose of examples is to run directly in codesandbox with `githubbox`.
> 
> See the doc https://github.com/pmndrs/jotai/blob/master/docs/introduction/showcase.md
> which has the link https://githubbox.com/pmndrs/jotai/tree/master/examples/text_length
> which works without react-app-rewired
> 
> Would you try it with mega-form and make it work?
> https://githubbox.com/pmndrs/jotai/tree/meris/mega-form-example/examples/mega-form

Ok I managed to make the link work, had to add `optics-ts` to deps. https://githubbox.com/pmndrs/jotai/tree/meris/mega-form-example/examples/mega-form

I just wanted to have a minimal amount of files in the example, to capture the essence of the example. It feels pretty intimidating to have some kind of custom-react-scripts setup plus some unfamiliar config file.

Are we sure we want to go with that? 

(Oh https://github.com/pmndrs/valtio/tree/master/examples/counter looks nice though, even though I have a bad impression from `react-scripts` always being broken somehow)

merisbahti:
Ok I don't really know why, but `react-script` keeps giving this error:
image
I just don't know what to do. It worked great in parcel, shouldn't we just keep parcel? Maybe we dont need all complexity that `react-scripts` brings for this?

dai-shi:
> Maybe we dont need all complexity that react-scripts brings for this?

No, we don't. I just thought it would work flawlessly if we dropped react-app-rewind.

> It worked great in parcel, shouldn't we just keep parcel?

Yeah, if it's less troublesome, we can go with parcel. Thanks for your investigation.

dai-shi:
@merisbahti has be keen about improving performance and created this example.
@Thisen might be interested in it too.

Let me merge this in. We can improve it as we go.

PR #303, opened 2021-02-15
chore: refactor with single context

dai-shi:
Having two contexts was somewhat for performance for the future.
But, if react supports use-context-selector natively in the future, we might not need this, furthermore, it doesn't improve performance as of now.
Having a single context simplifies code and possibly has room for improvements.


dai-shi:
https://codesandbox.io/s/jotai-r3f-forked-t9cie
jotai-r3f example is working fine.

PR #305, opened 2021-02-15
fix: promise value causing infinite loop

### LINKED ISSUE #304: Returning a promise from a reducer causes a memory leak
Code sandbox:
https://codesandbox.io/s/zealous-mountain-bkfnb?file=/src/App.js

Basically, I want to handle async logic in the reducer and then update state with the returned value.

dai-shi:
close #304

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.module.js": {
-    "bundled": 19580,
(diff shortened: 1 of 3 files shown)


dai-shi:
https://codesandbox.io/s/icy-star-m8q3x
It's working with `<Suspense>`

PR #316, opened 2021-02-24
fix(examples): use officially released version of jotai for mega form

merisbahti:
Since we now have `splitAtom` in mainline version of jotai, let's use that instead and not the CSB-build.

Not related to this PR:
Also, I tried upgrading to `^0.15.0` but it crashes (too many renders according to react). Is this something we should discover? Might be a bug from the provider-less mode.

### DIFF
--- a/examples/mega-form/package.json
+++ b/examples/mega-form/package.json
@@ -10,7 +10,7 @@
   "dependencies": {
     "fp-ts": "^2.9.1",
     "io-ts": "^2.2.13",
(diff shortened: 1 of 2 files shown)


merisbahti:
Please note that, if I memoize the focused atoms in the `Field`-component, it works fine:
image

dai-shi:
> Please note that, if I memoize the focused atoms in the `Field`-component, it works fine:
> image

Oh, is this about #317? Yes, useMemo is necessary. I didn't even know it worked without it.

dai-shi (approved):
Merging this. We probably have similar things in other does/examples...

PR #317, opened 2021-02-24
fix(tests): add breaking test for mega form

merisbahti:
Hey, I'm not 100% sure what we should do here. I noticed that with the most recent version mega form example crashes (but it did not crash in `0.14.5`.

So I'm wondering, is this test case useful for you to assert that mega-form works in `0.15.0` and later versions?

### DIFF
--- a/tests/utils/megaFormSplitAtom.test.tsx
+++ b/tests/utils/megaFormSplitAtom.test.tsx
@@ -1,0 +1,147 @@
+import { atom, Provider, useAtom, PrimitiveAtom } from 'jotai'
+import { focusAtom } from '../../src/optics'
+import React, { useMemo, useEffect, useRef } from 'react'
+import { render, fireEvent, waitFor } from '@testing-library/react'
+import { useAtomCallback } from '../../src/utils'
+import { splitAtom } from '../../src/utils/splitAtom'
+
+type TodoItem = { task: string; checked?: boolean }
+
+it('no unneccesary updates when updating atoms', async () => {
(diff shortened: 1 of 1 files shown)


dai-shi:
Yes, this is very helpful. I'm rewriting the core for the second iteration. Will investigate how it goes.

tests/utils/megaFormSplitAtom.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+})
+
+const Field = ({
+  fieldAtom,
+  removeField,
+}: {
+  fieldAtom: PrimitiveAtom<{ name: string; value: string }>
+  removeField: () => void
+}) => {
+  const [name, setName] = useAtom(focusAtom(fieldAtom, (o) => o.prop('name')))
```
dai-shi, line 31:
```suggestion
  const [name, setName] = useAtom(useMemo(() => focusAtom(fieldAtom, (o) => o.prop('name')), [fieldAtom]))
```

Does this fix the test?

merisbahti: Yes this fixes the test - (if applied to both useAtoms in that component) however I was hoping that we would not need to do `useMemo` since the value that the atom points to is referentially stable (`string`).

dai-shi: jotai's concept totally depends on atom referential identity. the atom value being string doesn't help, unfortunately.

merisbahti:
closing since this isn't supposed to be supported by jotai

PR #328, opened 2021-03-03
fix: provider debug state

dai-shi:
follow-up #326

state is really atomic, but this allows an opt-in listener so that we can see all atoms from outside. (for debug purpose only)

### DIFF
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -20,7 +20,7 @@ export const Provider: React.FC<{
   ) {
     /* eslint-disable react-hooks/rules-of-hooks */
     const atomsRef = useRef<AnyAtom[]>([])
-    if (storeRef.current == null) {
+    if (storeRef.current === null) {
       // lazy initialization
       storeRef.current = createStore(initialValues, (newAtom) => {
         atomsRef.current.push(newAtom)
@@ -32,7 +32,7 @@ export const Provider: React.FC<{
     )
     /* eslint-enable react-hooks/rules-of-hooks */
   } else {
-    if (storeRef.current == null) {
+    if (storeRef.current === null) {
       // lazy initialization
       storeRef.current = createStore(initialValues)
     }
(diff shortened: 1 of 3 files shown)


dai-shi:
@Thisen This is related with devtools. With the same technique, we can take snapshots.

src/core/Provider.ts
```diff
@@ -1,51 +1,87 @@
...
-  }
-  const store = storeRef.current as ReturnType<typeof createStore>
 
-  if (typeof process === 'object' && process.env.NODE_ENV !== 'production') {
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    useDebugState(useMutableSource(store[0], getState, subscribe))
+  if (
+    typeof process === 'object' &&
+    process.env.NODE_ENV !== 'production' &&
+    process.env.NODE_ENV !== 'test'
```
Thisen, line 19: I do see the assertion on `process`, but could this potentially screw with consumers test builds?

dai-shi:
https://codesandbox.io/s/react-typescript-forked-zww2r?file=/src/App.tsx
React devtools is working good now. 👍

src/core/Provider.ts
```diff
@@ -1,51 +1,87 @@
...
+        ],
       ]
     })
   )
 
-const useDebugState = (state: State) => {
-  useDebugValue(state, stateToPrintable)
+const getState = (state: State) => ({ ...state }) // shallow copy
+
+const useDebugState = (store: Store, atoms: AnyAtom[]) => {
```
Thisen, line 74: Maybe add a comment about that this makes sure we keep a reference to the atom in devmode, so it isn't GC'ed by the WeakMap

dai-shi: great idea. is it allowed for you to add a suggestion?

src/core/Provider.ts
dai-shi: This is newly added in this PR. Wish there would be a way to avoid this.
Without it, the test fails because the provider invokes extra subscribes.

src/core/Provider.ts
Thisen, line 74:
```suggestion
// This hook makes sure that we keep a reference to the atoms, in dev mode, so the atoms isn't garbage collected by the WeakMap of mounted atoms
const useDebugState = (store: Store, atoms: AnyAtom[]) => {
```

Thisen:
```suggestion
// This hook makes sure that we keep a reference to the atoms in dev mode, 
// so atoms aren't garbage collected by the WeakMap of mounted atoms
const useDebugState = (store: Store, atoms: AnyAtom[]) => {
```

PR #330, opened 2021-03-03
Fix/remove unnecessary await

### LINKED ISSUE #329: Uncaught ReferenceError: regeneratorRuntime is not defined
```
query.js:1007 Uncaught ReferenceError: regeneratorRuntime is not defined
    at query.js:1007
    at atomWithQuery (query.js:1037)
```

I'm getting this error when setting using `jotai/query`. I'm using it as shown below. 

```
const profileAtom = atomWithQuery((get) => ({
  queryKey: [],
  queryFn: async () => {
    return await getProfile();
  },
}));

export const useUserProfile = () => useAtom(profileAtom);
```

`getProfile` simply fetches the user's profile from the backend using the user's auth token. 

Googling the error message cites a few issues people encountered with Babel. 

For what it's worth, we're using `snowpack` for this repo. 

Anyways, I haven't encountered this error before, something to do with `async/await` not being polyfilled correctly but I'm not terribly sure. 

Any ideas would be much appreciated!

dai-shi:
We don't really need to use async/await in atomWithQuery.
It was there because there was a typing issue.
This fixes it with explicitly adding type annotation.

This should temporarily resolve #329 (But, we should explore fundamental resolution too.)

### DIFF
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -124,13 +124,16 @@ export function atomWithQuery<
       }
       return [options, observerAtom]
     },
-    async (get, set, action) => {
+    (get, set, action) => {
       if (action.type === 'refetch') {
         const [options] = get(queryAtom)
         set(pendingAtom, createPending<TData>()) // reset pending
-        getQueryClient(get, set).getQueryCache().find(options.queryKey)?.reset()
-        await getQueryClient(get, set).refetchQueries(options.queryKey)
+        const queryClient = getQueryClient(get, set)
+        queryClient.getQueryCache().find(options.queryKey)?.reset()
+        const p: Promise<void> = queryClient.refetchQueries(options.queryKey)
+        return p
       }
+      return
     }
   )
   const queryDataAtom = atom<TData, ResultActions>(


src/query/atomWithQuery.ts
```diff
@@ -124,13 +124,16 @@ export function atomWithQuery<
...
-    async (get, set, action) => {
+    (get, set, action) => {
       if (action.type === 'refetch') {
         const [options] = get(queryAtom)
         set(pendingAtom, createPending<TData>()) // reset pending
-        getQueryClient(get, set).getQueryCache().find(options.queryKey)?.reset()
-        await getQueryClient(get, set).refetchQueries(options.queryKey)
+        const queryClient = getQueryClient(get, set)
+        queryClient.getQueryCache().find(options.queryKey)?.reset()
+        const p: Promise<void> = queryClient.refetchQueries(options.queryKey)
```
Aslemammad, line 133: Aha, a great solution.

Aslemammad:
It looks great to me. Nice catch!

Thisen (approved):

PR #331, opened 2021-03-04
feat: useUpdateAtom in core [BREAKING CHANGE]

dai-shi:
This implements useUpdateAtom in core. This is a breaking change.

A few concerns:
- This is only because it need to use some internals for better implementation (more performance)
- We want to keep the simplicity of API. atom/useAtom is minimalistic.
- I never want to add useAtomValue in core, it's just a syntax sugar.
- If we change the core in the future, the new implementation could not be kept and can't stay in core. We will make a breaking change again.

Alternatives:
1. export `getStoreContext` officially and leave it in utils.
2. export `getStoreContext` unofficially (undocumented and only for internal use) and leave it in utils.

Hmm, I'm leaning to alternative 2.


Aslemammad:
@dai-shi I think we should document everything, and about A2, I think we should document it and say it's for internal usage and not recommended, maybe.

Thisen:
I think we should go with alternative 2 and leave some docs about this being internals and breaking changes to it can occur without notice.

dai-shi:
Thanks guys. Yeah, I'm not against documenting internals with cautions.

dai-shi:
Closing this in favor of #332.

PR #332, opened 2021-03-04
re-remplement useUpdateAtom with secret internals

dai-shi:
obsoletes #331 

I'm more comfortable with this. No breaking changes on public api.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,10 @@
 export { Provider } from './core/Provider'
 export { atom } from './core/atom'
 export { useAtom } from './core/useAtom'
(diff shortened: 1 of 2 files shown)


Thisen:
Maybe add a JSDoc on `getStoreContext` around:
```
/**
* This is internal and can break without notice. Don't use this in application code.
**/
```

dai-shi:
Oh, do you mean docs in code? Do we need some notes in the docs/*.md docs?

Thisen:
@dai-shi I mean as a doc comment on `getStoreContext`

Thisen (approved):

Thisen (approved):

PR #336, opened 2021-03-04
Fix useUpdateAtom for write-only atoms

Thisen:
This fixes `useUpdateAtom` for write without argument.

I'm not sure about the export of `SetAtom`, but I couldn't see any other way.

### DIFF
--- a/tests/utils/useUpdateAtom.test.tsx
+++ b/tests/utils/useUpdateAtom.test.tsx
@@ -119,3 +119,48 @@ it('useUpdateAtom with scope', async () => {
     getByText('count: 1')
   })
 })
+
+it('useUpdateAtom with write-only atom', async () => {
+  const countAtom = atom(0)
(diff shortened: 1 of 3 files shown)


dai-shi:
Ah, you are absolutely right because previously it was `atom()`. I did use `SetAtom` in #331. Let me tweak it a bit to avoid exporting it in the public api.

tests/utils/useUpdateAtom.test.tsx
```diff
@@ -119,3 +119,48 @@ it('useUpdateAtom with scope', async () => {
     getByText('count: 1')
   })
 })
+
+it('useUpdateAtom with write-only atom', async () => {
```
dai-shi, line 123:
```suggestion
it('useUpdateAtom with write without an argument', async () => {
```

dai-shi (approved):

PR #339, opened 2021-03-05
Refactor useResetAtom with secret internals

Thisen:
This refactors `useResetAtom` with almost identical implementation as `useUpdateAtom` - This saves an atom creation.

As commented, `_update` stays for compatibility changes, but I propose removing it before v1.

### DIFF
--- a/src/utils/useResetAtom.ts
+++ b/src/utils/useResetAtom.ts
@@ -6,8 +6,7 @@ import { RESET } from './atomWithReset'
 export function useResetAtom<Value>(anAtom: WritableAtom<Value, typeof RESET>) {
   const StoreContext = getStoreContext(anAtom.scope)
   const [, updateAtom] = useContext(StoreContext)
-  // FIXME Remove _update before v1, stays to not introduce breaking change
-  const setAtom = useCallback((_update) => updateAtom(anAtom, RESET), [
+  const setAtom = useCallback(() => updateAtom(anAtom, RESET), [
     updateAtom,
     anAtom,
   ])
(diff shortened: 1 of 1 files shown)


src/utils/useResetAtom.ts
```diff
@@ -1,12 +1,15 @@
...
-  const writeOnlyAtom = useMemo(
-    () => atom(null, (_get, set, _update) => set(anAtom, RESET)),
-    [anAtom]
-  )
-  writeOnlyAtom.scope = anAtom.scope
-  return useAtom(writeOnlyAtom)[1]
+  const StoreContext = getStoreContext(anAtom.scope)
+  const [, updateAtom] = useContext(StoreContext)
+  // FIXME Remove _update before v1, stays to not introduce breaking change
+  const setAtom = useCallback((_update) => updateAtom(anAtom, RESET), [
```
dai-shi, line 10:
```suggestion
  const setAtom = useCallback(() => updateAtom(anAtom, RESET), [
```
I assume removing it doesn't break anything. Or, is there any case this would fail?

Thisen: If, for no reason, you have added an argument, it would fail. That argument would do nothing, but the types would break. It took an `update?: Update` before.

dai-shi: Okay, thanks for clarification. I can say this improves types (previous one wasn't precise.) Let me remove it and merge it. Thanks for your work!

dai-shi (approved):

PR #342, opened 2021-03-06
Update the domain example

Aslemammad:


examples/hello/src/App.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import { Fragment } from 'react'
 import { Provider, atom, useAtom } from 'jotai'
```
dai-shi, line 2:
```suggestion
import { atom, useAtom } from 'jotai'
```
Can we remove this too?

dai-shi (approved):
Looks good. Thanks!

PR #348, opened 2021-03-11
refactor with invalidatedRevision

dai-shi:
This might not change much for now.
Hope this is better with real uMS and in CM.
At least the code is cleaner. 😁

### DIFF
--- a/src/core/vanilla.ts
+++ b/src/core/vanilla.ts
@@ -221,7 +221,7 @@ const setAtomReadPromise = <Value>(
   return new Map(wip).set(atom, atomState)
 }
 
-const setAtomInvalidatetd = <Value>(
+const setAtomInvalidated = <Value>(
   state: State,
   wip: WorkInProgress,
   atom: Atom<Value>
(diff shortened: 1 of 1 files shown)


src/core/vanilla.ts
```diff
@@ -212,10 +216,21 @@ const setAtomReadPromise = <Value>(
     atom,
     dependencies
   ) as AtomState<Value>
+  delete atomState.i
   atomState.p = promise
   return new Map(wip).set(atom, atomState)
 }
 
+const setAtomInvalidatetd = <Value>(
```
Thisen, line 224:
```suggestion
const setAtomInvalidated = <Value>(
```

src/core/vanilla.ts
```diff
@@ -212,10 +216,21 @@ const setAtomReadPromise = <Value>(
...
+  state: State,
+  wip: WorkInProgress,
+  atom: Atom<Value>
+): WorkInProgress => {
+  const atomState = wipAtomState(state, wip, atom) as AtomState<Value>
+  atomState.i = atomState.r
+  return new Map(wip).set(atom, atomState)
```
Thisen, line 231: Is `concatMap` better here?

src/core/vanilla.ts
dai-shi: 😓

src/core/vanilla.ts
dai-shi: It's just one set, so this should be good.

PR #358, opened 2021-03-14
Feature: waitFor utilities

Thisen:
Makes sure your async atoms aren't resolving in a waterfall.

Heavily inspired by Recoils [waitFor utilities](https://recoiljs.org/docs/api-reference/utils/waitForAll).

- [X] Implementation
- [X] Tests
- [X] Docs


src/utils/waitFor.ts
```diff
@@ -0,0 +1,46 @@
+import { Atom, Getter } from '../core/types'
+import { atomFamily } from './atomFamily'
+
+export const waitForAll = atomFamily(
+  (dependencies: ReadonlyArray<Atom<unknown>>) => (get) => {
```
dai-shi, line 5:
```suggestion
  (dependencies: readonly Atom<unknown>[]) => (get) => {
```
Is this the same?

src/utils/waitFor.ts
```diff
@@ -0,0 +1,46 @@
+import { Atom, Getter } from '../core/types'
+import { atomFamily } from './atomFamily'
+
+export const waitForAll = atomFamily(
```
dai-shi, line 4: I would like to implement this without atomFamily, which does extra things.

src/utils/waitFor.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+)
+
+function concurrentRequests(
+  get: Getter,
+  dependencies: ReadonlyArray<Atom<unknown>>
+) {
+  const results = Array<unknown | undefined>(dependencies.length).fill(
+    undefined
+  )
```
dai-shi, line 34:
```suggestion
  const results = []
```
Shouldn't this be fine?

Thisen:
@dai-shi, It's important to know that this is very much WIP, I just opened a draft PR to get early feedback.

src/utils/waitFor.ts
Thisen: No, because I want to keep the order of the atom results, that's why I fill two arrays and then stitch them together at the end to keep the order.

dai-shi:
It was just some feedbacks I noticed at a glance. Let me know when it's ready for another look. I like the idea pretty much and see room for improvement, especially avoiding atomFamily.

dai-shi:
I merged #361 branch and applied the same technique in #363.

src/utils/waitForAll.ts
```diff
@@ -0,0 +1,65 @@
...
+      } catch (e) {
+        if (e instanceof Promise) {
+          promises[index] = e
+        } else {
+          throw e
+        }
+      }
+    })
+    if (promises.length) {
+      throw Promise.all(promises)
```
dai-shi, line 37:
```suggestion
      return Promise.all(promises)
```
It would be nice if we can do this, because it's more idiomatic. (throwing promise is something we should hide as implementation details.

dai-shi (approved):
Other than that, it's super cool.

Thisen:
@dai-shi Please review again :)

dai-shi:
@Thisen Would you test it with a real app? It seems a bit weird, but tests are passing.

Thisen:
> @Thisen Would you test it with a real app? It seems a bit weird, but tests are passing.

https://codesandbox.io/s/react-typescript-forked-krwsv?file=/src/App.tsx

dai-shi (approved):
I'm not 100% sure why the first test passes without infinite loop, but implementation-wise it's clean and we can always avoid the said infinite loop with useMemo. Let's merge this and see if someone has issues with real use cases.

PR #360, opened 2021-03-15
Refactor uMS to avoid extra re-renders

### LINKED ISSUE #355: `setAtom` doesn't bail on same value if component reads from another atom
`setAtom(value)` doesn't bail on same value provided the component also uses another atom (it doesn't behave exactly like I just described in 100% of cases, but I think this is the gist).

csb: https://codesandbox.io/s/setatom-bail-failure-59tcw?file=/src/App.tsx

- jotai: 0.15.2
- strict mode: both strict & non-strict mode

Aslemammad:
This PR fixes #355. Actually, when we call the write function on an atom, it will change it correctly. Still, as we use uMS emulation, for now, it causes an unexpected extra re-render for some atoms that hold the previous value, and they don't change, but uMS cause a re-render for that.

### DIFF
--- a/src/core/useMutableSource.ts
+++ b/src/core/useMutableSource.ts
@@ -38,11 +38,22 @@ export const useMutableSource = (
   if (
     state[0] !== source ||
     state[1] !== getSnapshot ||
-    state[2] !== subscribe ||
-    (currentVersion !== state[3] && currentVersion !== lastVersion.current)
+    state[2] !== subscribe
   ) {
     currentSnapshot = getSnapshot(source[TARGET])
-    if (currentSnapshot !== state[4]) {
+    setState([
+      /* [0] */ source,
+      /* [1] */ getSnapshot,
+      /* [2] */ subscribe,
+      /* [3] */ currentVersion,
+      /* [4] */ currentSnapshot,
+    ])
+  } else if (
+    currentVersion !== state[3] &&
+    currentVersion !== lastVersion.current
+  ) {
+    currentSnapshot = getSnapshot(source[TARGET])
+    if (!Object.is(currentSnapshot, state[4])) {
       setState([
         /* [0] */ source,
         /* [1] */ getSnapshot,
@@ -70,7 +81,7 @@ export const useMutableSource = (
           ) {
             return prev
           }
-          if (prev[4] === nextSnapshot) {
+          if (Object.is(prev[4], nextSnapshot)) {
             return prev
           }
           return [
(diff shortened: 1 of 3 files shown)


src/core/useMutableSource.ts
```diff
@@ -42,13 +42,15 @@ export const useMutableSource = (
...
   ) {
     currentSnapshot = getSnapshot(source[TARGET])
-    setState([
-      /* [0] */ source,
-      /* [1] */ getSnapshot,
-      /* [2] */ subscribe,
-      /* [3] */ currentVersion,
-      /* [4] */ currentSnapshot,
-    ])
+    if (currentSnapshot !== state[4]) {
```
Aslemammad, line 45: I couldn't find the right solution that stays outside of this file, but this let the tests pass.

dai-shi: Thanks for investigating. I know what this means because I did this once.
At that time I found another solution in vanilla.ts. Maybe, this one is not avoidable.
One thing for sure is we might not be using the "version" correctly.
In any case, we don't know the final behavior of uMS, but pretty sure it doesn't rely on setState bail out like this emulation.
So, the expectation is it will be fixed in the future.
Let me think if this fix makes sense for other cases.

Aslemammad: Yea, ok, if you found another solution, push it, no problem. it's a draft for now.

src/core/useMutableSource.ts
```diff
@@ -49,6 +48,20 @@ export const useMutableSource = (
       /* [3] */ currentVersion,
       /* [4] */ currentSnapshot,
     ])
+  } else if (
```
dai-shi, line 51: @Aslemammad This is more correct fix. But, it might not be important for the jotai usage.

src/core/useMutableSource.ts
```diff
@@ -68,7 +81,7 @@ export const useMutableSource = (
           ) {
             return prev
           }
-          if (prev[4] === nextSnapshot) {
+          if (Object.is(prev[4], nextSnapshot)) {
```
dai-shi, line 84: Unlike valtio, jotai's snapshot can be anything, so use `Object.is` instead of `===`.

src/core/useMutableSource.ts
Aslemammad: No problem, it's great.

dai-shi (approved):
Great work, @Aslemammad 
Please update the PR title and description, and then mark the PR non-draft.

Aslemammad:
Done.

dai-shi:
Cool. Let me merge this.

PR #361, opened 2021-03-15
feat: waitForAll

dai-shi:
#358 has some issues, most notably memory leaks, and here's my alternative.

(This is for reference only for `waitForAll`, we are still working on a better shape in total.)


dai-shi:
This branch is merged into #358

PR #363, opened 2021-03-17
feat: weak cache to avoid infinite loop without useMemo

dai-shi:
While discussing with @Thisen, I noticed that we can create useMemo alternative for derived atoms. @merisbahti said he wanted to use `splitAtom` in render without useMemo. So, here it is.

(I hope I didn't miss something fundamental...

### DIFF
--- a/src/optics/focusAtom.ts
+++ b/src/optics/focusAtom.ts
@@ -1,10 +1,9 @@
 import { atom } from 'jotai'
 import * as O from 'optics-ts'
-
-import { getWeakCacheItem, setWeakCacheItem } from '../utils/weakCache'
+import { getWeakCacheItem, setWeakCacheItem, Cache } from '../utils/weakCache'
 import type { WritableAtom, SetStateAction, PrimitiveAtom } from '../core/types'
 
-const focusAtomCache = new WeakMap()
+const focusAtomCache: Cache<WritableAtom<any, any>> = new WeakMap()
 
 const isFunction = <T>(x: T): x is T & Function => typeof x === 'function'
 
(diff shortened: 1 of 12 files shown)


tests/optics/focus-lens.test.tsx
```diff
@@ -10,10 +11,10 @@ const succ = (input: number) => input + 1
 
 it('basic derivation using focus works', async () => {
   const bigAtom = atom({ a: 0 })
-  const aAtom = focusAtom(bigAtom, (optic) => optic.prop('a'))
+  const propA = (optic: O.OpticFor<{ a: number }>) => optic.prop('a')
```
dai-shi, line 14: @merisbahti what's the convention to name this function in optics-ts?

merisbahti: its called a lens

src/optics/focusAtom.ts
```diff
@@ -35,6 +39,11 @@ export function focusAtom<S, A>(
   | WritableAtom<A | undefined, SetStateAction<A>>
   | WritableAtom<Array<A>, SetStateAction<A>>
   | PrimitiveAtom<S> {
+  const deps: object[] = [baseAtom, callback]
+  const cachedAtom: any = getWeakCacheItem(focusAtomCache, deps)
```
Thisen, line 43: Can we improve the typings here?

tests/optics/focus-lens.test.tsx
dai-shi: What exactly would you name this function instead of `propA`.

Thisen (approved):

src/optics/focusAtom.ts
dai-shi: Please help. The issue is more in `any` is L48. I spent some time and didn't find a good way.

tests/optics/focus-lens.test.tsx
merisbahti: Oh, I think I would call it a `focus`, or `focusFunction`, or `focusCallback`.

merisbahti:
Suggested changes:
https://github.com/pmndrs/jotai/pull/365

merisbahti (approved):

PR #368, opened 2021-03-19
Update the api/core docs

Aslemammad:
I edited some parts of the file, and it's more doc-ish and less introduction. Feel free to review.


docs/api/core.md
```diff
@@ -2,7 +2,9 @@ This doc describes core `jotai` bundle.
 
 ## atom
 
-```ts
+`atom` is an object where we save data there. It's immutable and shouldn't get mutated/changed directly.
```
dai-shi, line 5: we want to clarify the `atom()` creator function, and the created atom config object. Provider holds "atom value" and atom config doesn't hold any values (thus can be used for multiple providers.) We should clarify the concept in `docs/introduction/concepts.md`.

docs/api/core.md
```diff
@@ -22,45 +24,32 @@ function atom<Value, Update>(
...
+- `initialValue`: as the name says, it's an initial value which is the atom's going to return unless its value doesn't get changed.
+- `read`: a function that's going to get called on every re-render. The signature of `read` is `(get) => Value | Promise<Value>`, and `get` is a function that takes an atom config and returns its value stored in Provider described below. Dependency is tracked, so if `get` is used for an atom at least once, the `read` will be reevaluated whenever the atom value is changed.
+- `write`: a function mostly used for mutating atom's values, for a better description; it gets called whenever we call the second returned value of `useAtom`, the `useAtom[1]`. The default value of this function in the primitive atom will change the value of that atom. The signature of `write` is `(get, set, update) => void | Promise<void>`. `get` is similar to the one described above, but it doesn’t track the dependency. `set` is a function that takes an atom config and a new value which then updates the atom value in Provider. `update` is an arbitrary value that we receive from the updating function returned by `useAtom` described below.
```
dai-shi, line 29:
```suggestion
- `write`: a function mostly used for mutating atom's values, for a better description; it gets called whenever we call the second returned value of `useAtom`, the `useAtom()[1]`. The default value of this function in the primitive atom will change the value of that atom. The signature of `write` is `(get, set, update) => void | Promise<void>`. `get` is similar to the one described above, but it doesn’t track the dependency. `set` is a function that takes an atom config and a new value which then updates the atom value in Provider. `update` is an arbitrary value that we receive from the updating function returned by `useAtom` described below.
```

docs/api/core.md
```diff
@@ -69,10 +58,9 @@ anAtom.onMount = (setAtom) => {
 }
 ```
 
-Invoking `setAtom` function will invoke the atom's `writeFunction`.
-Customizing `writeFunction` allows changing the behavior.
+Invoking `setAtom` function will invoke the atom’s `writeFunction`. Customizing `writeFunction` allows changing the behavior.
```
dai-shi, line 61:
```suggestion
Invoking `setAtom` function will invoke the atom’s `write`. Customizing `write` allows changing the behavior.
```

dai-shi (changes requested):
The first line sounds incorrect.

Wouldn't it be nice to add "Terminology" section?

docs/api/core.md
dai-shi:
```suggestion
`atom` is a function to create an atom config. The atom config is an immutable object. The atom config itself doesn't hold an atom value. The atom value is stored in a Provider state.
```

dai-shi (approved):
Let's merge this and continue improvements in follow-up PRs.

Issue #375, opened 2021-03-23
value from atom is not update and component is not rendered (v0.15.3)


caltamiranob:
Hi everybody, i hope you can help me with the following problem beacuse i don't know what i'm doing bad.

I have a "store.js" file where i have all my atoms for specific module:

```javascript
export const loading = atom(false);
export const document = atom(undefined);
export const loadDocument = atom(
    null,
    (_get, set, documentId) => {
        const fetch = async () => {
            try {
                set(loading, true);
                const response = await DocumentService.findById(documentId);
                log(response); //document is loaded ok
                set(document, response);
            } catch (error) {
                console.error(error);
            } finally {
                set(loading, false);
            }
        }
        fetch();
    }
);
```

In another file i have my component where use atoms, when this component load the the initial values from atom are: loading: false and document: undefined; next invoke to the loadDocument method (this method update the document state) in the useEffect hook, but my component continues showing "loading..." text, this happens because my _document state_ is not update and my component is not re-rendered. To make sure, i put a useEffect for _document state_, and this method just show in console _"document undefined"_. 

```javascript
import * as atoms from "./store";

export function ListPage() {
    const { documentId } = useParams(); //this value is get it from url
    const [loading, setLoading] = useAtom(atoms.loading);
    const [document] = useAtom(atoms.document);
    const [, loadDocument] = useAtom(atoms.loadDocument);

    useEffect(() => {
        loadDocument(documentId);
    }, []);

    useEffect(() => {
         log("document", document);
    }, [document]);

    return (
        <Provider>
            { loading && <h1>Loading...</h1>}
            { !loading && <h1>{document.name}</h1>}
       </Provider>
    )
}
```
When I changed the value of _document atom_ in the loadDocument method, my component should be re-rendered? is the right behavior? or I am doing wrong in my code?
Thanks for your support.

dai-shi:
The code looks good and the behavior sounds weird.
Would you be able to create a codesandbox reproduction so that we can dig into it?
Thanks for your help.

caltamiranob:
Hi @dai-shi, I share the sandbox:

https://codesandbox.io/s/test-jotai-async-method-41oyb

Thanks for your help.

dai-shi:
Confirmed this is a bug in v0.15.3.

It works ok in v0.15.2.
https://codesandbox.io/s/test-jotai-async-method-forked-4z27y

dai-shi:
#373 might be related. The root cause could be the same.

[...]

PR #376, opened 2021-03-24
Fix issue #373 - Infinite loop

### LINKED ISSUE #373: "Too many re-renders." when I set value to atom in useEffect (v0.15.3)
Thank you for awesome library!

I updated Jotai v0.15.2 to v0.15.3. in my project, and it starts to fall.
The error message is "Too many re-renders. React limits the number of renders to prevent an infinite loop.".

I uploaded the reproduction project to https://github.com/wancup/jotai-re-render-error-test. (It's ok in v0.15.2)
Main file is https://github.com/wancup/jotai-re-render-error-test/blob/main/src/App.js 

I think it's occurred by using an atom derived other derived atom and set a value in useEffect.

Is it a Jotai's bug?

Thisen:
Fixes #373


dai-shi (approved):
Looks perfect! Great work!

Thisen:
This won't fix #375 - I can look into that later though.

dai-shi:
Thanks! Let me merge this now then.

PR #377, opened 2021-03-24
[docs] No suspense

### LINKED ISSUE #269: [Question] Can jotai be used without Suspense?
In @JLarky 's [article](https://blog.logrocket.com/jotai-vs-recoil-what-are-the-differences/), it notes:

> The first set of features that stands out are functions that help to work with async selectors. For example, if you want to work with an async atom in a context where Suspense would not be appropriate, you can always consume it with useRecoilStateLoadable or useRecoilValueLoadable so it will no longer throw errors or promises.

The question is if we can use jotai without Suspense. Actually, we can under some constraints.

There are two parts it requires Suspense.

- Derived async atoms (async `read`)
image

- Async actions (async `write`)
image

---

The latter is easier.

The original code is this:
```js
const fetchCountAtom = atom(
  get => get(countAtom),
  async (_get, set, url) => {
    const response = await fetch(url)
    set(countAtom, (await response.json()).count)
  }
)
```
The `write` function (2nd arg) returns a promise, which will suspend.

So, We just don't return anything.
The modified code is this:
```js
const fetchCountAtom = atom(
  get => get(countAtom),
  (_get, set, url) => {
    const fetchData = async () => {
      const response = await fetch(url)
      set(countAtom, (await response.json()).count)
    }
    fetchData()
  }
)
```

---

Okay, the difficult one is async read.
The original code is this:
```js
const urlAtom = atom("https://json.host.com")
const fetchUrlAtom = atom(
  async get => {
    const response = await fetch(get(urlAtom))
    return await response.json()
  }
)
```

We can't make `read` function to return a promise.
So, we have to change the structure.
Here's the new code:
```js
const fetchResultAtom = atom({ loading: true, error: null, data: null })
const runFetchAtom = atom(null, (_get, set, url) => {
  const fetchData = async () => {
    set(fetchResultAtom, prev => ({ ...prev, loading: true }))
    try {
      const response = await fetch(url)
      const data = await response.json()
      set(fetchResultAtom, { loading: false, error: null, data })
    } catch (error) {
      set(fetchResultAtom, { loading: false, error, data: null })
    }
  }
  fetchData()
})

const Component = () => {
  const [, runFetch] = useAtom(runFetchAtom)
  useEffect(() => {
    runFetch("https://json.host.com")
  }, [])
```
Yes, we need useEffect for initialization. This would be mitigated once atom effects #211 lands.

Aslemammad:
close #269

### DIFF
--- a/docs/guides/no-suspense.md
+++ b/docs/guides/no-suspense.md
@@ -2,7 +2,7 @@
 
 > The content of this page would be easier if you have taken a look on the _introduction/async_ part of the docs.
 
-As you may know, sometimes we need to wrap our components in suspense to use asynchronous atoms. Two kinds of atoms need Suspense.
+Sometimes we need to wrap our components in suspense to use asynchronous atoms. Two kinds of atoms need Suspense.
 
 - Derived async atoms (async `read` function)
 - Asynchronous actions (async `write`)
--- a/docs/guides/no-suspense.md
+++ b/docs/guides/no-suspense.md
@@ -1,6 +1,6 @@
 # No Suspense
 
-> The content of this page would be easier if you have taken a look on the _introduction/async_ part of the docs.
+> The content of this page would be easier if you have taken a look on the [introduction/async](../introduction/async.md) part of the docs.
 
 Sometimes we need to wrap our components in suspense to use asynchronous atoms. Two kinds of atoms need Suspense.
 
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Looks great! Thanks for your work!
I will merge this tomorrow.

docs/guides/no-suspense.md
```diff
@@ -0,0 +1,87 @@
+# No Suspense
+
+> The content of this page would be easier if you have taken a look on the _introduction/async_ part of the docs.
+
+As you may know, sometimes we need to wrap our components in suspense to use asynchronous atoms. Two kinds of atoms need Suspense.
+
+- Derived async atoms (async `read` function)
+- Asynchronous actions (async `write`)
+
+But sometimes we're not allowed to wrap our components in Suspense, So we're going to do some hacks below and and make our async atoms syncronous.
```
Thisen, line 10: Typo
```suggestion
But sometimes we're not allowed to wrap our components in Suspense, So we're going to do some hacks below and and make our async atoms synchronous.
```

docs/guides/no-suspense.md
```diff
@@ -0,0 +1,87 @@
+# No Suspense
+
+> The content of this page would be easier if you have taken a look on the _introduction/async_ part of the docs.
+
+As you may know, sometimes we need to wrap our components in suspense to use asynchronous atoms. Two kinds of atoms need Suspense.
```
Thisen, line 5: Maybe remove the "As you may know", It doesn't provide value to the explanation.
```suggestion
Sometimes we need to wrap our components in suspense to use asynchronous atoms. Two kinds of atoms need Suspense.
```

Thisen (commented):
Looks good! Got a few suggestions.

PR #378, opened 2021-03-24
build: use esbuild and build concurrently

aulneau:
Last night I was playing around with building and running the project and I noticed it was taking a long time to build (around 100 seconds). 

This PR replaces the typescript rollup plugin with the [esbuild rollup plugin](https://github.com/egoist/rollup-plugin-esbuild) (esbuild is amazing and I love it). 

I've also broken out each build task into its own, and I'm running them concurrently. This brings the total build time on my machine from ~90-100 seconds to ~14 seconds. I think having the different exports as different build tasks is useful if someone is working on just one of the exports locally, they don't have to rebuild the whole project.

I've tested the package in a few of my local projects and nothing seems broken, but it would be good to do some more thorough testing in non ESM based apps.


package.json
```diff
@@ -57,7 +57,14 @@
   "sideEffects": false,
   "scripts": {
     "prebuild": "shx rm -rf dist",
-    "build": "rollup -c && rollup -c --config-utils && rollup -c --config-devtools && rollup -c --config-immer && rollup -c --config-optics && rollup -c --config-query && rollup -c --config-xstate",
+    "build": "concurrently 'yarn build:base' 'yarn build:utils' 'yarn build:devtools' 'yarn build:immer' 'yarn build:optics' 'yarn build:query' 'yarn build:xstate'",
```
dai-shi, line 60: Can we do something like `build:*`?
https://www.npmjs.com/package/concurrently 👀

rollup.config.js
```diff
@@ -32,7 +34,13 @@ function createESMConfig(input, output) {
     external,
     plugins: [
       resolve({ extensions }),
-      typescript(),
+      esbuild({
+        minify: process.env.NODE_ENV === 'production',
```
dai-shi, line 38: I'm not sure where this NODE_ENV is coming from. Is this `true` when we run `yarn build`?

rollup.config.js
```diff
@@ -32,7 +34,13 @@ function createESMConfig(input, output) {
...
+        target: 'esnext',
+        jsxFactory: 'React.createElement',
+        jsxFragment: 'React.Fragment',
+        tsconfig: path.resolve('./tsconfig.json'),
```
dai-shi, line 42: I thought esbuild just strips type information. Do they follow some tsconfigs? (Out of curiosity...

rollup.config.js
```diff
@@ -32,7 +34,13 @@ function createESMConfig(input, output) {
...
+      }),
       babel(getBabelOptions({ node: 8 })),
```
dai-shi, line 44: Do we still need to run babel? Doesn't esbuild accept target build?

dai-shi (commented):
I'm not very familiar with esbuild. Let me know if some of my questions don't make sense.

rollup.config.js
aulneau: You're right, i've removed it

rollup.config.js
aulneau: yep, removed

package.json
aulneau: nice! i didn't know about that, much better.

rollup.config.js
aulneau: https://esbuild.github.io/content-types/#tsconfig-json

rollup.config.js
aulneau: Actually, if jotai needs to support ie11, we need to keep it

Thisen:
Looks cool, I have a lot of experience with Webpack and Rollup, but not with ESBuild.

I got a couple of questions:

- Is it expected that the bundled size drops so significantly?
- How is type checking run?
- What's the rationale for splitting up the build for each unit?

aulneau:
> Is it expected that the bundled size drops so significantly?

I was surprised by this too! I think if we were to use a bundle analyzer, I'd bet that you'd find that there were a lot of things being bundled that should not have been. I saw warning around `babelhelpers` being bundled by default due to it not being explicitly set. I've changed that now to use `runtime` which reduced the overall size a bit. I don't have a good answer, but tests pass and I've loaded a few of my personal apps with this version and everything works.

> How is type checking run?

[This task will fail](https://github.com/pmndrs/jotai/blob/c3d685195b9db5ddb7c6f27335a402444033bd5f/package.json#L72) if type checking does not succeed, and we're still using `@rollup/plugin-typescript` to [emit declarations](https://github.com/pmndrs/jotai/blob/c3d685195b9db5ddb7c6f27335a402444033bd5f/rollup.config.js#L69). If the types are incorrect, this will fail.

> What's the rationale for splitting up the build for each unit?

This was here before I did any change, I didn't want to refactor this as it seems to be working fine. Maybe @dai-shi can comment on it.

aulneau:
> Is it expected that the bundled size drops so significantly?

Actually looking at this further, there is actually not that much size difference, it's just the output from the esbuild plugin is minified, where as previously it was not. The differences that are most relevant for "true size difference" would be to compare the `gzipped` key between each version.

rollup.config.js
dai-shi: > Only certain tsconfig.json fields are respected

I see. Thanks.

rollup.config.js
dai-shi: that makes sense.

dai-shi:
> What's the rationale for splitting up the build for each unit?

This is a bit controversial, and I hope you can just accept it for now.
We follow this from zustand and the main purpose is to explicitly split non-core utils from core.

dai-shi:
> it's just the output from the esbuild plugin is minified

Oh, yeah, we didn't minify before. Let's remove `minify: true` to keep consistency for now.

rollup.config.js
```diff
@@ -2,17 +2,25 @@ import path from 'path'
...
 const external = (id) => !id.startsWith('.') && !id.startsWith(root)
+
 const extensions = ['.js', '.ts', '.tsx']
 const getBabelOptions = (targets) => ({
   ...createBabelConfig({ env: (env) => env === 'build' }, targets),
   extensions,
 })
+function getEsbuild(target) {
+  return esbuild({
+    minify: true,
```
dai-shi, line 19:
```suggestion
```

Let's remove this.

dai-shi:
@aulneau Would love to hear you thoughts and plans on these:


>     - (like you said) some more thorough testing would be good
>     - Isn't it possible to compare the build output with and without this PR? Is it totally different?
>     - If we do this, we would like to do the same (incl. concurrently) for zustand and valtio, to keep the dev tool consistent

aulneau:
> > What's the rationale for splitting up the build for each unit?
> 
> This is a bit controversial, and I hope you can just accept it for now.
> We follow this from zustand and the main purpose is to explicitly split non-core utils from core.

I personally like this convention fwiw.

> Isn't it possible to compare the build output with and without this PR? Is it totally different?

I've made a temp commit to my fork here where it creates a diff of the output -> https://github.com/aulneau/jotai/commit/3c63f9f841ee26f838be3803bcbd0b0400cbff40

> - If we do this, we would like to do the same (incl. concurrently) for zustand and valtio, to keep the dev tool consistent

Sure, that makes total sense :) would you want me to make the PRs?

> - (like you said) some more thorough testing would be good

I'm not sure the best way, I've tested in multiple apps of my own, can we get others to do so as well?

dai-shi:
> I personally like this convention fwiw.

👍 

> I've made a temp commit to my fork here where it creates a diff of the output -> [aulneau@3c63f9f](https://github.com/aulneau/jotai/commit/3c63f9f841ee26f838be3803bcbd0b0400cbff40)

Thanks! I see some differences around object spread and iterators.
This helps, but still doesn't assure everything...

> > * If we do this, we would like to do the same (incl. concurrently) for zustand and valtio, to keep the dev tool consistent
> 
> Sure, that makes total sense :) would you want me to make the PRs?

Yes, please do that right away. I mean I would like to see them before merging this. (I don't mean any rush, please take your time.)

> > * (like you said) some more thorough testing would be good
> 
> I'm not sure the best way, I've tested in multiple apps of my own, can we get others to do so as well?

hm, once PRs are ready for all three, let's ask in discord and see if someone can help.

aulneau:
> hm, once PRs are ready for all three, let's ask in discord and see if someone can help.

I've made PRs in both:

https://github.com/pmndrs/zustand/pull/343
https://github.com/pmndrs/valtio/pull/129

rollup.config.js
```diff
@@ -46,7 +52,7 @@ function createCommonJSConfig(input, output) {
     external,
     plugins: [
       resolve({ extensions }),
-      typescript(),
+      getEsbuild('es2015'),
```
dai-shi, line 55:
```suggestion
```

according to https://github.com/pmndrs/zustand/pull/343#issuecomment-808745679 @aulneau

aulneau: fixed!

aulneau:
@dai-shi I've removed the esbuild step for the cjs process and rebased my branch. Should be ready for merge. The only difference I saw between using the esbuild/typescript step with cjs is that it stripped comments, so I added [this line](https://github.com/pmndrs/jotai/pull/378/files#diff-6814bf77564b4f1c92f5861e184e28fe217c080a047fefa8b73a728f755ec45cR18) to have babel do the same.

rollup.config.js
```diff
@@ -2,18 +2,31 @@ import path from 'path'
...
+const external = (id) =>
+  (!id.startsWith('.') && !id.startsWith(root)) ||
+  id.startsWith('@babel/runtime')
+
 const extensions = ['.js', '.ts', '.tsx']
 const getBabelOptions = (targets) => ({
   ...createBabelConfig({ env: (env) => env === 'build' }, targets),
   extensions,
+  comments: false,
+  babelHelpers: 'runtime',
```
dai-shi, line 19: what's the reason we need this? we don't use any async/await.

rollup.config.js
```diff
@@ -2,18 +2,31 @@ import path from 'path'
...
 
+function getEsbuild(target) {
+  return esbuild({
+    minify: false,
+    target,
+    tsconfig: path.resolve('./tsconfig.json'),
+  })
+}
+
 function createDeclarationConfig(input, output) {
```
dai-shi, line 30: Just noticed, we have function declarations and arrow functions mixed in rollup config. (for all z/j/v)
It's not super important, but it shows some inconsistencies while reviewing. We can make all arrow functions either in this PR or in follow-up PR.

package.json
```diff
@@ -127,10 +137,13 @@
       "tests/**/*.{js,ts,tsx}"
     ]
   },
-  "dependencies": {},
+  "dependencies": {
+    "@babel/runtime": "^7.13.10"
+  },
```
dai-shi, line 142:
```suggestion
  "dependencies": {},
```

I spent quite some effort to remove dependencies, so let's remove @babel/(plugin-transform-)runtime.

rollup.config.js
aulneau: Without it babel helpers would be included by default as the `bundled` setting (vs runtime). See here https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers. it gives this message: 

> It is recommended to configure this option explicitly (even if with its default value) so an informed decision is taken on how those babel helpers are inserted into the code.

I set it explicitly, and set it to `runtime` as I think that is the smallest footprint. I don't think helpers are only for async functions, but I might be wrong!

aulneau: Perhaps we could disable helpers all together?

dai-shi: I might be misunderstanding. Let me play with it a bit.

dai-shi: Okay, I misunderstood something. so the default is `bundled` and it's not recommended.

So, let's say we use `runtime` as recommended. My questions are:

- Can you eliminate it from dependenceis, as zustand doesn't have it?
  - My feeling is "no", in this case, we should put it in peerDependencies with optional meta, for all zustand/jotai/valtio
- Don't we need `@babel/plugin-transform-runtime` here as zustand has it?

dai-shi: I did some more experiments. Here's the cjs bundle size (not minified.)

| | `'runtime'` | `'bundled'` | diff |
|---|---|---|---|
|index.js|22123|22349|+226|
|utils.js|13887|12907|-908|
|devtools.js|2395|2395|0|
|immer.js|1964|1964|0|
|optics.js|1868|1868|0|
|query.js|3291|3291|0|
|xstate.js|3074|3062|-12|

So, given this resulting size, we could actually go with`bundled`, which is far less troublesome.

dai-shi: (btw, would it be possible for you to DM me in Discord? Just got a bit tired going back and forth in these three repos.)

dai-shi (approved):

PR #379, opened 2021-03-25
feat(util): atomWithDefault

### LINKED ISSUE #352: Writeable Atom with derived default value
I just recently went through a refactoring from recoil. Everything was pretty straightforward but one use case I couldn't build in jotai.

I have a list of countries I can fetch from the backend:

```js
const availableCountriesApi = atom<Array<Country>>(async () => fetchEntities("/core/countries"))
```

I built a defaultCountry atom based on this fetch

```js
const defaultCountry = atom<Country>(
  (get) => get(availableCountries).find(country => country.id === "US") || {id: "US", name: "United States"},
)
```

And now I have a `selectedCountry` atom which I want to initialize with the default country. But everything I tried resulted either in a type error or a readable only atom:

```js
const selectedCountryState = atom<Country>((get) => get(defaultCountry))
 
const [selectedCountry, setSelectedCountry] = useAtom(selectedCountryState)

setSelectedCountry({id: "DE", name: "Germany"}) // This expression is not callable, Type 'never' has no call signatures.
```

I think the async fetch is not super relevant, because when I change my defaultCoutry atom to this, I'll run in the same problem:

```js
const defaultCountry = atom<Country>({id: "US", name: "United States"})
```

So is there a way to build a writeable atom that I can initialize with a derived atom value?

dai-shi:
Close #352 

- [x] impl
- [x] tests
- [x] docs


src/utils/atomWithReset.ts
```diff
@@ -6,7 +6,7 @@ export const RESET = Symbol()
 
 export function atomWithReset<Value>(initialValue: Value) {
   type Update = SetStateAction<Value> | typeof RESET
-  const anAtom: any = atom<Value, Update>(initialValue, (get, set, update) => {
+  const anAtom = atom<Value, Update>(initialValue, (get, set, update) => {
```
dai-shi, line 9: just found we don't need any here.

docs/api/utils.md
```diff
@@ -180,6 +180,22 @@ const countReducerAtom = atomWithReducer(0, countReducer)
...
+Ref: https://github.com/pmndrs/jotai/issues/352
+
+This is a function to create a primitive atom.
+Its default value can be specified with a read function instead of a static initial value.
+
+```js
+import { atomWithDefault } from 'jotai/utils'
+
+const count1Atom = atom(1)
+const count2Atom = atomWithDefault((get) => get(count1Atom) * 2)
```
Thisen, line 194: Question:
Is it possible to add a variant, where you can add a default value in atom? Could be useful, if you want to base the default value on props.

src/utils/atomWithDefault.ts
```diff
@@ -0,0 +1,26 @@
...
+      if (get(overwrittenAtom)) {
+        return get(anAtom)
+      }
+      return getDefault(get)
+    },
+    (get, set, update) => {
+      set(overwrittenAtom, true)
+      set(
+        anAtom,
+        typeof update === 'function'
```
Thisen, line 19: Can it be a promise with this impl?

tests/utils/atomWithDefault.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  fireEvent.click(getByText('button1'))
+  await findByText('count1: 3, count2: 5')
+})
+
+it('simple async get default', async () => {
+  const count1Atom = atom(1)
+  const count2Atom = atomWithDefault(async (get) => {
+    await new Promise((r) => setTimeout(r, 10))
+    return get(count1Atom) * 2
```
Thisen, line 48: Maybe a test missing with async `set`

docs/api/utils.md
dai-shi: I'm not sure if I get "add a default value in atom".
Could you write some rough idea/code?

src/utils/atomWithDefault.ts
dai-shi: I don't think so. The created atom is just like a normal primitive atom.

tests/utils/atomWithDefault.test.tsx
dai-shi: Hadn't thought about it. Would it be possible for you to add such a test?

docs/api/utils.md
Thisen:
```
const DogInfo: FC = (props) => {
  const dogInfo = useAtom(dogAtom(props.dogName))
  // ...
}
```

tests/utils/atomWithDefault.test.tsx
Thisen: Sure, I'll do it

src/utils/atomWithDefault.ts
Thisen: Is that acceptable?

aulneau:
This is a great helper, I was wondering if it makes sense to have `atomFamilyWithDefault`, too?

src/utils/atomWithDefault.ts
dai-shi: I mean it should be the same as the normal `atom(...)`. As far as I remember, if we set atom with a promise, it works for some case, but had a limitation or pitfall. At least the type doesn't allow it, so we are not currently supporting such a pattern.

docs/api/utils.md
dai-shi: Ah, I want to support this pattern with this:
```js
const DogInfo = ({ dogName }) => {
  const dogAtom = useMemo(() => atom(dogName), [dogName])
  const dogInfo = useAtom(dogAtom)
  // ...
}
```

Oh, I see you want to give the default that can be overwritten.
```js
const DogInfo = ({ dogName }) => {
  const dogAtom = useRef(atom(dogName)).current
  const dogInfo = useAtom(dogAtom)
  // ...
}
```
I'm not a big fan of this useRef pattern.

```js
const DogInfo = ({ dogName }) => {
  const dogAtom = useMemo(() => atom(null), [])
  const dogInfo = useAtom(dogAtom) || dogName
  // ...
}
```

There might be some better ideas. I feel like using atomWithDefault for props is a misuse.

dai-shi:
> This is a great helper, I was wondering if it makes sense to have `atomFamilyWithDefault`, too?

tbh, i'm not very happy with it. we have atomWithRest too. we may have more helpers in the future.
I now think it's better if atomFamily takes atom creator instead of read/write.
```js
const myFamily = atomFamily(param => atom(...))
// then this is possible
const myFamily = atomFamily(param => atomWithDefault(...))
```

what do you think? worth a breaking change before v1?

aulneau:
> what do you think? worth a breaking change before v1?

I do think so. 

I have noticed that for any of the utility atoms, I've been needing to wrap them with other kinds. such as wishing i could do `atomFamilyWithQuery`. I could also see a need for something like this `atomFamilyWithDefaultWithQuery` lol (naming just for functional example)

dai-shi:
Okay, the type would be much better too. Let's do that.
Would you or anyone like to work on it? I'm happy to take it if nobody raises a hand.
Maybe we want to create a dedicated issue at first.

dai-shi:
> Maybe we want to create a dedicated issue at first.

Done: #385

tests/utils/atomWithDefault.test.tsx
dai-shi: I first thought it's a neat idea, but then this doesn't work... unfortunately. async just works more or less accidentally. the `write` may not usually write self (though technically possible except for types), so this is counter-idiomatic too.
Thanks for your work anyway. I got an idea to improve the impl. (I did something weird, to try something else.)

Thisen (approved):

dai-shi:
Thanks for the review. I will merge this later with others.

PR #380, opened 2021-03-25
Fix #375 - Atom state is not flushed when using sync invoked async setter

### LINKED ISSUE #375: value from atom is not update and component is not rendered (v0.15.3)

Thisen:
Fixes #375

### DIFF
--- a/tests/async.test.tsx
+++ b/tests/async.test.tsx
@@ -581,3 +581,47 @@ it('a derived atom from a newly created async atom (#351)', async () => {
   await findByText('loading')
   await findByText('derived: 13, commits: 3')
 })
+
+it('(#375)', async () => {
+  const loadingAtom = atom(false)
+  const documentAtom = atom<string | undefined>(undefined)
+  const loadDocumentAtom = atom(null, (_get, set) => {
+    const fetch = async () => {
+      set(loadingAtom, true)
+      const response = await new Promise<string>((resolve) =>
+        setTimeout(() => resolve('great document'), 10)
+      )
+      set(documentAtom, response)
+      set(loadingAtom, false)
+    }
+    fetch()
+  })
+
+  const ListDocuments: React.FC = () => {
+    const [loading] = useAtom(loadingAtom)
+    const [document] = useAtom(documentAtom)
+    const [, loadDocument] = useAtom(loadDocumentAtom)
+
+    useEffect(() => {
+      loadDocument()
+    }, [loadDocument])
+
+    return (
+      <>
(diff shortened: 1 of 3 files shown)


tests/async.test.tsx
```diff
@@ -581,3 +581,47 @@ it('a derived atom from a newly created async atom (#351)', async () => {
   await findByText('loading')
   await findByText('derived: 13, commits: 3')
 })
+
+it('(#375)', async () => {
```
dai-shi, line 585: Is it possible to have some descriptive name?

Thisen:
Your changes looks good @dai-shi - Thanks!

dai-shi (approved):
I'm not 100% comfortable with the current fix in terms of coding, which has some implications. Let us revisit this later when we have a chance.
We have the test now. Thanks @Thisen for great contribution!

PR #386, opened 2021-03-27
feat: add equality fn to atomWithQuery

aulneau:
This takes the same `equalityFn` signature from `selectAtom` and adds it to the `atomWithQuery` helper. I have found this very helpful in my apps to help control rerenders while the query revalidates. Currently, if the query is set to refetch on an interval, when it sets the data atom again it will cause any component that uses the data to rerender even if nothing in the data has changed.

Happy to explore other options, but this works for me currently.

### DIFF
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -30,7 +30,8 @@ export function atomWithQuery<
     | AtomQueryOptions<TQueryFnData, TError, TData, TQueryData>
     | ((
         get: Getter
-      ) => AtomQueryOptions<TQueryFnData, TError, TData, TQueryData>)
+      ) => AtomQueryOptions<TQueryFnData, TError, TData, TQueryData>),
+  equalityFn: (a: TData, b: TData) => boolean = Object.is
 ): WritableAtom<TData, ResultActions> {
   const pendingAtom = atom(createPending<TData>())
   const dataAtom = atom<TData | null>(null)
@@ -60,7 +61,10 @@ export function atomWithQuery<
             }
             action.initializer(getQueryClient(get, set))
           } else if (action.type === 'data') {
-            set(dataAtom, action.data)
+            const data = get(dataAtom)
+            if (!data || !equalityFn(data, action.data)) {
+              set(dataAtom, action.data)
+            }
             const pending = get(pendingAtom)
             if (!pending.fulfilled) {
               pending.resolve(action.data)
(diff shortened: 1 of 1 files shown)


dai-shi:
As we discussed in discord, I'm not a big fan of equalityFn. But, this sounds unavoidable.

> Currently, if the query is set to refetch on an interval, when it sets the data atom again it will cause any component that uses the data to rerender even if nothing in the data has changed.

As this means, react-query returns an different object with the same content.

src/query/atomWithQuery.ts
```diff
@@ -60,7 +61,10 @@ export function atomWithQuery<
...
             action.initializer(getQueryClient(get, set))
           } else if (action.type === 'data') {
-            set(dataAtom, action.data)
+            const data = get(dataAtom)
+            if (!data || !equalityFn(data, action.data)) {
+              set(dataAtom, action.data)
+            }
             const pending = get(pendingAtom)
             if (!pending.fulfilled) {
               pending.resolve(action.data)
```
dai-shi, line 70: It's probably better to keep the pair of dataAtom and pendingAtom in sync.
```suggestion
            const data = get(dataAtom)
            if (!data || !equalityFn(data, action.data)) {
              set(dataAtom, action.data)
              const pending = get(pendingAtom)
              if (!pending.fulfilled) {
                pending.resolve(action.data)
              }
```

src/query/atomWithQuery.ts
```diff
@@ -60,7 +61,10 @@ export function atomWithQuery<
             }
             action.initializer(getQueryClient(get, set))
           } else if (action.type === 'data') {
-            set(dataAtom, action.data)
+            const data = get(dataAtom)
+            if (!data || !equalityFn(data, action.data)) {
```
dai-shi, line 65:
```suggestion
            if (data === null || !equalityFn(data, action.data)) {
```
Let's make it strict here. (tbh, i'm not certain if `null` in L37 is valid to mean uninitialized. Can `TData` ever be `null`?)

dai-shi:
@Aslemammad thoughts?

aulneau:
> As this means, react-query returns an different object with the same content.

Yeah, I think it does. I know they have this feature: [`notifyOnChangeProps`](https://react-query.tanstack.com/guides/migrating-to-react-query-3#the-queryoptionsnotifyonstatuschange-option-has-been-superceded-by-the-new-notifyonchangeprops-and-notifyonchangepropsexclusions-options) however when I have tried to enable it, it breaks completely 🤷

dai-shi (approved):
LGTM

Would you like to add a note about equalityFn in docs?

dai-shi:
Merging this now.
Please open follow-up PRs to improve docs.

PR #390, opened 2021-03-29
Fix flushing state on delayed write (for self atom)

### LINKED ISSUE #389: Write function with delayed call to set
Hi,

I have an atom that I want to automatically be reset after a certain amount of time each time it is set.
I figured I could use a custom write function; something like this:
```javascript
const someAtom = atom(
	undefined,
	(get, set, update) => {
		set(someAtom, 'someData')
		window.setTimeout(() => set(someAtom, undefined) }, 5000)
	}
);
```
But this does not seem to work.
The second call to `set` has no effect. I suppose it has to do with the `setTimeout`.
Is there a simple way to achieve what I want?

Thanks for this library, I'm really enjoying it so far!

dai-shi:
close #389 

related #380 cc @Thisen

### DIFF
--- a/tests/async.test.tsx
+++ b/tests/async.test.tsx
@@ -626,6 +626,40 @@ it('Handles synchronously invoked async set (#375)', async () => {
   await findByText('great document')
 })
 
+it('async write self atom', async () => {
+  const countAtom = atom(0, async (get, set, _arg) => {
+    set(countAtom, get(countAtom) + 1)
+    await new Promise((r) => setTimeout(r, 100))
+    set(countAtom, -1)
+  })
+
+  const Counter: React.FC = () => {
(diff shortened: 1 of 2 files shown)


tests/async.test.tsx
```diff
@@ -625,3 +625,70 @@ it('Handles synchronously invoked async set (#375)', async () => {
   await findByText('loading')
   await findByText('great document')
 })
+
+it('async write self atom', async () => {
```
Thisen, line 629: I think we are missing two test:

- A sync write self atom (in basic.test.tsx ofc)
- Error handling for self write - What's the expected value of the atom after you have handled the error boundary?

dai-shi:
Merging this. It will be released after some more issues/PRs are resolved.

PR #392, opened 2021-03-30
New atom family implementation

### LINKED ISSUE #385: Generalize `atomFamily` (BREAKING CHANGE)
We noticed our current `atomFamily` has an issue in its API design.
It doesn't allow to use other atom creation utils like `atomWithReset`.

Ref: https://github.com/pmndrs/jotai/pull/379#issuecomment-808281883

Let's change it to the generalized one before we finalize the api as v1.

Aslemammad:
close #385 

- [x] implementation
- [x] tests
- [x] docs

### DIFF
--- a/src/utils/atomFamily.ts
+++ b/src/utils/atomFamily.ts
@@ -1,4 +1,3 @@
-import { atomFamily } from 'jotai/utils'
 import { Atom, WritableAtom } from 'jotai'
 
 type ShouldRemove<Param> = (createdAt: number, param: Param) => boolean
--- a/src/utils/atomFamily.ts
+++ b/src/utils/atomFamily.ts
@@ -7,7 +7,6 @@ type AtomFamily<Param, AtomType> = {
   remove(param: Param): void
   setShouldRemove(shouldRemove: ShouldRemove<Param> | null): void
 }
-type AnyAtomType<Value, Update> = Atom<Value> | WritableAtom<Value, Update>
 
 export function atomFamily<Param, Value, Update>(
   initializeAtom: (param: Param) => WritableAtom<Value, Update>,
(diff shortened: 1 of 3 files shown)


src/utils/atomFamily.ts
```diff
@@ -1,3 +1,4 @@
+import { AnyAtom } from './../core/types'
```
dai-shi, line 1: Types are imported in L3.

src/utils/atomFamily.ts
```diff
@@ -11,53 +12,53 @@ type AtomFamily<Param, AtomType> = {
...
-  areEqual?: (a: Param, b: Param) => boolean
-): AtomFamily<Param, PrimitiveAtom<Value>>
+// // primitive atom
+// export function atomFamily<Param, Value>(
+//   initializeRead: (param: Param) => Value,
+//   initializeWrite?: null,
+//   areEqual?: (a: Param, b: Param) => boolean
+// ): AtomFamily<Param, PrimitiveAtom<Value>>
 
+type AnyAtomType<Value, Update> = Atom<Value> | WritableAtom<Value, Update>
```
dai-shi, line 56: Let's use function overload.

```ts
export function atomFamily<Param, Value>(
  initializeAtom: (param: Param) => Atom<Value>,
  areEqual?: (a: Param, b: Param) => boolean
): : AtomFamily<Param, Atom<Value>>

export function atomFamily<Param, Value, Update>(
  initializeAtom: (param: Param) => WritableAtom<Value, Update>,
  areEqual?: (a: Param, b: Param) => boolean
): AtomFamily<Param, WritableAtom<Value>>
```

dai-shi (approved):
Looks good! Thank for your work!

PR #394, opened 2021-03-30
feat(util): atomWithStorage

sandren:
There are examples of how to add persistence to individual atoms in the docs, but I would like to create a robust `atomWithStorage` utility that covers most use cases, including `localStorage` or `sessionStorage` for React as well as `AsyncStorage` for React Native. Also note that the stored value is intentionally populated in the `onMount` method rather than the base `atom()` initialization to support `atomWithStorage` usage in SSR contexts such as Next.js and Gatsby. The first commit is a work in progress.

**Progress**
* [x]  impl
* [x]  tests
* docs -> a separate pr


src/utils/atomWithStorage.ts
```diff
@@ -0,0 +1,31 @@
+import { atom, WritableAtom } from 'jotai'
+import type { SetStateAction } from '../core/types'
+
+export function atomWithStorage<Value>(key: string, initialValue: Value) {
+  type Update = SetStateAction<Value>
```
dai-shi, line 5: Do we accept function update? If so, L24-L26 should be implemented so. There's no magic.

dai-shi:
The WIP version looks actually good, and is worth a guide in docs.

To make if a flexible util, we want to abstract the storage api.
```ts
type Storage<Value> = {
  getItem: (key: string) => Value | Promise<Value>
  setItem: (key: string, newValue: Value) => void | Promise<void>
}

export function atomWithStorage<Value>(key: string, default: Value, getStorage?: () => Storage<Value>): PrimitiveAtom<Value>
```

src/utils/atomWithStorage.ts
sandren: Can you elaborate?

dai-shi: It has to be:
```ts
    (get, set, update: Update) => {
      const newValue = typeof update === 'function' ? update(get(baseAtom)) : update
      set(baseAtom, newValue)
      window.localStorage.setItem(key, JSON.stringify(newValue))
    }
```

dai-shi (approved):

ghmeec:
this is awesome, i haven't been able to test in RN environment, would there be a problem since getItem and setItem methods are async?

dai-shi:
@ghmeec We hoped this would support AsyncStorage, but not thoroughly tested. Would you give it a try?

thinkdj:
Often, we need to store sensitive data like a token or basic authentication info for persistence, and storing such data plainly to localStorage seems less secure. Provision for a custom encode/decode method would be nice to have.

thinkdj:
Often, we need to store sensitive data like a token or basic authentication info for persistence, and storing such data plainly to localStorage seems less secure. Provision for a custom hook for encrypt/decrypt would be great to see

dai-shi:
You should be able to customize `storage`.

thinkdj:
> You should be able to customize `storage`.

Ah! Missed this. Thank you. 

Reference for others:  Customizing `storage`: [Docs](https://jotai.org/docs/utils/atom-with-storage#parameters) & [sample](https://github.com/pmndrs/jotai/discussions/1220).

PR #395, opened 2021-03-30
fix: waitForAll with throw

dai-shi:
https://github.com/pmndrs/jotai/pull/358#discussion_r602177739
Sorry, my suggestion back then was wrong. We are **catching** promises, so we should **throw** what we caught. This is correct.

### DIFF
--- a/tests/utils/waitForAll.test.tsx
+++ b/tests/utils/waitForAll.test.tsx
@@ -1,5 +1,5 @@
 import React, { Fragment, StrictMode, Suspense } from 'react'
-import { render } from '@testing-library/react'
+import { fireEvent, render } from '@testing-library/react'
 import { Provider as ProviderOrig, atom, useAtom } from '../../src/index'
 import { waitForAll } from '../../src/utils'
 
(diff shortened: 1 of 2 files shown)


tests/utils/waitForAll.test.tsx
```diff
@@ -85,12 +86,77 @@ it('waits for two async atoms', async () => {
 
   jest.runOnlyPendingTimers()
 
-  await findByText('num1: 1')
-  await findByText('num2: 2')
+  await findByText('num1: 1, num2: 2')
```
dai-shi, line 89: We should test with one `findByText` to avoid edge cases. (or, use `waitFor` and multiple `getByText`s.)

Thisen:
That's fair and all - I already got 90% of this code on my fix/issue-391 branch, so maybe lets go forward with my branch and close this?

Thisen:
You could also shave this PR down to the `throw` part and fixing the test - What you think @dai-shi ?

dai-shi:
Let me shave this pr to fix the issue in #358, because of my misunderstanding back then.
I need to revert the atom cache usage properly.

Thisen (approved):

PR #397, opened 2021-04-01
fix: atom creator types

dai-shi:
Avoids having `never` on function return type.


dai-shi (commented):
At least, this is better than before. Follow up suggestions are welcome.

Issue #398, opened 2021-04-01
`useAtomCallback` doesn't have types for 3-param signature


dwelle:
```ts
const App = () => {
  useAtomCallback(useCallback((get, set, arg) => {}, [])); // <------ untyped
  return null;
};
```

sandbox: https://codesandbox.io/s/epic-voice-vgjhr?file=/src/App.tsx

dai-shi:
Nice catch. That's similar to what you reported and we discussed about `setAtom` before. I'll make typing a bit looser likewise.

PR #399, opened 2021-04-01
fix types for 3-arg useAtomCallback

dwelle:
fix https://github.com/pmndrs/jotai/issues/398

It's a hack. I'm not exactly sure why the original overloads didn't work — that part TS (probably something to do with covariance/contravariance rules) is still something I haven't explored much.

This fix works, but it has bad UX. Could be improved by inlining the error type (right now the intellisense doesn't display the error message).

And of course, someone with more TS knowledge can probably make the overloads work properly. Feel free to close.

### DIFF
--- a/src/utils/useAtomCallback.ts
+++ b/src/utils/useAtomCallback.ts
@@ -3,16 +3,14 @@ import { atom, useAtom } from 'jotai'
 
 import type { Getter, Setter, Scope } from '../core/types'
 
-type UntypedArgError = {
-  error: 'you must manually assign type to the Arg argument'
-}
+type Callback<Result, Arg> = undefined extends Arg
+  ? (arg?: Arg) => Promise<Result>
+  : (arg: Arg) => Promise<Result>
 
-export function useAtomCallback<Result, Arg = UntypedArgError>(
+export function useAtomCallback<Result, Arg>(
   callback: (get: Getter, set: Setter, arg: Arg) => Result,
   scope?: Scope
-): [Arg] extends [UntypedArgError]
-  ? () => Promise<Result>
-  : (arg: Arg) => Promise<Result>
+): Callback<Result, Arg>
 
 export function useAtomCallback<Result, Arg>(
   callback: (get: Getter, set: Setter, arg: Arg) => Result,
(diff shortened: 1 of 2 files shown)


dai-shi:
Oh, I was about to push a branch.

Here's the diff:
```diff
diff --git a/src/utils/useAtomCallback.ts b/src/utils/useAtomCallback.ts
index 875473f..a0a494b 100644
--- a/src/utils/useAtomCallback.ts
+++ b/src/utils/useAtomCallback.ts
@@ -3,15 +3,14 @@ import { atom, useAtom } from 'jotai'
 
 import type { Getter, Setter, Scope } from '../core/types'
 
-export function useAtomCallback<Result>(
-  callback: (get: Getter, set: Setter) => Result,
-  scope?: Scope
-): () => Promise<Result>
+type Callback<Result, Arg> = undefined extends Arg
+  ? (arg?: Arg) => Promise<Result>
+  : (arg: Arg) => Promise<Result>
 
 export function useAtomCallback<Result, Arg>(
   callback: (get: Getter, set: Setter, arg: Arg) => Result,
   scope?: Scope
-): (arg: Arg) => Promise<Result>
+): Callback<Result, Arg>
 
 export function useAtomCallback<Result, Arg>(
   callback: (get: Getter, set: Setter, arg: Arg) => Result,
```

What do you think?

dai-shi:
> still something I haven't explored much.

Me neither.

> someone with more TS knowledge can probably make the overloads work properly.

Yeah, if this can be solved `SetAtom` in src/core/types.ts can be too.

dwelle:
Better. Commit under your name :).

dwelle:
Sadly when using with `useCallback` you always need to type the arg, but I think that's ok since you should do it anyway — the only problem is the unclear error (fyi, this behavior is identical to my original fix):

```ts
const update = useAtomCallback((get, set, arg /* unknown ok here */) => {})
const updateStable = useAtomCallback(useCallback((get, set, arg /* unknown NOT ok here */) => {}, []))
```

dai-shi:
Yeah, it might require more TS skills.

dai-shi (approved):
At least this fixes the issue and it's good to merge.
Thanks for your work! We hope eventually we get better understanding in TS and have better ideas. (or someone please help.)

PR #403, opened 2021-04-01
chore(docs): getting started

### LINKED ISSUE #401: [Docs] Improve Getting Started
`docs/introduction/getting-started.md`

dai-shi:
close #401 

There might be something missing that should be noted. I guess we would need to keep improving.
Suggestions are welcome before merging or even after.


docs/api/core.md
```diff
@@ -147,3 +147,46 @@ const [value, updateValue] = useAtom(anAtom)
...
+some hooks like `useRef` or `useMemo`.
+
+You can create an atom and store it wth `useState` or even in another atom.
+See an example in [issue #5](https://github.com/pmndrs/jotai/issues/5).
+
+You can cache atoms somewhere globally.
+See [this example](https://twitter.com/dai_shi/status/1317653548314718208) or
+[that example](https://github.com/pmndrs/jotai/issues/119#issuecomment-706046321).
+
+Check [`atomFamily`](../api/utils.md#atomfamily) in utils too.
```
Thisen, line 183:
```suggestion
Check [`atomFamily`](../api/utils.md#atomfamily) in utils for parameterized atoms.
```

docs/api/core.md
```diff
@@ -147,3 +147,46 @@ const [value, updateValue] = useAtom(anAtom)
...
+
+# Atoms can be created on demand
+
+Basic examples in readme only show defining atoms globally outside components.
+There is no restrictions about when we create an atom.
+As long as we know atoms are identified by their object referential identity,
+it's okay to create them at anytime.
+
+If you create atoms in render functions, you would typically want to use
+some hooks like `useRef` or `useMemo`.
```
Thisen, line 174:
```suggestion
a hook like `useRef` or `useMemo` for memoization. If not, the atom would be created everytime the component renders.
```

docs/api/core.md
```diff
@@ -147,3 +147,46 @@ const [value, updateValue] = useAtom(anAtom)
...
+# How atom dependency works
+
+To begin with, let's explain this. In the current implementation, every time we invoke the "read" function, we refresh dependencies. For example, If A depends on B, it means that B is a dependency of A, and A is a dependent of B.
+
+```js
+const uppercaseAtom = atom((get) => get(textAtom).toUpperCase())
+```
+
+The read function is the first parameter of the atom.
+The dependency will initially be empty. On first use, we run the read function and know that uppercaseAtom depends on textAtom. textAtom is the dependency of uppercaseAtom. So, add uppercaseAtom to the dependents of textAtom.
```
Thisen, line 162:
```suggestion
The dependency will initially be empty. On first use, we run the read function and know that `uppercaseAtom` depends on `textAtom`. `textAtom` has a dependency on `uppercaseAtom`. So, add `uppercaseAtom` to the dependents of `textAtom`.
```

docs/introduction/getting-started.md
```diff
@@ -1,100 +1,116 @@
...
+    // you can set as many atoms as you want at the same time
+  }
+)
 ```
 
-There are two kinds of atoms: a writable atom and a read-only atom.
-Primitive atoms are always writable.
-Derived atoms are writable if `writeFunction` is specified.
-The `writeFunction` of primitive atoms is equivalent to the setState of React.useState.
+`get` in the read function is to read atom value.
```
Thisen, line 50:
```suggestion
`get` in the read function is to read the atom value.
```

docs/introduction/getting-started.md
```diff
@@ -1,100 +1,116 @@
...
-`get` is similar to the one described above, but it doesn't track the dependency. `set` is a function that takes an atom config and a new value which then updates the atom value in Provider. `update` is an arbitrary value that we receive from the updating function returned by `useAtom` described below.
-
-## Provider
+`set` in the write function is to write atom value.
+It will invoke the write function of the target atom.
 
-Atom configs don't hold values. Atom values are stored in a Provider. A Provider can be used like React context provider. Usually, we place one Provider at the root of the app, however you could use multiple Providers, each storing different atom values for its component tree.
+Atom configs can be created anywhere, but referential equality is important.
+They can be created dynamically too.
+To create an atom in render function, `useMemo` is required to get a stable reference.
```
Thisen, line 62:
```suggestion
To create an atom in render function, `useMemo` or `useRef` is required to get a stable reference. If in doubt about using `useMemo` or `useRef` for memorization, use `useMemo`.
```

docs/introduction/getting-started.md
```diff
@@ -1,100 +1,116 @@
...
-it's okay to create them at anytime.
-
-If you create atoms in render functions, you would typically want to use
-some hooks like `useRef` or `useMemo`.
+## Provider
 
-You can create an atom and store it wth `useState` or even in another atom.
-See an example in [issue #5](https://github.com/pmndrs/jotai/issues/5).
+Provider is to provide a state for component sub tree.
+Multiple Providers can be used for multilple sub trees, even nested.
```
Thisen, line 98:
```suggestion
Multiple Providers can be used for multiple subtrees, even nested.
```

docs/api/core.md
```diff
@@ -147,3 +147,46 @@ const [value, updateValue] = useAtom(anAtom)
 ```
 
 The `updateValue` takes just one argument, which will be passed to the third argument of writeFunction of the atom. The behavior totally depends on how the writeFunction is implemented.
+
+---
+
+# How atom dependency works
+
+To begin with, let's explain this. In the current implementation, every time we invoke the "read" function, we refresh dependencies. For example, If A depends on B, it means that B is a dependency of A, and A is a dependent of B.
```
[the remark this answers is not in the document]
dai-shi: This is to explain "dependencies" and "dependents". Maybe, a diagram would help though.

docs/api/core.md
dai-shi: okay "has a dependency on"

docs/api/core.md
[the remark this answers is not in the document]
dai-shi:
```suggestion
To begin with, let's explain this. In the current implementation, every time we invoke the "read" function, we refresh dependencies and dependents. For example, If A depends on B, it means that A is a dependent of B.
```

dai-shi:
although this is far from perfect, let's keep moving. we have more to improve.
suggestions to improve docs are always welcome.

PR #407, opened 2021-04-04
feat(util): atomWithHash

dai-shi:
DOM only

- [x] impl
- [x] tests
- [x] docs -> make a PR to docs repo afterward

# Docs

https://github.com/pmndrs/website/pull/76

### Usage
```js
atomWithHash(key, initialValue): PrimitiveAtom
```
This creats a new atom that is connected with URL hash.
The hash must be in the URLSearchParams format.
It's two-way binding: changing atom value will change the hash and
changing the hash will change the atom value.
This function works only with DOM.
### Examples
```jsx
import { atomWithHash } from 'jotai/utils'
const countAtom = atomWithHash('count', 1)
const Counter: React.FC = () => {
  return (
    <>
      <div>count: {count}</div>
      <button onClick={() => setCount((c) => c + 1)}>button</button>
    </>
  )
}
```
### Codesandbox
https://codesandbox.io/s/react-typescript-forked-d5kn6


dai-shi:
https://codesandbox.io/s/react-typescript-forked-nto9h?file=/src/App.tsx
It turns out this has a pitfall (setCount is not sync, and setState is required.)
I should go with `atomWithHash`

dai-shi:
codesandbox example
https://codesandbox.io/s/react-typescript-forked-d5kn6?file=/src/App.tsx

PR #408, opened 2021-04-04
feat(wip): new docs with nextra

aulneau:
Hi there :) Here is a quick MVP for a new docs site using `nextra`, a next.js based docs site builder. It was really straightforward to set up, and I think does a good job at giving these docs a new home. I'm open to folks' suggestions around if this is the direction we'd like to go. I am also open to doing some more bespoke, however that would take more time and require more maintenance over time. This seems to be a good solution that allows for most people to not have to deal with all the more complex stuff behind making a docs site. 

I know @dai-shi really wanted to keep the `/docs` folder intact, but I don't think it will be possible if we use something like next.js for the website. I think it comes down to pros and cons of both. I'd love to hear more about what makes it important that the docs stay at the root of the repository?


dai-shi:
Thanks for opening this up!

I know putting docs inside website and link there from readme.md "works". (I did the same in [my repo](https://github.com/dai-shi/react-tracked) with a regret.)
This is probably just a preference, but what I wanted to see is the separation of docs and website as products. So, people can contribute to doc contents without worrying about website, and people can contribute website look&feel and features without worrying about contents.

So, basically we'd like to put efforts on website. It should be something maintainable, but that doesn't have to mean we create once and it works forever. We'd like to have website contributors.

Here's some random comments:

- next-mdx-remote suggested by @Thisen worked well for separating `docs`, which I like. It shows the flexibility of nextjs. But we need quite some work for the initial release.
- nextra looks good and simple, at least far better than the current bare website with next-mdx-remote.
- @Aslemammad had tried with docusaurus, but has the same docs issue, so we suspended.
- If nextra doesn't handle separating `docs`, docusaurus can be an option too.
- @sandren seems to have experience with gastby, which may solve the docs issue?
- Overall, I'm not good at website thing, so I'd like to count on someone.

In short, I'd like to have `docs` to be website-agnostic. It also feels good when you checkout the repo, having docs at the root (it's a rare use case though).

Thoughts, everyone?

Thisen:
Good work here @aulneau.

However, I agree with @dai-shi about keeping the docs seperate and use `next-mdx-remote`. I know that will be more work to create a proper website with Next.js, but long term I think it will be better.

aulneau:
I'm going to close this as it sounded like someone else will take this up and use a different framework :)

Issue #414, opened 2021-04-09
Error thrown in async write result in infinite loop


garrett-hopper:
Trying to debug this, it looks like the `Promise` wrapped test error is constantly alternating being thrown by [core/useAtom.ts#L33](https://github.com/pmndrs/jotai/blob/fa983b2b4d39cc123df1fc6e8e49f6e81f0b8e90/src/core/useAtom.ts#L33) and some internal React code.

This results in constant `Rendering Test` messages being logged.
```jsx
import { atom, Provider, useAtom } from 'jotai';
import { Suspense } from "react";
import { render } from "react-dom";
import { ErrorBoundary } from "react-error-boundary";

const testAtom = atom(undefined, (async () => {
  throw new Error('test');
}));

render(<App />, document.getElementById("app"));

function App() {
  return (
    <Provider>
      <ErrorBoundary fallbackRender={({ error }) => { return (<p>{error.toString()}</p>); }}>
        <Suspense fallback={"Loading..."}>
          <Test />
        </Suspense>
      </ErrorBoundary>
    </Provider>
  );
}

function Test() {
  console.log('Rendering Test');
  const [_, write] = useAtom(testAtom);

  return (
    <button onClick={() => write()}>Test</button>
  );
}

```

aulneau:
I've ran into this too, and it can be hard to debug. I think this will fix it:
```ts
const testAtom = atom(undefined, (async () => {
  await new Promise(resolve => resolve());
  throw new Error('test');
}));
```

Basically if you do an async atom that doesn't actually do anything async, it seems to break. Def not ideal, I wonder what can be done. @dai-shi what do you think?

dai-shi:
Throwing a write promise is something I invented, and not like a de-facto yet.
So, it's subject to change in the future. So far, it's handy to suspend with write atoms.

I'm not sure if I understand the issue of OP. Oh, it's infinite loop. That's unexpected.
So, how can you make this to cause infinite loop? Does clicking the "Test" button once makes infinite loop?

---
There's one design issue. We could remove the write promise when error occurred, and return a original value. Yeah, that makes sense actually. Let me do it.

garrett-hopper:
Yeah, a single click of `Test` results in infinite rendering calls.
Turning on `Pause on caught exceptions` shows that the `Promise` wrapped `Error` is bouncing between React and Jotai.

garrett-hopper:
[...]

PR #415, opened 2021-04-09
fix: remove write promise on error too

dai-shi:
ref: https://github.com/pmndrs/jotai/issues/414#issuecomment-817028501

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -168,9 +168,9 @@
     }
   },
   "index.js": {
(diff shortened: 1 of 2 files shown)


dai-shi:
I hoped to write a test for this, but didn't find a good way to handle infinite loop in tests.
If anyone has ideas to test it (see #414), please try and open a PR.

Let me merge this now, as the fix is rather obvious.

PR #417, opened 2021-04-10
fix: precise SetAtom type in useAtom

### LINKED ISSUE #416: SetAtom<Update> result should be one of void or Promise<void>, not both
Currently, the type of `SetAtom<Update>` is
```ts
export declare type SetAtom<Update> = undefined extends Update ? (update?: Update) => void | Promise<void> : (update: Update) => void | Promise<void>;
```
When I call `useAtom`, it seems to me all the information is available to determine if the result of the write function will be a `Promise` or not.

It looks like this would require `WritableAtom` containing this information as a type parameter.

dai-shi:
close #416

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -11,17 +11,27 @@ export type Setter = <Value, Update>(
   atom: WritableAtom<Value, Update>,
   update: Update
 ) => void
-export type Write<Update> = (
+export type SyncWrite<Update> = (
   get: Getter,
   set: Setter,
   update: Update
-) => void | Promise<void>
+) => void
+export type AsyncWrite<Update> = (
+  get: Getter,
+  set: Setter,
+  update: Update
+) => Promise<void>
+export type Write<Update> = SyncWrite<Update> | AsyncWrite<Update>
 
 export type Scope = symbol | string | number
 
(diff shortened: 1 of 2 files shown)


dai-shi:
@garrett-hopper How is this?

garrett-hopper:
@dai-shi, hmm, doesn't quite seem to work. :thinking:
```
Type 'WritableAtom<null, FileWithPath>' is not assignable to type 'AsyncWritableAtom<null, FileWithPath>'.
  Type 'SyncWritableAtom<null, FileWithPath>' is not assignable to type 'AsyncWritableAtom<null, FileWithPath>'.
    Type 'SyncWritableAtom<null, FileWithPath>' is not assignable to type '{ write: AsyncWrite<FileWithPath>; onMount?: OnMount<FileWithPath>; }'.
      The types returned by 'write(...)' are incompatible between these types.
        Type 'void' is not assignable to type 'Promise<void>'. [2322]
```

dai-shi:
Hmm, maybe it's not possible with this way.

dai-shi:
On second thought, this is not precise. It can return a promise even if the atom is sync writable. We could always return a promise, but I feel like it should be optional.

PR #419, opened 2021-04-14
feat: jotai/zustand

dai-shi:
following #405, but the implementation is much simpler as both zustand and jotai are based on immutable state.

> Is this worth adding?

Implementation-wise, it's simple. But, it would be worth to show the pattern explicitly.
It would be nice to elaborate docs to show real use cases.

### DIFF
--- a/tests/zustand/atomWithStore.test.tsx
+++ b/tests/zustand/atomWithStore.test.tsx
@@ -1,0 +1,43 @@
+import React from 'react'
+import { fireEvent, render, act } from '@testing-library/react'
+import create from 'zustand/vanilla'
+import { Provider, useAtom } from '../../src/index'
+import { atomWithStore } from '../../src/zustand'
+
+it('count state', async () => {
+  const store = create(() => ({ count: 0 }))
+  const stateAtom = atomWithStore(store)
+  const Counter: React.FC = () => {
+    const [state, setState] = useAtom(stateAtom)
+
+    return (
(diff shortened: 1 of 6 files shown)


sandren:
> Is this worth adding?

Yes! I already know of one project where we would replace the use of Zustand with `atomWithStore` once this is released. 😀

dai-shi:
> I already know of one project

Would you like to test if this PR works for you before releasing it? https://ci.codesandbox.io/status/pmndrs/jotai/pr/419/builds/119584

dai-shi:
https://codesandbox.io/s/react-typescript-forked-c6zi9?file=/src/App.tsx
An example.

tests/zustand/atomWithStore.test.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import { fireEvent, render, act } from '@testing-library/react'
+import create from 'zustand/vanilla'
+import { Provider, useAtom } from '../../src/index'
+import { atomWithStore } from '../../src/zustand'
+
+it('count state', async () => {
```
Thisen, line 7: Aren't there missing some tests, where Zustand store state is updated outside React and then you see those changes in the component?

dai-shi: That's what I did in L39. (I'd admit the test is minimal.)

PR #421, opened 2021-04-15
feat: jotai/redux

dai-shi:
If jotai/zustand #419 is a thing, why not `jotai/redux` ?

But then my question comes again. what should be in and out.
There could be a real use case though. Applying jotai gracefully to existing redux apps.

### DIFF
--- a/docs/api/redux.md
+++ b/docs/api/redux.md
@@ -15,7 +15,7 @@ npm install redux
 yarn add redux
 ```
 
-## atomWithProxy
+## atomWithStore
 
 `atomWithStore` creates a new atom with redux store.
 It's two-way binding and you can change the value from both ends.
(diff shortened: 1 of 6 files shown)


dai-shi:
An example: https://codesandbox.io/s/react-typescript-forked-cmlu5?file=/src/App.tsx

docs/api/redux.md
```diff
@@ -0,0 +1,52 @@
...
+
+You have to install `redux` to access this bundle and its functions.
+
+```
+npm install redux
+# or
+yarn add redux
+```
+
+## atomWithProxy
```
Thisen, line 18:
```suggestion
## atomWithStore
```

dai-shi: Thanks!

PR #422, opened 2021-04-15
fix: generalize typing in focusAtom

dai-shi:
from: https://github.com/pmndrs/jotai/issues/231#issuecomment-820163116

focusAtom can be improved to require not `PrimitiveAtom<T>` but just symmetric `WritableAtom<T, T>`.

### DIFF
--- a/src/optics/focusAtom.ts
+++ b/src/optics/focusAtom.ts
@@ -5,31 +5,31 @@ import {
   setWeakCacheItem,
   WeakCache,
 } from '../utils/weakCache'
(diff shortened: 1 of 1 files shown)


merisbahti:
Perhaps it should be able to accept both of them?

dai-shi:
> Perhaps it should be able to accept both of them?

Yes, but no worries. `PrimitiveAtom<T>` includes `WritableAtom<T, T>`, so it works. There could be better typing around `NonFunction`, but I'm not sure at the moment. It's in src/core/types.ts.

Issue #388, opened 2021-03-28
Snaphot-based devtools


Thisen:
Umbrella issue for snapshot-based devtools.

Things to consider/ideate:

- [x] Hooks for snapshots
- [ ] Babel plugin for adding debug labels
- [ ] Dedicated devtools?

Feel free to add ideas/thoughts.

dai-shi:
Thanks for opening this up.

So, snapshot-based means it allows time traveling? The first question that comes to my mind is how to support non serializable values. Probably keeping snapshot array in JS memory.

FWIW, we could do something similar with catchAllAtom and useAtomDevtools for serializable values.

```js
const atom1 = atom(...)
const atom2= atom(...)

const allAtoms = [atom1, atom2]
const catchAllAtom = atom(
  (get) => allAtoms.map(get),
  (get, set, values) => {
    allAtoms.forEach((a, i) => {
      set(a, values[i])
    })
  }
)

const Component = () => {
  useAtomDevtoools(catchAllAtom)
```

Thisen:
Closed, as babel plugin and dedicated devtools will be after v1. See #428  and #487

PR #425, opened 2021-04-15
Snapshot-based devtools (1st iteration)

Thisen:
See #388.

- [X] Impl
- [X] Tests

This is the first iteration of snapshot-based devtools. It's not ready yet for public use.


Thisen:
Everyone feel free to take a look.

@dai-shi with this current logic in `Provider.ts`: 

```
if (
    typeof process === 'object' &&
    process.env.NODE_ENV !== 'production' &&
    process.env.NODE_ENV !== 'test'
)
```

How do I create tests for the devtools stuff? I could find a simple solution with `NODE_ENV`, but lets aim for a good way to run these tests.

CC: @thiesl137

dai-shi:
Great work. So far, so good. I'd like to spend time to think about better internal apis for second iteration.
For this iteration, remainings are 1) `TODO` to create final snapshot map, 2) tests?

> `process.env.NODE_ENV !== 'test'`

This is something we are not happy with. I just added to pass some of tests (optimizations?), but happy to remove it and handle it in tests, if there's a solution (there must be).

src/core/Provider.ts
```diff
@@ -30,6 +41,11 @@ export const Provider: React.FC<{
       storeRef.current as ReturnType<typeof createStore>,
       atomsRef.current
     )
+    children = createElement(
+      RegisteredAtomsContext.Provider,
+      { value: atomsRef.current },
```
dai-shi, line 46: So, because this doesn't trigger useAtomsSnapshot to re-render, we can't capture atoms created on demand?

src/core/contexts.ts
```diff
@@ -39,3 +39,5 @@ export const getStoreContext = (scope?: Scope) => {
   }
   return StoreContextMap.get(scope) as StoreContext
 }
+
+export const RegisteredAtomsContext = createContext<AnyAtom[]>([])
```
dai-shi, line 43:
```suggestion
// only for DEV purpose
export const RegisteredAtomsContext = createContext<AnyAtom[]>([])
```

Thisen:
@dai-shi Feel free to review now. Make sure to look at imports aswell, not sure how to deal with those without exposing a bunch of internal APIs.

I think we should let me be undocumented, while we iterate on this API.

src/devtools/useAtomsSnapshot.ts
```diff
@@ -0,0 +1,38 @@
...
+      return () => {
+        unsubs.forEach((unsub) => unsub())
+      }
+    },
+    [atoms]
+  )
+  const state: State = useMutableSource(mutableSource, getState, subscribe)
+
+  return useMemo(() => {
+    const atomToAtomStateTuples = atoms
```
dai-shi, line 28:
```suggestion
    const atomToAtomValueTuples = atoms
```

I'd like to call it "atom value", because `AtomState` means something else.

tests/devtools/useAtomsSnapshot.test.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useState } from 'react'
+import { fireEvent, render } from '@testing-library/react'
+import { Provider, atom, useAtom } from '../../src/index'
+import { useAtomsSnapshot } from '../../src/devtools'
+
+beforeEach(() => {
+  process.env.NODE_ENV = 'development'
```
dai-shi, line 7: Oh, does this work? Let me try something.

tests/devtools/useAtomsSnapshot.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    return null
+  }
+
+  const SimpleDevtools: React.FC = () => {
+    const atoms = useAtomsSnapshot()
+
+    return (
+      <div>
+        {Array.from(atoms).map(([atom, atomState]) => (
+          <p key={atom.debugLabel}>{`${atom.debugLabel}: ${atomState}`}</p>
```
dai-shi, line 72:
```suggestion
        {Array.from(atoms).map(([atom, atomValue]) => (
          <p key={atom.debugLabel}>{`${atom.debugLabel}: ${atomValue}`}</p>
```

src/devtools/useAtomsSnapshot.ts
```diff
@@ -0,0 +1,38 @@
...
+      .filter((atom) => !!state.m.get(atom))
+      .map<[AnyAtom, unknown]>((atom) => {
+        const atomState = state.a.get(atom) ?? ({} as AtomState)
+        return [atom, atomState.e || atomState.p || atomState.w || atomState.v]
+      })
+    return new Map(atomToAtomStateTuples)
```
dai-shi, line 34:
```suggestion
    return new Map(atomToAtomValueTuples)
```

dai-shi (commented):
Overall, it looks good to me for the 1st iteration.

tests/devtools/useAtomsSnapshot.test.tsx
dai-shi: #446

dai-shi:
#446 is merged. Please adjust this PR a bit.

PR #438, opened 2021-04-20
fix: improve types without NonFunction in atom()

dai-shi:
https://github.com/pmndrs/jotai/discussions/433#discussioncomment-631065

> v0.16.1 introduces `NonFunction` which seems proper in types, but not good in DX. We'd like to improve it.

Here it is.


dai-shi:
I think this is an improvement, but not 100% sure if it's ideal. Let's release it and wait for feedbacks.

PR #440, opened 2021-04-21
RFC: allow to specify equalityFn for atoms

andreypopp:
This allows to define `equalityFn` for atoms:

    let v = atom(...)
    v.equalityFn = (a, b) => {
      ...
      return areAandBequal
    }

In case `equalityFn` returns `true` then it will bail out from updating
dependent atoms and subscribed components if any.

The main use case I have in mind is to be able to define derived atoms which
compute some non trivial structures. An example from my app:

    let $cellState = atomFamily((id: CellID) => {
      let a = atom((get) => {
        let state = get($editorState(id.notebookID));
        return {
          focus: state.focus?.cellID === id.cellID ? state.focus : null,
          result: state.notebook.cell_results[id.cellID],
          input: state.notebook.cell_inputs[id.cellID],
          inputRemoteVersion: state.cellInputRemoteVersions[id.cellID],
        };
      });
      a.debugLabel = "cellState";
      a.equalityFn = (a, b) => {
        return (
          a.focus === b.focus &&
          a.result === b.result &&
          a.input === b.input &&
          a.inputRemoteVersion === b.inputRemoteVersion
        );
      };
      return a;
    });

The alternative is to define those pieces (focus, result, input, ...) as
separate atoms but this is tedious as they are usually needed at once in my
case.

There's some weird thing with `tsc` which made thing spur `@ts-ignore`, needs to
be resolved before merging (if we agree on this).

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -28,8 +28,12 @@ export type OnMount<Update> = <S extends SetAtom<Update>>(
   setAtom: S
 ) => OnUnmount | void
 
+export type EqualityFn<Value> = (a: Value, b: Value) => boolean
+
 export type Atom<Value> = {
   toString: () => string
+  // This crashes TypeScript compiler...
+  //equalityFn?: EqualityFn<Value>
   debugLabel?: string
   scope?: Scope
   read: Read<Value>
--- a/src/core/vanilla.ts
+++ b/src/core/vanilla.ts
@@ -127,7 +127,18 @@ const setAtomValue = <Value>(
   delete atomState.e // read error
   delete atomState.p // read promise
   delete atomState.i // invalidated revision
-  if (!('v' in atomState) || !Object.is(atomState.v, value)) {
+  let areEqual: boolean
+  // @ts-ignore
+  if (atom.equalityFn != null) {
+    // @ts-ignore
+    let equalityFn = atom.equalityFn as any
+    areEqual =
+      'v' in atomState &&
+      (Object.is(atomState.v, value) || equalityFn(atomState.v, value))
+  } else {
+    areEqual = 'v' in atomState && Object.is(atomState.v, value)
+  }
+  if (!areEqual) {
     atomState.v = value
     ++atomState.r // increment revision
   }
--- a/tests/dependency.test.tsx
+++ b/tests/dependency.test.tsx
@@ -252,6 +252,129 @@ it('should bail out updating if not changed', async () => {
   expect(derivedFn).toHaveReturnedTimes(1)
 })
 
+it('should bail out updating if not changed (custom equalityFn)', async () => {
+  let commits: Record<string, number> = {}
+
+  const useCommitCount = (key: string) => {
+    useEffect(() => {
+      commits[key] = (commits[key] ?? 0) + 1
+    })
+  }
+
+  type V = { value: number }
+
+  const countAtom = atom<V>({ value: 0 })
+  const derivedFn = jest.fn().mockImplementation((get) => get(countAtom))
+  const derivedAtom = atom(derivedFn)
+  // @ts-ignore
+  derivedAtom.equalityFn = (a: V, b: V) => {
+    return a.value === b.value
+  }
+
+  const Counter: React.FC = () => {
+    const [count, setCount] = useAtom(countAtom)
+    useCommitCount('Counter')
+    return (
+      <>
+        <div>count: {count.value}</div>
+        <button onClick={() => setCount((c) => ({ value: 0 }))}>button</button>
+      </>
+    )
+  }
+
+  const DerivedCounter: React.FC = () => {
+    const [derived] = useAtom(derivedAtom)
+    useCommitCount('DerivedCounter')
+    return <div>derived: {derived.value}</div>
+  }
+
+  const { getByText } = render(
+    <Provider>
+      <Counter />
+      <DerivedCounter />
+    </Provider>
+  )
+
+  await waitFor(() => {
+    getByText('count: 0')
+    getByText('derived: 0')
+  })
+  expect(commits.Counter).toEqual(1)
+  expect(commits.DerivedCounter).toEqual(1)
+  expect(derivedFn).toHaveReturnedTimes(1)
+
+  fireEvent.click(getByText('button'))
+  await waitFor(() => {
+    getByText('count: 0')
+    getByText('derived: 0')
+  })
+  expect(commits.Counter).toEqual(2)
+  expect(commits.DerivedCounter).toEqual(1)
+  expect(derivedFn).toHaveReturnedTimes(2)
+})
+
+it('should bail out updating if not changed (custom equalityFn) 2', async () => {
+  let commits: Record<string, number> = {}
+
+  const useCommitCount = (key: string) => {
+    useEffect(() => {
+      commits[key] = (commits[key] ?? 0) + 1
+    })
+  }
+
+  type V = { value: number }
+
+  const countAtom = atom<V>({ value: 0 })
+  // @ts-ignore
+  countAtom.equalityFn = (a: V, b: V) => {
+    return a.value === b.value
+  }
+
+  const derivedFn = jest.fn().mockImplementation((get) => get(countAtom))
+  const derivedAtom = atom(derivedFn)
+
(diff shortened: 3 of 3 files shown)


dai-shi:
Hi, appreciate for the suggestion and actually trying the implementation.
However, we don't add this in core for the sake of simplicity. And, the way useState works is bailing out based on ref equality.

Such functionality could be added in utils, but it looks to me like your use case is already covered by `selectAtom` which has `equalityFn`.

Closing this. Feel free to leave some comments.

aulneau:
> Such functionality could be added in utils, but it looks to me like your use case is already covered by selectAtom which has equalityFn.

I am curious, is this use case exactly possible with `selectAtom`, if so, are there any cases in which it's not?

I personally actually like this ability to add an equalityFn to any atom.

andreypopp:
Hi! Thanks for checking out the PR!

> Such functionality could be added in utils, but it looks to me like your use case is already covered by selectAtom which has equalityFn.

This exact example I've provided — yes, `selectAtom` would suffice, but `selectAtom` is limited as it only allows a single atom as a source. Here I can define a derived atom which could depend on multiple other atoms and still have custom equality working.

From the simplicity point of view, I actually think that this functionality fits having a minimal expressive core — it adds nothing but generalises an already existent equality check.

dai-shi:
There should be no cases if referential equality is the issue. It only recalculates when dependencies changes. That's the whole point of the library (to avoid selector for optimization only and dealing with ref equality).

If one feels the need of equalityFn, I'd suspect they still have the mental model of selector for optimization, which is the whole concept I'd like to avoid. (exception `selectAtom` with the mental model, and `atomWithQuery` with the restriction of upstream lib.)

The OP example could also be implemented without equalityFn, and it's my recommendation.

```ts
let $cellState = atomFamily((id: CellID) => {
  let $focus = atom((get) => {
    let state = get($editorState(id.notebookID));
    return state.focus?.cellID === id.cellID ? state.focus : null;
  });
  // these can be improved further
  let $result = atom((get) => get($editorState(id.notebookID)).notebook.cell_results[id.cellID]);
  let $input = atom((get) => get($editorState(id.notebookID)).notebook.cell_inputs[id.cellID]);
  let $inputRemoteVersion = atom((get) => get($editorState(id.notebookID)).cellInputRemoteVersions[id.cellID]);
  let a = atom((get) => ({
    focus: get($focus),
    result: get($result),
    input: get($input),
    inputRemoteVersion: get$(inputRemoteVersion),
  }));
  a.debugLabel = "cellState";
  return a;
});
```

aulneau:
> If one feels the need of equalityFn, I'd suspect they still have the mental model of selector for optimization, which is the whole concept I'd like to avoid.

Can you elaborate why this is unfavorable?

> The OP example could also be implemented without equalityFn, and it's my recommendation.

Many things can be implemented in many ways, but I'm curious, why is your recommendation better?

andreypopp:
FWIW I don't think that custom equality function is about selector optimisation mental model but more about having a value semantics for user defined data types.

For example in my code I often have modules like `User` which export `User` type and `eq` function which compares two users:

```
import * as User from './User'

let u1 = User.make('Andrey', ...)
let u2 = User.make('John', ...)

console.log(User.eq(u1, u2)) # => false
```

So this custom equality function is not about selectors but rather useful in general.

As for the example you've provided above with multiple nested atoms — I think the end effect is exactly the same as with a single custom equality function — you are going to do multiple nested reference equality checks (same as the custom `equalityFn` I've provided above but with much more overhead for bookkeeping...)

dai-shi:
```js
const user1 = { name: 'John', age: 20 }
const user2 = { name: 'John', age: 30 }
const userEq = (u1, u2) => u1.name === u2.name
```
So, I guess this is a simplified example.

You could write JSX `<User name={user.name} age={user.age} />`.
But, because of the custom equality function, what this displays is unpredictable.

I'm sad, because I'm pretty sure this can't convince you guys.
Let me see if I can come up with better examples.
Meanwhile, a question would be why userState/useReducer don't have custom equality function? `useState(initialState, equalityFn)` (I might have to re-consider if React does it... which is happening in use-context-selector/useSelectedContext and I'm confused, tbh.)

aulneau:
I don't mean to press you very hard, I am mostly just very curious for my own understanding.

One element to consider too is developer experience. I totally agree that you can do it in the way you showed, however, one of the hardest concepts for me to understand about using jotai is how best to nest atoms and when too many atoms is too much. 

From DX alone, the examples with eq is so much simpler to understand and implement imo.

andreypopp:
Don’t be sad please! There’s no need to convince anyone and I’m not trying to do it, I’m just presenting my point of view here. Sorry if I was too insisting.

As for why React doesn’t allow to provide custom equality for useState/useReducer, I think it just doesn’t need to:

1. You can always bail out of updating the state if value equality holds. You can factor out this to a general utility with a minimal/reasonable overhead:

```
function useStateWithEq(init, eq) {
  let [state, setState_] = React.useState(init)
  let setState = React.useCallback(update => {
    setState_(prevState => {
      let nextState = update(prevState)
      return !eq(prevState, nextState) ? nextState : prevState
    })
  }, [setState_])
  return [state, setState]
}
```

2. React's memoization mechanism kinda allows custom equality checks when you specify a vector of dependencies to the `useMemo` hook. Why no proper custom equality? Well the `useMemo` is made for ad-hoc/local memoization and is not tied to any particular data type so you really need a granular control over what you are going to test for equality.

dai-shi:
> but I'm curious, why is your recommendation better?

Let me try this one. The problem with equalityFn is you can make it wrong. For example, if we remove one line,
```js
  a.equalityFn = (a, b) => {
    return (
      a.focus === b.focus &&
      a.result === b.result &&
      a.input === b.input
    );
  };
```
It's totally valid, and nothing is wrong. no errors. But, it can behave unexpectedly.
This is because the selector (or just an object) and equalityFn don't match. Impedance mismatch, so to say. EqualityFn gives too much freedom and thus error-prone.
Speaking of too much freedom, you can do like: `atomWithCustomEqlFn(0, (a, b) => a < b)` which is valid, but I feel it's tricky.

andreypopp:
Oh, and to add to

> As for why React doesn’t allow to provide custom equality for useState/useReducer, I think it just doesn’t need to:

3. `React.memo` actually allows to specify a custom comparator

dai-shi:
> You can factor out this to a general utility with a minimal/reasonable overhead:

I think we can do the same for atom:
```js
const atomWithEq = (init, eq) => {
  const anAtom = atom(init, (get, set, arg) => {
    if (!eq(get(anAtom), arg) {
      set(anAtom, arg)
    }
  })
  return anAtom
}
```

andreypopp:
> I think we can do the same for atom

That's correct but it doesn't address output of derived atoms unfortunately. My point was to demonstrate that there are cases where React allows custom equality but `useState` isn't that because it doesn't need to (you can easily "fix" it on update side and then there's no composable semantics for `useState` as there's for atoms — derived atoms).

dai-shi:
Okay, let's focus on "read" part.
If you define `isEqual({ name: 'John', age 20 }, { name: 'John', age 30 })`, we could technically return whichever. That's what "equal" means. So, semantically, it should behave just fine if we randomly returns, or if we only return the second argument. (If it does behave differently with two "equal" data, it's not equal.) If that holds, the custom equality function is only used for read/render optimization. And, the goal of this library is to replace that usage with atom dependency tracking.

andreypopp:
> If you define isEqual({ name: 'John', age 20 }, { name: 'John', age 30 }), we could technically return whichever.

To be honest, I don't understand this example as it feels like a programming error to me. Yes, technically you could define equality like this but then if you use `age` you'll be violating the that "equality" invariant. There are lots of way to introduce programming errors and especially with JavaScript with lack of static types and mutability all around... so would it be possible to right bad good with custom equality functions — yes, like with many other APIs and language constructs....

dai-shi:
> I don't understand this example as it feels like a programming error to me.

Oh, then, that was a bad example to express the point of view. I thought that showed what you wanted to achieve... (I think it's a programming error too.)

I know this is controversial. We had a big debate too when designing react-redux useSelector hook.

Issue #443, opened 2021-04-22
Endless API call with 0.16.2, strange behavior


leifg:
I just upgraded to jot 0.16.2 and I am seeing some very strange behavior.

I am using an atom to fetch the country of my users:

```js
const currentCountryCode = atom<string>(async () => fetchCurrentCountryCode())
const defaultCountryState = atom(async (get) => get(availableCountriesState).find(({id}) => id == get(currentCountryCode)) || {id: "US", name: "United States"} )
const selectedCountryState = atomWithDefault<Country>((get) => get(defaultCountryState))
```

Within fetchCurrentCountryCode I am making a request to an external service:

```js
const fetchCurrentCountryCode = async () => {
  try {
    const rawResponse = await fetch("https://extreme-ip-lookup.com/json/")
    return (await rawResponse.json()).countryCode
  }
  catch (error) {
    return null
  }
}
```

All of this works with jotai 0.16.1. But when i upgrade to 0.16.2 this request just hangs forever and my CPU goes crazy.

<img width="946" alt="Screen Shot 2021-04-22 at 16 26 33" src="https://user-images.githubusercontent.com/110321/115797354-96740e00-a388-11eb-81eb-5b3831fa1f2b.png">

What is particularly weird is that it doesn't seem an HTTP issue because I also fetch data from a backend that I host (different server, same domain). I also switched out the geolocation provider and the error stayed the same.

It also seems to have Heisenbug properties. When I add debug statements like this, it makes the request and the problem is gone.

```js
const fetchCurrentCountryCode = async () => {
  try {
    console.log("making the API call")
    const rawResponse = await fetch("https://extreme-ip-lookup.com/json/")
    console.log("finished the api call with", rawResponse)
    return (await rawResponse.json()).countryCode
  }
  catch (error) {
    console.log("api call failed", error)
    return null
  }
}
```

Even worse, I can't reproduce it on my local system.

The application is hosted on Netflify. You can see a working version with jotai 0.16.1 [here](https://app.filmswipe.io) but I also created a deployment where the error occurs [here](https://deploy-preview-170--naughty-babbage-bda88d.netlify.app/) - Click on "Create Session" to run into the infite request (open your developer tools before you do that, after that your browser doesn't react).

Reproducible in Chrome and Safari.

dai-shi:
Thanks for reporting. Sounds like a tough one. I'm pretty sure this is caused by #431.
We'd need to create a stable repro and test to investigate further.
[...]

PR #444, opened 2021-04-23
fix(core): async atom infinite loop bug

### LINKED ISSUE #442: Resolution of Promise atoms does not propagate correctly.
This example of atoms with Promise values works as expected: https://codesandbox.io/s/jotai-subscriber-a-xk21g. 

However, this small variation hangs: https://codesandbox.io/s/jotai-subscriber-a-forked-58nso?file=/src/App.tsx.  This is likely due to infoAtom depending on a chain of Promises: 
```infoAtom = atom((get) => get(userInfo(get(whoAmI))))```.

dai-shi:
- fix #442 
- improve async typing with overloads

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -1,6 +1,9 @@
 export type SetStateAction<Value> = Value | ((prev: Value) => Value)
 
-export type Getter = <Value>(atom: Atom<Value>) => Value
+export type Getter = {
+  <Value>(atom: Atom<Promise<Value>>): Value
+  <Value>(atom: Atom<Value>): Value
+}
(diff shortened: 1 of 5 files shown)


dai-shi:
@mhcoffin #442 should be fixed with this. Can you double check? You can install the codesandbox build. 

https://ci.codesandbox.io/status/pmndrs/jotai/pr/444/builds/122814

@leifg Can you try if this changes anything for #443?

leifg:
Unfortunately that doesn't fix it.

I created a completely new PR to make sure the build of the deployment started from scratch and the error is still there

dai-shi:
Let me merge this, as we would like to move forward for other issues.

PR #445, opened 2021-04-23
chore: replace docs with links to docs.pmnd.rs/jotai

sandren:


dai-shi:
Let's hope we recover `./website` soon.

PR #446, opened 2021-04-23
chore: fix the NODE_ENV workaround in Provider

dai-shi:
@Thisen does a trick in #425, which seems better than my workaround.
So, this PR fixes it.

However, this PR reveals some other issues. It now shows some warnings on tests. This should actually be addressed in the second iteration of snapshot-based devtools.

Also, fixed warnings in provider-less mode tests. some of tests are missing for provider-less mode. And found an issue with atomWithQuery, which should be addressed separtely (FIXME).

### DIFF
--- a/tests/async.test.tsx
+++ b/tests/async.test.tsx
@@ -1,8 +1,10 @@
-import React, { Fragment, StrictMode, Suspense, useRef, useEffect } from 'react'
+import React, { StrictMode, Suspense, useRef, useEffect } from 'react'
 import { fireEvent, render, waitFor } from '@testing-library/react'
 import { Provider as ProviderOrig, atom, useAtom, Atom } from '../src/index'
 
-const Provider = process.env.PROVIDER_LESS_MODE ? Fragment : ProviderOrig
+const Provider = process.env.PROVIDER_LESS_MODE
+  ? (props: any) => props.children
+  : ProviderOrig
 
 const useCommitCount = () => {
   const commitCountRef = useRef(1)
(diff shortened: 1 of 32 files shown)


tests/onmount.test.tsx
```diff
@@ -1,8 +1,19 @@
...
+  : ProviderOrig
+
+let savedNodeEnv: string | undefined
+beforeEach(() => {
+  savedNodeEnv = process.env.NODE_ENV
+  process.env.NODE_ENV = 'production'
+})
+afterEach(() => {
+  process.env.NODE_ENV = savedNodeEnv
+})
```
dai-shi, line 16: So, does this mean onMount has some issues with DEV mode? Or, is it just a testing thing?
Needs investigation. Help wanted.

tests/async.test.tsx
```diff
@@ -1,8 +1,10 @@
-import React, { Fragment, StrictMode, Suspense, useRef, useEffect } from 'react'
+import React, { StrictMode, Suspense, useRef, useEffect } from 'react'
 import { fireEvent, render, waitFor } from '@testing-library/react'
 import { Provider as ProviderOrig, atom, useAtom, Atom } from '../src/index'
 
-const Provider = process.env.PROVIDER_LESS_MODE ? Fragment : ProviderOrig
+const Provider = process.env.PROVIDER_LESS_MODE
```
Thisen, line 5: Maybe extract this to a util function and use it everywhere?

Thisen: @dai-shi I'll do that and commit - Then you can revert if you disagree :-)

dai-shi: I agree. Thanks for the suggestion.

PR #455, opened 2021-04-27
feat(core): atom component

artalar:
Sorry for the empty description, first of all, I need to know what do you think about this API and should I continue to implement it (add tests and docs)?

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -1,3 +1,5 @@
+import React from 'react'
+
 import {
   Read,
   Write,
@@ -5,7 +7,9 @@ import {
   WritableAtom,
   WithInitialValue,
   PrimitiveAtom,
+  SetAtom,
 } from './types'
+import { useAtom } from './useAtom'
 
 let keyCount = 0 // global key count for all atoms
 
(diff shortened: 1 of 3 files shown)


dai-shi:
Hi, I'm not sure if what is the use case of atom component, but whatever it is, we strongly want to make core minimal. Can you try this outside core?

artalar:
> the use case

There are three use cases
1. You shouldn't need to import an extra API (`useAtom`) and create an extra entity (atom value from an atom)
2. You don't need to think about hooks order and conditional dependencies, like
   ```jsx
   // render
   const data = useAtom(dataAtom)
   // here we have conditional branches, we subscribe to dataAtom changes in one of them, but not using it
   return props.some ? props.other : data
   ```
3. The key feature is a separation of data updates. A hook subscription forces the whole component to rerender, it is not friendly for big forms, for example. When we use the component API we separate render closures and subscriptions and archive better performance (for updates).

> we strongly want to make core minimal

The whole implementation is few lines of code, no more needed. I think it cool feature and the cost is nothing 👀

> Can you try this outside core?

As you may know, I prefer Reatom in regular development, but it's framework agnostic and it hard to archive the same features (we need to do some monkey patching and d.ts overriding or change all imports from `core` to `react` package, it is not friendly). Jotai already based on React and may include this feature transparent, it awesome, I think.

Honestly, I use Jotai in veeeeeery rare cases, but it is a good lib and I recommend it sometimes to people in my community and just want to do their experiences a little better.

dai-shi:
I don't go into details about the use case for now, but one thing is `useAtom` always subscribes to the atom, even if it doesn't use the value.

By minimal, it's not only about the bundle size, but the api surface.
We don't even include `useAtomValue` in the core.
https://github.com/pmndrs/jotai/blob/757d9e512d0dfbe9f4fc6bf3810dbc959d2992b0/src/utils/useAtomValue.ts#L3-L5
(This is controversial, but it is so now.)

By outside core, it's possible to have in `utils`. I feel like this is exactly a good case to wrap/extend an atom config to add a feature, which I hadn't thought about.

PR #458, opened 2021-04-30
Fix links to integrations in readme

TheoBr:
This threw me for a sec when hunting for the zustand integration - hope this is helpful :)


dai-shi (approved):
Nice catch. Thanks tons.

PR #463, opened 2021-05-05
feat: jotai/urql

### LINKED ISSUE #427: Create a bindings to the URQL GraphQL Client.
Jotai is an awesome local state management library, but especially with GraphQL usually, some specialized clients are way more preferable than Redux, Jotai, Recoil etc. although to get GraphQL client's data into the Jotai sometimes could be very helpful.
A common example would be a select options that are getting fetched from the GraphQL backend. Currently, the process is pretty annoying with `useEffect` and other things to be used in order to sync them together, which are very error-prone and hard to manage on a big scale, being able to wrap certain queries or mutations into atoms to have some calculated data based on it and other goodness, etc. will be simply awesome.

https://github.com/FormidableLabs/urql

If anyone will try to do such wrapper, I guess the best people to review and consult on PRs would be the main maintainers of URQL @kitten and @JoviDeCroock

Thanks!
I wish I could contribute a PR, but I'm still in the learning phase of Jotai and URQL too, so not an expert I would like to be to start such work :)

dai-shi:
close #427 

- [x] atomWithQuery
- [x] atomWithMutation
- [x] atomWithSubscription
- [x] tests
- [x] examples
- [x] docs

---

title: URQL
description: This doc describes `jotai/urql` bundle.

## Install

You have to install `@urql/core` and `wonka` to access this bundle and its functions.

```
yarn add @urql/core wonka
```

## atomWithQuery

`atomWithQuery` creates a new atom with a query. It internally uses [client.query](https://formidable.com/open-source/urql/docs/api/core/#clientquery).

```js
import { createClient } from "@urql/core";
import { atomWithQuery } from 'jotai/urql'

const client = createClient({ url: "..." });

const idAtom = atom(1)
const userAtom = atomWithQuery(
  (get) => ({
    query: '{ user { first_name last_name } }',
    variables: { id: get(idAtom) },
  }),
  () => client,
)

const UserData = () => {
  const [{ data }] = useAtom(userAtom)
  return <div>{JSON.stringify(data)}</div>
}
```

#### Basic demo

https://codesandbox.io/s/react-typescript-forked-zujf6?file=/src/App.tsx
`<Codesandbox id="zujf6" />`

## atomWithMutation

`atomWithMutation` creates a new atom with a mutation. It internally uses [client.mutation](https://formidable.com/open-source/urql/docs/api/core/#clientmutation).

```js
import { createClient } from "@urql/core";
import { atomWithMutation } from 'jotai/urql'

const client = createClient({ url: "..." });

const fooAtom = atomWithMutation(
  () => 'mutation Foo { text }',
  () => client,
)

const FooData = () => {
  const [{ data }, mutate] = useAtom(fooAtom)
  return <div>{JSON.stringify(data)} <button onClick={mutate}>Click me</button></div>
}
```

### Examples

TODO

## atomWithSubscription

`atomWithSubscription` creates a new atom with a mutation. It internally uses [client.subscription](https://formidable.com/open-source/urql/docs/api/core/#clientsubscription).

```js
import { createClient } from "@urql/core";
import { atomWithSubscription } from 'jotai/urql'

const client = createClient({ url: "..." });

const fooAtom = atomWithSubscription(
  () => 'subscription Foo { text }',
  () => client,
)

const FooData = () => {
  const [{ data }] = useAtom(fooAtom)
  return <div>{JSON.stringify(data)}</div>
}
```

### Examples

TODO


src/urql/atomWithQuery.ts
```diff
@@ -0,0 +1,132 @@
...
+            requestPolicy: args.requestPolicy,
+            ...args.context,
+            ...action.opts,
+          })
+          .toPromise()
+          .then((result) => {
+            if (!result.stale /* XXX can this happen? */) {
+              set(operationResultAtom, result)
+            }
+          })
```
dai-shi, line 111: @kitten I'm not sure if I understand this correctly. So, in imperative updates, can we use `toPromise()`? Or should we subscribe? (but then how can we unsubscribe)
Also, this should work both pause=true and pause=false, right?

src/urql/atomWithQuery.ts
```diff
@@ -0,0 +1,132 @@
...
+      get(initAtom) // use it here
+      const operationResult = get(operationResultAtom)
+      if (operationResult.error) {
+        throw operationResult.error
+      }
+      if (operationResult.data === undefined) {
+        // XXX can this happen?
+        throw new Error('no data in operation result')
+      }
+      return [operationResult.data, operationResult.extensions]
```
dai-shi, line 127: I changed to return a tuple with some regret.
What we lose here is `operationResult.operation`. Is it acceptable?

src/urql/atomWithQuery.ts
kitten: Well, not really; They're started imperatively but what you're changing is the source, so the most sure-fire way to make this work is to make the `source` (created using `client.executeQuery` or `client.query`) swappable itself, so that if it's set it swaps out the subscription and resubscribes to the new source. That'd make it very similar to how `@urql/vue` works. However, as discussed before, are we sure that we need to bring back these imperative triggers? In theory people could also swap out the entire atom themselves 🤔

src/urql/atomWithQuery.ts
kitten: I saw that you swapped this out; just on a side-note, I'd say that _every_ property on `OperationResult` is potentially interesting and important to users, so I'd hand the entire result through as-is.

src/urql/atomWithQuery.ts
dai-shi: > However, as discussed before, are we sure that we need to bring back these imperative triggers? In theory people could also swap out the entire atom themselves

I thought `pause` is for imperative triggers. Do you mean we can drop supporting `pause`?

src/urql/atomWithQuery.ts
dai-shi: Understand what you mean, but for example `stale` would never be true. Isn't it confusing to have a always-false property? Maybe not for you. We are looking at this from different perspectives.
More issue for me is, `Data` shouldn't be undefined. That's very important for TypeScript users.

kitten: @dai-shi `stale` would only ever be `false` if you decide to filter it (and I think you did?) but I'm not sure whether it's a good idea to hide this just by virtue of being in "jotai land", since that also means that users don't get access to certain pieces of data anymore, like for instance partial data with schema awareness.

But even then you'd probably want to at least match what's already in the type itself: https://github.com/FormidableLabs/urql/blob/12d49d3feccc830c95ef07f488f27655d4010b58/packages/core/src/types.ts#L63-L74

So you could skip `stale` but all the other properties should be made available.

src/urql/atomWithQuery.ts
kitten: It's a possibility! `pause` and the `executeQuery()` function on bindings (in the React/Preact/Vue bindings) exist because it's impossible to pause hooks once they're set up, however with Jotai's atoms that's not necessarily true unless they're only set up once. So it depends, but if a user, say, is encouraged to swap out the atom instead then there's no need for `pause`.

src/urql/atomWithQuery.ts
dai-shi: What about error? We already throw it, so it will never be defined.

So, it's like `Required<Pick<OperationResult, 'data'>> & Omit<OperationResult, 'data' | 'error' | 'stale'>`?

Well, I can just return OperationResult, but my strong desire is to make `Data` not optional.

kitten: Throwing it is unfortunately not an option because we may have partial data with some errors present. I can't find a long explanation for this, but here's a short one: https://github.com/FormidableLabs/urql/discussions/1456#discussioncomment-491031

dai-shi: Oh, that's true. Forgot it's graphql 😓 . I'm convinced.

(Curious if we can distinguish graphql error from others like 404, with which we can throw.)

dai-shi:
@kitten I think I'm done with `atomWithQuery`.

dai-shi:
@kitten 
https://github.com/FormidableLabs/urql/blob/00539ff282c8c61854cd2a805cbd9b8e5cae4d24/packages/react-urql/src/hooks/useMutation.ts#L53
Is it valid to use `toPromise` in atomWithMutation, and skip subscription.unsubscribe?

kitten:
@dai-shi it is, yes, mutations are entirely treated as fire-and-forget side-effects on the one hand, so unsubscribing wouldn't stop them either way, but `toPromise` is also special in that it'll internally subscribe & unsubscribe.

src/urql/atomWithMutation.ts
```diff
@@ -0,0 +1,42 @@
...
+        .toPromise()
+        .then((result) => {
+          set(operationResultAtom, result)
+          action.callback?.(result)
+        })
+        .catch(() => {
+          // TODO error handling
+        })
+    }
+  )
```
dai-shi, line 40: atomWithMutation is much simpler than atomWithQuery as it doesn't have onMount nor subscription.
It might be tricky to use because`useAtom` will suspend. So, either creating another atom or `useUpdateAtom` is necessary.

kitten: I think that's acceptable though. I'd expect that people would have to use `client.mutation` directly for their side-effects half the time, since atoms are really good for reactive values but in this case we're dealing with mutations, which are less reactive and more like an imperative side-effect

src/urql/atomWithSubscription.ts
```diff
@@ -0,0 +1,69 @@
...
+    const initAtom = atom(
+      null,
+      (get, set, cleanup: (callback: () => void) => void) => {
+        set(
+          operationResultAtom,
+          new Promise<OperationResult<Data, Variables>>(() => {}) // new fetch
+        )
+        const client = getClient(get, set)
+        const subscription = pipe(
+          client.subscription(args.query, args.variables, args.context),
```
dai-shi, line 36: atomWithSubscription is almost like atomWithQuery without `requestPolicy` in args.

dai-shi:
I think it's better to pass `client` directly to `atomWithQuery`.

```js
const client = createClient(...)

const anAtom = atomWithQuery(client, () => ({ query: ... }))
```

dai-shi:
https://codesandbox.io/s/react-typescript-forked-pqpof?file=/src/App.tsx

Hmm, having client static isn't very nice. Put it in the args???

dai-shi:
https://codesandbox.io/s/react-typescript-forked-yqboc?file=/src/App.tsx
I put separately as `getClient`.

dai-shi:
Hmm, I wonder how to create examples for mutation and subscription. Are there any dummy servers? Well, the initial release can be `atomWithQuery` only...

kitten:
@dai-shi Yes! we've got multiple example schemas up at trygql.formidable.dev; The examples folder in the urql repo shows us using some of them

dai-shi:
Okay, I will release this. I'm still not confident if this is stable and will note "may contain bugs". But, unless we release it we wouldn't get enough feedback.

PR #464, opened 2021-05-07
Devtools: Atoms in store in dev mode

Thisen:
This PR splits the store for production and development mode.
This brings us back to a single context, without breaking changes.

### DIFF
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -9,33 +9,23 @@ import React, {
 import type { AnyAtom, Scope } from './types'
 import { subscribeAtom } from './vanilla'
 import type { AtomState, State } from './vanilla'
-import {
-  createStore,
-  getStoreContext,
-  RegisteredAtomsContext,
-} from './contexts'
+import { createStore, getStoreContext, StoreForDevelopment } from './contexts'
 import type { Store } from './contexts'
 import { useMutableSource } from './useMutableSource'
 
 export const Provider: React.FC<{
   initialValues?: Iterable<readonly [AnyAtom, unknown]>
   scope?: Scope
-}> = ({ initialValues, scope, children: baseChildren }) => {
+}> = ({ initialValues, scope, children }) => {
   const storeRef = useRef(createStore(initialValues))
-  let children = baseChildren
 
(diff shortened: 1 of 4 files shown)


src/core/Provider.ts
```diff
@@ -1,54 +1,26 @@
...
 import { useMutableSource } from './useMutableSource'
 
 export const Provider: React.FC<{
   initialValues?: Iterable<readonly [AnyAtom, unknown]>
   scope?: Scope
-}> = ({ initialValues, scope, children: baseChildren }) => {
-  const storeRef = useRef<ReturnType<typeof createStore> | null>(null)
-  let children = baseChildren
+}> = ({ initialValues, scope, children }) => {
+  const storeRef = useRef(createStore(initialValues))
```
dai-shi, line 14: I think this should be lazy initialization. I'll make a quick fix.

src/core/Provider.ts
```diff
@@ -82,11 +54,25 @@ const stateToPrintable = ([state, atoms]: [State, AnyAtom[]]) =>
     })
   )
 
-const getState = (state: State) => ({ ...state }) // shallow copy
+export const getState = (state: State) => ({ ...state }) // shallow copy
```
dai-shi, line 57: If we export it, let's make it more descriptive names.

src/devtools/useAtomsSnapshot.ts
```diff
@@ -1,16 +1,28 @@
 import { useCallback, useContext, useMemo } from 'react'
 import { SECRET_INTERNAL_getStoreContext as getStoreContext } from 'jotai'
-import { AtomState, State, subscribeAtom } from '../core/vanilla'
-import { RegisteredAtomsContext } from '../core/contexts'
+import type { AnyAtom } from '../core/types'
+import type { AtomState, State } from '../core/vanilla'
+
+// FIXME across bundles, may or may not work
```
dai-shi, line 6:
```suggestion
// XXX across bundles, this is actually copying code
```

dai-shi (approved):
I'm not 100% confident if this is something ideal, but better than what is currently. Let's continue to improve it.

PR #467, opened 2021-05-09
chore(feat/query): refactor atomWithQuery

dai-shi:
While working on #463, I got a better idea around `initAtom`. This is to follow the pattern.


Aslemammad (commented):
Looks great to me!

PR #477, opened 2021-05-14
fix: debug store

dai-shi:
- [x] fix basic test
- fix onmount test -> follow-up pr
- fix quey/atomWithQuery test -> follow-up pr

### DIFF
--- a/src/core/Provider.ts
+++ b/src/core/Provider.ts
@@ -1,7 +1,6 @@
-import React, { createElement, useCallback, useRef, useDebugValue } from 'react'
+import React, { createElement, useRef, useDebugValue } from 'react'
 
 import type { AnyAtom, Scope } from './types'
-import { subscribeAtom } from './vanilla'
 import type { AtomState, State } from './vanilla'
 import { createStore, getStoreContext, StoreForDevelopment } from './contexts'
 import type { Store } from './contexts'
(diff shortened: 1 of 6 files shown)


src/core/Provider.ts
```diff
@@ -60,9 +59,14 @@ const stateToPrintable = ([state, atoms]: [State, AnyAtom[]]) =>
 const isDevStore = (store: Store): store is StoreForDevelopment => {
   return store.length > 2
 }
-export const getDevState = (state: State) => ({ ...state }) // shallow copy XXX might be better ways
-export const getDevAtoms = ({ atoms }: { atoms: AnyAtom[] }) => atoms
-export const subscribeDevAtoms = (
+export const getDebugStateAndAtoms = ({
```
Thisen, line 62: Instead of `debug*`, maybe call it `snapshot*`, because the atoms and their state is a snapshot?

Thisen (commented):
LGTM.

src/core/Provider.ts
dai-shi: To me, snapshots are immutable. `atoms` are immutable, but `state` is not.

dai-shi:
@Thisen
It looks like it would require more time on the remaining tasks.
As what we have now is probably better than `master`, do you think it's ok to merge this and tackle remaining issues separately?

Thisen (approved):

PR #478, opened 2021-05-15
fix(atomWithQuery): Error: Uncaught [TypeError: Cannot set property 'listeners' of undefined]

Aslemammad:
The destroy function loses the observer, So we save the observer and bind it later.


dai-shi (commented):
Thanks for the fix!
Can you simply do this instead of `.bind()`?
```js
cleanup(() => observer.destroy())
```

Aslemammad:
@dai-shi Done.

dai-shi (approved):
Thanks!!!

PR #483, opened 2021-05-18
feat(utils): atomWithToggle

astahmer:
Hey, I find myself duplicating this behaviour a lot across projects, having to setup an atom and then a `toggle` function, like I would with a basic useState, so here's a little helper that could come handy.

It supports 3 optional args: 
- `initialState`: boolean
- `key`: string, if present the boolean state will be saved in storage (using `atomWithStorage`)
- `storage`: Storage, the type of storage to pass to `atomWithStorage`

It returns an atom with `[boolean, toggleFn]`, the `toggleFn` can have an optional arg to force a particular state (if you want to make a `setActive` function out of it for example)

TL;DR:
```ts
import { atomWithToggle } from "jotai/utils";

// [...]

const withTwitchChatAtom = atomWithToggle(!isDev(), "coro/withTwitchChat");

// In a React component
const [withTwitchChat, toggleWithTwitchChat] = useAtom(withTwitchChatAtom);

// in the render

<Button onClick={() => toggleWithTwitchChat()}>Twitch Chat {withTwitchChat + ""}</Button>
```


dai-shi:
Hey, thanks for the suggestion.

..and, this is a tough one.

> I find myself duplicating this behaviour a lot across projects,

I really like this motivation, but my gut feeling is this shouldn't be added in `jotai/utils` for two reasons.

1. It's not domain specific, but maybe too specific. If we have this, we'd need to accept `atomWithIncrement`, `atomWithDecrement`, and many others.
2. We only provide isolated utils. A single util only depends on core. There might be some exceptions, but that's a general guideline.

> `key`: string, if present the boolean state will be saved in storage

This is mostly my preference, but I'd avoid such condition. I think it's better to have two utils. Or, maybe inject atom creator function (defaults to `atom`) in args.

---
So, my suggestions are:
- Create an unofficial package (which is the idea I like a lot. "jotai ecosystem")
- Come up with a possible generalized util without storage (not sure if that matches well)
- Make this as a recipe in docs

Hope it makes sense.

astahmer:
> It's not domain specific, but maybe too specific. If we have this, we'd need to accept atomWithIncrement, atomWithDecrement, and many others.

Ah, I get it, it totally makes sense I just didn't think about it !

I made it an unofficial package here: https://github.com/astahmer/pastable/tree/main/packages/atom-with-toggle

> This is mostly my preference, but I'd avoid such condition. I think it's better to have two utils. Or, maybe inject atom creator function (defaults to atom) in args.

Regarding that, for now I left the code as-is since having both features in 1 function fits my personal taste 😛

I'm curious tho, what would the `Or, maybe inject atom creator function (defaults to atom) in args.` the approach look like ?
Did you mean, passing an atom instead of the key+storage arg ? I like the idea and might implement it later on since that would make it quite composable !

Oh and for the docs, I initally wanted to contribute to it but I didn't find it in the code, it seems you moved it ?

dai-shi:
> I made it an unofficial package here: https://github.com/astahmer/pastable/tree/main/packages/atom-with-toggle

Nice! We should eventually have a list of 3rd party packages somewhere.

> > This is mostly my preference, but I'd avoid such condition. I think it's better to have two utils. Or, maybe inject atom creator function (defaults to atom) in args.
> 
> Regarding that, for now I left the code as-is since having both features in 1 function fits my personal taste 😛

Yeah, I can understand that from your usage example.

> I'm curious tho, what would the `Or, maybe inject atom creator function (defaults to atom) in args.` approach look like ?
> Did you mean, passing an atom instead of the key+storage arg ? I like the idea and might implement it later on since that would make it quite composable !

Yeah, something like this:
```ts
export function atomWithToggle(
  initialValue: boolean,
  createAtom: (initialValue: boolean) => PrimitiveAtom<boolean> = atom
) {
  // ...
}
```

> Oh and for the docs, I initally wanted to contribute to it but I didn't find it in the code, it seems you moved it ?

Docs are hosted here now: https://docs.pmnd.rs/jotai/introduction
Github is here: https://github.com/pmndrs/website/tree/docs/docs/jotai

If you are to add a recipe about `atomWithToggle`, please put a new file under `advanced-recipes` for now. We may re-organize them later. Maybe the recipe should cover not only `toggle` but also some other patterns, but we can revisit it later.

dai-shi:
I'm still interested in having this pattern in `advanced-recipes` in docs. Closing this as won't merge.

astahmer:
Yup, no problem ! I will get to it asap, probably tomorrow or this weekend
And thank you very much for this incredible piece of software that is jotai :)

PR #484, opened 2021-05-19
chore: fix warnings in tests

dai-shi:
It's more or less a hack. I don't know the root cause.

Also, fixes to prefer getByTestId to findByTestId, which should be good.


Aslemammad (commented):
It's Amazing, Good to go.

PR #489, opened 2021-05-22
Add expect type tests

Thisen:
This PR adds tests of our types. 
This requires that we start running `tsc` in CI.

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -17,7 +17,7 @@ type Write<Update> = (
   update: Update
 ) => void | Promise<void>
 
-type WithInitialValue<Value> = {
+export type WithInitialValue<Value> = {
   init: Value
 }
 
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -17,7 +17,7 @@ type Write<Update> = (
   update: Update
 ) => void | Promise<void>
 
-export type WithInitialValue<Value> = {
+type WithInitialValue<Value> = {
   init: Value
 }
 
--- a/tests/types.test.tsx
+++ b/tests/types.test.tsx
@@ -1,12 +1,11 @@
 import { expectType } from 'ts-expect'
-import { atom, Atom, WritableAtom } from '../src/index'
-import { WithInitialValue } from '../src/core/atom'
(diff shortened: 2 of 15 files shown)


src/core/atom.ts
```diff
@@ -17,7 +17,7 @@ type Write<Update> = (
   update: Update
 ) => void | Promise<void>
 
-type WithInitialValue<Value> = {
+export type WithInitialValue<Value> = {
```
dai-shi, line 20: not exporting this is intentional in #488 as being implementation details. Can we test it without exporting it? I think you can just define it in the test code.

dai-shi: Or, as it's not a public type, we don't need to test it?

tests/types.test.tsx
```diff
@@ -0,0 +1,32 @@
+import { expectType } from 'ts-expect'
+import { atom, Atom, WritableAtom } from '../src/index'
+import { WithInitialValue } from '../src/core/atom'
+
+it('atom() should return the correct types', () => {
+  function Component() {
+    // primitive atom
+    const primitiveAtom = atom(0)
+    expectType<Atom<number> & WithInitialValue<number>>(primitiveAtom)
```
dai-shi, line 9: Either,
```suggestion
    expectType<Atom<number> & { init: number }>(primitiveAtom)
```

Or,
```suggestion
    expectType<Atom<number>>(primitiveAtom)
```

What do you think?

dai-shi: Also, we want to expect with `PrimitiveAtom` which is public type.

src/core/atom.ts
Thisen: Then I can't test the type of `atom()`

dai-shi: I think what we want to test is `atom(0)` has `PrimitiveAtom<number>` type. That's it. WithInitialValue is implementation details and should be tested as behavior.

Thisen:
@dai-shi Ready for review. :-)

dai-shi (approved):
Cool. This will be a good addition to make it maintainable.
We could add more for async atoms.
Great work. Thanks.

PR #493, opened 2021-05-24
Fix typing for `useAtomCallback` returning a Promise.

### LINKED ISSUE #491: [useAtomCallback] Incorrect return type
As of now, `useAtomCallback` wraps everything in `Promise`, which is correct behaviour.

However, if callback returns promise, return type is double nested in `Promise`

```typescript
const valueAtom = atom("");

function updateValueAtom(get: Getter, set: Setter, value: string) {
  // do some async stuff
    set(valueAtom, value);
    return value;
  });
}

function Component() {
  const value = useAtomValue(valueAtom);
  const update = useAtomCallback(updateValueAtom); // Promise<Promise<string>

   // ...
}
```

Nested promise is awaited correctly, it is only typing that is incorrect.

Andreyco:
Fixes #491

### DIFF
--- a/src/utils/useAtomCallback.ts
+++ b/src/utils/useAtomCallback.ts
@@ -9,6 +9,11 @@ type Callback<Result, Arg> = undefined extends Arg
   : (arg: Arg) => Promise<Result>
 
 export function useAtomCallback<Result, Arg>(
+  callback: (get: Getter, set: Setter, arg: Arg) => Promise<Result>,
+  scope?: Scope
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Thanks! Does it work as expected?

Andreyco:
@dai-shi yes, it does. Thanks for prompt feedback!

dai-shi:
Good. Merging this. Thanks for your contribution.

PR #494, opened 2021-05-25
fix(useAtomsSnapshot): export useMutableSource as secret from core

### LINKED ISSUE #492: [BUG] TypeError: source[GET_VERSION] is not a function
jotai version: 0.16.6

I get this error when I try to get atoms snapshot. 

I implemented the official [example](https://docs.pmnd.rs/jotai/api/devtools#use-atoms-snapshot), and checked that **process.env.NODE_ENV** equals to development. However, the error still persists.

Thisen:
Fixes #492


src/devtools/useAtomsSnapshot.ts
```diff
@@ -1,12 +1,10 @@
...
-import type { Scope } from '../core/atom'
+import {
+  SECRET_INTERNAL_getStoreContext as getStoreContext,
+  SECRET_INTERNAL_useMutableSource as useMutableSource,
+} from 'jotai'
 import type { AtomState, State } from '../core/vanilla'
-
-// XXX across bundles, this is actually copying code
-import { useMutableSource } from '../core/useMutableSource'
+import { Atom, Scope } from '../core/atom'
```
dai-shi, line 7:
```suggestion
import type { Atom, Scope } from '../core/atom'
```
We are strict about this. This helps a bit to understand what code is across bundles.

Thisen: That's a mistake, will fix it right away

dai-shi (approved):
Thanks.

PR #495, opened 2021-05-25
chore: avoid using XXX in code comments

dai-shi:
We use the following keywords and avoid `XXX` which is ambiguous.

```ts
// FIXME this means, it's working but should be improved
// TODO this means, it's not implemented and should be done
// HACK this means, it's not very nice but it can't be helped
// NOTE this means, it's something that should be understood
```


Thisen (approved):
LGTM 👍🏼

PR #498, opened 2021-05-27
fix: remount atoms with dependencies

### LINKED ISSUE #490: [BUG]  derived atoms don't update depended on other atoms which update when router back
I use derived atoms filterTodoList, that depends on filterState, I don't use Provider, when I navigate to other page and go back, 
I can update filterState but filterTodoList doesn't update automatic, this is link for recurrent：https://codesandbox.io/s/jovial-shadow-4qnd8.  
动画

I find in source code:
image
not working!  
image
make `listeners.delete(callback)` and `delAtom(state, atom)` reverse order, it's working!

dai-shi:
close #490

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,43 +1,43 @@
 {
-  "redux.js": {
-    "bundled": 458,
-    "minified": 220,
-    "gzipped": 167,
+  "index.js": {
+    "bundled": 21553,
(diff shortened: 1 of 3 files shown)


dai-shi:
https://codesandbox.io/s/ancient-morning-gvkuk
Fixed!

PR #501, opened 2021-05-27
Update readme.md

Luccasoli:
just making it clearer where countAtom came from in this example

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -45,7 +45,9 @@ const mangaAtom = atom({ 'Dragon Ball': 1984, 'One Piece': 1997, Naruto: 1999 })
 It can be used like `React.useState`:
 
 ```jsx
-import { useAtom } from 'jotai'
+import { atom, useAtom } from 'jotai'
+
+const countAtom = atom(0)
 
 function Counter() {
   const [count, setCount] = useAtom(countAtom)


Thisen:
Not a bad idea. 👍🏼

Thisen (approved):

dai-shi:
@Luccasoli Hey, thanks for the suggestion.
That was actually somewhat intentional. (We want to keep readme compact if possible)
My question is: did you have a difficulty or confusion with the implicit example?
We could probably accept this suggestion, but I'd like to learn your opinion.

Luccasoli:
@dai-shi I like the idea of "copy-paste" examples provided for a library to test, and I believe that when I do this and the code doesn't work it already causes frustration for the dev

So the idea of this PR is to make this example "self-sufficient" to execute

(And I didn't actually use this lib, I just saw it on twitter and I showed up here now to take a look at Readme.md, and I only identified this example that wouldn't run just by copying and pasting)

dai-shi:
> the idea of "copy-paste" examples

Good point. But then, this alone doesn't accomplish the goal.
The current readme is really shaved to reduce size, so copy-paste is mostly impossible.
Let me hold this suggestion for now. We are working on new introduction web site where it would fit better with copy-paste examples idea.

> (And I didn't actually use this lib, I just saw it on twitter and I showed up here now to take a look at Readme.md, and I only identified this example that wouldn't run just by copying and pasting)

That's great to hear! Look forward to hearing more feedback.

Our docs repo is now at https://github.com/pmndrs/website/tree/docs/docs/jotai
where we need more help to improve docs.

Luccasoli:
@dai-shi Got it, and it really seems to have great examples in that repository! I already know then that the documentation site will be great!

I wish success with lib! And thanks for the quick responses, it is certainly a great attraction for any dev to know that those responsible are so active in the repository!

PR #502, opened 2021-05-30
fix: throw error for prohibited write operations

dai-shi:
ref: https://github.com/pmndrs/jotai/issues/231#issuecomment-851075102

I just found it was possible to set values to an atom that has `read` function.
This is not a supported pattern, and should be error to avoid misusing.

### DIFF
--- a/src/core/vanilla.ts
+++ b/src/core/vanilla.ts
@@ -449,13 +449,12 @@ const writeAtomState = <Value, Update>(
       ((a: AnyWritableAtom, v: unknown) => {
         const isPendingPromisesExpired = !pendingPromises.length
         if (a === atom) {
-          if (hasInitialValue(a)) {
-            setAtomValue(state, a, v)
-            invalidateDependents(state, a)
-          } else {
+          if (!hasInitialValue(a)) {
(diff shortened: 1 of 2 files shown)


src/core/vanilla.ts
```diff
@@ -449,6 +449,10 @@ const writeAtomState = <Value, Update>(
       ((a: AnyWritableAtom, v: unknown) => {
         const isPendingPromisesExpired = !pendingPromises.length
         if (a === atom) {
+          if (!hasInitialValue(a)) {
```
Thisen, line 452: Can we enforce this with types aswell?

dai-shi: Oh, I wish I could. The point is `a === atom`.
I think it's impossible, but I'm not sure. There might be a way. You can try as a type puzzle.

---
As a side note, I've been thinking, the jotai api is not super type friendly. It's still JS oriented. Especially function overload is tricky. See what they do in https://github.com/gaku-sei/re-jotai

dai-shi:
Thanks for the comment. Let's merge this in.

PR #503, opened 2021-06-03
fix: remount atoms with intermediates

dai-shi:
While I was looking into an issue discussed in #500, I found a bug.
This fixes it. It turns out that it's not directly related to #500.


Thisen (approved):
LGTM.

PR #505, opened 2021-06-04
fix(query): rewrite atomWithQuery

dai-shi:
resolves the issue discussed in #500.

It turns out that we can't assume onMount is invoked in async.
This is an entire rewrite of the function.


src/query/atomWithQuery.ts
```diff
@@ -22,83 +27,95 @@ export function atomWithQuery<
...
-          }
-          destroy = false
-        }
-      }
-      return [options, initAtom]
+      state.prevData = null
+      const queryClient = get(queryClientAtom)
+      const observer = new QueryObserver(queryClient, options)
+      state.observer?.destroy()
+      observer.subscribe((result) => {
```
Thisen, line 77: Doesn't this return an `unsubscribe`, which needs to be handled?

src/query/atomWithQuery.ts
```diff
@@ -22,83 +27,95 @@ export function atomWithQuery<
...
+        state.handle(result)
+      })
+      state.observer = observer
+      return { dataAtom, options }
     },
-    (get, set, action) => {
-      if (action.type === 'refetch') {
-        const [options] = get(queryAtom)
+    (get, set, action: Action | { type: 'mount' }) => {
```
Thisen, line 83: Maybe `{ type: 'mount' }` should be a part of the type aswell?

src/query/atomWithQuery.ts
```diff
@@ -22,83 +27,95 @@ export function atomWithQuery<
...
+      if (action.type === 'mount') {
```
Thisen, line 84: If you use a reducer type pattern, might aswell use a switch

Thisen (changes requested):

src/query/atomWithQuery.ts
dai-shi: We have been using `observer.destroy()`. Is it different?

src/query/atomWithQuery.ts
dai-shi: this is not exposed. oh actually, we should check it...

src/query/atomWithQuery.ts
dai-shi: do you prefer switch?

Thisen: For reducer type code, I do

src/query/atomWithQuery.ts
Thisen: If the observer has listeners, those won't be handled. See here: https://github.com/tannerlinsley/react-query/blob/69d88cca55/src/core/queryObserver.ts#L96

dai-shi: So, if it has listeners, unsubscribe does nothing, but `destroy()` does the job? I'm not sure how this thing is working. We only use one subscription, so it can just be the same. I can change it to `unsubscribe`.

src/query/atomWithQuery.ts
dai-shi: maybe it's harmless.

src/query/atomWithQuery.ts
dai-shi: 0336383

src/query/atomWithQuery.ts
Thisen: I think that's safer tbh.

dai-shi: a794ee0

Thisen (approved):
Looks great!

Aslemammad (commented):
That's great!

dai-shi:
Let me merge this as our next step might require this.

PR #507, opened 2021-06-05
fix(core): write atom promise value

dai-shi:
I happen to notice a failing test and fixed it. I feel like there can be a better fix, though...

### DIFF
--- a/src/core/vanilla.ts
+++ b/src/core/vanilla.ts
@@ -455,8 +455,8 @@ const writeAtomState = <Value, Update>(
           }
           if (v instanceof Promise) {
             const promise = v
-              .then((vv) => {
-                setAtomValue(state, a, vv)
+              .then((resolvedValue) => {
+                setAtomValue(state, a, resolvedValue)
                 invalidateDependents(state, a)
                 flushPending(state)
               })
--- a/src/core/vanilla.ts
+++ b/src/core/vanilla.ts
@@ -161,7 +161,7 @@ const setAtomReadError = <Value>(
   state: State,
   atom: Atom<Value>,
   error: Error,
-  dependencies: Set<AnyAtom>,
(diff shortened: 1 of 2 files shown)


src/core/vanilla.ts
```diff
@@ -161,7 +161,7 @@ const setAtomReadError = <Value>(
   state: State,
   atom: Atom<Value>,
   error: Error,
-  dependencies: Set<AnyAtom>,
+  dependencies?: Set<AnyAtom>,
```
Thisen, line 164: Isn't it better to send in an empty set as argument, when there's no dependencies?

I'm scared we'll forget to parse it at a later point.

src/core/vanilla.ts
```diff
@@ -453,7 +453,25 @@ const writeAtomState = <Value, Update>(
             // NOTE technically possible but restricted as it may cause bugs
             throw new Error('no atom init')
           }
-          setAtomValue(state, a, v)
+          if (v instanceof Promise) {
+            const promise = v
+              .then((vv) => {
```
Thisen, line 458: Maybe some better naming than v, vv and a.

src/core/vanilla.ts
dai-shi: Unfortunately, it can't be an empty set. It's different.
This is already the pattern used everywhere, not just in this PR change.
Let's discuss separately for better readability, when we get time.

src/core/vanilla.ts
dai-shi: Thanks. `a` and `v` are used for other places, so let's tackle it separately. (I agree to fix because it doesn't harm to bundle size. suggestions are welcome.)
For `vv`, let me fix in this PR.

dai-shi:
Let me merge this as we want to move forward to another PR.

PR #508, opened 2021-06-05
feat: drop returning promise by writeAtom/setAtom

dai-shi:
The API is confusing especially in TypeScript and the implementation is too tricky.
We remain error handling as `TODO`.

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -25,8 +25,8 @@ export type Scope = symbol | string | number
 
 // Are there better typings?
 export type SetAtom<Update> = undefined extends Update
-  ? (update?: Update) => void | Promise<void>
-  : (update: Update) => void | Promise<void>
+  ? (update?: Update) => void
+  : (update: Update) => void
 
 type OnUnmount = () => void
 type OnMount<Update> = <S extends SetAtom<Update>>(
--- a/src/core/vanilla.ts
+++ b/src/core/vanilla.ts
@@ -396,129 +396,80 @@ const invalidateDependents = <Value>(state: State, atom: Atom<Value>): void => {
 const writeAtomState = <Value, Update>(
   state: State,
   atom: WritableAtom<Value, Update>,
-  update: Update,
-  pendingPromises: Promise<void>[]
+  update: Update
 ): void => {
-  const isPendingPromisesExpired = !pendingPromises.length
   const writePromise = getAtomState(state, atom)?.w
   if (writePromise) {
-    const promise = writePromise.then(() => {
-      writeAtomState(state, atom, update, pendingPromises)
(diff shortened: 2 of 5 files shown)


src/core/atom.ts
```diff
@@ -25,8 +25,8 @@ export type Scope = symbol | string | number
 
 // Are there better typings?
 export type SetAtom<Update> = undefined extends Update
-  ? (update?: Update) => void | Promise<void>
-  : (update: Update) => void | Promise<void>
+  ? (update?: Update) => void
```
Thisen, line 28: Arent we losing async for write then? Maybe I'm missing something.

src/core/vanilla.ts
```diff
@@ -396,129 +396,80 @@ const invalidateDependents = <Value>(state: State, atom: Atom<Value>): void => {
...
   const writePromise = getAtomState(state, atom)?.w
   if (writePromise) {
-    const promise = writePromise.then(() => {
-      writeAtomState(state, atom, update, pendingPromises)
-      if (isPendingPromisesExpired) {
-        flushPending(state)
-      }
+    writePromise.then(() => {
+      writeAtomState(state, atom, update)
+      flushPending(state)
```
Thisen, line 405: Don't we need to guard for expired promises anymore?

src/core/vanilla.ts
```diff
@@ -396,129 +396,80 @@ const invalidateDependents = <Value>(state: State, atom: Atom<Value>): void => {
...
+        throw aState.p // read promise
+      }
+      if ('v' in aState) {
+        return aState.v // value
+      }
+      if (
+        typeof process === 'object' &&
+        process.env.NODE_ENV !== 'production'
+      ) {
+        console.warn(
```
Thisen, line 434: Typo: this is probably a bug.

src/core/atom.ts
dai-shi: The behavior doesn't change. This is more like `dispatch`ing `update`.

src/core/vanilla.ts
dai-shi: No, because we don't return a promise. There's no concept of expiration.
So, this was really confusing in code. Nobody understands what's going on. Hence I'm willing to make this change before v1, and if we have real requirement, let's consider a possible solution with better code.

src/core/vanilla.ts
dai-shi: oops. thanks. it has been a typo for a long time.

Thisen (approved):
LGTM

dai-shi:
Let me merge this to move forward.

@aulneau you are still encouraged to try this to see if it works for your use case.

Issue #434, opened 2021-04-18
async read atoms keep waiting the first promise to be resolved and can't be overriden by the second promise


dai-shi:
from: https://github.com/pmndrs/jotai/issues/430#issuecomment-822082692

repro: https://codesandbox.io/s/winter-bash-rp2rr?file=/src/App.js

PR #509, opened 2021-06-06
feat: mount dependencies on commit

dai-shi:
This is a big change. It's something that has been concerning me since introducing provider-less mode refactoring, which does some "side effects" in the render phase.
This should fix it. It now runs `mountDependencies` in the commit phase.
This drops what I added in #434, which was not possible in this new method.


Aslemammad (commented):
Looks great to me, Do all the tests we have cover this (big) change?

dai-shi:
> Looks great to me, Do all the tests we have cover this (big) change?

I need to change tests with atoms snapshot because of timing (not 100% sure).
Also, one test #434 is dropped because it's not valid usage per se.

Aslemammad:
@dai-shi Great!

src/core/contexts.ts
```diff
@@ -15,12 +15,14 @@ type UpdateAtom = <Value, Update>(
 
 type StoreForProduction = [
   stateMutableSource: MutableSource<State>,
-  updateAtom: UpdateAtom
+  updateAtom: UpdateAtom,
+  commit: () => void
```
Thisen, line 19: Maybe type for Commit callback

src/core/useAtom.ts
```diff
@@ -61,8 +61,9 @@ export function useAtom<Value, Update>(
   )
 
   const StoreContext = getStoreContext(atom.scope)
-  const [mutableSource, updateAtom] = useContext(StoreContext)
+  const [mutableSource, updateAtom, commit] = useContext(StoreContext)
   const value: Value = useMutableSource(mutableSource, getAtomValue, subscribe)
+  useEffect(() => commit())
```
Thisen, line 66: What if it's unmounting, while calling commit? You might need a guard here.

src/core/contexts.ts
dai-shi: Do you have a better name? `Commit` can be too simple.

src/core/useAtom.ts
dai-shi: if this atom is unmouting, it doesn't do anything. I'm not sure if I get your question.
btw, we should explicitly ignore the return value of commit here.

Thisen: If the effect runs while it's unmounting, it will throw something like this: 
```
Warning: Can't perform a React state update on an unmounted component.
This is a no-op, but it indicates a memory leak in your application.
```

src/core/contexts.ts
Thisen: `CommitCallback`

src/core/useAtom.ts
dai-shi: No worries about it. uMS handles it already.

dai-shi:
Merging this. If anyone notices anything, please let me know even afterward.

PR #513, opened 2021-06-11
fix: handle update on commit

### LINKED ISSUE #512: atomFamily not updating with param from another atom
I've got a derived atomFamily that uses a string atom as a parameter in another atomFamily. The problem is that when the first atomFamily updates, the derived atomFamily doesn't seem to notice.

Here's the full reproduction, which can be pasted into App.tsx of a project created with `npx create-react-app jotclock --template typescript`.

This app just has an atomFamily of clock times, and an atom with the ID of the selected clock. When you press 'run test' it just updates the time of one clock on an interval and displays it onscreen, but the atomFamily atom doesn't re-run so the timestamp doesn't update.

```typescript
import { atom } from "jotai";
import { useUpdateAtom, atomFamily, useAtomValue } from "jotai/utils";

import { useCallback, useEffect } from "react";

type ClockId = string;
type Milliseconds = number;

// This atom is true if the test is running
const runningAtom = atom<boolean>(false);

// The ID of the selected clock
// NOTE: The issue does not reproduce if we have the initial value as: "a"
const selectedClockIdAtom = atom<ClockId | null>(null);

// The current time shown on each clock
const clockTimeFamily = atomFamily<
  ClockId,
  Milliseconds | null,
  Milliseconds | null
>(() => atom<Milliseconds | null>(null));

//
// The time of the selected clock.
//
// THE ISSUE IS HERE:
// It updates whenever selectedClockIdAtom changes, but it does NOT update
// whenever the clockTimeFamily atoms have changed, even if it has already
// looked at that exact atom (and should have it in its dependencies).
//
const selectedClockTimeAtom = atom<Milliseconds | null>((get) => {
  // this code doesn't run as often as it should

  const clockId = get(selectedClockIdAtom);
  if (!clockId) {
    return null;
  }
  console.log("derived atom clockId", clockId);
  console.log(
    "derived atom selected clock time atom",
    clockTimeFamily(clockId)
  );

  const selectedClockTime = get(clockTimeFamily(clockId));
  console.log("derived atom selected clock time", selectedClockTime);
  return selectedClockTime;
});

// if test is running, update the clock time every 100ms on the given clock ID
function Incrementer({ clockId }: { clockId: ClockId }) {
  const running = useAtomValue(runningAtom);
  const setTime = useUpdateAtom(clockTimeFamily(clockId));

  // set the time of this clock to the current time
  const setCurrentTime = useCallback(() => {
    const now = Date.now();
    console.log(`setting clockTimeFamily[${clockId}] atom to:`, now);
    setTime(now);
  }, [setTime, clockId]);

  // if the test is running, periodically call setCurrentTime
  useEffect(() => {
    if (!running) {
      return;
    }
    const intervalId = setInterval(setCurrentTime, 100);
    return () => {
      clearInterval(intervalId);
    };
  }, [setCurrentTime, running]);

  return null;
}

// display running and current player time
function Display() {
  const time = useAtomValue(selectedClockTimeAtom);
  const running = useAtomValue(runningAtom);
  return (
    <div>
      <p>Running: {running ? "yes" : "no"}</p>
      <p>Time: {time}</p>
    </div>
  );
}

  const setRunning = useUpdateAtom(runningAtom);
  const setSelectedClockIdAtom = useUpdateAtom(selectedClockIdAtom);

  const runTest = useCallback(() => {
    setRunning((x) => !x);
  }, [setRunning]);

  useEffect(() => {
    // this will trigger selectedClockTimeAtom to re-evaluate once, at which point it will
    // get clockTimeFamily("a") ONE TIME but it won't re-run when clockTimeFamily("a") changes
    console.log("setting selected clock ID to 'a'");
    setSelectedClockIdAtom("a");
  });

  return (
    <div className="App">
      <button onClick={runTest}>Run Test</button>
      <Incrementer clockId="a" />
      <Display />
    </div>
  );
}
```

dai-shi:
close #512.

Most likely, this is a regression with #509.

### DIFF
--- a/src/core/vanilla.ts
+++ b/src/core/vanilla.ts
@@ -596,7 +596,9 @@ const commitAtomState = <Value>(
 }
 
 export const flushPending = (state: State): void => {
-  state.p.forEach((prevDependencies, atom) => {
+  const pending = [...state.p]
+  state.p.clear()
+  pending.forEach(([atom, prevDependencies]) => {
     const atomState = getAtomState(state, atom)
     if (atomState) {
       if (prevDependencies) {
@@ -611,7 +613,6 @@ export const flushPending = (state: State): void => {
     const mounted = state.m.get(atom)
     mounted?.l.forEach((listener) => listener())
   })
-  state.p.clear()
 }
 
 export const subscribeAtom = (
--- a/tests/dependency.test.tsx
+++ b/tests/dependency.test.tsx
@@ -539,3 +539,62 @@ it('can remount atoms with intermediate atom', async () => {
     getByText('derived: 8')
   })
 })
+
+it('can update dependents with useEffect (#512)', async () => {
+  const enabledAtom = atom(false)
+  const countAtom = atom(1)
+
+  const derivedAtom = atom((get) => {
+    const enabled = get(enabledAtom)
+    if (!enabled) {
+      return 0
+    }
+    const count = get(countAtom)
+    return count * 2
+  })
+
+  const Counter: React.FC = () => {
+    const [count, setCount] = useAtom(countAtom)
+    return (
+      <>
+        <div>count: {count}</div>
+        <button onClick={() => setCount((c) => c + 1)}>button</button>
+      </>
+    )
+  }
+
+  const DerivedCounter: React.FC = () => {
+    const [derived] = useAtom(derivedAtom)
+    return <div>derived: {derived}</div>
+  }
+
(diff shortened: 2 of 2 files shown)


dai-shi:
https://codesandbox.io/s/strange-voice-84bzc
Fixed!
Thanks @ahfarmer for reporting the issue.

Thisen (approved):

PR #515, opened 2021-06-11
examples/todos ts version

tell-y:
I created ts version of examples/todos.

codesandbox which is created by the branch:
https://codesandbox.io/s/jotai-examplestodos-ts-version-0kmsw

Thank you for your review.

### DIFF
--- a/examples/todos/src/App.tsx
+++ b/examples/todos/src/App.tsx
@@ -1,12 +1,17 @@
-import * as React from 'react'
-import { Provider, atom, useAtom } from 'jotai'
+import { FC, FormEvent } from 'react'
+import { Provider, atom, useAtom, PrimitiveAtom } from 'jotai'
 import { Radio } from 'antd'
 import { CloseOutlined } from '@ant-design/icons'
 import { a, useTransition } from '@react-spring/web'
 
+type Todo = {
+  title: string
+  completed: boolean
+}
+
 const filterAtom = atom('all')
-const todosAtom = atom([])
-const filteredAtom = atom((get) => {
+const todosAtom = atom<PrimitiveAtom<Todo>[]>([])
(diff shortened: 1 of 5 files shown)


examples/todos/src/App.tsx
```diff
@@ -1,12 +1,17 @@
...
+type Todo = {
+  title: string
+  completed: boolean
+}
+
 const filterAtom = atom('all')
-const todosAtom = atom([])
-const filteredAtom = atom((get) => {
+const todosAtom = atom<PrimitiveAtom<Todo>[]>([])
+const filteredAtom = atom<PrimitiveAtom<Todo>[]>((get) => {
```
dai-shi, line 14: This is not correct, because it's read-only atom. Doesn't it error?
It should work without explicit annotation.
```suggestion
const filteredAtom = atom((get) => {
```

Or, if you want to, the following:
```suggestion
const filteredAtom = atom<Atom<Todo>[]>((get) => {
```

dai-shi:
https://githubbox.com/tell-y/jotai/tree/feat/todos-ts/examples/todos
The behavior looks good.

examples/todos/src/App.tsx
tell-y: todosAtom type is Atom<PrimitiveAtom<Todo>[]>, so it seems filteredAtom is also Atom<PrimitiveAtom<Todo>[]> ? 🤔 
image

tell-y:
> FAIL tests/utils/waitForAll.test.tsx
Why???

examples/todos/src/App.tsx
dai-shi: Sorry, that was my bad. I was confused with the filteredAtom type.

dai-shi (approved):
Passed tests. Was weird.
Thanks for your contribution!

PR #518, opened 2021-06-14
fix: flush pending undefined error

### LINKED ISSUE #517: [Bug] atom state not found in flush undefined (introduced in 0.16.11)
I had been using 0.16.7 for a while, and decided to upgrade my dependencies tonight.
This updated my app to 0.16.11, where Jotai started failing left and right for me with multiple errors popping up in the console: 
```
[Bug] atom state not found in flush undefined
```
I reverted back to 0.16.7 and started climbing the release versions to isolate the culprit.
0.16.10 still works fine for me... it seems 0.16.11 is problematic in my situation.
Let me know if you need more details.

dai-shi:
close #517

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 20625,
-    "minified": 9844,
-    "gzipped": 3195,
+    "bundled": 20632,
+    "minified": 9851,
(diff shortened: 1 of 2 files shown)


dai-shi:
I confirmed, locally with removing `"module"` in package.json, that this PR fixes the issue.

PR #526, opened 2021-06-19
chore(jotai/query): refactor atomWithQuery

dai-shi:
I found `useQuery` in react-query uses `fetchOptimistic` for suspense.
This is undocumented, but it should solve some potential issues in atomWithQuery.
(Previously, there's a risk of memory leak. I'm so glad to get rid of stateAtom which uses mutable state.)

### DIFF
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -40,7 +40,7 @@ export function atomWithQuery<
         })
       )
       let setData: (data: TData) => void = () => {
-        throw new Error('setting data without mount')
+        throw new Error('atomWithQuery: setting data without mount')
       }
       let prevData: TData | null = null
       const listener = (result: QueryObserverResult<TData, TError>) => {
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -8,7 +8,7 @@ import { atom } from 'jotai'
 import type { WritableAtom, Getter } from 'jotai'
 import { queryClientAtom } from './queryClientAtom'
 
-type Action = { type: 'refetch' } | { type: 'cleanup' }
+type Action = { type: 'refetch' }
 
 type AtomQueryOptions<TQueryFnData, TError, TData, TQueryData> =
   QueryObserverOptions<TQueryFnData, TError, TData, TQueryData> & {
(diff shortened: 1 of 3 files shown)


aulneau:
@dai-shi could you also export the `AtomQueryOptions` types? I have a `atomFamilyWithQuery` that i maintain for my projects and would like to leverage the type

src/query/atomWithQuery.ts
```diff
@@ -28,76 +28,60 @@ export function atomWithQuery<
...
+      const options =
+        typeof createQuery === 'function' ? createQuery(get) : createQuery
+      let resolve: ((data: TData) => void) | null = null
+      const dataAtom = atom<TData | Promise<TData>>(
+        new Promise<TData>((r) => {
+          resolve = r
+        })
+      )
+      let setData: (data: TData) => void = () => {
+        throw new Error('setting data without mount')
```
aulneau, line 43:
```suggestion
        throw new Error('atomWithQuery: setting data without mount')
```

aulneau: I think the more intention we put into these, the more helpful it will be for users.

When is it likely this error will happen?

src/query/atomWithQuery.ts
```diff
@@ -28,76 +28,60 @@ export function atomWithQuery<
...
-      const options =
-        typeof createQuery === 'function' ? createQuery(get) : createQuery
-      const state = get(stateAtom)
-      const dataAtom = atom<TData | Promise<TData>>(
-        new Promise<TData>((resolve) => {
-          state.resolve = resolve
+      }
+      const defaultedOptions = queryClient.defaultQueryObserverOptions(options)
+      if (typeof defaultedOptions.staleTime !== 'number') {
+        defaultedOptions.staleTime = 1000
```
aulneau, line 64: why is this important?

src/query/atomWithQuery.ts
```diff
@@ -28,76 +28,60 @@ export function atomWithQuery<
...
+      }
+      const observer = new QueryObserver(queryClient, defaultedOptions)
+      observer
+        .fetchOptimistic(defaultedOptions)
+        .then(listener)
+        .catch(() => {
+          // TODO error handling
         })
```
aulneau, line 72: can we pass `options.onError` here?
what happens when folks use the `useErrorBoundary` react query option?

src/query/atomWithQuery.ts
dai-shi: It's meant for detecting bugs, not really for users.

Hm, it could happen when you refetch before mount. So, users would actually sees it if they invoke `refetch` in render. (we could avoid the error by setting `setData` before refetch, but that would only cause infinite loop. 🤔 Maybe the suggestion is an easy fix.)

src/query/atomWithQuery.ts
dai-shi: Without this, it will invoke fetchFn again on mount. `1000` is just a rough estimate, I guess. I just took the value from `useQuery`.

src/query/atomWithQuery.ts
dai-shi: So, my question is what we should do with error handling in `atomWithQuery`.
Should we follow `onError` or `ErrorBoundary` or both?

src/query/atomWithQuery.ts
```diff
@@ -28,76 +28,60 @@ export function atomWithQuery<
...
-      state.prevData = null
-      const queryClient = get(queryClientAtom)
-      const observer = new QueryObserver(queryClient, options)
-      state.unsubscribe?.()
-      state.unsubscribe = observer.subscribe((result) => {
-        state.handle(result)
-      })
+      dataAtom.onMount = (update) => {
+        setData = update
+        const unsubscribe = observer.subscribe(listener)
```
Thisen, line 75: Can this throw? If so, it needs error handling.

Thisen (approved):

src/query/atomWithQuery.ts
Thisen, line 75: Can this throw? If so, it needs error handling.

dai-shi: Looking at RQ code, it looks okay.
https://github.com/tannerlinsley/react-query/blob/9aed040364796cdd5cab0f53123e02a6c6f769de/src/react/useBaseQuery.ts#L85

src/query/atomWithQuery.ts
dai-shi: Let's work on error handling separately. I need to understand how it should behave to implement error handling.

PR #532, opened 2021-06-19
refactor(devtools): remove unused code

Gusted:
`const action` is never used and removing that make `const actions` useless.


dai-shi (approved):
Thanks for the refactor.

PR #537, opened 2021-06-20
fix(utils/atomWithDefault): support refresh

### LINKED ISSUE #521: async set atom remains cached
I have an async atom that looks like follows

```js
export const someAsyncAtom = atom(
    async (get) => {
        const {data} = await getData();
        return data;
    },
    async (get, set, update) => {
        if (update) {
            set(someAsyncAtom, update);
        } else {
            const {data} = await getData();
            set(someAsyncAtom, data);
        }
    }
);
```

Now up until `v0.16.8` if I wanted to imperatively refresh my data I could use something like

```js
const updateData = useUpdateAtom(someAsyncAtom)
```

However starting with `v.0.16.9` after calling `updateData` every component that has 

```js
const data = useAtomValue(someAsyncAtom)
```

has now stale data that dates back to before calling the imperative refresh. Any ideas? This is a breaking change for our application.

dai-shi:
close #521 

- [x] impl
- [x] tests
- [x] docs https://github.com/pmndrs/website/pull/108


dai-shi:
@fkocovski Would you try this? (you can npm/yarn install the codesandbox build.)

LucaColonnello:
@dai-shi I tested this with different scenarios [here](https://codesandbox.io/s/jotai-atomwithdefault-refresh-ceuw9).

**Tested the following:**

- dependency tracking refreshing default value (every time you click the `Change Id` button, which changes a dependency)
- default value (async in this case)
- set override value (done in the useEffect)
- refresh the default value, removing the override (which calls again the getDefault function)

It seems to work perfectly fine for me in all these scenario

dai-shi:
@LucaColonnello

Great to hear! and the csb example looks good too.
Would you be willing to contribute to this PR? There are two remaining tasks.

LucaColonnello:
Hey @dai-shi,
I'd love to yeah!

I just need to familiarise myself with the testing setup here.

I'll look around and give it a try.

Is there a discord or slack channel (or equivalent) I can join if I get stuck where I can ask for help? (I can just write here otherwise)

dai-shi:
> Is there a discord

Yes, please find an invitation link in readme.
There's `#jotai` channel in the Poimandres discord server. You can DM me too.

LucaColonnello:
Amazing thank you!

I'll let you know how it goes, I think I have an idea based on the existing atomWithDefault tests I've seen.

LucaColonnello:
About the above failing test, I wasn't sure it was an environment issue (my local machine), but the `waitForAll` tests seem to be intermittently failing, and it seems unrelated to the changes on `atomWithDefault`.

dai-shi:
> the waitForAll tests seem to be intermittently failing

Yeah, I've seen this a few times. @Thisen can you investigate it? (but the reproduction is gone...)

dai-shi:
Thanks to @LucaColonnello , this is ready for review and merge. Anyone, feel free to leave some comments.

dai-shi:
> > the waitForAll tests seem to be intermittently failing
> 
> Yeah, I've seen this a few times. @Thisen can you investigate it? (but the reproduction is gone...)

Okay, this is probably my bad. I commented out `jest.useFakeTimers()`.
@Thisen Will you still help me? How can we comment this in?

Thisen:
Should be fixed now.

Thisen (approved):

PR #540, opened 2021-06-22
fix(utils): improve atomWithStorage covering atomWithHash

dai-shi:
This improves atomWithStorage.

- support optional `subscribe` function.
- fix a behavior of dual render on mount.
- add optional `delayInit` boolean to update storage value on mount.
- re-implement `atomWithHash` with the new atomWithStorage using `subscribe`.


dai-shi:
The `subscribe` option should allow @TheoBr 's `atomWithSyncedLocalStorage` use case too.
https://gist.github.com/TheoBr/b558fd85e671395620d41fece5e2334c

```js
const syncedLocalStorage = {
  getItem: (key) => {
    const storedValue = localStorage.getItem(key)
    if (storedValue === null) {
      throw new Error('no value stored')
    }
    return JSON.parse(storedValue)
  },
  setItem: (key, newValue) => {
    localStorage.setItem(key, JSON.stringify(newValue))
  },
  subscribe: (key, callback) => {
    const listener = (e) => {
      if (e.key === key && e.newValue) {
        callback(JSON.parse(e.newValue))
      }
    }
    window.addEventListener("storage", listener)
    return () => widnow.removeEventListener("storage", listener)
  }
}

const atomWithSyncedLocalStorage = (key, initialValue) => atomWithStorage(key, initialValue, syncedLocalStorage)
```

(hm, does it cause dual render on update?)

aulneau (approved):
This is great, really love the addition of the subscription element. I think it'd be cool to have an `atomWithSubscription` util someday

dai-shi:
Would `atomWithSubscription` be just `atom.onMount`?

Anyway, let's merge this.

PR #543, opened 2021-06-22
fix: disable useDebugState in test env

aulneau:
While starting to test our state, I noticed a warning from jest/hooks-testing-library about needing to wrap a function in Provider with `act`. This is the change @dai-shi recommended and I've tested it locally.


Thisen (approved):

dai-shi (approved):
Thanks! I think this is reasonable.

PR #547, opened 2021-06-23
fix(query): queryClientAtom for initialValues

dai-shi:
With #516, setting initialValue in Provider for derived atoms is not recommended.
This PR provides a primitive atom for queryClient. This allows to set the atom in useEffect too.


aulneau:
Can you provide examples of how to use this and why it's better?

dai-shi:
Well, the usage is not changed and not documented.
This *fixes* for something that used work before #516.

dai-shi:
I know there are doc issues (help wanted), but let me merge.

PR #548, opened 2021-06-23
feat(query): atomWithQuery to use options.initialData to avoid suspense

aulneau:
This updates the initial atom used for data for the query atom to be `atomWithDefault` and uses the initialData value if present for that. This is helpful because if data is passed, the query atom will not be suspense/async by default. This is _very_ helpful in the context of SSR where suspense is not well supported yet.

@dai-shi curious what you think, I've tested this locally and it works well. The last thing I'd like to do is pass `get` to the initialData function, what do you think?

Also this type error I am not sure how to fix yet:
```
src/query/atomWithQuery.ts(42,23): error TS2349: This expression is not callable.
  Not all constituents of type 'InitialDataFunction<TQueryData> | (TQueryData & Function)' are callable.
    Type 'TQueryData & Function' has no call signatures.
```

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,43 +1,29 @@
 {
-  "index.js": {
-    "bundled": 21126,
-    "minified": 10113,
-    "gzipped": 3334,
+  "valtio.js": {
+    "bundled": 1235,
+    "minified": 598,
+    "gzipped": 335,
+    "treeshaked": {
+      "rollup": {
+        "code": 37,
+        "import_statements": 37
+      },
+      "webpack": {
+        "code": 1054
+      }
+    }
+  },
+  "redux.js": {
+    "bundled": 458,
+    "minified": 220,
+    "gzipped": 167,
     "treeshaked": {
       "rollup": {
         "code": 14,
         "import_statements": 14
       },
       "webpack": {
-        "code": 1220
-      }
-    }
-  },
-  "utils.js": {
-    "bundled": 13390,
-    "minified": 6561,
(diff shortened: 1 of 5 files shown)


dai-shi:
> error TS2349: This expression is not callable.

Currently, it's TS limitation. https://github.com/microsoft/TypeScript/issues/37663

dai-shi:
@aulneau Pushed a commit.

Note that this breaks the note in the official docs.
https://react-query.tanstack.com/reference/useQuery
> If set to a function, the function will be called **once** during the shared/root query initialization, and be expected to synchronously return the initialData

It will be **twice**.

aulneau:
Nice, just tested this out and works the same as what I had. 

For my own curiosity, why did you prefer to do this over the atomWithDefault?

dai-shi:
atomWithDefault has two atoms internally (which overlaps with `dataAtom` here), and does potentially extra work.

Furthermore, the grand rule of mine is `jotai/*` only depends on `jotai`. Something more complex should be out of the scope. They would be 3rd party libs or recipes in docs.

dai-shi (approved):
If you are fine with breaking the official note. It's OK for me.

aulneau:
Ah makes sense!

I'm fine with it, I personally have never used the function version of initial data. I'd be curious to see how it might be problematic if it fires twice. Let's cross that bridge when it if we come to it.

Thanks for updating the implementation!

aulneau:
@dai-shi would a merge be possible soon? Thanks 🙏

PR #552, opened 2021-06-23
breaking(utils): drop REFRESH symbol in atomWithDefault and use common RESET symbol

### LINKED ISSUE #549: useResetAtom does not work on atomWithDefault
`atomWithDefault` can be reset by sending the `REFRESH` symbol.
`atomWithReset` can be reset by sending the `RESET` symbol or with `useResetAtom`.

`useResetAtom` does not work on `atomWithDefault`.

Should we get rid of the `REFRESH` symbol, and instead use the `RESET` symbol everywhere including `atomWithDefault`? Then `useResetAtom` would work on `atomWithDefault`.

dai-shi:
close #549


LucaColonnello (approved):
Amazing!!

Neat pick: the RESET Symbol could come from a constants file, so we won't need to import it from the useResetAtom, which sounds like a consumer rather than an owner for such a constant.

It will also avoid us to depend on useResetAtom for every feature that requires the Symbol to be compatible, making it less coupled.

Other than that, LGTM!!

dai-shi:
Thanks! Done.

LucaColonnello (approved):

PR #554, opened 2021-06-24
fix(types): overload for atom(null | undefined)

### LINKED ISSUE #550: null/undefined initial atom value confuses typescript
Whenever I set null or undefined as the initial value on an atom, TypeScript infers the type as `Atom` instead of `WritableAtom`.

Working case: initial value is `false` so type is shown as `WritableAtom`:
Screen Shot 2021-06-23 at 3 12 58 PM

Non-working case: initial value is `null` so type is just `Atom`:

Screen Shot 2021-06-23 at 3 13 22 PM

This shows up as a type error on `useUpdateAtom`:
Screen Shot 2021-06-23 at 3 13 15 PM

My current workaround: caste to `PrimitiveAtom`:
```typescript
const myAtomConfig = atom<boolean | null>(null) as PrimitiveAtom<
  boolean | null
>;
```

ahfarmer:
Fixes #550

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -75,6 +75,14 @@ export function atom<Value, Update>(
   ? never
   : WritableAtom<Value, Update> & WithInitialValue<Value>
 
+// primitive atom, specifically handling null | undefined
+// to avoid misclassification as read-only derived atom
+export function atom<Value extends unknown>(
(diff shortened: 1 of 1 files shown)


dai-shi:
https://github.com/pmndrs/jotai/issues/550#issuecomment-867977452

PR #564, opened 2021-06-27
chore(core): flushPending just once on write

dai-shi:
While working on https://github.com/Thisen/js-framework-benchmark/pull/5,
I found flushPending is invoked multiple times.
Not sure how it would affect performance, but I feel it's better to run flushPending at the end or writeAtom.

The difference in this PR is shown as the comparison between B and C.

image

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 21126,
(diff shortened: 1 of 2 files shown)


dai-shi:
Hmm, this doesn't pass the tests. Something doesn't work with async. Giving up for now.

PR #573, opened 2021-06-30
fix(query): use observer for refetch

aulneau:
This takes the pattern of using the `observer` from the `atomWithInfiniteQuery` implementation to refetch vs the queryClient method that was used previously. Suspense behavior has not been impacted.

### DIFF
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -98,16 +98,19 @@ export function atomWithQuery<
         const unsubscribe = observer.subscribe(listener)
         return unsubscribe
       }
-      return { dataAtom, options }
+      return {
+        dataAtom,
+        observer,
+      }
     },
     (get, set, action: AtomWithQueryAction) => {
       switch (action.type) {
         case 'refetch': {
-          const { dataAtom, options } = get(queryDataAtom)
+          const { dataAtom, observer } = get(queryDataAtom)
           set(dataAtom, new Promise<TData>(() => {})) // infinite pending
-          const queryClient = get(getQueryClientAtom)
-          queryClient.getQueryCache().find(options.queryKey)?.reset()
-          const p: Promise<void> = queryClient.refetchQueries(options.queryKey)
+          const p: Promise<void> = observer
+            .refetch({ cancelRefetch: true })
+            .then(() => {})
           return p
         }
       }
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -6,7 +6,7 @@ import {
   InitialDataFunction,
 } from 'react-query'
 import { atom } from 'jotai'
-import type { WritableAtom, Getter } from 'jotai'
+import type { WritableAtom, PrimitiveAtom, Getter } from 'jotai'
 import { getQueryClientAtom } from './queryClientAtom'
 
 export type AtomWithQueryAction = { type: 'refetch' }
@@ -29,7 +29,13 @@ export function atomWithQuery<
       ) => AtomWithQueryOptions<TQueryFnData, TError, TData, TQueryData>),
   equalityFn: (a: TData, b: TData) => boolean = Object.is
 ): WritableAtom<TData | TQueryData, AtomWithQueryAction> {
-  const queryDataAtom = atom(
+  const queryDataAtom: WritableAtom<
+    {
+      dataAtom: PrimitiveAtom<TData | TQueryData | Promise<TData | TQueryData>>
+      observer: QueryObserver<TQueryFnData, TError, TData, TQueryData>
+    },
+    AtomWithQueryAction
+  > = atom(
     (get) => {
       const queryClient = get(getQueryClientAtom)
       const options =
@@ -98,19 +104,14 @@ export function atomWithQuery<
         const unsubscribe = observer.subscribe(listener)
         return unsubscribe
       }
-      return {
-        dataAtom,
-        observer,
-      }
+      return { dataAtom, observer }
     },
     (get, set, action: AtomWithQueryAction) => {
       switch (action.type) {
         case 'refetch': {
           const { dataAtom, observer } = get(queryDataAtom)
           set(dataAtom, new Promise<TData>(() => {})) // infinite pending
-          const p: Promise<void> = observer
-            .refetch({ cancelRefetch: true })
-            .then(() => {})
+          const p = observer.refetch({ cancelRefetch: true }).then(() => {})
           return p
         }
       }
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -70,9 +70,9 @@
     }
   },
   "query.js": {
-    "bundled": 5246,
-    "minified": 2261,
-    "gzipped": 694,
+    "bundled": 5170,
+    "minified": 2240,
+    "gzipped": 686,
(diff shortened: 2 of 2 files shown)


src/query/atomWithQuery.ts
```diff
@@ -98,16 +98,17 @@ export function atomWithQuery<
...
       switch (action.type) {
         case 'refetch': {
-          const { dataAtom, options } = get(queryDataAtom)
+          const { dataAtom, observer, options } = get(queryDataAtom)
           set(dataAtom, new Promise<TData>(() => {})) // infinite pending
-          const queryClient = get(getQueryClientAtom)
-          queryClient.getQueryCache().find(options.queryKey)?.reset()
-          const p: Promise<void> = queryClient.refetchQueries(options.queryKey)
+          const wrappedPromise = async () => {
+            await observer.refetch({ cancelRefetch: true })
```
aulneau, line 109: > If `cancelRefetch` is true, then the current request will be cancelled before a new request is made

seems to make sense?

more context: https://react-query.tanstack.com/reference/useQuery#_top

aulneau: @tkdodo curious what you think?

src/query/atomWithQuery.ts
```diff
@@ -98,16 +98,17 @@ export function atomWithQuery<
...
         return unsubscribe
       }
-      return { dataAtom, options }
+      return { dataAtom, observer, options }
     },
     (get, set, action: AtomWithQueryAction) => {
       switch (action.type) {
         case 'refetch': {
-          const { dataAtom, options } = get(queryDataAtom)
+          const { dataAtom, observer, options } = get(queryDataAtom)
```
dai-shi, line 106: Do we still use `options`?

src/query/atomWithQuery.ts
```diff
@@ -98,16 +98,17 @@ export function atomWithQuery<
...
           set(dataAtom, new Promise<TData>(() => {})) // infinite pending
-          const queryClient = get(getQueryClientAtom)
-          queryClient.getQueryCache().find(options.queryKey)?.reset()
-          const p: Promise<void> = queryClient.refetchQueries(options.queryKey)
+          const wrappedPromise = async () => {
```
dai-shi, line 108: Please be reminded that in jotai code, there's no `async` anywhere. 🤯 
(well, you should know because you did all bundling tooling.)

If `const p: Promise<void> = ...` hack doesn't work, please use `Promise.resolve(...)`.

src/query/atomWithQuery.ts
dai-shi: seems to make sense from jotai point of view, but I'm not sure how RQ cache works. There might be a reason why it's not default.

dai-shi (commented):
thanks for working on this. overall looks good.

src/query/atomWithQuery.ts
```diff
@@ -98,16 +98,22 @@ export function atomWithQuery<
...
           set(dataAtom, new Promise<TData>(() => {})) // infinite pending
-          const queryClient = get(getQueryClientAtom)
-          queryClient.getQueryCache().find(options.queryKey)?.reset()
-          const p: Promise<void> = queryClient.refetchQueries(options.queryKey)
+          const p: Promise<void> = new Promise((resolve, reject) =>
+            observer
+              .refetch({ cancelRefetch: true })
+              .then(() => resolve())
+              .catch((error) => {
+                set(dataAtom, Promise.reject<TData>(error))
```
dai-shi, line 113: Hm, are you sure we need this? You mean L69 doesn't work for refetch?

You don't need `new Promise`, something like this would do.
```js
return observer.refetch({ cancelRefetch: true }).catch((error) => {
  set(dataAtom, Promise.reject<TData>(error))
})
```
or
```js
return observer.refetch({ cancelRefetch: true }).then(() => {}).catch((error) => {
  set(dataAtom, Promise.reject<TData>(error))
})
```

aulneau: should be fixed, and I removed the error setting, was not needed

src/query/atomWithQuery.ts
aulneau: removed!

src/query/atomWithQuery.ts
```diff
@@ -98,16 +98,19 @@ export function atomWithQuery<
         const unsubscribe = observer.subscribe(listener)
         return unsubscribe
       }
-      return { dataAtom, options }
+      return {
+        dataAtom,
+        observer,
+      }
```
dai-shi, line 104:
```suggestion
      return { dataAtom, observer }
```
Does prettier complain with this?

src/query/atomWithQuery.ts
```diff
@@ -98,16 +98,19 @@ export function atomWithQuery<
...
         case 'refetch': {
-          const { dataAtom, options } = get(queryDataAtom)
+          const { dataAtom, observer } = get(queryDataAtom)
           set(dataAtom, new Promise<TData>(() => {})) // infinite pending
-          const queryClient = get(getQueryClientAtom)
-          queryClient.getQueryCache().find(options.queryKey)?.reset()
-          const p: Promise<void> = queryClient.refetchQueries(options.queryKey)
+          const p: Promise<void> = observer
+            .refetch({ cancelRefetch: true })
+            .then(() => {})
```
dai-shi, line 113: If I understand it correctly, either `p: Promise<void> =` or `.then(() => {})` does the hack of returning `Promise<void>`. It doesn't need both.

dai-shi (commented):
Just some minor comments. Can you check please?

dai-shi:
BTW, rebase & force-push is not necessary, as commits will be squashed when merging the pr.

src/query/atomWithQuery.ts
aulneau: this is auto formatted via prettier

src/query/atomWithQuery.ts
aulneau: as it is now:
Screen Shot 2021-07-01 at 7 55 36 AM

no explicit typing:

Screen Shot 2021-07-01 at 7 55 48 AM

no .then:

Screen Shot 2021-07-01 at 8 22 29 AM

aulneau:
> BTW, rebase & force-push is not necessary, as commits will be squashed when merging the pr.

just my preferred style of `git`'ing :) hope that's ok

dai-shi:
> just my preferred style of `git`'ing :) hope that's ok

If it's your preference, just fine.

src/query/atomWithQuery.ts
dai-shi: let me try.

src/query/atomWithQuery.ts
dai-shi: oh, the queryDataAtom is not properly typed... let me fix.

dai-shi:
done.

dai-shi (approved):

dai-shi:
@aulneau If you are fine, I will merge.

@Aslemammad You might want to take a look.

aulneau:
@dai-shi ah perfect!! yes please ship it :) would love a release when possible too if that's ok

dai-shi:
Good. I will probably cut a release tomorrow.

PR #574, opened 2021-07-01
chore(core): write getter default parameter instead of optional parameter

### LINKED ISSUE #572: Version 1.0.1 broke ReScript bindings
Working on ReScript bindings for Jotai, I didn't come up with these bindings but have been trying to help maintain them. https://github.com/eccentric-j/re-jotai/tree/stable. They're not complete, or perfect, but the core functionality is there and they pair quite well in so far. As of v1.0.1 though, the get function within a writable atom accepts an optional unstable_promise boolean which broke the bindings as ReScript was returning a function instead of the value of the dependent readable atom. I have a work around by defining a get wrapper that takes two arguments but it does require me to change all my existing writable atoms to call it. It's not ideal but sufficient to unblock progress.

ReScript bindings are likely far from anyone's priority with this project but it's worth recommending a process change to constrain future API changes to major version updates if possible to avoid issues like this. Please note overall I really like Jotai and very much appreciate the work required to make and maintain this library, I don't wish to curb innovation. Since Jotai hit 1.0.0 I am hoping we can find a compromise to avoid breaking dependent libs and bindings from minor version changes.

dai-shi:
possibly fix #572.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "index.js": {
-    "bundled": 21254,
-    "minified": 10166,
-    "gzipped": 3349,
+    "bundled": 21262,
+    "minified": 10169,
+    "gzipped": 3353,
     "treeshaked": {
       "rollup": {
         "code": 14,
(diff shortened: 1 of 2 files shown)


dai-shi:
Merging this. I don't think we dare to note this in summary in the release note.

PR #575, opened 2021-07-02
feat(devtools): devtool atom

aulneau:
This is a WIP

### DIFF
--- a/src/devtools/devtoolsAtom.ts
+++ b/src/devtools/devtoolsAtom.ts
@@ -1,0 +1,133 @@
+import { atom, PrimitiveAtom } from 'jotai'
+
+type Config = {
+  instanceID?: number
+  name?: string
+  serialize?: boolean
+  actionCreators?: any
+  latency?: number
+  predicate?: any
+  autoPause?: boolean
+}
+
+type Message = {
+  type: string
+  payload?: any
+  state?: any
+}
+
+type ConnectionResult = {
+  subscribe: (dispatch: any) => () => void
+  unsubscribe: () => void
+  send: (action: string, state: any) => void
+  init: (state: any) => void
+  error: (payload: any) => void
+}
+
+type Extension = {
+  connect: (options?: Config) => ConnectionResult
+}
+
+type Devtools = ConnectionResult & { shouldInit?: boolean }
+
+const IS_DEVTOOL_ENV =
+  typeof process === 'object' &&
+  process.env.NODE_ENV === 'development' &&
+  typeof window !== 'undefined' &&
+  '__REDUX_DEVTOOLS_EXTENSION__' in window
+
+const getExtension = (): Extension => {
+  try {
+    return (window as any).__REDUX_DEVTOOLS_EXTENSION__ as Extension
+  } catch (e) {
+    throw new Error('Please install or enable Redux Devtools')
+  }
+}
+
+const devtoolAtomCache = new WeakMap()
+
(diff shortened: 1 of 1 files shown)


src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,133 @@
...
+    const lastValueAtom = atom(undefined)
+    const atomName = anAtom.debugLabel || anAtom.toString()
+    const extension = getExtension()
+    const cachedDevtools = devtoolAtomCache.get(anAtom)
+    const devtools: Devtools =
+      cachedDevtools || extension.connect({ name: atomName })
+
+    let unsubscribe: undefined | (() => void)
+
+    const subscribeAtom = atom<void, { type: 'mount' }>(
```
dai-shi, line 65: this is creating a new atom every time. oh that's why you need the cache. this could be improved.

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,133 @@
...
+        }
+      }
+    )
+
+    subscribeAtom.onMount = (setAtom) => {
+      setAtom({ type: 'mount' })
+      return () => unsubscribe?.()
+    }
+
+    get(subscribeAtom)
```
dai-shi, line 128: this is probably a hack, but this works like it subscribe and unsubscribe every time?

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    if (!isWriteable) {
+      devtools.send(`${atomName} - ${new Date().toLocaleString()}`, get(anAtom))
```
dai-shi, line 131: this is a side effect in `read`.
conceptually `read` must be pure, and it may be double invoked in the future or render without commit (if async somewhere).

I think it would be nice to send in `write` which solved double invocation, but render without commit can still happen. If we want to see something committed, `useAtomDevtools` would be better.

dai-shi (commented):
It looks like this is trying too many challenges at once. I'd suggest to try one by one.

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+const devtoolAtomCache = new WeakMap()
+
+// TODO: this was an atom family, but i've changed it to a fn
+// this will probably result in an infinite loop when used unless memoized
+export const devtoolAtom = (anAtom: PrimitiveAtom<any>) =>
+  atom((get) => {
+    if (!IS_DEVTOOL_ENV) return
+    const isWriteable = !!anAtom.write
```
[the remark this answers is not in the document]
aulneau: yeah either read only or writable atoms should be accepted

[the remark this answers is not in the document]
dai-shi: If this is the primary goal, I'd suggest to work on `useAtomDevtools` for a start, to fix only this.
In this case, we don't allow time traveling, right? How does it look like in the Redux DevTools Extension? Do users notice if time traveling is unsupported?

aulneau:
@dai-shi seems like this might be more work than I thought -- I have a local version we use in our projects and works for my team. Seems like maybe this is out of scope? Feel free to re-open if you think otherwise

dai-shi:
not out of scope, but needs more work to put it in a library.
this pr was addressing too many challenges at once.
my suggestion is to tackle one by one.
if anyone is interested, please file an issue for one.

PR #582, opened 2021-07-07
fix(query): gracefully handle CancelledError

aulneau:
So this was a strange bug that I came upon. It would happen where I had an `atomWithInfiniteQuery` already loaded, and then navigated to a page that had that same query. It seems there is some [default cancellation behavior](https://github.com/tannerlinsley/react-query/blob/16b7d290c70639b627d9ada32951d211eac3adc3/src/core/infiniteQueryBehavior.ts#L118) in the way infiniteQueries work in react query.

I think this solution makes sense, and it fixes my bug (without this, the application breaks).

### DIFF
--- a/src/query/atomWithInfiniteQuery.ts
+++ b/src/query/atomWithInfiniteQuery.ts
@@ -5,6 +5,7 @@ import {
   InfiniteData,
   InitialDataFunction,
   QueryObserverResult,
+  isCancelledError,
 } from 'react-query'
 import { atom } from 'jotai'
 import type { WritableAtom, Getter } from 'jotai'
@@ -87,7 +88,7 @@ export function atomWithInfiniteQuery<
           | QueryObserverResult<InfiniteData<TData>, TError>
           | { data?: undefined; error: TError }
       ) => {
-        if (result.error) {
+        if (result.error && !isCancelledError(result.error)) {
           if (settlePromise) {
             settlePromise(null, result.error)
             settlePromise = null
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -70,9 +70,9 @@
     }
   },
   "query.js": {
-    "bundled": 5170,
-    "minified": 2240,
-    "gzipped": 686,
+    "bundled": 5330,
+    "minified": 2351,
+    "gzipped": 746,
     "treeshaked": {
       "rollup": {
         "code": 105,


src/query/atomWithInfiniteQuery.ts
```diff
@@ -88,6 +89,15 @@ export function atomWithInfiniteQuery<
           | { data?: undefined; error: TError }
       ) => {
         if (result.error) {
+          if (isCancelledError(result.error)) {
+            // fetch has been canceled,
+            // revert to prevData
+            if (settlePromise) {
+              if (prevData) settlePromise(prevData)
+              settlePromise = null
```
dai-shi, line 97:
```suggestion
              if (prevData) {
                settlePromise(prevData)
                settlePromise = null
              }
```
settlePromise should be cleared only if it's invoked.

src/query/atomWithInfiniteQuery.ts
```diff
@@ -88,6 +89,15 @@ export function atomWithInfiniteQuery<
           | { data?: undefined; error: TError }
       ) => {
         if (result.error) {
+          if (isCancelledError(result.error)) {
+            // fetch has been canceled,
+            // revert to prevData
```
dai-shi, line 94:
```suggestion
            // fetch has been canceled, revert to prevData
```
unless prettier complains.

src/query/atomWithInfiniteQuery.ts
dai-shi: resolving with prevData is required for refetch? hm but this file is atomWithInfiniteQuery, not atomWithQuery.
I feel this a bit weird because settlePromise and setData are always paired.

Can we do
```js
        if (result.error && !isCancelledError(result.error)) {
```
in L91?

dai-shi (commented):
I wasn't aware about the cancellation of promise. I think it's better if we can simply ignore it.
Is it the issue with atomWithQuery too?

src/query/atomWithInfiniteQuery.ts
aulneau: Yep that works too, wasn't sure if we needed to do the settledPromise stuff (tbh not clear on what it does exactly lol)

aulneau:
> Is it the issue with atomWithQuery too?

I have not run into it, and looking at the source code of react query, it seems that there isn't this same kind of CancelledError usage in a non-infinite query.

src/query/atomWithInfiniteQuery.ts
aulneau: I removed the comments

dai-shi (approved):
Cool. This looks just fine.

src/query/atomWithInfiniteQuery.ts
dai-shi: I know. This is because of very tricky suspense behavior. I'm not even 100% sure, so adding tests to check the behavior would be important.

PR #583, opened 2021-07-08
fix: package.json exports to have import

dai-shi:
We had some discussions in https://github.com/pmndrs/valtio/pull/169.
Like valtio, jotai uses lots of module states.
So, this change suffers from dual package hazard.
However, we feel the benefit is bigger because cjs build is built for traditional browsers and esm build is always recommended if possible.

Would anyone please try this, and if it works fine? (You can install the codesandbox build with npm/yarn.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,56 +21,67 @@
     ".": {
       "types": "./index.d.ts",
       "module": "./esm/index.js",
+      "import": "./esm/index.js",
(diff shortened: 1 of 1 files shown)


aulneau:
I've tested this in a next.js project and works fine (no difference)

dai-shi:
Thanks!

PR #586, opened 2021-07-09
fix(query): subscribe instead of fetch if initialData

aulneau:
While doing some more testing, I noticed that on mount a query would automatically refetch even though I had provided `initialData` to it. I could not stop this behavior -- based off of reading https://react-query.tanstack.com/guides/initial-query-data#_top, it is the case that initialData will be treated as fresh until the staleTime has been met, or based on other user options passed (meaning a fetch should not happen until it become stale, and there have been some refetch options passed)

Because we were automatically running the fetchOptimistic method regardless of if initialData was passed, we were causing an unneeded fetch to happen. React query will pick up and take over and do all the fetching if needed when there is initialData. We are already subscribing on mount, so there is no need to fetch during the initial get.

### DIFF
--- a/src/query/atomWithInfiniteQuery.ts
+++ b/src/query/atomWithInfiniteQuery.ts
@@ -120,7 +120,7 @@ export function atomWithInfiniteQuery<
 
       const observer = new InfiniteQueryObserver(queryClient, defaultedOptions)
 
-      if (!defaultedOptions.initialData) {
+      if (!getInitialData()) {
         observer
           .fetchOptimistic(defaultedOptions)
           .then(listener)
--- a/src/query/atomWithInfiniteQuery.ts
+++ b/src/query/atomWithInfiniteQuery.ts
@@ -61,11 +61,14 @@ export function atomWithInfiniteQuery<
               >
             )()
           : options.initialData
+
+      const initialData = getInitialData()
+
       const dataAtom = atom<
         | InfiniteData<TData | TQueryData>
         | Promise<InfiniteData<TData | TQueryData>>
       >(
-        getInitialData() ||
+        initialData ||
           new Promise<InfiniteData<TData>>((resolve, reject) => {
             settlePromise = (data, err) => {
               if (err) {
@@ -120,7 +123,7 @@ export function atomWithInfiniteQuery<
 
       const observer = new InfiniteQueryObserver(queryClient, defaultedOptions)
 
-      if (!getInitialData()) {
+      if (!initialData) {
         observer
           .fetchOptimistic(defaultedOptions)
           .then(listener)
--- a/src/query/atomWithInfiniteQuery.ts
+++ b/src/query/atomWithInfiniteQuery.ts
@@ -120,10 +120,12 @@ export function atomWithInfiniteQuery<
 
       const observer = new InfiniteQueryObserver(queryClient, defaultedOptions)
 
-      observer
-        .fetchOptimistic(defaultedOptions)
-        .then(listener)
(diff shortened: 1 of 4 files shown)


src/query/atomWithQuery.ts
```diff
@@ -95,10 +95,12 @@ export function atomWithQuery<
         defaultedOptions.staleTime = 1000
       }
       const observer = new QueryObserver(queryClient, defaultedOptions)
-      observer
-        .fetchOptimistic(defaultedOptions)
-        .then(listener)
-        .catch((error) => listener({ error }))
+      if (!defaultedOptions.initialData) {
```
dai-shi, line 98: More precisely, you want to check if `getInitialData()` returns non-falsy value. Like, `options.initialData = () => null`. We may drop the support of it. Just needs to make it consistent.

dai-shi (commented):
The approach seems valid: if we don't store the initial promise, it will mount, so fetchOptimistic is not required.
Is it possible to add failing tests? I'm not super confident to keep the behavior when we do some refactors.

src/query/atomWithQuery.ts
aulneau: Sure, that is a good catch. Do you think this would be sufficient?
```suggestion
      if (!getInitialData()) {
```

tests/query/atomWithQuery.test.tsx
```diff
@@ -301,5 +302,7 @@ it('query with initialData test', async () => {
 
   // NOTE: the atom never suspends
   await findByText('count: 0')
+  expect(mockFetch).toHaveBeenCalledTimes(0)
```
aulneau, line 305: @dai-shi does this work for you with regards to testing? I can't think immediately a way to create a failing test

src/query/atomWithQuery.ts
dai-shi: I think we should call `getInitialData()` just once in our code.

tests/query/atomWithQuery.test.tsx
dai-shi: Do you mean this doesn't fail before the fix?

aulneau: ah i see what you mean, yes it does fail:

Screen Shot 2021-07-09 at 6 41 09 PM

src/query/atomWithQuery.ts
aulneau: yep, agreed!

dai-shi (approved):
Thanks for the change. It looks good to me.

Is anyone else interested in trying and/or reviewing?

src/query/atomWithInfiniteQuery.ts
```diff
@@ -120,10 +123,12 @@ export function atomWithInfiniteQuery<
...
 
-      observer
-        .fetchOptimistic(defaultedOptions)
-        .then(listener)
-        .catch((error) => listener({ error }))
+      if (!initialData) {
+        observer
+          .fetchOptimistic(defaultedOptions)
+          .then(listener)
+          .catch((error) => listener({ error }))
```
aulneau, line 130: @dai-shi I don't think we need this bit of code either way it seems -- can you tell me more about it?

Because we are subscribing on mount, the subscription kicks off the fetch initially. Seems like we're doing double work, unless I am missing something

dai-shi: When an atom suspends, the component suspends and it will not be mounted, which means `onMount` will not be called. Unless we resolve the promise, the component will not be mounted.

PR #589, opened 2021-07-11
fix: support atom scope outside core

dai-shi:
An internal discussion with @aulneau revealed that many of utils/integrations don't support atom scope properly, especially when we create internal atoms in a function.
This PR addresses the issue. Note this affects only the case using `scope`, which is mainly for library usage.

### DIFF
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -1,4 +1,5 @@
 import {
+  QueryClient,
   QueryKey,
   QueryObserver,
   QueryObserverOptions,
@@ -7,7 +8,7 @@ import {
 } from 'react-query'
 import { atom } from 'jotai'
 import type { WritableAtom, PrimitiveAtom, Getter } from 'jotai'
-import { getQueryClientAtom } from './queryClientAtom'
+import { queryClientAtom } from './queryClientAtom'
 
 export type AtomWithQueryAction = { type: 'refetch' }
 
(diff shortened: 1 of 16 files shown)


src/query/atomWithQuery.ts
```diff
@@ -27,7 +28,8 @@ export function atomWithQuery<
     | ((
         get: Getter
       ) => AtomWithQueryOptions<TQueryFnData, TError, TData, TQueryData>),
-  equalityFn: (a: TData, b: TData) => boolean = Object.is
+  equalityFn: (a: TData, b: TData) => boolean = Object.is,
+  getQueryClient: (get: Getter) => QueryClient = (get) => get(queryClientAtom)
```
aulneau, line 32: I know there is a preference to avoid objects being passed, but I think this will result in a lot of `atomWithQuery(createQuery, null, (get) => get(myQueryClientAtom))`, as I think passing a custom query client will be used more often than an equalityFn. and maybe passing null is ok, too :)

dai-shi: I see your point and yes to the preference.
I'm not sure how often people use `myQueryClientAtom` in app code. Unlike `jotai/urql` which require "URL", the default client in `jotai/query` would work for most cases?
We still can use `<Provider initialValues={[[queryClientAtom, myQueryClient]]}>`.

To allow `null` for equalityFn, we can't use default parameter, right? we can change that.

aulneau: > We still can use <Provider initialValues={[[queryClientAtom, myQueryClient]]}>.

If we did this though, it would not allow you to modify the scope, right?

I think folks might use the custom query client often, esp if they have an app that is using react-query already as they might have defined defaults or other settings they want to use.

dai-shi: Yes, for the custom atom scope, we need to use the third argument.
Do you think it's a good idea to swap the second and third arguments? (=breaking change)

aulneau: yeah, I do. I'm honestly not sure how much use or importance the equalityFn provides -- i find modifying the query client to only update on certain changes is more effective https://react-query.tanstack.com/guides/migrating-to-react-query-3#the-queryoptionsnotifyonstatuschange-option-has-been-superceded-by-the-new-notifyonchangeprops-and-notifyonchangepropsexclusions-options

dai-shi: If we can remove `equalityFn` completely, it seems better to me. Is it possible?

dai-shi: #592

PR #592, opened 2021-07-14
breaking(query): remove optional equalityFn from atomWith(Infinite)Query

dai-shi:
This is basically a commit reverting #386.

https://github.com/pmndrs/jotai/pull/589#discussion_r668949661 shows an alternative method, so I think this is not necessary.


aulneau (approved):
yep this makes sense :) sorry for the original change lol

PR #596, opened 2021-07-16
breaking(rxjs): move atomWithObservable in utils

dai-shi:
Thanks to #593, atomWithObservable doesn't depend on rxjs, so moving to utils. Modified a bit to fit with coding styles.


dai-shi:
I don't feel comfortable to have `jotai/rxjs` to stay, so let me merge this and cut a release.

PR #598, opened 2021-07-16
fix(examples, tests): use `useUpdateAtom` and write-only atoms to avoid re-render

liby:
Response to the call for this [issue](https://github.com/pmndrs/jotai/issues/26#issuecomment-691588119)
> The utility hooks can get big. I need someone to volunteer to help it, especially for documentation & examples.

### DIFF
--- a/tests/items.test.tsx
+++ b/tests/items.test.tsx
@@ -94,13 +94,18 @@ it('add an item with filtered list', async () => {
     text: string
     checked: boolean
   }
+  type ItemAtoms = PrimitiveAtom<Item>[]
+  type Update = (prev: ItemAtoms) => ItemAtoms
 
   let itemIndex = 0
-  const itemsAtom = atom<PrimitiveAtom<Item>[]>([])
+  const itemAtomsAtom = atom<ItemAtoms>([])
+  const setItemsAtom = atom<null, Update>(null, (_get, set, update) =>
+    set(itemAtomsAtom, update)
+  )
   const filterAtom = atom<'all' | 'checked' | 'not-checked'>('all')
   const filteredAtom = atom((get) => {
     const filter = get(filterAtom)
-    const items = get(itemsAtom)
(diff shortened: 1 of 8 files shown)


dai-shi:
Thanks for your help.
Having best practices in examples is nice.
Using utils in utils tests is ok.

However, I would like to keep `tests/*.test.tsx` to depend only on core (no utils).
It's intentional/acceptable to have extra re-renders there (not important for tests).
If it were issues, we would make write-only atoms (that's the suggestion for core-only solution).

If you can adjust it, please go ahead. I can take it over if you like.

liby:
@dai-shi
I've revised it. Could you take another review of it for me?

> It's intentional/acceptable to have extra re-renders there (not important for tests).
> If it were issues, we would make write-only atoms (that's the suggestion for core-only solution).

If necessary, I can undo changes other than utils tests and examples.

tests/items.test.tsx
```diff
@@ -94,9 +94,14 @@ it('add an item with filtered list', async () => {
     text: string
     checked: boolean
   }
+  type ItemsAtom = PrimitiveAtom<Item>[]
+  type Update = (prev: ItemsAtom) => ItemsAtom
 
   let itemIndex = 0
-  const itemsAtom = atom<PrimitiveAtom<Item>[]>([])
+  const itemsAtom = atom<ItemsAtom>([])
```
dai-shi, line 101: FWIW, I name it like this in my last tutorial.
```ts
  const itemAtomsAtom = atom<ItemAtoms>([])
```

dai-shi (approved):
Looks good. As I said, I don't think it's very important to avoid re-renders in some of core tests, but someone may learn coding pattern from tests, so it'd be good as you suggest.

liby:
> but someone may learn coding pattern from tests

Yes, that's why I kept the comments in the sample code.

PR #599, opened 2021-07-17
chore(utils): refactor atomWithObservable

dai-shi:
This makes the implementation close to atomWithQuery in jotai/query.
My hope was to support error retry, but I suppose it's not yet supported with atomWithQuery either.


dai-shi:
#601 is filed. Let me merge this as this is just refactoring.

PR #602, opened 2021-07-18
chore(imports): enforce a convention in module import order

liby:
Specific changes and motives:
1. Add rules for `js,jsx,ts,tsx` files in `overrides`
    1.1 `alphabetize`: Sort the order within each group in alphabetical manner based on **import path**.(autofixable)
    1.2 `groups`: [Details](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#groups-array)(autofixable)
    1.3 `newlines-between`: Enforces or forbids new lines between import groups.(autofixable)
    1.4 `pathGroups`: Force sort order of certain packages within group.(autofixable)
    1.5 `pathGroupsExcludedImportTypes `: Configure the `pathGroups` option to use the.(autofixable)
    1.6 `sort-imports`: Order members of import with multiple members.(autofixable)
2. Move `"react/jsx-uses-react": "off",` and `"react/react-in-jsx-scope": "off",` into the `overrides`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,7 +12,14 @@
     "plugin:import/errors",
     "plugin:import/warnings"
   ],
-  "plugins": ["@typescript-eslint", "react", "prettier", "react-hooks", "import", "jest"],
+  "plugins": [
+    "@typescript-eslint",
+    "react",
+    "prettier",
+    "react-hooks",
+    "import",
+    "jest"
+  ],
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
     "ecmaVersion": 2018,
@@ -31,15 +38,22 @@
       "import/namespace": "off",
       "import/default": "off",
       "@typescript-eslint/explicit-module-boundary-types": "off",
-      "no-unused-vars": ["warn", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }],
-      "@typescript-eslint/no-unused-vars": ["warn", { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }],
+      "no-unused-vars": [
+        "warn",
+        { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }
+      ],
+      "@typescript-eslint/no-unused-vars": [
+        "warn",
+        { "argsIgnorePattern": "^_", "varsIgnorePattern": "^_" }
+      ],
       "@typescript-eslint/no-use-before-define": "off",
       "@typescript-eslint/no-empty-function": "off",
       "@typescript-eslint/no-empty-interface": "off",
       "@typescript-eslint/no-explicit-any": "off",
-      "jest/consistent-test-it": ["error", { "fn": "it", "withinDescribe": "it" }],
-      "react/jsx-uses-react": "off",
-      "react/react-in-jsx-scope": "off"
+      "jest/consistent-test-it": [
+        "error",
+        { "fn": "it", "withinDescribe": "it" }
+      ]
     }
   },
   "settings": {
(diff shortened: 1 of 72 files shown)


.eslintrc.json
```diff
@@ -67,6 +81,44 @@
...
+              "builtin",
+              "external",
+              "internal",
+              "parent",
+              "sibling",
+              "index",
+              "object",
+              "type"
+            ],
+            "newlines-between": "never",
```
dai-shi, line 102: How would it be like if we make this "always"? Something too weird?

dai-shi: ah, will it require newlines between `import { foo } from './foo'` and `import { bar } from './bar'`?

src/urql/atomWithMutation.ts
```diff
@@ -1,12 +1,12 @@
...
   Client,
-  TypedDocumentNode,
   OperationContext,
   OperationResult,
+  TypedDocumentNode,
 } from '@urql/core'
 import { atom } from 'jotai'
-import type { Getter } from 'jotai'
 import { clientAtom } from './clientAtom'
+import type { Getter } from 'jotai'
```
dai-shi, line 9: This change looks a little unfortunate...

.eslintrc.json
liby: > How would it be like if we make this "always"? Something too weird?

If set to always, at least one new line between each group will be enforced, and new lines inside a group will be forbidden.

```ts
/* eslint import/order: ["error", {"newlines-between": "always"}] */
import fs from 'fs';
import path from 'path';

import index from './';

import sibling from './foo';
```

> ah, will it require newlines between import { foo } from './foo' and import { bar } from './bar'?

The following is only a personal opinion:
1. Now everyone is using auto import
2. Configuring eslint rules will help us sort  imports

In summary, there are now very few cases where the imports section needs to be managed manually, then it is not so necessary to add blank lines just for the sake of readability.

dai-shi: Okay, it's just unfortunate for me that the rule doesn't allow adding one line only between absolute path imports and relative path imports.

src/core/Provider.ts
```diff
@@ -1,9 +1,9 @@
-import { createElement, useRef, useDebugValue, FC } from 'react'
+import { FC, createElement, useDebugValue, useRef } from 'react'
```
dai-shi, line 1: It might be out of the scope of this PR, but as we strictly use `import type` for types, this would be:
```suggestion
import { createElement, useDebugValue, useRef } from 'react'
import type { FC } from 'react'
```

src/devtools/useGotoAtomsSnapshot.ts
```diff
@@ -1,9 +1,8 @@
 import { useCallback, useContext } from 'react'
 import { SECRET_INTERNAL_getStoreContext as getStoreContext } from 'jotai'
-
 import type { Scope } from '../core/atom'
-// NOTE this is across bundles and actually copying code
 import { isDevStore } from '../core/contexts'
+// NOTE this is across bundles and actually copying code
```
dai-shi, line 5:
```suggestion
// NOTE importing from '../core/contexts' is across bundles and actually copying code
```

src/devtools/useAtomsSnapshot.ts
```diff
@@ -3,11 +3,10 @@ import {
   SECRET_INTERNAL_getStoreContext as getStoreContext,
   SECRET_INTERNAL_useMutableSource as useMutableSource,
 } from 'jotai'
-
 import type { Atom, Scope } from '../core/atom'
+import { getDebugStateAndAtoms, subscribeDebugStore } from '../core/Provider'
 import type { AtomState, State } from '../core/vanilla'
 // NOTE this is across bundles and actually copying code
```
dai-shi, line 9:
```suggestion
// NOTE importing from '../core/Provider' is across bundles and actually copying code
```

tests/async.test.tsx
```diff
@@ -1,6 +1,6 @@
-import { StrictMode, Suspense, useRef, useEffect, FC } from 'react'
+import { FC, StrictMode, Suspense, useEffect, useRef } from 'react'
```
dai-shi, line 1:
```suggestion
import { StrictMode, Suspense, useEffect, useRef } from 'react'
import type { FC } from 'react'
```

dai-shi (approved):
Thanks for your work!
We would be removing `FC`, so I'd just leave it for now.

.eslintrc.json
liby: I will study this again, and if it can be realized, I’ll launch a PR.

PR #603, opened 2021-07-19
chore(examples, tests): remove `React.FC` from the TSX file

liby:
> We would be removing FC, so I'd just leave it for now.

It's done.

### DIFF
--- a/examples/todos/src/App.tsx
+++ b/examples/todos/src/App.tsx
@@ -1,4 +1,4 @@
-import { FC, FormEvent } from 'react'
+import { FormEvent } from 'react'
 import { CloseOutlined } from '@ant-design/icons'
 import { a, useTransition } from '@react-spring/web'
 import { Radio } from 'antd'
@@ -26,7 +26,7 @@ type TodoItemProps = {
   atom: PrimitiveAtom<Todo>
   remove: RemoveFn
 }
-const TodoItem: FC<TodoItemProps> = ({ atom, remove }) => {
+const TodoItem = ({ atom, remove }: TodoItemProps) => {
   const [item, setItem] = useAtom(atom)
   const toggleCompleted = () =>
     setItem((props) => ({ ...props, completed: !props.completed }))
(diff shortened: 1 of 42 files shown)


dai-shi (approved):
Thanks again for you work.

Although this might be controversial if we ask someone, but let's assume it's the trend/recommendation and merge this.

liby:
> Although this might be controversial if we ask someone, but let's assume it's the trend/recommendation and merge this.

Thank you for your indulgence.  I’ll continue to follow up on the issue of adding only one line between absolute path import and relative path import.

dai-shi:
@liby Would you mind if I ask another chore? For consistency, we'd want to use `import type {}` for TS types in `src` and `tests`, (maybe not in `examples`).

liby:
@dai-shi 
Yes, I'd love to, but it may be a little later. I will first see if there is an easy way to distinguish between `type`. If there is no good way, I may need to find it one by one in the clumsiest way.

liby:
@dai-shi 
The use of  `Add all missing imports` in the `.tsx` file in the "tests" folder will always import React, See below:

Kapture 2021-07-20 at 00 27 10
Rename the `Provider.ts` file to `Provider.tsx`, consistent with the behavior in the gif.

---

Changing `jsx` value in the  `tsconfig.json` file to `react-jsx` solves the above problem:

Kapture 2021-07-20 at 00 31 32

---

The problem is solved, but I am still confused as to why the behavior is inconsistent when `"jsx": "preserve"`, I checked a lot of related issue, but I couldn't get an answer.(e.g., [TypeScript/issues/41958](https://github.com/microsoft/TypeScript/issues/41958), [TypeScript/issues/41762](https://github.com/microsoft/TypeScript/issues/41762), [TypeScript/pull/41950](https://github.com/microsoft/TypeScript/pull/41950))

You built this project, maybe you can give me an answer? I've been obsessed with it all day. Looking forward to your reply

dai-shi:
...unfortunately, I don't know. I usually type imports before writing code...
Changing tsconfig is acceptable as long as the final build files are identical.

PR #606, opened 2021-07-20
chore(imports): update the imports declarations of type annotations and declarations to `import type`

liby:
To be honest, when I was doing this PR, I saw a lot of related issues and pull.
e.g.:
- [Type-only imports and exports](https://github.com/microsoft/TypeScript/pull/35200)
- [Add option to disable TS using inline import types in quick fixes](https://github.com/microsoft/TypeScript/issues/35078)
- [Feedback on 'import type' UX for editors](https://github.com/microsoft/TypeScript/issues/36038)
- [Update type-only import semantics to allow type queries](https://github.com/microsoft/TypeScript/pull/36092)
- [Type-only auto imports](https://github.com/microsoft/TypeScript/pull/36412)
- [Auto import adds types to existing value imports under --importsNotUsedAsValues=error](https://github.com/microsoft/TypeScript/issues/39432)

Now I am beginning to think, is this really necessary?

### DIFF
--- a/examples/mega-form/src/App.tsx
+++ b/examples/mega-form/src/App.tsx
@@ -1,5 +1,6 @@
 import { useCallback, useMemo } from 'react'
-import { PrimitiveAtom, Provider, atom, useAtom } from 'jotai'
+import { Provider, atom, useAtom } from 'jotai'
+import type { PrimitiveAtom } from 'jotai'
 import { focusAtom } from 'jotai/optics'
 import { useAtomCallback } from 'jotai/utils'
 import initialValue from './initialValue'
--- a/examples/mega-form/src/useAtomSlice.ts
+++ b/examples/mega-form/src/useAtomSlice.ts
@@ -1,5 +1,6 @@
 import { useMemo } from 'react'
-import { PrimitiveAtom, useAtom } from 'jotai'
+import { useAtom } from 'jotai'
+import type { PrimitiveAtom } from 'jotai'
 import { splitAtom } from 'jotai/utils'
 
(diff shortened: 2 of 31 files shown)


dai-shi:
> is this really necessary?

This would be just a preference, but it helps to me to understand dependencies across bundles. So, it might not be really necessary, but make things consistent.

dai-shi:
I also confirmed, tsconfig change doesn't affect bundles.
```
$ git diff tsconfig.json
diff --git a/tsconfig.json b/tsconfig.json
index 9f5dd48..3f3d107 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "es2019",
     "strict": true,
-    "jsx": "preserve",
+    "jsx": "react-jsx",
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "moduleResolution": "node",
$ diff -r dist.jsx_preserve dist.jsx_react-jsx
$ 
```

dai-shi (approved):
Thanks for the work. Much appreciated.
Yeah, it's not 100% certain if we can continue the style in the future.
For now, it looks good to me. (examples may not work with old TS, but let's hope no one complain.)

liby:
> This would be just a preference, but it helps to me to understand dependencies across bundles. So, it might not be really necessary, but make things consistent.

I see what you mean.

> Thanks for the work. Much appreciated.

That's alright, glad to help!

> For now, it looks good to me. (examples may not work with old TS, but let's hope no one complain.)

If there are any problems in the future, let's communicate at that time to see how to modify them.

Issue #624, opened 2021-07-31
atomWithProxy triggers infinite loop


israelidanny:
When using `atomWithProxy` merely referencing any value from the state will cause an infinite loop. I created a minimal reproduction of the bug:
https://codesandbox.io/s/jotaivaltio-infinite-loop-yphdz?file=/src/App.js

It has a proxy atom and it uses it like so:
`const [atomState] = useAtom(proxyAtom);`

If you just print out something like `JSON.stringify(atomState)` everything is great, but the moment you do a simple change to `JSON.stringify(atomState.someNestedKey)` everything freezes in an infinite loop (even though the key exists and prints out successfully without being referenced directly.

dai-shi:
If I don't misunderstand your example, you use async/suspense feature of valtio.

https://github.com/pmndrs/jotai/blob/2f142913e0af72fbfcf9c40cca1fed58c5dfa305/src/valtio/atomWithProxy.ts#L35
It's not currently supported due to typing issue.
But, I didn't know about the infinite loop.
Maybe, we can support the promise usage along with fixing the infinite loop.

dai-shi:
I wish some contributors work on this, but it will require fairly deep understanding of jotai and valtio.

At least, our documentation should note about the limitation with promises.

israelidanny:
Do I understand correctly that the typing issue is that once you set a state key as a promise, overwriting it would fail TypeScript checks because it expects a promise and not whatever promise returns?

dai-shi:
Exactly.

israelidanny:
I'd be interested to at least give it a small stab. Will report what, if anything, came out of it.

Off the top of your head: any guesses as to what could be triggering an infinite loop? We're not actually mutating the object but rather accessing a snapshot key, it's starange that merely accessing it would cause that.

dai-shi:
If it doesn't touch the key, it will not throw a promise. so, that's probably why.

dai-shi:
So, typing issue and infinite loop are probably different problems.

israelidanny:
Alright, I'm giving it a stab now. Let's see if anything useful comes out of it :)

israelidanny:
So, as far as I understand the following happens to cause the infinite loop:
1. `atomWithProxy` creates a snapshot of the proxy
2. The following logic executes and sets the prop to a getter that throws a promise: https://github.com/pmndrs/valtio/blob/master/src/vanilla.ts#L101-L110
3. The promise resolves
[...]

PR #625, opened 2021-07-31
test(valtio): infinite loop bug

israelidanny:
This is a test for the fix of the infinite loop bug (https://github.com/pmndrs/jotai/issues/624) that happens when working with promises and atomWithProxy.

- [x] create a failing test for the issue (passing now with the fix below)
- [x] fix the infinite loop bug (fixed by https://github.com/pmndrs/valtio/pull/202)
- [x] ~sort out the types (needs https://github.com/pmndrs/valtio/pull/203)~ if we accept the fact that we follow the strict typing of the initial state (promise can only be replaced by promise) then the types look OK as it is
- [x] update `valtio` dependency once the type PR is merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,7 +219,7 @@
     "ts-node": "^10.1.0",
     "tslib": "^2.3.0",
     "typescript": "^4.3.5",
-    "valtio": "^1.1.2",
+    "valtio": "^1.1.3",
     "wonka": "^4.0.15",
     "xstate": "^4.17.1",
     "zustand": "^3.5.7"
--- a/package.json
+++ b/package.json
@@ -219,7 +219,7 @@
     "ts-node": "^10.1.0",
     "tslib": "^2.3.0",
     "typescript": "^4.3.5",
-    "valtio": "^1.0.3",
+    "valtio": "^1.1.2",
     "wonka": "^4.0.15",
     "xstate": "^4.17.1",
     "zustand": "^3.5.7"
--- a/src/valtio/atomWithProxy.ts
+++ b/src/valtio/atomWithProxy.ts
@@ -32,8 +32,6 @@ const applyChanges = <T extends object>(proxyObject: T, prev: T, next: T) => {
   })
 }
 
-// No support for promises in proxy as it's not symmetric
-// Should we type it precisely?
 export function atomWithProxy<Value extends object>(proxyObject: Value) {
   const baseAtom = atom(snapshot(proxyObject))
   baseAtom.onMount = (setValue) => {
--- a/src/valtio/atomWithProxy.ts
+++ b/src/valtio/atomWithProxy.ts
@@ -32,6 +32,13 @@ const applyChanges = <T extends object>(proxyObject: T, prev: T, next: T) => {
   })
 }
 
+// Currently atomWithProxy does not support overwriting Promise() with a primitive
+// due to the requirement of valtio types to always be symmetric.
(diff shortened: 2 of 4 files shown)


israelidanny:
Dependencies are updated to valtio v1.1.3

israelidanny:
@dai-shi as far as I'm concerned, this is ready.
The types behave fine. The only thing is that I think it's going to be worth documenting the design philosophy you mentioned here:
https://github.com/pmndrs/valtio/pull/203#issuecomment-890523149

In this regard, I think it's fair to request `atomWithProxy` to respect that design pattern, and as long as it does - the promises now work great, as long as they are only ever overwritten with another promise (and never directly with the underlying resolved type)

dai-shi:
> as long as they are only ever overwritten with another promise (and never directly with the underlying resolved type)

but, this is my point of the typing issue.
I would like to merge this, but I want to revert the comment (maybe making it more precise.)

israelidanny:
@dai-shi alright. What I find myself doing in my code is:
```javascript
// we have to do it this way because of valtio typing
setStatusState({...state, status: Promise.resolve('newStatus')});
```

I see three potential ways around this:
1) We could introduce something under the hood of `atomWithProxy` that would introduce this as "magic" (turn values into Promise.resolve behind the scenes). 

Pros: Use doesn't have to do this weird Promise.resolve() thing
Cons: This behavior is different from valtio itself. Why is it different? We need a proper mental model explanation both in-code and in docs.

2) We could override this on the underlying typing level (atomWithProxy somehow overriding valtio's types to "force" it on valtio). Same pros & cons as above. But I actually find this one odd, because if you override types to do something valtio doesn't support - why doesn't valtio support it? It's a design discussion here.

3) We could canonize this behavior (Once a promise always a promise) and explain it well in both comment and docs.

For now I am making a commit that introduces a comment again because this is faster for merging. But I would like to hear your thoughts on the above and perhaps we should open a separate issue to continue this discussion?

israelidanny:
I updated the comment with something more specifically informative:
https://github.com/pmndrs/jotai/pull/625/files#diff-779f01aba33b2f379dc27b8fdeb28d51d518f37a4c21ae3f7d0119feb003c7cdR35-R41

How does this look?

dai-shi:
> `setStatusState({...state, status: Promise.resolve('newStatus')});`

I agree this is annoying.

> I see three potential ways around this:

Thanks for considerations.

I may add option 4: Only accept proxy type with `Promise<T> | T`.

```ts
atomWithProxy(proxy({ n: Promise.resolve(1) })) // this will cause TS error
atomWithProxy(proxy({ n: Promise.resolve(1) as Promise<number> | number })) // this will pass
```

I think this is precise typing for the current behavior.
Two issues: (a) not sure if we can really implement this, (b) we should document this well because the TS error message wouldn't be so friendly.

> perhaps we should open a separate issue to continue this discussion?

Yes, would you please open one?

---
A side note: I'd admit the way valtio handles a promise is like cheating. I didn't know there were use cases of `snapshot` other than `useSnapshot` back then. Ideally, in terms of valtio/vanilla, `snapshot` should keep a promise without resolving it, and only `useSnapshot` should behave for React convention. It would be nice it we have idea for this, but a) it will require major version up which will not happen any time soon, and b) the current method is very efficient and performant as I understand.

dai-shi (approved):
This is good. Thanks for working on this.

PR #633, opened 2021-08-02
fix(weakCache): Replace recursion with while loop to avoid too much recursion

### LINKED ISSUE #630: waitForAll inside selector throws with a lot of atoms
https://codesandbox.io/s/hopeful-newton-0p6q3?file=/src/App.js

Running locally it starts throwing after 6225 atoms, on codesandbox 8000+.

I've also added a failing test in a fork.
https://github.com/fc1943s/jotai/commit/db74e3ccb01aced42a448c447741c852e40f8209

image

Thisen:
Fixes #630

### DIFF
--- a/src/utils/weakCache.ts
+++ b/src/utils/weakCache.ts
@@ -1,9 +1,6 @@
 export type WeakCache<T> = WeakMap<object, [WeakCache<T>] | [WeakCache<T>, T]>
 
-export const getWeakCacheItem = <T>(
-  cache: WeakCache<T>,
-  deps: readonly object[]
-): T | undefined => {
+const _getWeakCacheItem = <T>(cache: WeakCache<T>, deps: readonly object[]) => {
   const [dep, ...rest] = deps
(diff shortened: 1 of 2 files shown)


dai-shi:
Great you found a solution!
I'd like to understand the issue and possibly come up with alternative solution.

So, is this only the issue with the weakCache? If we didn't use getWeakCacheItem by commenting out, would the test pass? (well, I can try...)

Thisen:
The issue seems to be too much recursion, JS doesn't handle that very well. Another solution would be a for loop, but I'm not sure how you would do that, given the current algorithm, where you push the item into the entry in `setWeakCacheItem`>

dai-shi:
Oh, I see. It's recursion. Yeah, I kind of knew this would happen.
I think we can implement with for-loop. Would you like to try or want me to try?

dai-shi (approved):
Yay! Thanks for finding the solution and fixing it!

PR #634, opened 2021-08-02
test(urql): add tests for client switching

israelidanny:
Introduce passing tests for the use case of switching clients for both subscriptions and queries. The tests are passing and are meant for better coverage of this as it is a rather widespread use-case (authentication involves switching clients) that we don't want breaking.

**This PR is a partial of https://github.com/pmndrs/jotai/pull/635 (in case 635 needs further discussion). Only merge it if 635 will become delayed. If 635 is OK, merge that instead**

- [x] test for `atomWithQuery` to work properly when `client` is switched 
- [x] test for `atomWithSubscription` to work properly when `client` is switched 
- [ ] throw a promise in `atomWithQuery` if `client === null`
- [ ] throw a promise in `atomWithSubscription` if `client === null`
- [ ] test for `atomWithQuery` to work properly when `client` is toggled to null
- [ ] test for `atomWithSubscription` to work properly when `client` is toggled to null

### DIFF
--- a/tests/urql/atomWithSubscription.test.tsx
+++ b/tests/urql/atomWithSubscription.test.tsx
@@ -1,24 +1,21 @@
 import { Suspense } from 'react'
-import { render } from '@testing-library/react'
+import { fireEvent, render } from '@testing-library/react'
 import type { Client, TypedDocumentNode } from '@urql/core'
 import { interval, map, pipe, take, toPromise } from 'wonka'
-import { useAtom } from '../../src/'
+import { atom, useAtom } from '../../src/'
 import { atomWithSubscription } from '../../src/urql'
 import { getTestProvider } from '../testUtils'
 
-const withPromise = (source$: any) => {
-  source$.toPromise = () => pipe(source$, take(1), toPromise)
-  return source$
-}
-const clientMock = {
-  subscription: () =>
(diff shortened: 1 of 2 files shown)


tests/urql/atomWithSubscription.test.tsx
```diff
@@ -1,24 +1,21 @@
...
-import { render } from '@testing-library/react'
+import { fireEvent, render } from '@testing-library/react'
 import type { Client, TypedDocumentNode } from '@urql/core'
 import { interval, map, pipe, take, toPromise } from 'wonka'
-import { useAtom } from '../../src/'
+import { atom, useAtom } from '../../src/'
 import { atomWithSubscription } from '../../src/urql'
 import { getTestProvider } from '../testUtils'
 
-const withPromise = (source$: any) => {
```
israelidanny, line 9: Removed as per `urql` docs:
https://formidable.com/open-source/urql/docs/advanced/subscriptions/#one-off-subscriptions

> This method will always returns a Wonka stream and doesn't have a .toPromise() shortcut method, since promises won't return the multiple values that a subscription may deliver.

dai-shi (approved):
Looks good. Thanks for adding and improving tests.

Issue #622, opened 2021-07-30
jotai/urql client getter get a wrong atom value


kalvin807:
Hi, I am trying to use jotai to manage the state of an app with a login system. When a user login, app will receive a token and the app will use it to create an urql client with the token as the bearer key.

However, the app will crash whenever the login state change like this: `login -> logout -> login (crash!)`, and saying `Cannot read property 'query' of null` . It looks like the getter in atomWithQuery somehow fetched an old value, but when I try to log the client they get, it all look fine (showing a client object). I wonder if I missed something, or it is a bug.

A small demo to the problem, the app will crash when you attempt to login the second time: [Link](https://codesandbox.io/s/test-jotai-urql-crash-3mi3y?file=/src/App.js)  
Stack-trace
```
TypeError
Cannot read property 'query' of null
Object.eval [as read]
https://codesandbox.io/static/js/sandbox.68fd4509e.js:1:100695
readAtomState
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:258:33
eval
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:261:25
Object.eval [as read]
https://codesandbox.io/static/js/sandbox.68fd4509e.js:1:100695
readAtomState
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:258:33
eval
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:241:13
readAtomState
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:237:19
readAtom
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:319:21
eval
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:643:23
checkForUpdates
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:69:30
eval
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:521:63
eval
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:521:43
flushPending
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:511:11
eval
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:406:5
Object.config.write
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:632:7
writeAtomState
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:384:30
writeAtom
https://3mi3y.csb.app/node_modules/jotai/esm/index.js:417:3
updateAtom
[...]

PR #635, opened 2021-08-03
allow edge case of client being null on urql atoms

israelidanny:
This solves https://github.com/pmndrs/jotai/issues/622 and adds a test for this edge case.

- [x] test for `atomWithQuery` to work properly when `client` is switched 
- [x] test for `atomWithSubscription` to work properly when `client` is switched 
- [x] throw a promise in `atomWithQuery` if `client === null`
- [x] throw a promise in `atomWithSubscription` if `client === null`
- [x] test for `atomWithQuery` to work properly when `client` is toggled to null
- [x] test for `atomWithSubscription` to work properly when `client` is toggled to null 

This PR expands on https://github.com/pmndrs/jotai/pull/634, and if this one is merged then 634 can be closed as this one has all of those tests as well.

### DIFF
--- a/src/urql/atomWithQuery.ts
+++ b/src/urql/atomWithQuery.ts
@@ -19,10 +19,13 @@ type QueryArgs<Data, Variables extends object> = {
 
 export function atomWithQuery<Data, Variables extends object>(
   createQueryArgs: (get: Getter) => QueryArgs<Data, Variables>,
-  getClient: (get: Getter) => Client = (get) => get(clientAtom)
+  getClient: (get: Getter) => Client | null = (get) => get(clientAtom)
 ) {
   const queryResultAtom = atom((get) => {
     const client = getClient(get)
+    if (client === null) {
+      throw new Promise(() => {})
+    }
     const args = createQueryArgs(get)
     let resolve: ((result: OperationResult<Data, Variables>) => void) | null =
       null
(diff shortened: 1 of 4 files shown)


dai-shi:
Sorry, if I rushed you with a hacky solution. To me, this is a last resort, and I would like to avoid this if possible. People wouldn't expect an atom to suspend if client is null. Or would they.

So, I think we should explore two things before this.

1. can we somehow pass non-null client if that's recommended in urql? (this is a doc issue)
2. I hope the guard originally proposed in #622 to work, or understand why it can't be fixed. (this is probably jotai core issue)

if 2) is really jotai core issue rather than jotai/urql issue, it's a broader issue we should investigate.

israelidanny:
As far as (1) goes, I am fairly confident that setting client to null is
not a urql pattern.

Rather, the expectation is to keep the client but avoid mounting a
component that does `useSubscription` or `useQuery`. Alternatively there is
also a `pause` prop on the hooks. We could perhaps implement the `pause`?

dai-shi:
Yeah, `pause=true` is much better than `client=null`.

We had some discussions about it: https://github.com/pmndrs/jotai/pull/463#discussion_r628814676

israelidanny:
I'm not totally convinced that this is very urgent to build. I'm actually doing quite fine in my app without ever having to set `client` to null (by guarding against displaying any components / screens that require auth). If someone desperately needs to set client to null, there is the hack you proposed. 

Do you think this is really important to be solved on library level?

I think this could be closed & we could focus on investigating why the original guard fails (potential core bug)

dai-shi:
> I think this could be closed & we could focus on investigating why the original guard fails (potential core bug)

Yes, that's what I meant in https://github.com/pmndrs/jotai/pull/635#issuecomment-891492347.

PR #636, opened 2021-08-03
failing test for urql client null

israelidanny:
These failing tests are to help debug https://github.com/pmndrs/jotai/issues/622
They are replicating the situation in that issue


Thisen:
I've updated #622 - Take a look

dai-shi:
Merged in #733.

Issue #340, opened 2021-03-05
Discussion: Provide jotai/nextjs utils


Thisen:
We already provide some examples about how to use Jotai with Next.js and SSR in general.

This issue looks to start a discussion if it's possible for Jotai for provide a better story for using Jotai in a SSR scenario - We believe we could start with Next.js.

I believe this would be mostly useful in the scenario, where you have async atoms, which fetch their state using e.g. `fetch`.

This is mostly inspired Next.js story for react-query: https://react-query.tanstack.com/guides/ssr

Let me think what you think - Should we provide _something_, or should we just document how to with our current API for "hydration" of atoms.

Aslemammad:
We have something like initialData in react-query, which works in Provider; I think we update it to work with provider-less mode.

Thisen:
As described in the issue, I'm aware of the current API. However, I believe there's room for improvement. 🙂

aulneau:
I would agree, I think there is a lot of room for improvement and I bet it could be such that ssr can work kind of just out of the box, eg I define an atom that has an ability to fetch its own state, and it _just works_ with loading the initial data out of the box.

Thisen:
I was thinking along the same lines @aulneau, but I didn't want to come with any solutions, as this should be a discussion.

Thisen:
Btw, I'm happy to work on any ideas that comes out of this.

Aslemammad:
Let's do more discussions on that with @dai-shi and other folks; then, we'll work on it. Thanks for the issue!

aulneau:
For sure :~) I don't mean to imply any correct solutions, and would be happy to help out too!

Thisen:
I believe a great place to start would be `getServerSideProps` and `getStaticProps` and give them a simplistic API to hydrate atoms. 

Are there other use cases I'm not thinking of?

dai-shi:
Thanks for opening this issue up @Thisen .

> Let me think what you think - Should we provide something, or should we just document how to with our current API for "hydration" of atoms.

I'm pretty sure there would be something to provide, but instead of thinking what is the thing, we would like to work on the best practices (especially because I'm not super familiar with next, yet). So the steps are:
1) Discuss in this issue
2) Improve docs and examples
3) And then, add a new util if it makes sense.

dai-shi:
> I think we update it to work with provider-less mode.

[...]

PR #637, opened 2021-08-05
Feature: useHydrateAtoms

Thisen:
Adds a new hook, that can hydrate an atom with a initial value.

Primary use case is SSR/Next.js, but it's not bound to anything.

Please supply your use cases, so we can experiment/test this hook properly.

### DIFF
--- a/src/core/contexts.ts
+++ b/src/core/contexts.ts
@@ -15,22 +15,20 @@ type UpdateAtom = <Value, Update>(
 type CommitCallback = () => void
 
 type StoreForProduction = [
-  stateMutableSource: MutableSource<State>,
-  updateAtom: UpdateAtom,
-  commitCallback: CommitCallback
+  MutableSource<State>, // stateMutableSource
+  UpdateAtom, // updateAtom
+  CommitCallback, // commitCallback
+  (values: Iterable<readonly [Atom<unknown>, unknown]>) => void // restore
 ]
 
 export type StoreForDevelopment = [
-  stateMutableSource: MutableSource<State>,
-  updateAtom: UpdateAtom,
-  commitCallback: CommitCallback,
-  debugMutableSource: MutableSource<{
+  ...StoreForProduction,
+  MutableSource<{
     version: number
     atoms: Atom<unknown>[]
     state: State
     listeners: Set<() => void>
-  }>,
-  restore: (values: Iterable<readonly [Atom<unknown>, unknown]>) => void
+  }> // debugMutableSource
 ]
 
 export type Store = StoreForProduction | StoreForDevelopment
@@ -45,7 +43,9 @@ const createStoreForProduction = (
     atom: WritableAtom<Value, Update>,
     update: Update
   ) => writeAtom(state, atom, update)
-  return [stateMutableSource, updateAtom, commitCallback]
+  const restore = (values: Iterable<readonly [Atom<unknown>, unknown]>) =>
+    restoreAtoms(state, values)
+  return [stateMutableSource, updateAtom, commitCallback, restore]
 }
 
 const createStoreForDevelopment = (
@@ -85,8 +85,8 @@ const createStoreForDevelopment = (
     stateMutableSource,
     updateAtom,
     commitCallback,
-    debugMutableSource,
     restore,
+    debugMutableSource,
   ]
 }
 
--- a/src/core/contexts.ts
+++ b/src/core/contexts.ts
@@ -86,5 +86,5 @@ export const getStoreContext = (scope?: Scope) => {
 }
 
 export const isDevStore = (store: Store): store is StoreForDevelopment => {
-  return store.length > 3
+  return store.length > 4
 }
--- a/src/core/contexts.ts
+++ b/src/core/contexts.ts
@@ -15,20 +15,22 @@ type UpdateAtom = <Value, Update>(
 type CommitCallback = () => void
 
 type StoreForProduction = [
-  MutableSource<State>, // stateMutableSource
-  UpdateAtom, // updateAtom
-  CommitCallback, // commitCallback
-  (values: Iterable<readonly [Atom<unknown>, unknown]>) => void // restore
+  stateMutableSource: MutableSource<State>,
+  updateAtom: UpdateAtom,
+  commitCallback: CommitCallback
 ]
 
 export type StoreForDevelopment = [
-  ...StoreForProduction,
-  MutableSource<{
+  stateMutableSource: MutableSource<State>,
+  updateAtom: UpdateAtom,
+  commitCallback: CommitCallback,
+  debugMutableSource: MutableSource<{
     version: number
     atoms: Atom<unknown>[]
     state: State
     listeners: Set<() => void>
-  }> // debugMutableSource
+  }>,
+  restore: (values: Iterable<readonly [Atom<unknown>, unknown]>) => void
 ]
 
 export type Store = StoreForProduction | StoreForDevelopment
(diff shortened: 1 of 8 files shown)


src/core/contexts.ts
```diff
@@ -15,22 +15,20 @@ type UpdateAtom = <Value, Update>(
 type CommitCallback = () => void
 
 type StoreForProduction = [
-  stateMutableSource: MutableSource<State>,
-  updateAtom: UpdateAtom,
-  commitCallback: CommitCallback
+  MutableSource<State>, // stateMutableSource
+  UpdateAtom, // updateAtom
+  CommitCallback, // commitCallback
+  (values: Iterable<readonly [Atom<unknown>, unknown]>) => void // restore
```
dai-shi, line 21: You don't like named tuples? I think it would show `restore` in the editor. The comment doesn't help to tell what this function is at the caller end.

src/core/contexts.ts
```diff
@@ -85,8 +85,8 @@ const createStoreForDevelopment = (
     stateMutableSource,
     updateAtom,
     commitCallback,
-    debugMutableSource,
     restore,
+    debugMutableSource,
```
dai-shi, line 89: Maybe we should now use createStoreForProduction and extend the result?
```js
const storeForProduction = ...
const debugMutableSource = ...
return [...storeForProduction, debugMutableSource]
```

src/core/contexts.ts
Thisen: I love named tuples, but I can't reuse the types, if it's a named tuple.

dai-shi: oh gosh.

Nazzanuk:
@Thisen looks good I have a couple questions, which I think are essentially the same.

If two components both use `useHydrateAtoms` will the atom value be initialised once or twice?
If the prop is externally updated atom value be overwritten?

Thisen:
> @Thisen looks good I have a couple questions, which I think are essentially the same.
> 
> If two components both use `useHydrateAtoms` will the atom value be initialised once or twice?
> If the prop is externally updated, will the atom value be overwritten?

The current implementation is:
`useHydrateAtoms` will run once per use of the hook. Therefore, if you use it multiple places, it will override.
It doesn't care about updates, as it only run once.

Do you see any issues with this?

dai-shi:
> It doesn't care about updates, as it only run once.

Yes. So, for @Nazzanuk 's use case in https://github.com/pmndrs/jotai/issues/340#issuecomment-894193393, useEffect is still required.

Thisen:
> > It doesn't care about updates, as it only run once.
> 
> Yes. So, for @Nazzanuk 's use case in https://github.com/pmndrs/jotai/issues/340#issuecomment-894193393, useEffect is still required.

Yup. We might need a better mechanism.

Nazzanuk:
I think it's ok to run only once on the first render, the value will come directly from the atom afterwards anyway, so updating can be handled in a useEffect. 

Just thinking about my scenario, I'd only want the atom to be updated once / or maybe only if the value is nullish.

I'd be hesitant to use the current hook in a deeper component with wider reuse as I may not be sure of which instance will render first/last

I could be thinking of a different scenario from what was intended, [I still have this in mind](https://github.com/pmndrs/jotai/discussions/581#discussioncomment-970662)

dai-shi:
> I think it's ok to run only once on the first render,

I also think it's ok. useEffect on user end seems like a valid use case.



> I'd be hesitant to use the current hook in a deeper component with wider reuse as I may not be sure of which instance will render first/last

Could you elaborate your concern? I almost understand what this means, but would like to have a better understanding.

dai-shi:
@Thisen I have one concern. If we restore atoms, atoms onMount won't fire...

Nazzanuk:
I think this is a bit contrived but it might help illustrate. 

```typescript
const userIdAtom = atom<number | null>(null);

const MyUserWidget = ({ userId }) => {
  useHydrateAtoms([[userIdAtom, userId]]); // what happens if userId is empty?

  const [userId, setUserId] = useAtom(userIdAtom);

  return <>...widget</>;
};

const App = <>
  <Header/>
  <Container>
    <MyUserWidget userId={23}/>
    <ChangeUserButton /> {/* Updates userIdAtom */}
    
    <LazyComponent> {/* Rendered lazily or asynchronously */}
      <MyUserWidget userId={23}/>  {/* What should userId be here? Will it reset the first widget just by rendering? */}
    </LazyComponent>
  </Container>
  <Footer/>
</>;
```

Thinking of an example was quite difficult actually, and this can all be fixed by hoisting `useHydrateAtoms` to a higher component.

src/core/contexts.ts
dai-shi: My bad. I found it not easy because of stateListener.

dai-shi:
> @Thisen I have one concern. If we restore atoms, atoms onMount won't fire...
> (It's the same if we use Provider's initialValues.)

@Thisen My bad. If we use `restoreAtoms`, the onMount works. Sorry about that.
We should probably prefer the original implementation. btw, you want to work on it after we merge #646.

dai-shi:
> But is there a use case where an atom would need to be hydrated more than once?

@Nazzanuk I'm not quite sure how people would like to use it, but `useHydrateAtoms` would work just once only for atoms that are not mounted.



> this can all be fixed by hoisting `useHydrateAtoms` to a higher component.

I agree with this. My concern is that it's hard to detect misusage from the library, so we would need a good doc about the correct usage.

Nazzanuk:
> @Nazzanuk I'm not quite sure how people would like to use it, but `useHydrateAtoms` would work just once only for atoms that are not mounted.

Ah ok, yeah I think the current implementation works really well. It will make the integration with Next.js much cleaner and more straightforward. If needed I can wrap it in another hook for my particular use case (check first if nullish).

src/utils/useHydrateAtoms.ts
```diff
@@ -0,0 +1,25 @@
...
+  const StoreContext = getStoreContext(scope)
+  const restoreAtoms = useContext(StoreContext)[3]
+
+  useMemo(() => {
+    const tuplesToRestore = []
+    for (const tuple of values) {
+      const atom = tuple[0]
+      if (atom.hydrated !== hydratedSymbol) {
+        tuplesToRestore.push(tuple)
+        atom.hydrated = hydratedSymbol
```
dai-shi: This is changing the atom config, which is shared by other stores/providers.
If we would like to check this, we should use `WeakMap`.
```ts
type HydratedAtoms = WeakSet<Atom<unknown>>
type HydratedMap = WeakMap<Store, HydratedAtoms>
```

Thisen: Check new commit.

src/utils/useHydrateAtoms.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+  if (tuplesToRestore.length) {
+    restoreAtoms(tuplesToRestore)
+  }
+}
+
+function getHydratedSet(store: Store) {
+  let hydratedSet = hydratedMap.get(store)
+  if (!hydratedSet) {
+    hydratedSet = new Set()
```
dai-shi, line 33:
```suggestion
    hydratedSet = new WeakSet()
```
I wonder if TS can yell at this.

Thisen: Ohh, that's a typo after I refactored it. It's fixed now.
I wonder why TS didn't anything...

dai-shi (approved):
Looks great. Thanks for your work!

PR #640, opened 2021-08-07
breaking: useAtom scope instead of atom scope

### LINKED ISSUE #639: Introduce `useAtom(atom, scope?)` and drop `atom.scope`
https://github.com/pmndrs/jotai/issues/619#issuecomment-888037369
I realized that probably `atom.scope` was a bad abstraction.
We have really weird scope copying/validation in various utils, which is hard to maintain and increases bundle size.
The atom scope is only needed when we get a context, so having it in the second argument of `useAtom` seems the right way. We already do this for `useAtomCallback`, `useAtomsSnapshot`, and `useGotoAtomsSnapshot`.

[Migration Plan]
We simply drop `atom.scope` in TS, but have a fallback behavior in `useAtom` to avoid breaking change in JS (and we show the deprecation message).

dai-shi:
close #639 

This changes how to use atom scope. It's not a new feature, but fixing the existing feature. (The previous atom scope abstraction was problematic.)

This makes `atom.scope` type deprecated.
`useAtom` falls back to the previous behavior, so there's no breaking changes in core.

For non-core functions, changes are technically breaking. We expect this is not too confusing as atom scope is mainly for library use. If it's too confusing, let's consider creating a preliminary patch release to only warn the upcoming change.

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -46,6 +46,9 @@ type OnMount<Update> = <S extends SetAtom<Update>>(
 export type Atom<Value> = {
   toString: () => string
   debugLabel?: string
+  /**
+   * @deprecated
+   */
   scope?: Scope
   read: Read<Value>
 }
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -47,7 +47,7 @@ export type Atom<Value> = {
   toString: () => string
   debugLabel?: string
   /**
-   * @deprecated
+   * @deprecated Instead use `useAtom(atom, scope)`
    */
   scope?: Scope
   read: Read<Value>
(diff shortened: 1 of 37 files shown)


src/core/atom.ts
```diff
@@ -46,6 +46,9 @@ type OnMount<Update> = <S extends SetAtom<Update>>(
 export type Atom<Value> = {
   toString: () => string
   debugLabel?: string
+  /**
+   * @deprecated
```
Thisen, line 50:
```suggestion
   * @deprecated Instead use `useAtom(atom, scope)`
```

Thisen (approved):

PR #644, opened 2021-08-07
fix(utils): improve useAtomCallback performance

### LINKED ISSUE #641: `useAtomCallback` creates a lot of atoms, resulting in performance issues
`useAtomCallback` dynamically creates an atom whenever it is called, once per call site. Because of that, code with a number of components, where each component uses just a couple of callbacks, can result in a very large number of atoms being created. This appears to have a _serious_ impact on performance.

I made a recording where I show the extent of the problem (sorry for crappy quality, I had to compress it because it was too big for Github). I also have a link to the Codesandbox I used.

https://codesandbox.io/s/demonstration-of-useatomcallback-performance-issues-29qtj?file=/src/App.tsx.

https://user-images.githubusercontent.com/10849003/128598500-a75f8984-a1f9-4da0-8f6b-5e7dd4207c84.mov

dai-shi:
close #641

### DIFF
--- a/src/utils/useAtomCallback.ts
+++ b/src/utils/useAtomCallback.ts
@@ -1,7 +1,9 @@
 import { useCallback, useMemo } from 'react'
-import { atom, useAtom } from 'jotai'
+import { atom } from 'jotai'
 import type { Setter, WritableAtom } from 'jotai'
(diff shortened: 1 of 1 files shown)


Thisen (approved):
Looks good.

PR #646, opened 2021-08-09
refactor(core): useMutableSource emulation without symbol and any

dai-shi:
This is to follow https://github.com/pmndrs/valtio/pull/208 (we use the same code.)

@Thisen once suggested to have precise typing in useMutableSource.ts. This covers it.

### DIFF
--- a/src/core/useMutableSource.ts
+++ b/src/core/useMutableSource.ts
@@ -9,20 +9,28 @@ export {
 
 import { useEffect, useRef, useState } from 'react'
 
-const TARGET = Symbol()
-const GET_VERSION = Symbol()
+const TARGET = '_uMS_T'
+const GET_VERSION = '_uMS_V'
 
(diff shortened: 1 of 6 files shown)


src/core/useMutableSource.ts
```diff
@@ -9,20 +9,28 @@ export {
 
 import { useEffect, useRef, useState } from 'react'
 
-const TARGET = Symbol()
-const GET_VERSION = Symbol()
+const TARGET = '_uMS_T'
```
Thisen, line 12: Why remove the use of `Symbol`?

Thisen (approved):

src/core/useMutableSource.ts
dai-shi: Because, with internal symbols, TS can't emit types properly. That's why `any` was used.

PR #648, opened 2021-08-11
refactor: rename store to scope container

dai-shi:
This is just a refactoring to rename `Store` to `ScopeContainer` and so on.


Thisen (approved):
Looks good!

PR #650, opened 2021-08-12
refactor: new store.ts based on old vanilla.ts

dai-shi:
This refactors core to be more like idiomatic `createStore`.
`contexts.ts` becomes simpler.

<details>
<summary>git diff -w main:src/core/vanilla.ts src/core/store.ts</summary>

```diff
diff --git a/src/core/vanilla.ts b/src/core/store.ts
index d01988e..d885a4b 100644
--- a/src/core/vanilla.ts
+++ b/src/core/store.ts
@@ -54,8 +54,6 @@ export type AtomState<Value = unknown> = {
   d: ReadDependencies
 }
 
-type AtomStateMap = WeakMap<AnyAtom, AtomState>
-
 type Listeners = Set<() => void>
 type Dependents = Set<AnyAtom>
 type Mounted = {
@@ -64,35 +62,28 @@ type Mounted = {
   u: OnUnmount | void
 }
 
-type MountedMap = WeakMap<AnyAtom, Mounted>
-
 // for debugging purpose only
 type StateListener = (updatedAtom: AnyAtom, isNewAtom: boolean) => void
 
-type StateVersion = number
-
-type PendingMap = Map<AnyAtom, ReadDependencies | undefined>
-
-// mutable state
-export type State = {
-  l?: StateListener
-  v: StateVersion
-  a: AtomStateMap
-  m: MountedMap
-  p: PendingMap
-}
-
-export const createState = (
+// store methods
+export const GET_VERSION = 'v'
+export const READ_ATOM = 'r'
+export const WRITE_ATOM = 'w'
+export const FLUSH_PENDING = 'f'
+export const SUBSCRIBE_ATOM = 's'
+export const RESTORE_ATOMS = 'h'
+export const DEV_GET_ATOM_STATE = 'a'
+export const DEV_GET_MOUNTED = 'm'
+
+export const createStore = (
   initialValues?: Iterable<readonly [AnyAtom, unknown]>,
   stateListener?: StateListener
-): State => {
-  const state: State = {
-    l: stateListener,
-    v: 0,
-    a: new WeakMap(),
-    m: new WeakMap(),
-    p: new Map(),
-  }
+) => {
+  let version = 0
+  const atomStateMap = new WeakMap<AnyAtom, AtomState>()
+  const mountedMap = new WeakMap<AnyAtom, Mounted>()
+  const pendingMap = new Map<AnyAtom, ReadDependencies | undefined>()
+
   if (initialValues) {
     for (const [atom, value] of initialValues) {
[...]

PR #652, opened 2021-08-12
Fix(useAtomsSnapshot): Only use values in snapshots

Thisen:
Related to #650


dai-shi (approved):
Thanks for the fix. I think it's pretty safe to merge this only with the code review.

PR #653, opened 2021-08-12
fix(query): improve initial data method

aulneau:
While experimenting with some of the new features in react query, specifically the [`persistQueryClient`](https://react-query.tanstack.com/plugins/persistQueryClient) stuff, I found a bug where if a query has been hydrated via methods exported by react-query (for SSR, or for this new persistence feature), our atom would not handle it correctly.

This fixes that by checking to see if there is data for the queryKey, if there is, treat it the same as if `initialData` was passed.

### DIFF
--- a/src/query/atomWithInfiniteQuery.ts
+++ b/src/query/atomWithInfiniteQuery.ts
@@ -50,14 +50,23 @@ export function atomWithInfiniteQuery<
       let settlePromise:
         | ((data: InfiniteData<TData> | null, err?: TError) => void)
         | null = null
-      const getInitialData = () =>
-        typeof options.initialData === 'function'
-          ? (
-              options.initialData as InitialDataFunction<
-                InfiniteData<TQueryData>
-              >
-            )()
-          : options.initialData
+
+      const getInitialData = () => {
+        let data: InfiniteData<TQueryData> | InfiniteData<TData> | undefined =
+          queryClient.getQueryData<InfiniteData<TData>>(options.queryKey)
+
+        if (!data && options.initialData)
+          data =
+            typeof options.initialData === 'function'
+              ? (
+                  options.initialData as InitialDataFunction<
+                    InfiniteData<TQueryData>
+                  >
+                )()
+              : options.initialData
+
+        return data
+      }
 
       const initialData = getInitialData()
 
@@ -114,7 +123,7 @@ export function atomWithInfiniteQuery<
 
       const observer = new InfiniteQueryObserver(queryClient, defaultedOptions)
 
-      if (!initialData) {
+      if (initialData === undefined) {
         observer
           .fetchOptimistic(defaultedOptions)
           .then(listener)
--- a/src/query/atomWithInfiniteQuery.ts
+++ b/src/query/atomWithInfiniteQuery.ts
@@ -55,7 +55,7 @@ export function atomWithInfiniteQuery<
         let data: InfiniteData<TQueryData> | InfiniteData<TData> | undefined =
           queryClient.getQueryData<InfiniteData<TData>>(options.queryKey)
 
-        if (!data && options.initialData)
+        if (data === undefined && options.initialData) {
           data =
             typeof options.initialData === 'function'
               ? (
@@ -64,6 +64,7 @@ export function atomWithInfiniteQuery<
                   >
                 )()
               : options.initialData
+        }
 
         return data
       }
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -43,10 +43,19 @@ export function atomWithQuery<
         typeof createQuery === 'function' ? createQuery(get) : createQuery
       let settlePromise: ((data: TData | null, err?: TError) => void) | null =
         null
-      const getInitialData = () =>
-        typeof options.initialData === 'function'
-          ? (options.initialData as InitialDataFunction<TQueryData>)()
-          : options.initialData
+
+      const getInitialData = () => {
+        let data: TQueryData | TData | undefined =
+          queryClient.getQueryData<TData>(options.queryKey)
+
+        if (!data && options.initialData)
+          data =
+            typeof options.initialData === 'function'
+              ? (options.initialData as InitialDataFunction<TQueryData>)()
+              : options.initialData
+
+        return data
+      }
 
       const initialData = getInitialData()
 
@@ -94,7 +103,7 @@ export function atomWithQuery<
         defaultedOptions.staleTime = 1000
       }
       const observer = new QueryObserver(queryClient, defaultedOptions)
-      if (!initialData) {
+      if (initialData === undefined) {
         observer
           .fetchOptimistic(defaultedOptions)
           .then(listener)
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -48,12 +48,12 @@ export function atomWithQuery<
         let data: TQueryData | TData | undefined =
           queryClient.getQueryData<TData>(options.queryKey)
 
-        if (!data && options.initialData)
+        if (data === undefined && options.initialData) {
           data =
             typeof options.initialData === 'function'
               ? (options.initialData as InitialDataFunction<TQueryData>)()
               : options.initialData
-
+        }
         return data
       }
 
--- a/tests/query/atomWithQuery.test.tsx
+++ b/tests/query/atomWithQuery.test.tsx
@@ -269,7 +269,6 @@ it('query with enabled (#500)', async () => {
   await findByText('hidden')
 
   fireEvent.click(getByText('toggle'))
-  await findByText('loading')
   await findByText('count: 1')
 })
 
@@ -277,7 +276,7 @@ it('query with initialData test', async () => {
   const mockFetch = jest.fn(fakeFetch)
 
   const countAtom = atomWithQuery(() => ({
-    queryKey: 'count1',
+    queryKey: 'initialData_count1',
     queryFn: async () => {
       return await mockFetch({ count: 10 })
     },
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -70,16 +70,16 @@
     }
   },
   "query.js": {
-    "bundled": 5106,
-    "minified": 2143,
-    "gzipped": 684,
+    "bundled": 5988,
+    "minified": 2530,
+    "gzipped": 786,
     "treeshaked": {
       "rollup": {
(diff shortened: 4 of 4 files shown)


src/query/atomWithInfiniteQuery.ts
```diff
@@ -1,3 +1,5 @@
+import { atom } from 'jotai'
```
aulneau, line 1: ESlint said these needed to move up

aulneau: ignore -- my deps were stale

src/query/atomWithInfiniteQuery.ts
```diff
@@ -24,6 +24,8 @@ export type AtomWithInfiniteQueryOptions<
   queryKey: QueryKey
 }
 
+const getInitialDataCache = new WeakMap()
```
aulneau, line 27: I noticed that `getInitialData` will run many times when an atom is used -- this seems to not be desirable. @dai-shi curious what you think about this method.

tests/query/atomWithQuery.test.tsx
```diff
@@ -5,307 +5,309 @@ import { atomWithQuery } from '../../src/query'
 import { getTestProvider } from '../testUtils'
 import fakeFetch from './fakeFetch'
 
-const Provider = getTestProvider()
+describe(atomWithQuery.name, function () {
```
aulneau, line 8: wrapping in `describe` is generally best practice

dai-shi: I don't disagree about using `describe`. (Not sure about `.name` pattern).

Would you create a separate PR for this change? I mean if we prefer `describe` generally, should we change all tests?

src/query/atomWithInfiniteQuery.ts
```diff
@@ -114,7 +128,7 @@ export function atomWithInfiniteQuery<
 
       const observer = new InfiniteQueryObserver(queryClient, defaultedOptions)
 
-      if (!initialData) {
+      if (typeof initialData === 'undefined') {
```
dai-shi, line 131: Would you help me understand the background of this change? If this change is necessary, I may misunderstand something and feel checking `=== 'undefined'` might not be enough.

src/query/atomWithInfiniteQuery.ts
dai-shi: At a glance, it feels weird. Do you know why `getinitialData` runs many times? I expect there should be a better fix.

aulneau: I'm not sure -- but when I ran a single test, and did `console.log('running')` in `getInitialData`, it ran 4 times. This could be a test-only phenomena though. I'll test in an app

src/query/atomWithInfiniteQuery.ts
aulneau: `initialData` could be of `null` or `0` type, which would be a defined value, but the previous check would have assumed it was equal to undefined

tests/query/atomWithQuery.test.tsx
aulneau: > Would you create a separate PR for this change? I mean if we prefer describe generally, should we change all tests?
Sure! makes sense to be consistent 

> (Not sure about .name pattern)
I've found this is nice if ever a refactor happens and the method changes.

src/query/atomWithInfiniteQuery.ts
aulneau: this was also to fix the tests
see this commit failing:
https://github.com/pmndrs/jotai/pull/653/commits/b1a931530cbd19215bd25ec945131ecdac48577e
and this passes when ensuring `getInitialData` runs only once per atom
https://github.com/pmndrs/jotai/pull/653/commits/8d3f49ae08e33268c23d35503f9488b69852994b

src/query/atomWithInfiniteQuery.ts
dai-shi: > `initialData` could be of `null` or `0` type

OK, that makes sense now. I wasn't aware of it. But then, a user can do `initialData = undefined`, right? I would introduce a new variable without relying of `undefined`.

tests/query/atomWithQuery.test.tsx
dai-shi: Sounds good.

src/query/atomWithInfiniteQuery.ts
dai-shi: OK, if it's test failing, it's easier for me to look into it.

> This could be a test-only phenomena though. I'll test in an app

Let me know how it goes, and if I can jump in to investigate.

src/query/atomWithInfiniteQuery.ts
aulneau: >  a user can do initialData = undefined, right? I would introduce a new variable without relying of undefined.

Can you give an example, or rephrase? I think if a user were to pass `initialData: undefined` it would be equal to not passing it.

dai-shi: Hm, if a user knows `undefined` to lead to Suspense, I think it's fine as is.

I thought about a function:
```js
// case A
options.initialData = () => {
  let something = null
  if (...) {
    something = ...
  }
  return something
}

// case B
options.initialData = () => {
  let something
  if (...) {
    something = ...
  }
  return something
}
```

I expect case A and case B to behave the same.

aulneau: yeah, I think the argument for `null == undefined` is valid. There are cases in which it could be that someone has fetched data on the server where the result is `0` and it's a valid value, but if we just did `!initialData` it would suspend (where the user would expect it to not).

Additionally, if we did support `null` as a valid value for initialData, it would be a way for folks to get around suspense by default for query atoms. It is more of a side effect/hack and I can see why it might not be desirable.

src/query/atomWithInfiniteQuery.ts
aulneau: This does seem to _only_ happen in the testing context. I'm curious if it would still render the same atom multiple times if we used the hooks testing library instead.

aulneau: If you'd like to look, that'd be helpful. The issue is the function runs 4 times very quickly, and as the queryClient updates over time, the 3rd time it runs, it receives a value thus never suspends (or suspends very very quickly).

src/query/atomWithInfiniteQuery.ts
dai-shi: okay, my concern was too subtle. I don't mean to do `!initialData`.
Let's just make it 
```js
      if (initialData === undefined) {
```
to be consistent with other code. (We don't assume the `undefined` global variable to be overwritten.)

tests/query/atomWithQuery.test.tsx
```diff
@@ -269,15 +269,14 @@ it('query with enabled (#500)', async () => {
   await findByText('hidden')
 
   fireEvent.click(getByText('toggle'))
-  await findByText('loading')
```
aulneau, line 272: @dai-shi this was the issue previously -- because we are remounting the same query atom, the previous value in the cache is used immediately, thus there is no suspend.

I'll work to improve this test in a follow up PR related to `enabled` functionality

tests/query/atomWithQuery.test.tsx
```diff
@@ -269,15 +269,14 @@ it('query with enabled (#500)', async () => {
...
   await findByText('count: 1')
 })
 
 it('query with initialData test', async () => {
   const mockFetch = jest.fn(fakeFetch)
 
   const countAtom = atomWithQuery(() => ({
-    queryKey: 'count1',
+    queryKey: 'initialData_count1',
```
aulneau, line 279: naming collision from previous test above

src/query/atomWithInfiniteQuery.ts
aulneau: This was a mistake, feel free to ignore :)

src/query/atomWithInfiniteQuery.ts
aulneau: fixed!

aulneau:
@dai-shi ready for your 👀  anytime :)

src/query/atomWithQuery.ts
```diff
@@ -43,10 +43,19 @@ export function atomWithQuery<
...
+
+      const getInitialData = () => {
+        let data: TQueryData | TData | undefined =
+          queryClient.getQueryData<TData>(options.queryKey)
+
+        if (!data && options.initialData)
+          data =
+            typeof options.initialData === 'function'
+              ? (options.initialData as InitialDataFunction<TQueryData>)()
+              : options.initialData
```
dai-shi, line 55:
```suggestion
        if (!data && options.initialData) {
          data =
            typeof options.initialData === 'function'
              ? (options.initialData as InitialDataFunction<TQueryData>)()
              : options.initialData
        }
```
sorry for the tiny styling thing. this is mostly preference but to make style consistent.
(I think there's an eslint rule. maybe we can try to enable it.)

src/query/atomWithInfiniteQuery.ts
```diff
@@ -50,14 +50,23 @@ export function atomWithInfiniteQuery<
...
-                InfiniteData<TQueryData>
-              >
-            )()
-          : options.initialData
+
+      const getInitialData = () => {
+        let data: InfiniteData<TQueryData> | InfiniteData<TData> | undefined =
+          queryClient.getQueryData<InfiniteData<TData>>(options.queryKey)
+
+        if (!data && options.initialData)
```
dai-shi, line 58: Doesn't this have to be `data === undefined && ...`? For data can be null or 0?

dai-shi (commented):
Looks so clean now. Thanks for the investigation in tests. Added a few minor comments.

src/query/atomWithInfiniteQuery.ts
aulneau: great catch

aulneau: fixed in https://github.com/pmndrs/jotai/pull/653/commits/d2599e20e2eea00b1b23f100b725da5b463236ce

src/query/atomWithQuery.ts
aulneau: yep, fixed in https://github.com/pmndrs/jotai/pull/653/commits/d2599e20e2eea00b1b23f100b725da5b463236ce

dai-shi (approved):

PR #665, opened 2021-08-18
feat(query): enabled functionality

aulneau:
This adds the `enabled` functionality described here -> https://react-query.tanstack.com/guides/disabling-queries#_top

When the user passes any value for `enabled`, the data value becomes potentially undefined. I've added two tests, one for disabled at start, and one for toggling enabled functionality.

@dai-shi lmk what you think of this implementation, and if it's good I'll do it to the infinity query atom too!

### DIFF
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -1,18 +1,16 @@
 import { QueryObserver } from 'react-query'
 import type {
   InitialDataFunction,
-  QueryClient,
   QueryKey,
   QueryObserverOptions,
   QueryObserverResult,
 } from 'react-query'
 import { atom } from 'jotai'
-import type { Getter, PrimitiveAtom, WritableAtom } from 'jotai'
+import type { PrimitiveAtom, WritableAtom } from 'jotai'
 import { queryClientAtom } from './queryClientAtom'
+import type { CreateQueryOptions, GetQueryClient } from './types'
 
 export type AtomWithQueryAction = { type: 'refetch' }
-export type CreateQueryOptions<Options> = Options | ((get: Getter) => Options)
-export type GetQueryClient = (get: Getter) => QueryClient
 export type AtomWithQueryOptions<TQueryFnData, TError, TData, TQueryData> =
   QueryObserverOptions<TQueryFnData, TError, TData, TQueryData> & {
     queryKey: QueryKey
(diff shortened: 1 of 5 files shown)


src/query/atomWithQuery.ts
```diff
@@ -11,28 +11,62 @@ import type { Getter, PrimitiveAtom, WritableAtom } from 'jotai'
 import { queryClientAtom } from './queryClientAtom'
 
 export type AtomWithQueryAction = { type: 'refetch' }
-
+export type CreateQueryOptions<Options> = Options | ((get: Getter) => Options)
+export type GetQueryClient = (get: Getter) => QueryClient
 export type AtomWithQueryOptions<TQueryFnData, TError, TData, TQueryData> =
```
dai-shi, line 16: Types seem to work correctly now, so this is ok as is.
I just expected to omit `'enabled'` from this `AtomWithQueryOptions` type like in L27, which seems safer? what do you think?

dai-shi (commented):
It looks great actually. #665 is taken into account, right?

src/query/atomWithQuery.ts
aulneau: I played around with that and a few other options, but this was the result that allowed for all the previous atoms in the tests to not change their typings, but correctly change the return value of the new test atoms from a guaranteed return value to one that is potentially `undefined`. 

My understanding of how this is currently set up is that this type provides the option for `enabled` to be passed, while `AtomWithQueryOptionsWithEnabled` then is used when `enabled` is passed (for any value, true/false). When I remove it from AtomWithQueryOptions, it causes type errors on L144 and another, thinking `enabled` is never a correct option to pass.

aulneau:
> It looks great actually. #665 is taken into account, right?

#665 is this PR, did you mean to reference something else?

aulneau:
@dai-shi added in the changes for infinite too, should be good to go!

dai-shi:
> > It looks great actually. #665 is taken into account, right?
> 
> #665 is this PR, did you mean to reference something else?

I meant #655.

src/query/atomWithQuery.ts
dai-shi: Thanks. I didn't dig into it, and if you did it's good to go.

dai-shi (approved):
LGTM

PR #670, opened 2021-08-20
feat(query): infinite query page refresh

aulneau:
This enables support of the new feature where you can refetch a given page for an infinite query, see https://github.com/tannerlinsley/react-query/pull/2557

### DIFF
--- a/src/query/atomWithInfiniteQuery.ts
+++ b/src/query/atomWithInfiniteQuery.ts
@@ -7,13 +7,22 @@ import type {
   QueryObserverResult,
 } from 'react-query'
 import { atom } from 'jotai'
-import type { Getter, WritableAtom } from 'jotai'
+import type { WritableAtom } from 'jotai'
 import { queryClientAtom } from './queryClientAtom'
 import { CreateQueryOptions, GetQueryClient } from './types'
 
-export type AtomWithInfiniteQueryAction = {
+export interface AtomWithInfiniteQueryRefetchPageAction<TData> {
+  type: 'refetchPage'
+  payload: (page: TData, index: number, allPages: TData[]) => boolean
+}
+
+export interface AtomWithInfiniteQueryActionBase {
   type: 'refetch' | 'fetchNextPage' | 'fetchPreviousPage'
 }
+
+type AtomWithInfiniteQueryAction<TData> =
+  | AtomWithInfiniteQueryActionBase
+  | AtomWithInfiniteQueryRefetchPageAction<TData>
 
 export type AtomWithInfiniteQueryOptions<
   TQueryFnData,
@@ -52,7 +61,7 @@ export function atomWithInfiniteQuery<
   getQueryClient?: GetQueryClient
 ): WritableAtom<
   InfiniteData<TData | TQueryData> | undefined,
-  AtomWithInfiniteQueryAction
+  AtomWithInfiniteQueryAction<TData>
 >
 export function atomWithInfiniteQuery<
   TQueryFnData,
@@ -64,7 +73,10 @@ export function atomWithInfiniteQuery<
     AtomWithInfiniteQueryOptions<TQueryFnData, TError, TData, TQueryData>
   >,
   getQueryClient?: GetQueryClient
-): WritableAtom<InfiniteData<TData | TQueryData>, AtomWithInfiniteQueryAction>
+): WritableAtom<
+  InfiniteData<TData | TQueryData>,
+  AtomWithInfiniteQueryAction<TData>
+>
 export function atomWithInfiniteQuery<
   TQueryFnData,
   TError,
@@ -77,7 +89,7 @@ export function atomWithInfiniteQuery<
   getQueryClient: GetQueryClient = (get) => get(queryClientAtom)
 ): WritableAtom<
   InfiniteData<TData | TQueryData> | undefined,
-  AtomWithInfiniteQueryAction
+  AtomWithInfiniteQueryAction<TData>
 > {
   const queryDataAtom = atom(
     (get) => {
@@ -183,11 +195,15 @@ export function atomWithInfiniteQuery<
       }
       return { dataAtom, observer, options }
     },
-    (get, _set, action: AtomWithInfiniteQueryAction) => {
+    (get, _set, action: AtomWithInfiniteQueryAction<TData>) => {
       const { observer } = get(queryDataAtom)
       switch (action.type) {
         case 'refetch': {
           void observer.refetch()
+          break
+        }
+        case 'refetchPage': {
+          void observer.refetch({ refetchPage: action.payload as any })
           break
         }
         case 'fetchPreviousPage': {
@@ -204,7 +220,7 @@ export function atomWithInfiniteQuery<
 
   const queryAtom = atom<
     InfiniteData<TData | TQueryData> | undefined,
-    AtomWithInfiniteQueryAction
+    AtomWithInfiniteQueryAction<TData>
   >(
     (get) => {
       const { dataAtom } = get(queryDataAtom)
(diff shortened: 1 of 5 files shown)


src/query/atomWithInfiniteQuery.ts
```diff
@@ -183,11 +197,12 @@ export function atomWithInfiniteQuery<
...
     },
-    (get, _set, action: AtomWithInfiniteQueryAction) => {
+    (get, _set, action: AtomWithInfiniteQueryAction<TData>) => {
       const { observer } = get(queryDataAtom)
-      switch (action.type) {
+      const { type, ...options } = action
+      switch (type) {
         case 'refetch': {
-          void observer.refetch()
+          void observer.refetch(options as any)
```
dai-shi, line 205: If we destruct `options` within the case block, can we avoid `any`?

src/query/atomWithInfiniteQuery.ts
```diff
@@ -5,16 +5,27 @@ import type {
...
+  type: 'refetch'
 }
 
+export interface AtomWithInfiniteQueryPageActions {
+  type: 'fetchNextPage' | 'fetchPreviousPage'
+}
+
+export type AtomWithInfiniteQueryAction<TData> =
+  | AtomWithInfiniteQueryPageActions
+  | AtomWithInfiniteQueryRefetchAction<TData>
```
dai-shi, line 27: If I were to do this, I'd do this:
```ts
export type AtomWithInfiniteQueryAction<TData> =
  | { type: 'refetch', options: Partial<RefetchOptions & RefetchQueryFilters<TData>> }
  | { type: 'fetchNextPage' }
  | { type: 'fetchPreviousPage' }
```

This will be easy to avoid `any`.
It's more or less preference, so it's up to you.

---

```ts
export type AtomWithInfiniteQueryAction<TData> =
  | { type: 'refetch' } & Partial<RefetchOptions & RefetchQueryFilters<TData>>
  | { type: 'fetchNextPage' }
  | { type: 'fetchPreviousPage' }
```
This is the just a syntax difference from yours. I prefer this to reduce exported types, but again I leave it to you as some prefer more flexible types. (zustand is like so.)

dai-shi (approved):
It looks good. Left comments in case you are interested.

src/query/atomWithInfiniteQuery.ts
aulneau: nice yeah this is much cleaner!

src/query/atomWithInfiniteQuery.ts
aulneau: I don't quite understand the issue that is happening. I think it's some how picking up an inferred type. It wants the data to be `InfiniteData<TData>`, but that is not correct, as it will be just the unwrapped `TData`. I cast this: 
```ts
        case 'refetch': {
          const { type: _type, ...options } =
            action as AtomWithInfiniteQueryAction<InfiniteData<TData>>
          void observer.refetch(options)
          break
        }
```

which is better than `any`, but not sure where it's picking up the incorrect typing. If you'd like to look, feel free!

This makes it so the typing for users is correct, but need to cast behind the scenes.

dai-shi: Fixed db8cef4

We could move `InfiniteData<>` into `AtomWithInfiniteQueryAction` type alias.

aulneau: Right, that's what I thought too, but it's technically incorrect. Doing this will show the end user the incorrect types for what is returned. See this line https://github.com/pmndrs/jotai/pull/670/files#diff-23e85af4de4d11957f0e23e4ecb338798f2d748ea3ffaee35ffcc83162eefab6R246 the first value is would be `TData` in practice, but the change you added would make it the paginated version which is incorrect.

aulneau: What I'm stumped on is where the Observer is picking up on the types such that the refetch method expects the paginated types. I think it's being inferred somehow, but not sure where.

dai-shi: Do you mean, we want to receive `action: AtomWithInfiniteQueryAction<TData>`, but need to pass `Partial<RefetchOptions & RefetchQueryFilters<InfiniteData<TData>>>` to `observer.refetch()`? That doesn't seem to work from the code, so there might be some bugs somewhere.

dai-shi: de9b177 seems better?

but I needed to patch this to pass types:
```diff
--- node_modules/react-query/types/core/queryObserver.d.ts.bak	2021-08-21 12:20:21.000000000 +0900
+++ node_modules/react-query/types/core/queryObserver.d.ts	2021-08-21 13:21:42.000000000 +0900
@@ -40,7 +40,7 @@
     getNextResult(options?: ResultOptions): Promise<QueryObserverResult<TData, TError>>;
     getCurrentQuery(): Query<TQueryFnData, TError, TQueryData, TQueryKey>;
     remove(): void;
-    refetch(options?: RefetchOptions & RefetchQueryFilters<TData>): Promise<QueryObserverResult<TData, TError>>;
+    refetch(options?: RefetchOptions & RefetchQueryFilters<TQueryFnData>): Promise<QueryObserverResult<TData, TError>>;
     fetchOptimistic(options: QueryObserverOptions<TQueryFnData, TError, TData, TQueryData, TQueryKey>): Promise<QueryObserverResult<TData, TError>>;
     protected fetch(fetchOptions?: ObserverFetchOptions): Promise<QueryObserverResult<TData, TError>>;
     private executeFetch;
```

I didn't dig into it. So, I might be wrong, but if this is RQ's bug, type assertion on our end is unavoidable.

aulneau: Yeah, that's what I'm trying to understand too, I can't determine if it is in fact a RQ bug, or if some how the type is inferred from something else in our implementation.

aulneau: What you have looks good though, so we can ship this anytime and come back to it. curious if @TkDodo has any ideas?

TkDodo: ugh maybe I screwed up the types for refetch. Can take a look after my vacation.

aulneau: @TkDodo no rush at all, enjoy your vacation 👌✨

dai-shi: Okay, meanwhile, let me merge this.
If it turns out that it's not a bug in RQ, let's revisit the `FIXME` part again.

TkDodo: can confirm that the types are wrong on our side. need to take a closer look though, not an easy fix 😅

TkDodo: okay, this is really tricky. partly because `refetch` exists on normal useQuery as well as useInfiniteQuery, but the types are different, and more so because the `ObserverResult` type doesn't know about the type of the queryFn. If `select` is added to the mix, it gets really complex, and I think I'd need to add a third generic to the ObserverResult and that escalates quickly. I thought for `refetch` it would be possible to have the types be inferred correctly 😔.

The easiest fix would be to remove the type and just go back to `unknown`. It's the same as `queryClient.refetchQueries` also gets, and it would be on the users to type-narrow. Would that work for you as well?

TkDodo: I created a PR here: https://github.com/tannerlinsley/react-query/pull/2638

can you help me verify if that fixes your issue please :)

dai-shi: @aulneau can you?

aulneau: sure! i'll take a look today or tomorrow and get back to ya :) thanks @TkDodo!

TkDodo: I merged it today. Please let me know if it helped your use-case, thanks

PR #674, opened 2021-08-23
atomWithStorageAtom

jonaswre:
I had the situation where the storage was dependant on something I had to retrieve from an atom. Thats why i changed atomWithStorage to receive a storage from an atom and get the initialValue from a Read<Value> function.

It worked for me, but I thought i might be good be be discussed open and add to the official utils.

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -1,10 +1,10 @@
-type Getter = {
+export type Getter = {
   <Value>(atom: Atom<Value | Promise<Value>>): Value
   <Value>(atom: Atom<Promise<Value>>): Value
   <Value>(atom: Atom<Value>): Value
 }
 
-type WriteGetter = Getter & {
+export type WriteGetter = Getter & {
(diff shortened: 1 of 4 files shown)


dai-shi:
Interesting use case.
I wonder if your use case can be solved with composition? something like this?

```js
const storageAtomAtom = atom((get) => atomWithStorage(...))
const anAtom = atom(
  (get) => {
    const storageAtom = get(storageAtomAtom)
    return get(storageAtom)
  },
  (_get, set, update) => {
    const storageAtom = get(storageAtomAtom)
    set(storageAtom, update)
  }
}
```

---
It's often the case you want to open an issue first to discuss about the idea. It's totally fine to open a PR directly though, if you are working on coding as an experiment or study.

dai-shi:
While this is not going to be merged for several reasons, the idea seems interesting. Please open a new issue or discussion and we hope to discuss what we can do.

PR #686, opened 2021-09-04
wip: use-sync-external-store

dai-shi:
😕


dai-shi:
Let's go with #687 for now.
We might be going to revisit this, once uSES becomes stable, and if it really solves something.

PR #689, opened 2021-09-05
fix(core): return promise void on write atom and set

### LINKED ISSUE #607: Thrown errors swallowed when updating state with async function
With a read-write or write only atom that take an async function for the write argument (for using the "last" value), the errors seem to be swallowed.

```tsx
const fooAtom = atom(0, async (get, set, update) => {
  const oldValue = get(fooAtom);
  const newValue = await update(oldValue);
  set(fooAtom, newValue);
}

// Then in some component...
const [foo, setFoo] = useAtom(fooAtom);

const onClick = () => {
  try {
    await setFoo(async (old) => await someFailableAsyncThing()) 
  } catch (e) {
    console.log("This will never be run");
  }
}

return <button onClick={onClick}>Foo</button>
```

As pointed out by @dai-shi, handling errors in this codepath is [not yet implemented](https://github.com/pmndrs/jotai/blob/2c3e3c31a9491ab5032e1d71db6abb5800e2be53/src/core/vanilla.ts#L493). From a DX point of view, it may make sense for the `setState` writer that is provided by `useAtom` to always re-throw any errors that are thrown by an update function, but there may be other consequences to how this is handled due to the mechanisms of `Suspense`.

### LINKED ISSUE #685: Unable to await an async writable atom from another async writable atom
i am using jotai for my auth flow and have a number of async writable (write-only) atoms. some of them depend on other async writable atoms. i tried to use `await set(otherAsyncAtom);` to trigger the other async atom and wait for it to complete, but execution continues immediately (no awaiting).

here’s a simple (contrived) demonstration of the issue, using `console.log` to show when each part of the async writable atoms execute: https://codesandbox.io/s/jotai-await-async-set-2esbb

is this the expected behavior? and if so, is there a different way to accomplish what i am looking for? thanks for all of your assistance and for an excellent library!

dai-shi:
close #685 
close #607 

This is partially reverting #508 change.


dai-shi:
@acusti @mdtusz Could you please test you use cases with this PR.
https://ci.codesandbox.io/status/pmndrs/jotai/pr/689
Open this and find "Local Install Instructions".

dai-shi:
@aulneau We discussed once about TypeScript issue regarding this, AFAIR.
[typescript playground](https://www.typescriptlang.org/play?#code/DYUwLgBATiDOCuwwC4IAUoHsC2BLWIAPAG6a4AmAfBAD4SkUQC8EAFKwJTPUDeAvh04AoIQENYATwB2AYwgAzeLLC5MUhVM4QeQiHuhxEYXftEB3UbkgwESIXyFA)
This looks okay. Would you double check if this PR doesn't break your use case?

acusti:
@dai-shi i just installed the version from that link (`https://pkg.csb.dev/pmndrs/jotai/commit/d8ea2e48/jotai`) and it seems to do just what i need! i’m not sure about the implication of these changes and React.Suspense. should awaiting the result of calling the async write atoms write function from a useEffect hook or useCallback trigger the react suspense fallback? it didn’t in my testing when using it with useEffect, but i don’t know the expected behavior.

dai-shi:
@acusti Thanks for the confirmation!

> should awaiting the result of calling the async write atoms write function from a useEffect hook or useCallback trigger the react suspense fallback?

No, it will only suspend if you `useAtom` the async write atoms.

Thisen (approved):
Great change, made it more readable aswell.

Issue #672, opened 2021-08-22
`loadable` API for using async atoms without suspense


Pinpickle:
Prior discussion: https://github.com/pmndrs/jotai/issues/269

Recoil has a `useRecoilValueLoadable` for bypassing suspense and returning a "loadable" object. This is useful in some cirumstances when suspense is not desired, and is also much easier than implementing it via the method recommended in the [Jotai docs](https://docs.pmnd.rs/jotai/guides/no-suspense).

<details>
  <summary>No-suspense code example</summary>

  ```
  const fetchResultAtom = atom({ loading: true, error: null, data: null })
  const runFetchAtom = atom(
    (get) => get(fetchResultAtom),
    (_get, set, url) => {
      const fetchData = async () => {
        set(fetchResultAtom, (prev) => ({ ...prev, loading: true }))
        try {
          const response = await fetch(url)
          const data = await response.json()
          set(fetchResultAtom, { loading: false, error: null, data })
        } catch (error) {
          set(fetchResultAtom, { loading: false, error, data: null })
        }
      }
      fetchData()
    }
  )
  runFetchAtom.onMount = (runFetch) => {
    runFetch('https://json.host.com')
  }

  const Component = () => {
    const [result] = useAtom(runFetchAtom)

    console.log(result) // { loading: ..., error: ..., data: ... }
    return <div>...</div>
  }
  ```
</details>

I'm proposing a read-only `loadable` util which returns an atom that converts an async atom into a sync atom, just like Recoil's loadable utils.

Depending on context, consumers may want errors to still be thrown or captured into the loadable object. There are two ways of handling this.

### 1. Separate `suspendable`/`loadable` APIs

```ts
const SUSPENDED = Symbol("jotai/atomSuspended");

type Loadable<T, TError = any> = { type: "loading" } | { type: "value", value: T } | { type: "error", value: TError };

// will still throw on error
const suspendable = <T>(atom<Promise<T> | T>) => T | SUSPENDED;

// will not throw on error
const loadable = <T>(atom<Promise<T> | T>) => Loadable<T>;
```

This has the advantage of calling a single function without parameters whether you want errors to be thrown or not. The `suspendable` type signature is also much simpler, as we don't need to be able to disambiguate between value and error types.

### 2. Param in the `loadable` function

```ts
function loadable<Value, Update>(
  atom: WritableAtom<Value, Update>
): WritableAtom<{ loading: true; data?: Value } | { data: Value } | { error: unknown }, Update>

[...]

Issue #601, opened 2021-07-18
atomWithObservable can't recover from error (and also with any atoms?)


dai-shi:
While working on #599, I noticed retrying from ErrorBoundary doesn't work (there's a failing test skipped). It's probably because the atom value (error) is stored in Provider, so re-render with the atom doesn't trigger a retry. If this is the case, we'd need to find a fundamental solution. 🤔 

Better to start with investigating how useState + useContext would work.

tsanyqudsi:
jotai v.1.2.1 apparently have this :

```
node_modules/jotai/utils/atomWithObservable.d.ts:13:5 - error TS1170: A computed property name in a type literal must refer to an expression whose type is a literal type or a 'unique symbol' type.

13     [Symbol.observable]?: () => ObservableLike<T>;
       ~~~~~~~~~~~~~~~~~~~

node_modules/jotai/utils/atomWithObservable.d.ts:13:13 - error TS2339: Property 'observable' does not exist on type 'SymbolConstructor'.

13     [Symbol.observable]?: () => ObservableLike<T>;
```

dai-shi:
@tsanyqudsi Sorry about the inconvenience, #605 will fix it.

tsanyqudsi:
my problem is confirmed : fixed with #605

dai-shi:
@tsanyqudsi v1.2.2 is published with #605.

tsanyqudsi:
updated and fixed. Thanks @dai-shi :D

PR #698, opened 2021-09-09
feat(utils): Add loadable

Pinpickle:
Fixes: https://github.com/pmndrs/jotai/issues/672 (or at least tries to)

This implements loadable as discussed in the mentioned issue.

I can't get all edge-cases to work. In particular, recovering from error, when an atom goes through these states:

`suspense -> error -> suspense -> resolved`

Going from `error -> suspense` does not trigger a re-render.

I think this might need some changes to core, as I suspect by catching the error/promise, we are preventing some state-keeping logic from happening.

@dai-shi could you have a look? It's the last test in `loadable.test.tsx` that's failing.

### DIFF
--- a/src/utils/loadable.ts
+++ b/src/utils/loadable.ts
@@ -1,0 +1,52 @@
+import { atom } from 'jotai'
+import type { Atom } from 'jotai'
+
+const loadableAtomCache = new WeakMap<Atom<unknown>, Atom<Loadable<unknown>>>()
+const errorLoadableCache = new WeakMap<object, Loadable<never>>()
+
+type Loadable<Value> =
+  | { state: 'loading' }
+  | { state: 'hasError'; error: unknown }
+  | { state: 'hasData'; data: Value }
+
+const LOADING_LOADABLE: Loadable<never> = { state: 'loading' }
+
+export function loadable<Value>(anAtom: Atom<Value>): Atom<Loadable<Value>> {
+  const cachedAtom = loadableAtomCache.get(anAtom)
+  if (cachedAtom) {
+    return cachedAtom as Atom<Loadable<Value>>
+  }
+
+  const derivedAtom = atom((get): Loadable<Value> => {
+    try {
+      const value = get(anAtom)
+
+      return {
+        state: 'hasData',
+        data: value,
+      }
+    } catch (error) {
+      if (error instanceof Promise) {
+        return LOADING_LOADABLE
+      }
+
+      const cachedErrorLoadable = errorLoadableCache.get(error as Error)
+
+      if (cachedErrorLoadable) {
+        return cachedErrorLoadable
+      }
+
+      const errorLoadable: Loadable<never> = {
+        state: 'hasError',
+        error,
+      }
+
+      errorLoadableCache.set(error as Error, errorLoadable)
+      return errorLoadable
+    }
+  })
+
+  loadableAtomCache.set(anAtom, derivedAtom)
+
+  return derivedAtom
+}
--- a/tests/utils/loadable.test.tsx
+++ b/tests/utils/loadable.test.tsx
@@ -1,0 +1,149 @@
+import { fireEvent, render } from '@testing-library/react'
+import { Atom, atom } from '../../src/index'
+import { loadable, useAtomValue, useUpdateAtom } from '../../src/utils'
+import { getTestProvider } from '../testUtils'
+
+const Provider = getTestProvider()
+
+it('loadable turns suspense into values', async () => {
+  let resolveAsync!: (x: number) => void
+  const asyncAtom = atom(() => {
+    return new Promise<number>((resolve) => (resolveAsync = resolve))
+  })
+
+  const { findByText, getByText } = render(
+    <Provider>
+      <LoadableComponent asyncAtom={asyncAtom} />
+    </Provider>
+  )
+
+  getByText('Loading...')
+  resolveAsync(5)
+  await findByText('Data: 5')
+})
+
+it('loadable turns errors into values', async () => {
+  let rejectAsync!: (error: Error) => void
+  const asyncAtom = atom(() => {
+    return new Promise<number>((resolve, reject) => (rejectAsync = reject))
+  })
+
+  const { findByText, getByText } = render(
+    <Provider>
+      <LoadableComponent asyncAtom={asyncAtom} />
+    </Provider>
+  )
+
+  getByText('Loading...')
+  rejectAsync(new Error('An error occurred'))
+  await findByText('Error: An error occurred')
+})
+
+it('loadable turns primitive throws into values', async () => {
+  let rejectAsync!: (errorMessage: string) => void
+  const asyncAtom = atom(() => {
+    return new Promise<number>((resolve, reject) => (rejectAsync = reject))
+  })
+
+  const { findByText, getByText } = render(
+    <Provider>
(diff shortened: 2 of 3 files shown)


dai-shi:
Nice work.
I have some coding preference to apply, reducing bundle size, fixing refAtom usage, but in general, it looks good.

However:

> I can't get all edge-cases to work. In particular, recovering from error

Oh, right... This is not resolved yet, if it's same as #601.
Yeah, I agree it's probably core responsibility.
I think it's going to take some time for this.
Should we fix it first, or could `loadable` be useful with this limitation for now?

Pinpickle:
@dai-shi I wasn't aware that it was a bug for all atoms

Given that this is a bug with all atoms, I _think_ it's OK to release as-is. It should probably be documented somewhere that error recovery doesn't work. Should I mark the test as skipped or remove altogether?

dai-shi:
I'm not 100% sure if it's the same bug as #601. (tbh, I haven't spent much time on it yet.)

Please leave the tests as skipped. Let's tackle it separately, if this is useful as is.
I'll request some changes on code.

src/utils/loadable.ts
```diff
@@ -0,0 +1,67 @@
+import { atom } from 'jotai'
+import type { Atom } from 'jotai'
+import { getWeakCacheItem, setWeakCacheItem } from './weakCache'
+
+const loadableAtomCache = new WeakMap()
+
+export type Loadable<Value> =
+  | LoadableLoading
+  | LoadableError
+  | LoadableValue<Value>
```
dai-shi, line 10: We prefer simpler (highly subjective though) type aliases, and avoid exporting types unless it's a huge pain. Let's see how it goes without exporting it.
```suggestion
type Loadable<Value> =
  | { state: 'loading' }
  | { state: 'hasError'; error: unknown }
  | { state: 'hasData'; data: Value }
```

src/utils/loadable.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface LoadableLoading {
+  state: 'loading'
+}
+
+export interface LoadableValue<Value> {
+  state: 'hasData'
+  data: Value
+}
+
+const LOADING_LOADABLE: LoadableLoading = { state: 'loading' }
```
dai-shi, line 26: I'm guessing we don't need this if we have refAtom.

src/utils/loadable.ts
```diff
@@ -0,0 +1,67 @@
...
+export function loadable<Value>(anAtom: Atom<Value>): Atom<Loadable<Value>> {
+  const deps: object[] = [anAtom]
+  const cachedAtom = getWeakCacheItem(loadableAtomCache, deps)
+  if (cachedAtom) {
+    return cachedAtom as Atom<Loadable<Value>>
+  }
+
+  const refAtom = atom<{ error: LoadableError | null }>({
+    error: null,
+  })
```
dai-shi, line 37: refAtom must be a read-only atom with a function for lazy initialization, otherwise the state could be shared among different providers.
I guess we can just store the entire `Loadable<Value>` type in the ref.
```suggestion
  const loadableRefAtom = atom((): [Loadable<Value>] => ([{ state: 'loading' }]))
```

Hmm, the tuple might be not very readable. In such a case, we can make it an object.

dai-shi: Let me try writing the new derivedAtom with this ref.

```js
  const derivedAtom = atom((get): Loadable<Value> => {
    const loadableRef = get(loadableRefAtom)
    try {
      const value = get(anAtom)
      loadableRef[0] = { state: 'hasData', data: value }
    } catch (error) {
      if (error instanceof Promise) {
        if (loadableRef[0].state !== 'loading') {
          loadableRef[0] = { state: 'loading' }
        }
      } else {
        if (loadableRef[0].state !== 'hasError' || !Object.is(loadableRef[0].error, error)) {
          loadableRef[0] = { state: 'hasError', error }
        }
      }
    }
    return loadableRef[0]
  })
```

Now, I'm not sure if this is more readable, maybe your version is better.

---
Now, I have another idea. We can create a WeakMap at module level.
The key of the WeakMap is `error` object and the value is Loadable<Value> (= `{ state: 'hasError', error }`)
This allows us to eliminate the hacky refAtom.
Yeah, I think is far cleaner. What do you think?

dai-shi (commented):
My final recommendation is to use WeakMap instead of refAtom, but let me shamelessly leave my intermediate comments.

src/utils/loadable.ts
Pinpickle: Will do

src/utils/loadable.ts
Pinpickle: I like it!

Pinpickle:
@dai-shi I've addressed your comments, it's a lot cleaner now.

Let me know what you think.

I also looked into #601 and it is a core bug, opened a PR at #705

dai-shi (commented):
Looks very nice!
Added a small suggestion to refactor.

Pinpickle:
@dai-shi done! I did the same for the loadable cache as well, it's only one key

tests/utils/loadable.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+      <LoadableComponent asyncAtom={asyncAtom} />
+    </Provider>
+  )
+
+  getByText('Loading...')
+  rejectAsync(new Error('An error occurred'))
+  await findByText('Error: An error occurred')
+})
+
+it('loadable turns primitive throws into values', async () => {
```
Pinpickle, line 42: As we're using a `WeakMap`, I added this test case. If a primitive value ever became a key for `error`, it wouldn't work.

dai-shi:
> I did the same for the loadable cache as well, it's only one key

Interestingly, I didn't notice it. Thanks. (There might be other utils we could do the same.)

src/utils/loadable.ts
```diff
@@ -0,0 +1,52 @@
+import { atom } from 'jotai'
+import type { Atom } from 'jotai'
+
+const loadableAtomCache = new WeakMap<Atom<unknown>, Atom<Loadable<any>>>()
```
dai-shi, line 4: Can `any` be `unknown`? Or, TS complains?

Pinpickle: It can be unknown, it'll just require a cast further down. I'll update

dai-shi (approved):
Looks good! Thanks so much.

dai-shi:
Would you also work on docs? For now, docs are in a separate repo for now.
https://github.com/pmndrs/website/tree/docs/docs/jotai/api/utils.mdx
(and, we should probably split utils page in to some pages.

Pinpickle:
@dai-shi yep I can have a look at that, will probably be next week

TwistedMinda:
@dai-shi Can't find any docs on loadable! It's available right?
I guess it's time to use the "Async" section of the docs to replace the good old "No Suspense".

Async section would look like:

- kick-off: atomWithPending (async then normal, as many times as you want)
  - quick example with suspense
  - quick example with loadable, mention that it can be used the same way for any async atom
  - quick mention (+ link) to the atomWithAsyncStorage that uses the same concept

- full async methods
   - async get example (use word "async selector" somehow so reader can relate)
   - async write example (use word "action" somehow so reader can relate)
   - async get-write example (i can't even relate to this)

- common practices
  - suspense-everywhere
  - preloading (mention it fits well with atomWithAsyncStorage)

What u think?

PR #687, opened 2021-09-05
refactor(core): re-implement useAtom with useReducer

dai-shi:
uMS won't land, #686 doesn't work so far.
I feel like simpler solution with useReducer would work better.
This simplifies a lot. Also, fixes weird issues in devtools.

PR #700, opened 2021-09-10
refactor(core,quey): batch flush pending and useState based useAtom

dai-shi:
While working on #687, I thought there should be an easier impl with useState.
This seems to be working. A few tests have to be tweaked.
Still, it reads a mutable store in render. 🤔


dai-shi:
Okay, I don't think this is worth it. We would still need to migrate to uSES eventually.
I will file a smaller PR to just fix some issues found in this PR.

Issue #702, opened 2021-09-11
Suspense issue with atomWithQuery refetch


dazweeja:
While trying to get atomWithQuery refetch working with Suspense, I noticed that even in the Basic Demo for atomWithQuery, refetch does not work with Suspense for any version of jotai later than v1.1.1. By that I mean that Suspense never sees that the Promise has resolved after a refetch. It may be related to https://github.com/pmndrs/jotai/pull/573

To test, update the dependencies in the atomWithQuery Basic Demo (https://codesandbox.io/s/jotai-query-demo-ij2sd) to v3.23.2 for react-query and v1.1.2 or later for jotai, and click the refetch button, The Suspense fallback will be shown indefinitely.

dai-shi:
https://codesandbox.io/s/jotai-query-demo-forked-t1bqi?file=/src/App.tsx
Issue reproduced. cc @aulneau 

I hoped #701 fixes it, but unfortunately not.

aulneau:
Oh very strange, I'll dig in and get this fixed!

dai-shi:
Is #705 somehow related by chance?

dai-shi:
@aulneau Does #747 change anything?

PR #703, opened 2021-09-11
fix: refetch

aulneau:
WIP to fix https://github.com/pmndrs/jotai/issues/702

### DIFF
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -163,10 +163,17 @@ export function atomWithQuery<
       switch (action.type) {
         case 'refetch': {
           const { dataAtom, observer } = get(queryDataAtom)
-          set(dataAtom, new Promise<TData>(() => {})) // infinite pending
+          if (!observer?.options?.keepPreviousData) {
+            set(dataAtom, new Promise<TData>(() => {})) // infinite pending
+          }
           const p = Promise.resolve()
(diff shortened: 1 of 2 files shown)


src/query/atomWithQuery.ts
```diff
@@ -162,8 +162,7 @@ export function atomWithQuery<
     (get, set, action: AtomWithQueryAction) => {
       switch (action.type) {
         case 'refetch': {
-          const { dataAtom, observer } = get(queryDataAtom)
-          set(dataAtom, new Promise<TData>(() => {})) // infinite pending
```
aulneau, line 166: @dai-shi I wish I knew why this fixes it, but it does. would be curious if you had any ideas

aulneau: I see this change now makes it impossible for atoms that reference a query atom to know that it's suspended, hence the failing test

dai-shi: > I wish I knew why this fixes it

Me too.

Hmm, no, I don't have any ideas instantly. I'd need to dig into it.

dai-shi: Your first approach would work with resolving the promise instead of infinite pending promise.
However, we should probably first need to reproduce the issue behavior in tests, otherwise we can't make sure to work in the future.

PR #705, opened 2021-09-12
fix(core): Allow retries for errored async reads

### LINKED ISSUE #601: atomWithObservable can't recover from error (and also with any atoms?)

Pinpickle:
When changing a dependency for an atom in an error state, the derived atom's read function is re-run.

We'd expect that if that re-run is successful, `useAtom` will return the new value.

This is currently the case with sync atoms, but not async atoms. This PR fixes that, and adds tests (the async variant was failing, but the sync one wasn't).

Fixes #601

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -177,6 +177,7 @@ export const createStore = (
       return
     }
     atomState.c?.() // cancel read promise
+    delete atomState.e // read error
     if (isInterruptablePromise(promise)) {
       atomState.p = promise // read promise
       delete atomState.c // this promise is from another atom state, shouldn't be canceled here
--- a/tests/error.test.tsx
+++ b/tests/error.test.tsx
@@ -492,7 +492,7 @@ describe('error recovery', () => {
     await findByText('Value: 1')
   })
 
-  it.only('recovers from async errors', async () => {
+  it('recovers from async errors', async () => {
     const asyncAtom = atom(async (get) => {
       const value = get(counter)
       await new Promise((resolve) => {
--- a/tests/error.test.tsx
+++ b/tests/error.test.tsx
@@ -26,7 +26,12 @@ class ErrorBoundary extends Component<
   }
   render() {
     return this.state.hasError ? (
-      <div>{this.props.message || 'errored'}</div>
+      <div>
+        {this.props.message || 'errored'}
+        <button onClick={() => this.setState({ hasError: false })}>
+          retry
+        </button>
+      </div>
     ) : (
       this.props.children
     )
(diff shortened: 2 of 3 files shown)


tests/utils/atomWithObservable.test.tsx
```diff
@@ -95,32 +70,3 @@ it('writable count state', async () => {
   fireEvent.click(getByText('button'))
   await findByText('count: 9')
 })
-
-// FIXME we would like to support retry
-it.skip('count state with error', async () => {
```
Pinpickle, line 100: This test wasn't valid - an observable can't recover from an error state. For the notion of "retrying" observables to work, the error would never reach `atomWithObservable`, so the retrying logic is out of scope for jotai.

src/core/store.ts
```diff
@@ -177,6 +177,7 @@ export const createStore = (
       return
     }
     atomState.c?.() // cancel read promise
+    delete atomState.e // read error
```
Pinpickle, line 180: This was being done when writing to a sync atom, but not an async atom.

dai-shi: Wow! Nice catch!
Feels like you have fairly deeper understanding of core.

tests/error.test.tsx
```diff
@@ -447,3 +452,79 @@ describe('throws an error while updating in effect cleanup', () => {
...
+    )
+
+    await findByText('errored')
+
+    fireEvent.click(getByText('increment'))
+    fireEvent.click(getByText('retry'))
+    await findByText('Value: 1')
+  })
+
+  it.only('recovers from async errors', async () => {
```
dai-shi, line 495:
```suggestion
  it('recovers from async errors', async () => {
```

tests/utils/atomWithObservable.test.tsx
dai-shi: So, I didn't know about the observable error behavior, and created this test. But, it's an expected behavior.
Now, you find the actual issue of error async reads, and fix it. Very very nice.

dai-shi (approved):
Amazing. You did great work here.

tests/error.test.tsx
```diff
@@ -492,7 +492,7 @@ describe('error recovery', () => {
     await findByText('Value: 1')
   })
 
-  it.only('recovers from async errors', async () => {
```
Pinpickle: Oops - thank you!

Pinpickle:
Thanks @dai-shi 🎉 

Hopefully we can get this merged before #698, which means I can bring back the failing test

dai-shi:
> Hopefully we can get this merged before #698, which means I can bring back the failing test

Yes, that's my plan too.

Seems like a test fails. Is this because our tests are not stable enough because we don't properly use fake timers?

Pinpickle:
> Seems like a test fails. Is this because our tests are not stable enough because we don't properly use fake timers?

This was because of less mode and which meant the atom state was shared between tests

I've fixed it now - could you run the workflow again?

dai-shi:
> This was because of less mode and meant the atom state was shared between tests

I see.

PR #710, opened 2021-09-15
refactor: with memoizeAtom

dai-shi:
putting typical pattern in a function. better typing.

### DIFF
--- a/src/utils/weakCache.ts
+++ b/src/utils/weakCache.ts
@@ -1,6 +1,8 @@
+import type { Atom } from 'jotai'
+
 export type WeakCache<T> = WeakMap<object, [WeakCache<T>] | [WeakCache<T>, T]>
 
-export const getWeakCacheItem = <T>(
+const getWeakCacheItem = <T>(
   cache: WeakCache<T>,
   deps: readonly object[]
 ): T | undefined => {
@@ -18,7 +20,7 @@ export const getWeakCacheItem = <T>(
   }
 }
 
-export const setWeakCacheItem = <T>(
+const setWeakCacheItem = <T>(
   cache: WeakCache<T>,
   deps: readonly object[],
   item: T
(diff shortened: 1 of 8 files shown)


src/utils/weakCache.ts
```diff
@@ -38,3 +40,20 @@ export const setWeakCacheItem = <T>(
     deps = rest
   }
 }
+
+export const createMemoizeAtom = () => {
```
Thisen, line 44: Nitpick: I feel like `cacheAtom` is better naming, but that's obviously subjective.
 
```suggestion
export const createCacheAtomFn = () => {
```

src/utils/weakCache.ts
```diff
@@ -38,3 +40,20 @@ export const setWeakCacheItem = <T>(
...
+  const cache: WeakCache<Atom<unknown>> = new WeakMap()
+  const memoizeAtom = <AtomType extends Atom<unknown>, Deps extends object[]>(
```
Thisen, line 46:
```suggestion
  const cacheAtom = <AtomType extends Atom<unknown>, Deps extends readonly object[]>(
```

src/utils/weakCache.ts
```diff
@@ -38,3 +40,20 @@ export const setWeakCacheItem = <T>(
...
+  ) => {
+    const cachedAtom = getWeakCacheItem(cache, deps)
+    if (cachedAtom) {
+      return cachedAtom as AtomType
+    }
+    const createdAtom = createAtom()
+    setWeakCacheItem(cache, deps, createdAtom)
+    return createdAtom
+  }
+  return memoizeAtom
```
Thisen, line 58:
```suggestion
  return cacheAtom
```

Thisen (approved):

dai-shi:
Thanks. 

I don't have much preference on naming the function `memoizeAtom()` or `cacheAtom()`. Let's hear what others think. @aulneau @Aslemammad or anyone seeing this pr.

Aslemammad:
It's remarkably subjective, but here, I think it's not just a simple caching; it's more about memoizing. Am I wrong? so I'd like to go with `memoizeAtom`

aulneau:
I agree it feels more like memoization than caching to me. The distinction is I just want one instance of this atom, where if I wanted something cached, I'd assume the atom value would be involved. But again, very subject as others have said.

dai-shi:
Thanks everyone. I will keep the current naming. It feels less confusing to me.

PR #713, opened 2021-09-16
breaking(types): refine atom type and other types

dai-shi:
There are quite a few issues around types. This is to address fix typing issues from the ground.

## What were the pains

Originally, I thought async would be hidden thanks to Suspense, so normal atoms and async atoms are intentionally not distinguishable.
However, that becomes a pain.
a) We can't know if an atom is an async-read atom, so we can't give api with fine grained types. It can't support complex union type either.
b) Likewise, we can't know if an atom is an async-write atom, so `set` returns `void` or `Promise<void>`. This was known issue and hidden before v1, but #689 reveals it again
c) We sort of support `atom(Promise.resolve('foo'))` usage, but types are not fully compatible with it.

## What are changing

We don't hide promises behind types. If an atom returns a number,  it's `Atom<number>`. If an atom returns a number asynchronously, it's `Atom<Promise<number>>` (previously it was `Atom<number>`). A writable atom is now `WritableAtom<Value, Update, Return>` and `Return` is either `void` or `Promise<void>` for now. Thus, we can distinguish async writable atoms.

The entire types are updated to follow the core change.
There is another change which I wanted to fix: atomFamily generics are simplified. 

Basically, if people use type inference, the change is trivial. For explicit typing, people need to modify types.

## Release Plan

This is a huge breaking change in **types**. But, we plan a minor version up.
JS code is basically **unchanged**, except for some small fixes to aligning with types.
For this release, we should be more careful to get feedback in advance as much as possible.
We can consider alpha release too. (with/without publish tag.)
We should prepare some migration guide. Not sure about codemod.

## Getting feedback

Everyone is encouraged to try the codesandbox build in this PR.
Feel free to report here how it goes with your projects.
If the breaking change is too much pain rather than benefit, we could give this up.

### DIFF
--- a/tests/utils/waitForAll.test.tsx
+++ b/tests/utils/waitForAll.test.tsx
@@ -311,7 +311,7 @@ it('large atom count', async () => {
     atom((getter) => {
       const data = createArray(count)
       const atoms = data.map(chunksFamily)
-      const values = waitForAll(atoms)
+      const values = waitForAll(atoms as any /* FIXME better typing? */)
       return getter(values)
     })
   )
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -1,15 +1,15 @@
 type ResolveType<T> = T extends Promise<infer V> ? V : T
 
 type Getter = {
-  <Value>(atom: Atom<Promise<Value>> & WithInitialValue<Promise<Value>>):
-    | Promise<Value>
-    | Value
   <Value>(atom: Atom<Value | Promise<Value>>): Value
   <Value>(atom: Atom<Promise<Value>>): Value
   <Value>(atom: Atom<Value>): ResolveType<Value>
 }
 
 type WriteGetter = Getter & {
+  <Value>(atom: Atom<Value | Promise<Value>>, unstable_promise: true):
+    | Promise<Value>
+    | Value
   <Value>(atom: Atom<Promise<Value>>, unstable_promise: true):
     | Promise<Value>
     | Value
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -1,34 +1,37 @@
+type ResolveType<T> = T extends Promise<infer V> ? V : T
+
 type Getter = {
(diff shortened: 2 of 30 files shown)


Aslemammad:
Hmm, it works well here: https://codesandbox.io/s/react-typescript-forked-gogyp?file=/src/App.tsx

tests/utils/waitForAll.test.tsx
```diff
@@ -311,7 +311,7 @@ it('large atom count', async () => {
     atom((getter) => {
       const data = createArray(count)
       const atoms = data.map(chunksFamily)
-      const values = waitForAll(atoms)
+      const values = waitForAll(atoms as any /* FIXME better typing? */)
```
Aslemammad, line 314: I don't think numbers go well (`i` & `count`) with Promises in `waitForAll`!
So I think it's better to delete the FIXME comment!

Aslemammad: Or changing the test behaviour to async, which broke the test for me (basically)!

Aslemammad (commented):
LGTM!

tests/utils/waitForAll.test.tsx
dai-shi: Thanks for the hint.
72333c3 looks good?

Aslemammad: Now I think it's great!

tell-y:
My one project has no problem with the below settings.

.eslintrc.json
```
    "@typescript-eslint/no-floating-promises": ["error", { "ignoreIIFE": true }],
```

package.json
```
    "@typescript-eslint/eslint-plugin": "^4.31.2",
    "@typescript-eslint/parser": "^4.31.2",
    "eslint": "^7.32.0",
    "typescript": "^4.4.3"
```

dai-shi:
Made writable atom return type defaults to void. This should mitigate the migration.

tests/utils/waitForAll.test.tsx
dai-shi: That was my bad: fixed in b587a6d and c9bf5a3

PR #716, opened 2021-09-17
Switch back to Node 12

Thisen:
Since Vercel is default on Node 12, lets use that for now to avoid errors on default deployments.


dai-shi (commented):
Thanks! I will cut a release soon with other approved PRs.

dai-shi (approved):

PR #717, opened 2021-09-17
refactor(core,utils,devtools): add dev methods in store

dai-shi:
This simplifies scope container a lot. Also, this fixes memory leak issue in DEV.

### DIFF
--- a/src/core/contexts.ts
+++ b/src/core/contexts.ts
@@ -2,65 +2,18 @@ import { createContext } from 'react'
 import type { Context } from 'react'
 import type { Atom, Scope } from './atom'
 import { createStore } from './store'
+import type { Store } from './store'
 
-const createScopeContainerForProduction = (
+export type ScopeContainer = {
+  s: Store
+}
+
+export const createScopeContainer = (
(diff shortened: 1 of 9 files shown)


src/core/contexts.ts
```diff
@@ -2,66 +2,19 @@ import { createContext } from 'react'
...
+import type { Store } from './store'
 
-const createScopeContainerForProduction = (
-  initialValues?: Iterable<readonly [Atom<unknown>, unknown]>
-) => {
-  const store = createStore(initialValues)
-  return [store] as const
+export type ScopeContainer = {
+  s: Store
 }
```
dai-shi, line 9: Keeping this a wrapper object, so that we can add some more properties in the future.

dai-shi (commented):
self-reviewed. oh, i'm so happy to fix the state listener behavior. merging this.

dai-shi:
https://github.com/pmndrs/jotai/commit/957ad82fcdfa142adacaa97c735725fa55c4de94 forgot to update size snapshot. 🤦‍♂️ (yeah, let's use a github action for it.)

Issue #709, opened 2021-09-14
Upgrading from 1.3.3 to 1.3.4 gives worrying build-time warnings


tomhicks:
I'm using jotai in a Next.js project, with nothing special in particular with the setup.

When I upgrade to 1.3.4 I get this in the build output, which seems reasonably important. My guess is that it's to do with the way the exports were changed to use wildcards?

Reverting back to 1.3.3 removes the warning.

```
warn  - ./node_modules/jotai/esm/index.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

./node_modules/jotai/esm/utils.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

./node_modules/jotai/esm/utils.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

dai-shi:
Thanks for reporting. It sounds like so... Exports is hard...
Would you help me reproduce this? Any steps to follow?

tomhicks:
hang on let me try to get a minimal repro

tomhicks:
```
yarn create-next-app
yarn add jotai
```

```typescript
import {atom, useAtom} from "jotai"
const numberAtom = atom(5)

export default function Home() {
  const [value, setValue] = useAtom(numberAtom)

  React.useEffect(() => {
    setInterval(() => {
      setValue((old) => old + 1)
    }, 1000)
  }, [setValue])

  return (
    <div className={styles.container}>
      <Head>
        <title>Create Next App</title>
        ... rest of the Home component here
```

`npm run dev`

```
warn  - ./node_modules/jotai/esm/index.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

tomhicks:
^^ Essentially... use jotai 😂

dai-shi:
Thanks for the steps. It's reproducible.
It has nothing to do with wildcards.

It seems the following code in `jotai/esm/index.js` causes the message.
```js
typeof require !== "undefined" ? require : (x) => {
  throw new Error('Dynamic require of "' + x + '" is not supported');
};
```

I don't think it's important. You can ignore for now.

I guess some of build tool updates caused this. 🤔

dai-shi:
This seems to fix the issue.
```diff
diff --git a/package.json b/package.json
index d8b6b96..3d12b2e 100644
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "benny": "^3.6.15",
[...]

PR #719, opened 2021-09-18
chore(build): pin esbuild for now and replace size plugin with github action

aulneau:
This fixes an issue with the esbuild/esm config, where when using jotai in next.js, this is given in the console:
```
info  - ready on http://localhost:3000
event - build page: /
wait  - compiling...
warn  - ./node_modules/jotai/esm/index.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

./node_modules/jotai/esm/query.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

./node_modules/jotai/esm/utils.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

./node_modules/jotai/esm/utils.js
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

I determined it was due to the target being set to a node version, and the platform defaulting to `browser`. esbuild injects a helper for `require` `typeof require !== "undefined" ? require : (x) => { throw new Error('Dynamic require of "' + x + '" is not supported'); };` that causes this nasty warning in webpack.

I've changed the target from `node12` to `esnext`. I've tested it both in a vite react-ts project, and a new next.js project, everything seems to work fine. Not sure what we should do to test it out further.

Additionally, I've pinned all our dependencies to their exact versions, rather than `^x`, as this makes it so there is less of a chance of having different versions installed between each other, it's also more secure. 

NB: @dai-shi -> with the change to `esnext`, the size snapshot plugin no longer works with the esm build. I'd recommend we move over to a github action i've used before, [`compressed-size-action`](https://github.com/preactjs/compressed-size-action) which gives a comment like this for each PR:

Screen Shot 2021-09-18 at 10 47 01 AM


dai-shi:
> This fixes an issue with the esbuild/esm config, where when using jotai in next.js, this is given in the console:

Oh, this is nice. Are you aware of #709? It think the warning is a false positive.

> I determined it was due to the target being set to a node version, and the platform defaulting to `browser`. esbuild injects a helper for `require` `typeof require !== "undefined" ? require : (x) => { throw new Error('Dynamic require of "' + x + '" is not supported'); };` that causes this nasty warning in webpack.

Yeah, that helper isn't help anything. It's just an unused expression.

> I've changed the target from `node12` to `esnext`. I've tested it both in a vite react-ts project, and a new next.js project, everything seems to work fine. Not sure what we should do to test it out further.

I still think this is the issue with esbuild. Changing to `esnext` may cause issues with older node versions. As you reported and fixed in #716, some envs still use node12. So, I guess it's safe to stay on node12 target.

If this warning is too annoying, I would suggest we go back to node v0.12.25 for the moment, and file an issue in esbuild to investigate.

> Additionally, I've pinned all our dependencies to their exact versions, rather than `^x`, as this makes it so there is less of a chance of having different versions installed between each other, it's also more secure.

This doesn't help much, because yarn.lock does it and I generally upgrade most of dependencies. I would suggest to remove `^` only for packages we intentionally pin, like esbuild in this case.

> NB: @dai-shi -> with the change to `esnext`, the size snapshot plugin no longer works with the esm build. I'd recommend we move over to a github action i've used before, [`compressed-size-action`](https://github.com/preactjs/compressed-size-action) which gives a comment like this for each PR:

This is a great idea. Actually, we looked for such a github action, but I didn't find any. If that one works, well, we can mover over.

aulneau:
> > This fixes an issue with the esbuild/esm config, where when using jotai in next.js, this is given in the console:
> 
> Oh, this is nice. Are you aware of #709? It think the warning is a false positive.
> Yeah, that helper isn't help anything. It's just an unused expression.

Yeah it does seem like a false positive because it's not really used. It's likely a bug in esbuild. I can make an issue over there and see what they think. Do you want to not modify jotai to remove it?

> > I've changed the target from `node12` to `esnext`. I've tested it both in a vite react-ts project, and a new next.js project, everything seems to work fine. Not sure what we should do to test it out further.
> 
> I still think this is the issue with esbuild. Changing to `esnext` may cause issues with older node versions. As you reported and fixed in #716, some envs still use node12. So, I guess it's safe to stay on node12 target.

Yes -- i wonder though what features we use that won't be supported in node12. Here is the diff between `node12` and `esnext` https://github.com/pmndrs/jotai/commit/a96f7515cf9ac41b94fdfc5db541b84d0c333a40

> If this warning is too annoying, I would suggest we go back to node v0.12.25 for the moment, and file an issue in esbuild to investigate.

Yeah! if that works it's likely a better choice.


> This doesn't help much, because yarn.lock does it and I generally upgrade most of dependencies. I would suggest to remove `^` only for packages we intentionally pin, like esbuild in this case.

Sure. It is possible that when someone runs yarn on a fresh pull without node_modules already present, the use of `^` can allow the version that is ultimately installed to change, without it be explicit in package.json. Since they are all dev deps, it likely does not matter much. It's much more important for direct dependencies.


> > NB: @dai-shi -> with the change to `esnext`, the size snapshot plugin no longer works with the esm build. I'd recommend we move over to a github action i've used before, [`compressed-size-action`](https://github.com/preactjs/compressed-size-action) which gives a comment like this for each PR:
> 
> This is a great idea. Actually, we looked for such a github action, but I didn't find any. If that one works, well, we can mover over.

Yeah, I really love it and have had great success with it. Want me to add the github action?

aulneau:
@dai-shi okay much more simple -> https://github.com/pmndrs/jotai/pull/719/files

- removed the size plugin
- added the compressed-size-action
- removed the duplicate mjs bundle step (now uses the multi-output feature)
- pinned esbuild which removes the warning

dai-shi:
> Yeah it does seem like a false positive because it's not really used. It's likely a bug in esbuild. I can make an issue over there and see what they think.

Yes, please. (Then, we know when we can unpin esbuild.)

> Yes -- i wonder though what features we use that won't be supported in node12. Here is the diff between `node12` and `esnext` [a96f751](https://github.com/pmndrs/jotai/commit/a96f7515cf9ac41b94fdfc5db541b84d0c333a40)

For example, optional chaining is not supported in node12.

> Yeah! if that works it's likely a better choice.

Okay, please make it sure we unpin if the issue is gone.

> Sure. It is possible that when someone runs yarn on a fresh pull without node_modules already present, the use of `^` can allow the version that is ultimately installed to change, without it be explicit in package.json. Since they are all dev deps, it likely does not matter much. It's much more important for direct dependencies.

AFAIU, yarn1 installs locked versions with yarn.lock.

> Yeah, I really love it and have had great success with it. Want me to add the github action?

Yes.

---

>     * removed the size plugin
> 
>     * added the compressed-size-action
> 
>     * removed the duplicate mjs bundle step (now uses the multi-output feature)
> 
>     * pinned esbuild which removes the warning

Looks great! How can we activate the github workflow in this PR?

.github/compressed-size-action.yml
```diff
@@ -0,0 +1,14 @@
+name: Compressed Size
+
+on: [ pull_request ]
```
dai-shi, line 3: maybe, we need to create a new pull request?

rollup.config.js
```diff
@@ -26,6 +25,8 @@ function getEsbuild(target) {
   return esbuild({
     minify: false,
     target,
+    external,
+    platform: 'neutral',
```
dai-shi, line 29: How do these change? I mean, why it was working okay before the change (expecially `external` one, not sure what `platform` does.)

rollup.config.js
```diff
@@ -50,9 +51,12 @@ function createDeclarationConfig(input, output) {
 function createESMConfig(input, output) {
   return {
     input,
-    output: { file: output, format: 'esm' },
+    output: [
+      { file: `${output}.js`, format: 'esm' },
+      { file: `${output}.mjs`, format: 'esm' },
```
dai-shi, line 56: wow, we don't even need rollup copy plugin.

dai-shi:
@barelyhuman You want to review this too. Related: https://github.com/pmndrs/zustand/pull/546

@aulneau I'd like to have the same change in zustand and valtio too.

rollup.config.js
barelyhuman: we can still use it for the type definition copying if we want to get rid of the scripts in `package.json` https://github.com/pmndrs/zustand/pull/546

barelyhuman:
2 cents

1. the `platform:neutral` makes sense since the target can change based on what's being used an SSR or a Web Framework 
2. Multi Output adds a little more abstraction but that's okay, not a big deal 
3. I like the compressed action thing, We might want to check `brotli` size for each as well , if the action supports that

.github/compressed-size-action.yml
barelyhuman: Yeah, only pulls would show this data, 
though don't you think a file on the repo works better since we know the state of every commit with updated `size.json` cause PR's will get hard to search after a point

rollup.config.js
barelyhuman: I think it's a little more explicitly defined but this is already handled by `tsconfig` ? 
platform `neutral` make sure the esm is not specifically targeted towards browser or node

barelyhuman (approved):

.github/compressed-size-action.yml
dai-shi: It has pros and cons, but so far this has been more pain with multiple contributors. I wanted to switch to github actions.

rollup.config.js
dai-shi: ah, that's right..

.github/compressed-size-action.yml
barelyhuman: Make sense, we can kinda have github actions create a size file and push to the repo as well, I have something similar working for https://github.com/barelyhuman/custom-rom-index , so if we want to look into that angle, let me know

dai-shi:
Testing on main branch:

Adding `external`: No difference
Adding `platform: 'neutral'`: No difference

dai-shi:
> We might want to check brotli size for each as well , if the action supports that

@aulneau ?

aulneau:
> > We might want to check brotli size for each as well , if the action supports that
> 
> @aulneau ?

It does not support it

> Testing on main branch:
>
> Adding external: No difference
> Adding platform: 'neutral': No difference

Yeah, I think we should keep the `platform` change as maybe something in the future would use it, but the `externals` can go

dai-shi:
> but the `externals` can go

does it accept a function?
Any other docs than https://esbuild.github.io/api/#external ?

@aulneau would you be able to open PRs in zustand and valtio?

dai-shi (approved):

Issue #728, opened 2021-09-22
splitAtom(arrayAtom) returns inconsistent length temporarily with multiple renderers since v1.3.3


sxy8469:
const itemsAtom = atom([])
const itemAtomsAtom = splitAtom(itemsAtom)

const items= useAtomValue(itemsAtom );
const itemAtoms= useAtomValue(itemAtomsAtom);
**items.length ≠ itemAtoms.length ???**

dai-shi:
Thanks for reporting. This seems unexpected.
Would you be able to create a minimal reproduction in codesandbox?

It works in v1.3.2, but not v1.3.3, right?
Either #678 or #689 might cause an issue. 🤔

sxy8469:
> Thanks for reporting. This seems unexpected.
> Would you be able to create a minimal reproduction in codesandbox?
> 
> It works in v1.3.2, but not v1.3.3, right?
> Either #678 or #689 might cause an issue. 🤔

https://codesandbox.io/s/react-typescript-forked-gi301?file=/src/App.tsx

dai-shi:
> It works in v1.3.2, but not v1.3.3, right?
> Either #678 or #689 might cause an issue. 

Either #687 or #689, and it's #687 causing this.

dai-shi:
Hm, `removePoint` removes the original array. A bit of surprising it's working pre v1.3.3.
Let me see if I can repro in tests.

dai-shi:
Okay, I found the issue is not reproducible if we don't use react-konva.

sxy8469:
> Okay, I found the issue is not reproducible if we don't use react-konva.

Yes, but the "react-konva" is required.
I still hope this problem can be solved in the next version

dai-shi:
But, we don't know what is the problem, thus we can't investigate a possible solution. We would like to fix it for sure. Do you happen to know how react-konva works? Or, can you try to reproduce the issue without react-konva?
Meanwhile, this is a workaround: https://codesandbox.io/s/react-typescript-forked-6dqs8

sxy8469:
> But, we don't know what is the problem, thus we can't investigate a possible solution. We would like to fix it for sure. Do you happen to know how react-konva works? Or, can you try to reproduce the issue without react-konva?
> Meanwhile, this is a workaround: https://codesandbox.io/s/react-typescript-forked-6dqs8

Thanks,  I temporarily reverted to 1.3.2 version   : )

sxy8469:
[...]

PR #729, opened 2021-09-23
add test for #728

dai-shi:
for #728


dai-shi:
It turns out this is not what is reported in #728.

PR #730, opened 2021-09-23
fix(core,utils): support unknown error type

dai-shi:
I think the library shouldn't modify thrown errors.
Unfortunately, the implementation of `loadable` gets hacky, but it passes the test (it's refactored, but specs are not changed. cc @Pinpickle )

### DIFF
--- a/src/utils/loadable.ts
+++ b/src/utils/loadable.ts
@@ -15,7 +15,15 @@ const LOADING_LOADABLE: Loadable<never> = { state: 'loading' }
 
 export function loadable<Value>(anAtom: Atom<Value>): Atom<Loadable<Value>> {
   return memoizeAtom(() => {
+    const refAtom = atom(
+      () =>
+        ({} as {
+          err: unknown
+          obj: object
+        })
+    )
     const derivedAtom = atom((get): Loadable<Value> => {
+      const ref = get(refAtom)
       try {
         const value = get(anAtom)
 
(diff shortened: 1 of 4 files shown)


src/utils/loadable.ts
```diff
@@ -28,7 +36,17 @@ export function loadable<Value>(anAtom: Atom<Value>): Atom<Loadable<Value>> {
...
         }
 
-        const cachedErrorLoadable = errorLoadableCache.get(error as Error)
+        let errorObject: object
+        if (typeof error === 'object') {
+          errorObject = error as object
+        } else if (ref.err === error) {
+          errorObject = ref.obj
+        } else {
+          errorObject = new Error(error as string)
```
Pinpickle, line 45: This seems a bit unintuitive that everywhere else Jotai keeps primitive values in tact, but here it converts them into error objects.

This also won't get cached because it's making a new error on every call.

Maybe this should be replaced with a refAtom after all? Although after the fix in #705, the caching may not be strictly necessary anymore?

Pinpickle:
@dai-shi I generally agree with this as a move! I'm not 100% convinced on how errors are still transformed for loadables (I left a comment).

This is an edge-case of an edge-case though, so it isn't a major issue.

src/utils/loadable.ts
dai-shi: > Maybe this should be replaced with a refAtom after all?

This is true. I just wanted to keep the nice WeakMap approach, but didn't make sense...
Okay, let me refactor it again. btw, `refAtom` is a real hack and may not work in concurrent rendering. We need to revisit it. There are about three functions using it.

> Although after the fix in #705, the caching may not be strictly necessary anymore?

I don't think this has something to do with #705.

PR #732, opened 2021-09-24
chore(build): upgrade and unpin esbuild and fix compressed size action

dai-shi:
#719 pinned esbuild version for #709, but as esbuild v0.12.29 solves it, let's upgrade it to the latest.


dai-shi:
@aulneau the compressed size action doesn't seem to work.

Oh! The yml is not in the workflows folder.

aulneau:
> @aulneau the compressed size action doesn't seem to work.
> 
> Oh! The yml file is not in the workflows folder.

Whoops!! Good catch

PR #734, opened 2021-09-26
refactor(utils): loadable util

dai-shi:
From: https://github.com/pmndrs/jotai/pull/730#discussion_r716109779

cc @Pinpickle


Pinpickle (approved):
LGTM

I agree that refAtom probably won't work in concurrent mode... I'll be interested to see what you come up with to get around that!

PR #740, opened 2021-09-28
refactor(tests): checking console error messages

dai-shi:
to avoid testing with other console error messages. there can be room for improvement.


Aslemammad (commented):
Nice new way! LGTM

PR #746, opened 2021-09-30
test: add test for #742

### LINKED ISSUE #742: Suspense doesn't resume on derived atom with an async dependency resolving to the same value
I'm facing an issue where Suspense doesn't resume when the async atom dependency of a derived atom resolves to the same value it had.

Here is a codesandbox highlighting the issue: https://codesandbox.io/s/kev5t?file=/src/App.tsx
By clicking on `fetch data`, Suspense will get stuck even though the underlying async atom has resolved.
The problem seems to occur when the atom resolves to the same value it had.

I'm not familiar enough with Jotai's code base to find the root cause, perhaps I'm declaring the derived atom incorrectly, but it seems to be inline with the documentation. (https://docs.pmnd.rs/jotai/basics/async)

citolen:
This adds a failing test to help resolve #742.

### DIFF
--- a/tests/async.test.tsx
+++ b/tests/async.test.tsx
@@ -75,6 +75,77 @@ it('does not show async stale result', async () => {
   expect(committed).toEqual([0, 2])
 })
 
+it('does not show async stale result on derived atom', async () => {
+  const committed: null[] = []
+  const countAtom = atom(0)
+  const asyncAlwaysNullAtom = atom(async (get) => {
+    get(countAtom)
+    await new Promise((r) => setTimeout(r, 10))
+    return null
+  })
+  const derivedAtom = atom((get) => get(asyncAlwaysNullAtom))
+
+  const DisplayDerivedValue = () => {
+    const [derivedValue] = useAtom(derivedAtom)
+    useEffect(() => {
+      committed.push(derivedValue)
+    })
+    return <div>derived value:{derivedValue}</div>
+  }
+
+  const DisplayAsyncValue = () => {
+    const [asyncValue] = useAtom(asyncAlwaysNullAtom)
+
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Looks clean!! Let me merge this into a branch.

tests/async.test.tsx
```diff
@@ -75,6 +75,77 @@ it('does not show async stale result', async () => {
...
+  fireEvent.click(getByText('button'))
+
+  await findByText('count: 1')
+  await findByText('loading async value')
+  await findByText('loading derived value')
+  await waitFor(() => {
+    expect(queryByText('loading async value')).toBeNull()
+    expect(queryByText('loading derived value')).toBeNull()
+  })
+  expect(committed).toEqual([null])
```
dai-shi, line 146: I don't think we can ensure this. DisplayDerivedValue is hidden and shown again, which triggers useEffect. Let me remove this. (if this is a wrong assumption, let's revisit it separately.)

citolen: Ok, I'm not sure either.
My assumption was that `DisplayDerivedValue` would not re-render because the derived atom resolves to the same value.
Once the issue is fixed I'm expecting Suspense would resume, removing the fallback and `display: none` from the content, but not necessarily re-rendering the component as nothing has changed.
Hence `expect(committed).toEqual([null])`.
You're right let's revisit this later once the initial issue is fixed.

dai-shi: AFAIK, in react 17, it's re-rendering and committing. in react 18, it might be more like re-mounting. I'm not sure if `display: none` and firing the effect is the same thing. You can play with it even without jotai and see how it behaves.

PR #747, opened 2021-09-30
fix(core): resume from suspense on derived atom with async atom dependency

### LINKED ISSUE #742: Suspense doesn't resume on derived atom with an async dependency resolving to the same value

dai-shi:
fix #742.
This is actually something I wasn't very confident with in the logic. Happy to have a fix 🎉 .

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -166,17 +166,22 @@ export const createStore = (
       return
     }
     atomState.c?.() // cancel read promise
-    delete atomState.e // read error
-    delete atomState.p // read promise
-    delete atomState.c // cancel read promise
-    delete atomState.i // invalidated revision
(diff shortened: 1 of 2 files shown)


dai-shi:
https://codesandbox.io/s/red-pond-lniin
The smaller repro is also fixed!

PR #704, opened 2021-09-12
chore(build): simplify exports and support nested subpath

dai-shi:
- simplifies exports in package.json
- supports nested subpath in rollup.config.js
- update node version from 12 to 14

PR #748, opened 2021-09-30
fix(package): avoid subpath pattern in exports

### LINKED ISSUE #743: Using jotai/utils in v1.3.6 on Next.js app (webpack 5) causes errors in build
`Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './utils' is not defined by "exports" in /Users/tomhicks/Developer/orbital/webapp/node_modules/jotai/package.json`

Sorry for the pretty vague report.

Here's the extent of my investigation:

## 1.3.3
👍 no build errors 👍 

## 1.3.4 onwards
```
warn  - ./node_modules/jotai/esm/index.js - Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

then later on

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './utils' is not defined by "exports"
```

This stops the app from rendering. I reported the first error a while back, and it didn't seem to prevent the app from rendering, so I'm a bit confused as to why it now won't render, even on 1.3.4, where before it seemed to render fine despite the warnings.

I don't think anything else in the build chain has changed (e.g. babel) so I might be misremembering the fact that it rendered on 1.3.4 before.

dai-shi:
close #743
partly reverting #704

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,11 +23,77 @@
       "import": "./esm/index.mjs",
       "default": "./index.js"
     },
-    "./*": {
-      "types": "./*.d.ts",
-      "module": "./esm/*.js",
-      "import": "./esm/*.mjs",
-      "default": "./*.js"
+    "./utils": {
+      "types": "./utils.d.ts",
+      "module": "./esm/utils.js",
+      "import": "./esm/utils.mjs",
+      "default": "./utils.js"
+    },
+    "./devtools": {
+      "types": "./devtools.d.ts",
+      "module": "./esm/devtools.js",
+      "import": "./esm/devtools.mjs",
+      "default": "./devtools.js"
+    },
+    "./immer": {
+      "types": "./immer.d.ts",
+      "module": "./esm/immer.js",
+      "import": "./esm/immer.mjs",
+      "default": "./immer.js"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -61,7 +127,7 @@
     "copy": "shx cp -r dist/src/* dist/esm && shx mv dist/src/* dist && shx rm -rf dist/{src,tests} && downlevel-dts dist dist/ts3.4 && shx cp package.json readme.md LICENSE dist && json -I -f dist/package.json -e \"this.private=false; this.devDependencies=undefined; this.optionalDependencies=undefined; this.scripts=undefined; this.prettier=undefined; this.jest=undefined; this['lint-staged']=undefined;\""
   },
   "engines": {
-    "node": ">=12"
+    "node": ">=12.7.0"
```
Thisen, line 130: Shouldn't it be v12.13.0?

dai-shi: subpath pattern requires v12.13.0, whereas just subpath requires v12.7.0 according to the docs.
As this is confimred to work in v12.8.0, it should be correct.

PR #749, opened 2021-09-30
refactor(tests): checking console error messages, following up #740

dai-shi:
following up #740

### DIFF
--- a/tests/error.test.tsx
+++ b/tests/error.test.tsx
@@ -236,7 +236,7 @@ it('can throw an error in write function', async () => {
   const errorAtom = atom(
     (get) => get(countAtom),
     () => {
-      throw new Error('err_in_write_function')
+      throw new Error('error_in_write_function')
     }
   )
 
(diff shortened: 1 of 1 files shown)


tests/error.test.tsx
```diff
@@ -265,18 +265,18 @@ it('can throw an error in write function', async () => {
   )
 
   await findByText('no error')
-  expect(errorMessages).not.toContain('Error: err_in_write_function')
+  expect(errorMessages).not.toContain('Error: error_in_write_function')
```
aulneau, line 268: more of a style thing, but having errors as `Enum` really helps with this

dai-shi: Maybe, maybe not. I don't have strong preference on tests compared to src, but copy-pasting in tests doesn't seem so bad if it's fairly readable. For me, using enum in tests seems less readable with unnecessary indirection, but it's subjective. Let me merge this as is, and if you want to improve it, please feel free.

PR #752, opened 2021-09-30
Add option for atomWithProxy to be synchronous

tarngerine:
I've found that when some of my state is in a jotai atom, and some of my state is in valtio, it can become out of sync and cause some errors

```
const selectedIds = atom<string[]>([])

const state = proxy({elements: {}})
const stateAtom = atomWithProxy(state);

const createElement = atom(null, (_, _, element) => {
  // set to jotai — synchronous and causes render immediately
  set(selectedIds, [element.id]);

  // set to proxy, not synchronous
  state.elements[element.id] = element;
});

function SelectedElements() {
  const selection = useAtomValue(selectedIds);
  const state = useAtomValue(stateAtom);
  return <>
    { selection.map(id => <Element element={state.elements[id]}>) } // this can be undefined
  </>;
}
```

so this PR just adds an option for atomWithProxy to be synchronous, modeled the Options after the one in valtio/useSnapshot.


tarngerine:
you may ask the question — why am i not just using valtio for everything? i would answer: idk, im in the middle of a migration :p

src/valtio/atomWithProxy.ts
```diff
@@ -32,20 +32,24 @@ const applyChanges = <T extends object>(proxyObject: T, prev: T, next: T) => {
...
 // setStatusState({ ...state, status: Promise.resolve('newStatus') })
-export function atomWithProxy<Value extends object>(proxyObject: Value) {
+export function atomWithProxy<Value extends object>(proxyObject: Value, options?: Options) {
   const baseAtom = atom(snapshot(proxyObject))
   baseAtom.onMount = (setValue) => {
     const callback = () => {
       setValue(snapshot(proxyObject))
     }
-    const unsub = subscribe(proxyObject, callback)
+    const unsub = subscribe(proxyObject, callback, options?.boolean)
```
dai-shi, line 52: You mean:
```suggestion
    const unsub = subscribe(proxyObject, callback, options?.sync)
```

dai-shi (commented):
Nice catch. I kind of knew this, but thought no use cases. It's good you find one.
If you are motivated, please feel free to add tests and docs (separate repo).

tarngerine:
Added docs PR: https://github.com/pmndrs/website/pull/190

do you have advice for how to set up a test here? should i extend one of the existing tests? haven't written one before :0 thanks!

dai-shi:
> do you have advice for how to set up a test here? should i extend one of the existing tests? haven't written one before :0 thanks!

roughly like this?

1. learn the basics of `@testing-library/react`
2. read `tests/valtio/atomWithProxy.test.tsx`
3. add a new test in the file, that fails without the change in this PR
4. confirm the test passes with the change

tarngerine:
i see, thanks! i gave it a shot and added a test. it fails if you remove the sync option/set it to false.

i wasn't sure if this was the best way to write the test, but i used getbyText which is synchronous to make the text query fail. 

image

tests/valtio/atomWithProxy.test.tsx
```diff
@@ -130,3 +130,49 @@ it('state with a promise', async () => {
...
+
+  const { findByText, getByText } = render(
+    <Provider>
+      <Elements />
+    </Provider>
+  )
+
+  await findByText('selected element: none')
+  fireEvent.click(getByText('create and select element'))
+  getByText('selected element: element') // synchronous
```
dai-shi, line 177: I never tried this test, but it looks good. Let's do this for now, and revisit it in the future if something is necessary.

dai-shi (approved):
Great addition! Thanks for your contribution.
I will merge this later possibly with some other fixes.

tests/valtio/atomWithProxy.test.tsx
tarngerine: the other way i had it working was with await but it was hard to read IMO bc i had to "invert" the test

https://github.com/pmndrs/jotai/pull/752/commits/ce31bd264dda39be6aa204ae86a39fabb29d8baf#diff-25a143e1f82ca88bb0daf47b0a2910c9974fbca024f2921fdbeebd321e205a2aR181-R183

tarngerine: happy to whatver makes sense for u tho!

tarngerine:
thanks for making 2 amazing libraries! super cool and useful stuff :)

docs/integrations/valtio.mdx
```diff
@@ -40,14 +40,31 @@ const Counter: React.FC = () => {
   return (
     <>
       count: {state.count}
-      <button onClick={() => setState((prev) => ({ ...prev, count: prev.count + 1 }))}>
+      <button
+        onClick={() =>
+          setState((prev) => ({ ...prev, count: prev.count + 1 }))
+        }>
```
tarngerine, line 46:
```suggestion
      <button onClick={() => setState((prev) => ({ ...prev, count: prev.count + 1 }))}>
```

oops, my prettier hit this, sorry

tarngerine:
docs have been moved here!

Issue #485, opened 2021-05-19
Is there a way to use jotai synced with NextJs router?


ErfanAsadi:
At first let me thank you for your awsome store management, 
There is a point, storing some data (state) on the query field of the URL is necessary some times, 
But I don't like to handle this directly, I think syncing store with the URL will be a good idea. 
In this situation we will have for example a `NextRouterAtom` which is connected to the URL and changing query or state will cause changing the other one. 
What is your idea about that?

Thisen:
We want to tackle NextJS stuff after v1, however feel free to add ideas in #340.

ErfanAsadi:
Thanks @dai-shi ,
I read about `atomWithHash`, Unfortunately It is just supporting hash in the URL, 
I think We need a more general atom which is completely connected to the router (Not just in NextJS).

dai-shi:
I'm not sure how we can completely connect to the router without two way bindings.
Hm, I think you can always assume changing the route through the atom.
I guess you can create one with `atomWithStorage`, but as it's not "storage", `atomWithRouter` can be implemented for specific router purpose.
Will you work on it? I'm not sure what API people expect. Also, not certain how/if it can be used in nextjs.

ErfanAsadi:
Sure I will work on that,, 
Just let me to read the existing packages which are for connecting router to the Redux and other store managements to see how they work exactly, 
I will inform you soon.

ptbrowne:
Just to add an experience here, we successfully used a modified version of atomWithHash, made directly from atomWithStorage, so that the atom state gets updated when the next route changes, the only change is the subscription to next's router.

```
    subscribe: (key, setValue) => {
      const callback = () => {
        const searchParams = new URLSearchParams(location.hash.slice(1))
        const str = searchParams.get(key)
        if (str !== null) {
          setValue(deserialize(str))
        } else {
          setValue(initialValue)
        }
      }
      Router.events.on('routeChangeComplete', callback) // Subscribe to next router
      window.addEventListener('hashchange', callback)
      return () => {
        Router.events.off('routeChangeComplete', callback)  // Unsubscribe to next router
        window.removeEventListener('hashchange', callback)
      }
    },
 ```

We have to do this because the history.{pushState,replaceState} used by next's router do not fire hashchange events.

>  Note that pushState() never causes a hashchange event to be fired, even if the new URL differs from the old URL only in its hash.
https://developer.mozilla.org/en-US/docs/Web/API/History/pushState

[...]

PR #758, opened 2021-10-05
fix(utils): atomWithHash options

### LINKED ISSUE #660: Feature request: add option for atomWithHash replace History instead of push
We can save state in URL but not push a new URL to history.
In my case, I can't use atomWithHash if it pushes a new URL.  And I think this is useful for navigating on the mobile web.

### LINKED ISSUE #739: atomWithHash is initialized asynchronously
Since atomWithHash is built with `delayInit`, it is initialized asynchronously = it will first have the initialValue and then after mount, will pick up the value from the URL.

My problem is that I am checking inside a useEffect that the value is the initialValue and if this is the case I populate it with values that are fetched from the server. Since initally, the atom *always* have the initialValue, my atom state is overwritten with new values, even if there is already state in the URL. 

1. I wonder why is the initialization of the atom from the URL async ? Is it to avoid any slow operation (presuming that reading from the URL is slow) during initialization ?
2. Would you be open to have `delayInit` be an option of the constructor of atomWithHash ?

```
atomWithHash("myAtom", [], serialize, deserialize, delayInit);
```

3. Do you see any other solution ? I am thinking of using a sentinel initialization value that could be checked inside the useEffect to differentiate between a real value and a initialization value.

My current work around has been to copy the atomWithHash code and change the delayInit to `false`.

Anyway, thanks for jotai !

dai-shi:
close #660 
close #739

### DIFF
--- a/src/utils/atomWithStorage.ts
+++ b/src/utils/atomWithStorage.ts
@@ -82,12 +82,56 @@ export function atomWithStorage<Value>(
   return anAtom
 }
 
+// atomWithHash is implemented with atomWithStorage
+
 export function atomWithHash<Value>(
   key: string,
   initialValue: Value,
-  serialize: (val: Value) => string = JSON.stringify,
-  deserialize: (str: string) => Value = JSON.parse
+  options?: {
+    serialize?: (val: Value) => string
+    deserialize?: (str: string) => Value
+    delayInit?: boolean
+    replaceState?: boolean
+  }
+): PrimitiveAtom<Value>
+
+/**
+ * @deprecated Use options
+ */
+export function atomWithHash<Value>(
+  key: string,
+  initialValue: Value,
+  deprecatedSerialize?: (val: Value) => string,
+  deprecatedDeserialize?: (str: string) => Value
+): PrimitiveAtom<Value>
+
+export function atomWithHash<Value>(
+  key: string,
+  initialValue: Value,
+  options?:
+    | {
+        serialize?: (val: Value) => string
+        deserialize?: (str: string) => Value
+        delayInit?: boolean
+        replaceState?: boolean
+      }
+    | ((val: Value) => string),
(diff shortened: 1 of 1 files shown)


dai-shi:
@ptbrowne ready for your review!

ptbrowne:
LGTM, thanks 🙇 I think one point of improvement could be to export the default subscribe behavior, or to attach it as a static method to atomWithHash so that it could be called from a custom subscribe.

Anyway, it looks good to me like this and could be a further improvement if enough users demand it.

meotimdihia:
@dai-shi commit https://github.com/pmndrs/jotai/commit/d7ed74635f88e0f550d3c8405fe563371596df74 is not work for my website. 
Is there any way to debug this problem? I want to help you to debug it 😀.
I am using Next.js, tried to print console in `/node_modules/jotail/ultis.js` but it does not work.

dai-shi:
So, it sounds like `hashchange` doesn't fire on location change.
Does `popstate` fire? Otherwise, we'd need to pull location.href change periodically.
https://stackoverflow.com/questions/3522090/event-when-window-location-href-changes
In other words, how do you "goes to another gallery: /gallery/view/dog-album"?

dai-shi:
> in my Next.js website

Oh, you are using nextjs.
So, @ptbrowne 's suggestion should work. https://github.com/pmndrs/jotai/issues/485#issuecomment-936215117

```js
const pageAtom = atomWithHash("page", 1, {
  replaceState: true,
  subscribe: (callback) => {
    Router.events.on('routeChangeComplete', callback)
    window.addEventListener('hashchange', callback)
    return () => {
      Router.events.off('routeChangeComplete', callback)
      window.removeEventListener('hashchange', callback)
    }
  }
})
```

meotimdihia:
@dai-shi I have tried with the code and it works. 😀

PR #759, opened 2021-10-06
feat(website): add shelf navigation on mobile and add attribution

sandren:


dai-shi:
Please check the layout with width=1024px.

image

PR #737, opened 2021-09-27
chore(ci): multiple version tests

dai-shi:

Issue #273, opened 2021-01-22
[Bug] useAtom(atom) does not re-render when the input atom changes.


kueychuj:
Issue updated in https://github.com/pmndrs/jotai/issues/273#issuecomment-765399339

Item Component won't re-render when ```id``` changes.

```ts
import * as React from "react";
import { atomFamily } from "jotai/utils";
import { useAtom } from "jotai";

const todoFamily = atomFamily<number, any>((id) => ({
  id
}));
const Item: React.FC<{ id: number }> = ({ id }) => {
  const todo = useAtom(todoFamily(id));
  return <h1>{JSON.stringify(todo)}</h1>;
};

export default function App() {
  const [a, setA] = React.useState(1);
  return (
    <div className="App">
      <button onClick={() => setA((a) => a + 1)}>+</button>
      <button onClick={() => setA((a) => a - 1)}>-</button>
      <Item id={a} />
    </div>
  );
}

```

Sandbox Example
https://codesandbox.io/s/keen-haslett-cnjz9

any thought?

merisbahti:
what's the definition of todofamily? I think that might give it away :)

kueychuj:
> what's the definition of todofamily? I think that might give it away :)

So fast. :p

Thanks for quick response. I have just created a minified example on this.
https://codesandbox.io/s/react-typescript-forked-8zfrn

merisbahti:
Hey, just 1 thing (and we might need to improve our typings to fix this), but `atomFamily` should return an `Atom<T>` and not `T`, take a look at the tests here: https://github.com/pmndrs/jotai/blob/master/tests/utils/atomFamily.test.ts

I haven't found how to reproduce your error in the UI yet, I keep checking.

dai-shi:
Wow, nice catch. Thanks for the sandbox. I confirmed that this is broken since v0.12.0.

dai-shi:
OK, I added a failing spec: 033b2e8
Not sure yet where the bug is...

kueychuj:
Thank Both, look forward to the bug fix :-)

kueychuj:
It appears unrelated to atomFamily but the useAtom hook.

The example below shows that useAtom(atom) did not resolve the correct value after the input atom is changed.
It is broken since v0.12.0.

Example: https://codesandbox.io/s/zen-snyder-rwqx9

```ts
import * as React from "react";
import { atom, useAtom } from "jotai";

const atomA = atom({ id: "a" });

[...]

PR #761, opened 2021-10-06
fix: useReducer's stale value in React@16.8.0

Aslemammad:
Fixes the small failing test in #737


dai-shi:
@Aslemammad Thanks for investigating this.
I wonder why `it('changes atom from parent (#273, #275)'` test doesn't fail.
Anyway, this seems like a bug in useReducer https://github.com/facebook/react/pull/15124,
So, I will simply change the version in ci for now. We may change the implementation of useAtom in the future and may try changing it back.
Thanks again for your work!

PR #766, opened 2021-10-08
fix(core): resolve infinite loop with chained async dependencies

### LINKED ISSUE #751: Chain of async dependencies with immediate resolution causes infinite loop
This ones a bit weird, here's the reproduction: https://codesandbox.io/s/jotai-crashing-suspend-bug-vsr2b

Basically, it's possible to construct a series of atoms that creates an infinite loop.

```js
const asyncAtom = atom(async () => {
  await new Promise((resolve) => setTimeout(resolve, 2000));

  return null;
});
const derivedAtom1 = atom(async (get) => {
  // Uncomment this line to stop crashing
  // await new Promise((resolve) => setTimeout(resolve, 10));
  return get(asyncAtom);
});

const derivedAtom2 = atom(async (get) => {
  return get(derivedAtom1);
});
```

Calling `useAtom(derivedAtom2)` will cause the infinite loop.

From what I can tell, the derived atoms have to be async, but they have to resolve immediately. This is happening in an app I'm working on at the moment.

The linked codesandbox includes the fix in #747, so it's different to #742

dai-shi:
close #751

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -16,7 +16,7 @@ const IS_EQUAL_PROMISE = Symbol()
 const INTERRUPT_PROMISE = Symbol()
 type InterruptablePromise = Promise<void> & {
   [IS_EQUAL_PROMISE]: (p: Promise<void>) => boolean
-  [INTERRUPT_PROMISE]?: () => void
+  [INTERRUPT_PROMISE]?: () => void // defined if interruptable
 }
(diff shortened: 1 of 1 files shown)


dai-shi:
https://codesandbox.io/s/jotai-crashing-suspend-bug-wb5t0 yay!

PR #777, opened 2021-10-13
fix(utils): resolve undefined observable

### LINKED ISSUE #760: Error when using `atomWithObservable` with Redux DevTools enabled
When testing `atomWithObservable` with Redux DevTools extension enabled, I get the following error:

`Cannot read properties of undefined (reading 'subscribe').`

Looks that the problem is happening here: 

https://github.com/pmndrs/jotai/blob/770dcee8b06ff833fa82a680bd82bddf499ea7e8/src/utils/atomWithObservable.ts#L47-L50

I noticed that when the DevTools extension is enabled, the `Symbol.observable` is defined so the if condition is truthy. But after running the `returnItself()` function, the observable subscribe becomes undefined.

When I disable the extension, the `Symbol.observable` is undefined so the if is skipped and the code works fine. If I comment these lines it works fine too.

I'm not sure about what these lines of code exactly do but I suppose that they are there due to some requirement, so I want to understand better about it before propose a fix.

dai-shi:
close #760

### DIFF
--- a/src/utils/atomWithStorage.ts
+++ b/src/utils/atomWithStorage.ts
@@ -94,44 +94,7 @@ export function atomWithHash<Value>(
     replaceState?: boolean
     subscribe?: (callback: () => void) => () => void
   }
-): PrimitiveAtom<Value>
-
-/**
- * @deprecated Use options
- */
-export function atomWithHash<Value>(
-  key: string,
-  initialValue: Value,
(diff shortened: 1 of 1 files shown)


dai-shi:
https://codesandbox.io/s/react-typescript-forked-forked-pbue6
Fixed! cc @ricmello

Issue #796, opened 2021-10-24
Non-hook devtools util


dai-shi:
`useAtomDevtools` is nice, but it is a hook.
In jotai mental model, we'd want it outside React.

Prior art by @aulneau: #575 

Now, I think we should revisit it. In non-hook use case, we can't automatically unsubscribe. And, an atom can be used in multiple providers.

```js
devtoolsAtom(anAtom: Atom<Value> | WritableAtom<Value, Value>, options?: { name?: string }): Unsubscribe
```

Related: https://github.com/pmndrs/jotai/issues/449#issuecomment-950248452

dai-shi:
Or, we can enable on mount and disable on unmount. In this case, we don't need to unsubscribe.
```ts
type Options = { name?: string }
function devtoolsAtom<AtomType extends Atom<Value>>(anAtom: AtomType, options?: Options): AtomType
function devtoolsAtom<AtomType extends WritableAtom<Value, Value>>(anAtom: AtomType, options?: Options): AtomType
```

Yeah, this looks more handy. Actually, it has to be, because we need to return a new atom.

dai-shi:
After working on #807 and discussing with @Aslemammad internally,
we decided to go another route.
The current `devtoolsAtom` has some limitations that I'm not happy with:
1. It only works for `PrimitiveAtom<Value>`. We can't use it for read-only atoms.
2. It doesn't cover multi provider use case. It conflicts and shows combined values in the devtools ext, without any warnings.

I will open another issue shortly for the new idea.

PR #807, opened 2021-10-28
feat: devtoolsAtom

Aslemammad:
Feat #796 

This PR is also a WIP version of #575. Special thanks to @aulneau for his helpful code!

### DIFF
--- a/src/devtools/devtoolsAtom.ts
+++ b/src/devtools/devtoolsAtom.ts
@@ -1,4 +1,4 @@
-import { atom, Atom, WritableAtom } from 'jotai'
+import { atom, PrimitiveAtom } from 'jotai'
 
 type Config = {
   instanceID?: number
@@ -38,20 +38,16 @@ const getExtension = (): Extension => {
   }
 }
 
-export function devtoolsAtom<Value, AtomType extends Atom<Value>>(
-  anAtom: AtomType,
+export function devtoolsAtom<Value>(
+  anAtom: PrimitiveAtom<Value>,
   options?: Options
-): AtomType
-
-export function devtoolsAtom<
-  Value,
-  AtomType extends WritableAtom<Value, Value>
->(anAtom: AtomType, options?: Options): AtomType {
-  const extension = getExtension()
+): PrimitiveAtom<Value> {
   const isWriteable = !!anAtom.write
   const atomName = options?.name || anAtom.debugLabel || anAtom.toString()
 
+  const extension = getExtension()
   const devtools = extension.connect({ name: atomName })
+
   let isTimeTraveling = false
 
   const listenerAtom = atom(
@@ -60,6 +56,7 @@ export function devtoolsAtom<
       if (message.type === 'init') {
         return devtools.init(get(anAtom))
       }
+
       if (message.type === 'ACTION' && message.payload) {
         try {
           if (isWriteable) {
@@ -95,7 +92,7 @@ export function devtoolsAtom<
           message.payload.nextLiftedState?.computedStates || []
         computedStates.forEach(({ state }: { state: any }, index: number) => {
           if (index === 0) {
-            devtools?.init(state)
+            devtools.init(state)
           } else {
             if (isWriteable) {
               set(subscribeAtom, state)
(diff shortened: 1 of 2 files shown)


src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export function devtoolsAtom<Value>(
+  anAtom: PrimitiveAtom<Value>,
+  options?: Options
+): PrimitiveAtom<Value> {
+  const isWriteable = !!anAtom.write
+  const atomName = options?.name || anAtom.debugLabel || anAtom.toString()
+
+  const extension = getExtension()
+  const devtools = extension.connect({ name: atomName })
```
dai-shi, line 49: We should do this in `onMount` or `write` of listenerAtom.

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  let isTimeTraveling = false
```
dai-shi, line 51: This is outside atom, which is shared among multiple instances. Please use `refAtom`. Refer `selectAtom`.

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+    null,
+    (get, set, message: Message | { type: 'init' }) => {
+      if (message.type === 'init') {
+        return devtools.init(get(anAtom))
+      }
+
+      if (message.type === 'ACTION' && message.payload) {
+        try {
+          if (isWriteable) {
+            set(subscribeAtom, JSON.parse(message.payload))
```
dai-shi, line 63: Is it intentional that this is not `anAtom`?

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+        }
+      } else if (message.type === 'DISPATCH' && message.state) {
+        if (
+          message.payload?.type === 'JUMP_TO_ACTION' ||
+          message.payload?.type === 'JUMP_TO_STATE'
+        ) {
+          isTimeTraveling = true
+
+          if (isWriteable) {
+            set(subscribeAtom, JSON.parse(message.state))
```
dai-shi, line 79: Can this be `anAtom`? Then, we don't need `isTimeTraveling` at all?

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+        message.payload?.type === 'IMPORT_STATE'
+      ) {
+        const computedStates =
+          message.payload.nextLiftedState?.computedStates || []
+        computedStates.forEach(({ state }: { state: any }, index: number) => {
+          if (index === 0) {
+            devtools.init(state)
+          } else {
+            if (isWriteable) {
+              set(subscribeAtom, state)
```
dai-shi, line 98: Same here. But, I may misunderstand something.

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+    setAtom({ type: 'init' })
+
+    return unsubscribe
+  }
+
+  const subscribeAtom: PrimitiveAtom<Value> = atom(
+    (get) => {
+      get(listenerAtom)
+
+      return anAtom.read(get)
```
dai-shi, line 120: We should not use the internal `read` here. Can this be `get(anAtom)`?

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+    },
+    (get, set, update) => {
+      const writeReturn = anAtom.write(get, set, update)
```
dai-shi, line 123: Please do `set(anAtom, update)`.

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+
+      if (isTimeTraveling) {
+        isTimeTraveling = false
+      } else {
+        devtools.send(
+          `${atomName} - ${new Date().toLocaleString()}`,
+          anAtom.read(get)
```
dai-shi, line 130: Just send `writeReturn`.

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+        )
+      }
+
+      return writeReturn
+    }
+  )
+
+  subscribeAtom.onMount = anAtom.onMount
+  subscribeAtom.debugLabel = anAtom.debugLabel
```
dai-shi, line 139: Don't copy these stuffs. onMount should just work. debugLabel should be set by a user.

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  return subscribeAtom
```
dai-shi, line 141: Not sure if there's a better name than `subscribeAtom`. 🤔

src/devtools/devtoolsAtom.ts
```diff
@@ -0,0 +1,142 @@
...
+  } catch (e) {
+    throw new Error('Please install/enable Redux devtools extension')
+  }
+}
+
+export function devtoolsAtom<Value>(
+  anAtom: PrimitiveAtom<Value>,
+  options?: Options
+): PrimitiveAtom<Value> {
+  const isWriteable = !!anAtom.write
```
dai-shi, line 45: We don't need this flag, as we can assume `anAtom` is always writable.

dai-shi (commented):
Overall, it looks good.
We can't write jest tests for this, can we?
Maybe mock devtools api?

src/devtools/devtoolsAtom.ts
Aslemammad: newAtom?

src/devtools/devtoolsAtom.ts
Aslemammad: sure! thinking about where write doesn't return anything.

Aslemammad: this doens't work, so `get(anAtom)` is better.

Aslemammad:
@LuckyPigeon Could you test it to see if it works for atomFamily?

LuckyPigeon:
@Aslemammad Sure. I can do that!

dai-shi:
https://github.com/pmndrs/jotai/issues/796#issuecomment-956015136

Sorry @LuckyPigeon, we will start over.

PR #817, opened 2021-11-04
feat(devtools): read-only atom support in useAtomDevtools

### LINKED ISSUE #808: devtools: useAtomDevtools to support read-only atoms
Currently, useAtomDevtools only supports `WritableAtom<Value, Value>`. While this is intentional, we found #807 can't support read-only atoms, technically, because it's not a hook. useAtomDevtools can do, so it would be nice to support `Atom<Value>`. In this case, we can't allow time traveling, so it would need a clear warning message.

Aslemammad:
Previously, `useAtomDevtools` only accepts primitive atoms (more precisely `WritableAtom<Value, Value>` type). Now, with this PR, it supports read-only atoms (`Atom<Value>` type).

close #808

### DIFF
--- a/src/devtools/useAtomDevtools.ts
+++ b/src/devtools/useAtomDevtools.ts
@@ -75,7 +75,7 @@ export function useAtomDevtools<Value>(
       return
     }
     console.warn(
-      '[Warn] you cannot do write operations (Time-travelling, etc) in read-only atoms ',
+      '[Warn] you cannot do write operations (Time-travelling, etc) in read-only atoms\n',
       anAtom
     )
   }
--- a/src/devtools/useAtomDevtools.ts
+++ b/src/devtools/useAtomDevtools.ts
@@ -1,7 +1,7 @@
 import { useEffect, useRef } from 'react'
 import { useAtom } from 'jotai'
-import type { WritableAtom } from 'jotai'
-import type { Scope } from '../core/atom'
+import type { WritableAtom } from '../'
+import type { Atom, Scope, SetAtom } from '../core/atom'
 
 type Config = {
   instanceID?: number
@@ -31,11 +31,28 @@ type Extension = {
   connect: (options?: Config) => ConnectionResult
 }
 
+const isWritable = <Value, Update, Result extends void | Promise<void>>(
+  atom: Atom<Value> | WritableAtom<Value, Update, Result>
+): atom is WritableAtom<Value, Update, Result> =>
+  !!(atom as WritableAtom<Value, Update, Result>).write
+
 export function useAtomDevtools<Value>(
   anAtom: WritableAtom<Value, Value>,
   name?: string,
   scope?: Scope
-) {
+): void
+
+export function useAtomDevtools<Value>(
+  anAtom: Atom<Value>,
+  name?: string,
+  scope?: Scope
+): void
+
+export function useAtomDevtools<Value>(
+  anAtom: WritableAtom<Value, Value> | Atom<Value>,
+  name?: string,
+  scope?: Scope
+): void {
   let extension: Extension | undefined
   try {
     extension = (window as any).__REDUX_DEVTOOLS_EXTENSION__ as Extension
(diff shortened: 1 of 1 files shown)


Aslemammad:
@dai-shi I think it's ready for review now! https://codesandbox.io/s/react-typescript-forked-q3fwd?file=/src/App.tsx

src/devtools/useAtomDevtools.ts
```diff
@@ -1,7 +1,6 @@
 import { useEffect, useRef } from 'react'
 import { useAtom } from 'jotai'
-import type { WritableAtom } from 'jotai'
-import type { Scope } from '../core/atom'
+import type { Atom, Scope, SetAtom, WritableAtom } from '../core/atom'
```
dai-shi, line 3:
```suggestion
import type { Atom, WritableAtom } from 'jotai'
import type { Scope, SetAtom } from '../core/atom'
```
Let's make it sure what's public and what's internal.

src/devtools/useAtomDevtools.ts
```diff
@@ -31,11 +30,28 @@ type Extension = {
...
   name?: string,
   scope?: Scope
-) {
+): void
+
+export function useAtomDevtools<Value>(
+  anAtom: Atom<Value>,
+  name?: string,
+  scope?: Scope
+): void
```
dai-shi, line 48: Now, I'm not sure if we need function overload for `useAtomDevtools`.
If not, let's remove it.

src/devtools/useAtomDevtools.ts
```diff
@@ -61,14 +89,24 @@ export function useAtomDevtools<Value>(
     if (extension) {
       devtools.current = extension.connect({ name: atomName })
       const unsubscribe = devtools.current.subscribe((message: Message) => {
-        if (message.type === 'DISPATCH' && message.state) {
+        if (message.type === 'ACTION' && message.payload) {
```
dai-shi, line 92: nice!

src/devtools/useAtomDevtools.ts
```diff
@@ -51,6 +67,18 @@ export function useAtomDevtools<Value>(
   }
 
   const [value, setValue] = useAtom(anAtom, scope)
+
+  const setAnyAtom = (value: Value) => {
```
dai-shi, line 71: I'm not sure what `Any` means here. Can we name something like `setValueSafely`?

src/devtools/useAtomDevtools.ts
```diff
@@ -51,6 +67,18 @@ export function useAtomDevtools<Value>(
...
+    if (isWritable(anAtom)) {
+      ;(setValue as SetAtom<Value, void>)(value)
```
dai-shi, line 73: I guess you want to type guard `setValue`, not `anAtom`.
What's the type of `setValue`? Let me know if you need help.

dai-shi (changes requested):
overall looks good. just minor suggestions.

src/devtools/useAtomDevtools.ts
dai-shi: Did you see any concern removing the function overload?
Only what is useful would be explicit typing.

src/devtools/useAtomDevtools.ts
```diff
@@ -59,16 +72,36 @@ export function useAtomDevtools<Value>(
...
+        } else if (message.type === 'DISPATCH' && message.state) {
           if (
             message.payload?.type === 'JUMP_TO_ACTION' ||
             message.payload?.type === 'JUMP_TO_STATE'
           ) {
             isTimeTraveling.current = true
+
+            setValueIfWritable(JSON.parse(message.state))
           }
-          setValue(JSON.parse(message.state))
```
dai-shi, line 71: we don't need to set value if the type is not JUMP_TO_ACTION/STATE?

dai-shi (commented):
A few questions before finishing.

src/devtools/useAtomDevtools.ts
Aslemammad: In other libs, we don't do that tbh, and they work well! Do we need it?

dai-shi: Oh, I see. So, it was kind of mistake only in jotai. No, I was just curious about the change.

dai-shi (approved):
looks good to me. Will merge it when we release it.

Aslemammad:
Thanks!

PR #818, opened 2021-11-04
fix(core): mount self atom before mounting dependencies

### LINKED ISSUE #813: derived atom not updating in specific circumstances
I have found that derived atoms sometimes do not update in response to changes to atoms that they depend on. The circumstances are very difficult to describe, but I have a short 72-line reproduction of the issue.

In either repro, notice that the `itemBName` atom does not update when the upstream atom is updated.

## Repro Option 1: Stackblitz

You can see the problem on stack blitz if you want. Notice the `bName: ` field is blank.

https://stackblitz.com/edit/react-ts-fpopsd?file=index.tsx

## Repro Option 2: Zip File / create-react-app

Or if its easier, download this create-react-app sample app, and then follow the directions below:
[atom-not-updating.zip](https://github.com/pmndrs/jotai/files/7456120/atom-not-updating.zip)

1. Install and start up
```
yarn install
yarn start
```

2. Open `http://localhost:3000/`

3. Notice that the 'bName' value is missing upon initial page load

## The Code

Providing the App.tsx code here from the above zip file for reference.

```typescript
import { atom } from "jotai";
import { useAtomValue } from "jotai/utils";

const INIT = Symbol();

interface ServerResponse {
  items: {
    uuid: string;
    name: string;
  }[];
}

const responseBaseAtom = atom<ServerResponse | null>(null);

function asyncFetch(set) {
  // imagine a network request here
  setTimeout(() => {
    set(responseBaseAtom, {
      items: [
        { name: "alpha", uuid: "a" },
        { name: "beta", uuid: "b" },
      ],
    });
  }, 100);
}

const responseAtom = atom(
  (get) => get(responseBaseAtom),
  (get, set, arg) => {
    if (arg === INIT) {
      asyncFetch(set);
    } else {
      set(responseBaseAtom, arg);
    }
  }
);
responseAtom.onMount = (dispatch) => {
  dispatch(INIT);
};

const mapAtom = atom((get) => {
  const response = get(responseAtom);
  if (!response) {
    return null;
  }
  return new Map(response.items.map((x) => [x.uuid, x]));
});

const itemA = atom(get => get(mapAtom)?.get('a'));
const itemB = atom(get => get(mapAtom)?.get('b'));

// FYI: For some reason if you grab this data direct from responseAtom (insstead of mapAtom) everything works.
// const itemA = atom(get => get(responseAtom)?.items.find(x => x.uuid === 'a'));
// const itemB = atom(get => get(responseAtom)?.items.find(x => x.uuid === 'b'));

const itemAName = atom(get => get(itemA)?.name);
const itemBName = atom(get => get(itemB)?.name);

  const aName = useAtomValue(itemAName);
  const bName = useAtomValue(itemBName);
  const responseValue = useAtomValue(responseAtom);
  const mapValue = useAtomValue(mapAtom);

  return (
    <div className="App">
      <div>aName: {aName}</div>
      <div>bName: {bName}</div>
      <div>Response: {JSON.stringify(responseValue)}</div>
      <div>Map: {mapValue && JSON.stringify(Array.from(mapValue.entries()))}</div>
    </div>
  );
}
```

Thisen:
close #813

- [x] Add test case from (#813)
- [x] Fix the issue


dai-shi:
@ahfarmer 
Fixed!! https://codesandbox.io/s/react-typescript-forked-nh9iw

PR #819, opened 2021-11-07
fix(test): improve valtio test

dai-shi:
to be consistent with other tests.


dai-shi:
Oh, my bad, this was because it doesn't pass in react 18.

dai-shi:
testing...

Issue #811, opened 2021-11-01
devtools: useAtomsDevtools


dai-shi:
I was hoping #796 (and #807) would be a nice debugging tool and possibly fix #449, which turned out it's not a silver bullet.

So, here's a new idea. `useAtomsDevtools`.
We already have `useAtomDevtools` which is an atomic debugging tool using Redux DevTools Ext. (We'd like to improve it with #808.)
`useAtomsDevtools` is a catch-all version of it. It still uses Redux DevTools Ext, and it might cover some use cases that #428 would cover.

`useAtomsDevtools` will show all atoms in store (= Provider), as a single source in Redux DevTools Ext. It might be too many, but will see.

Implementation-wise, it will be similar to what we have in `useAtomsSnapshot` and `useGotoAtomsSnapshot`, and of course what we did in `useAtomDevtools`.

Time traveling is a bit of challenge, because we need to keep all snapshots locally while sending it to Redux Devtools Ext. It requires some string keys of the snapshots.

Aslemammad:
Let me work on it now!

PR #837, opened 2021-11-13
feat: useAtomsDevtools

Aslemammad:
useAtomsDevtools #811 

- [x] initial trial
- [x] pass tests
- [x] batch actions
- [x] avoid duplicated logs
- [x] make dependencies work (they work in the tests)
- [x] overall behavior checks
- [x] make dependencies viewable
- [x] investigate to avoid set timeout
- [x] investigate to avoid create microtask
- [x] fix time-travel feature (JUMP_TO_STATE)
- [x] add test for JUMP_TO_STATE
- [x] add docs
- [x] refactor code
- [x] test with `<StrictMode>`
- [x] instead of dependencies, use dependents like `<Provider>`
- [x] create a demo with many atoms with changing dependencies
- [x] fix failing tests (re-implement)
- [x] fix JUMP_TO_STATE in [codesandbox](https://codesandbox.io/s/react-typescript-forked-2dhhh?file=/src/App.tsx) 👉 [a slight modification](https://codesandbox.io/s/react-typescript-forked-3fpx4?file=/src/App.tsx) works with `<AtomsDevtools>`
- [x] fix failing tests (unstable_versionedWrite)
- [x] modify docs with `<AtomsDevtools>` practice


src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,142 @@
+import {
+  AtomState,
+  DEV_GET_ATOM_STATE,
+  DEV_SUBSCRIBE_STATE,
+  WRITE_ATOM,
```
dai-shi, line 5: Please see how `useAtomsSnapshot` is implemented.
We should use DEV_GET_MOUNTED_ATOMS. We should not keep unmounted atoms.
A snapshot should hold all mounted atoms and it shouldn't be mutated after creation.
And, please use RESTORE_ATOMS, not WRITE_ATOM.

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,142 @@
...
+      console.warn('Please install/enable Redux devtools extension')
+    }
+  }
+
+  const ScopeContext = getScopeContext(scope)
+  const scopeContainer = useContext(ScopeContext)
+  const store = scopeContainer.s
+
+  if (!store[DEV_SUBSCRIBE_STATE]) {
+    throw new Error('useAtomsSnapshot can only be used in dev mode.')
```
dai-shi, line 59:
```suggestion
    throw new Error('useAtomsDevtools can only be used in dev mode.')
```

src/devtools/useAtomsDevtools.ts
Aslemammad: I get what you mean. but I don't find a solution with using snapshots (as you say). So let me try with atom.toString and push my findings. then let me know your opinion.

tests/devtools/useAtomsDevtools.test.tsx
```diff
@@ -0,0 +1,284 @@
...
+  fireEvent.click(getByText('button'))
+  await findByText('count: 2')
+  await findByText('double: 4')
+  expect(extension.send).toBeCalledTimes(5)
+})
+
+describe('when it receives an message of type...', () => {
+  describe('DISPATCH and payload of type...', () => {
+    describe('JUMP_TO_STATE | JUMP_TO_ACTION...', () => {
+      it('time travelling', async () => {
```
dai-shi, line 206: Can we add a little bit more complex scenario?

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+  const snapshots = useRef<AtomsSnapshot[]>([])
+
+  useEffect(() => {
+    let devtoolsUnsubscribe: () => void | undefined
+    if (extension) {
+      devtools.current = extension.connect({ name })
+
+      devtoolsUnsubscribe = devtools.current.subscribe((message: Message) => {
+        switch (message.type) {
+          case 'ACTION':
```
dai-shi, line 109: maybe we don't need to support this. there's no such notion.

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+          case 'DISPATCH':
+            switch (message.payload?.type) {
+              case 'RESET':
+                // Todo
+                return
+              case 'COMMIT':
+                devtools.current!.init(undefined)
+                return
+              case 'JUMP_TO_STATE':
+              case 'JUMP_TO_ACTION':
```
dai-shi, line 120: should we only support JUMP_TO_ACTION? JUMP_TO_STATE doesn't make sense, and we don't have actionId, do we?

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+      setTimeout(() => {
+        isTimeTraveling.current = false
+      }, 0)
+    } else if (isRecording.current) {
+      snapshots.current.push(snapshot)
+
+      const parsedSnapshot = serializeSnapshot(snapshot)
+
+      devtools.current.send(
+        `action:${snapshots.current.length} - ${new Date().toLocaleString()}`,
```
dai-shi, line 166: Should this be just
```
`${snapshots.current.length}`
```
to make L124 work?

If we want the date, put it in the action object.
```js
{
  type: `${snapshots.current.length}`,
  updatedAt: new Date().toLocaleString(),    
}
```

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+      devtools.current.shouldInit = false
+      return
+    }
+
+    if ([...snapshot.keys()].length === 0) {
+      return
+    }
+
+    if (isTimeTraveling.current) {
+      // better solution to stop logging on JUMP_TO_ACTION & JUMP_TO_STATE ?
```
dai-shi, line 156: why does it work in `useAtomDevtools`?

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (!devtools.current) {
+      return
+    }
+    if (devtools.current.shouldInit) {
+      devtools.current.init(undefined)
+      devtools.current.shouldInit = false
+      return
+    }
+
+    if ([...snapshot.keys()].length === 0) {
```
dai-shi, line 151:
```suggestion
    if (!snapshot.size) {
```

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  const isTimeTraveling = useRef(false)
+  const isRecording = useRef(true)
+  const devtools = useRef<ConnectionResult & { shouldInit?: boolean }>()
+  const snapshots = useRef<AtomsSnapshot[]>([])
+
+  useEffect(() => {
+    let devtoolsUnsubscribe: () => void | undefined
+    if (extension) {
+      devtools.current = extension.connect({ name })
```
dai-shi, line 105:
```suggestion
      const connection = extension.connect({ name })
      devtools.current = connection // move this to just above L134
```
can avoid L117 non-null assertion.

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+                return (isRecording.current = !isRecording.current)
+            }
+        }
+      })
+
+      devtools.current.shouldInit = true
+    }
+    return () => {
+      devtoolsUnsubscribe?.()
+    }
```
dai-shi, line 138:
```suggestion
    return devtoolsUnsubscribe
```
can work?

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    if (isTimeTraveling.current) {
+      // better solution to stop logging on JUMP_TO_ACTION & JUMP_TO_STATE ?
+      setTimeout(() => {
+        isTimeTraveling.current = false
+      }, 0)
+    } else if (isRecording.current) {
+      snapshots.current.push(snapshot)
+
+      const parsedSnapshot = serializeSnapshot(snapshot)
```
dai-shi, line 163:
```suggestion
      const serializedSnapshot = serializeSnapshot(snapshot)
```

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  snapshot.forEach((v, atom) => {
+    result[atomToDebuggingString(atom)] = v
+  })
+
+  return result
+}
+
+const atomToDebuggingString = (atom: Atom<unknown>) =>
+  `${atom}:${atom.debugLabel ? atom.debugLabel : atom}`
```
dai-shi, line 53:
```suggestion
  `${atom}:${atom.debugLabel || atom}`
```

dai-shi: Hm, but this is still weird.
Maybe this?
```js
  atom.debugLabel ? `${atom}:${atom.debugLabel}` : `${atom}`
```

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,174 @@
...
+  const result: Record<string, unknown> = {}
+
+  snapshot.forEach((v, atom) => {
+    result[atomToDebuggingString(atom)] = v
+  })
+
+  return result
+}
+
+const atomToDebuggingString = (atom: Atom<unknown>) =>
```
dai-shi, line 52: btw, we call it `atomToPrintable` in Provider.ts.

dai-shi (commented):
Please fix lint errors and tests too.

src/devtools/useAtomsDevtools.ts
Aslemammad: i guess it didn't work

src/devtools/useAtomsDevtools.ts
Aslemammad: hmm, but doesn't work well with types

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,180 @@
...
+  return result
+}
+
+export function useAtomsDevtools(name: string, scope?: Scope) {
+  const snapshot = useAtomsSnapshot()
+  const goToSnapshot = useGotoAtomsSnapshot()
+  let extension: Extension | undefined
+  try {
+    extension = (window as any).__REDUX_DEVTOOLS_EXTENSION__ as Extension
+  } catch {}
```
dai-shi, line 71:
```suggestion
  } catch {
    // ignored
  }
```

src/devtools/useAtomsDevtools.ts
dai-shi: Oh, this is not correct:
```ts
let devtoolsUnsubscribe: () => void | undefined
```
It has to be:
```ts
let devtoolsUnsubscribe: (() => void) | undefined
```

Aslemammad:
I'm a hell of a test runner. 😅

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,212 @@
+import { useContext, useEffect, useRef } from 'react'
+import { useAtomsSnapshot, useGotoAtomsSnapshot } from 'jotai/devtools'
+import { SECRET_INTERNAL_getScopeContext as getScopeContext } from 'jotai'
+import type { Atom, Scope } from '../core/atom'
+import { DEV_GET_ATOM_STATE, DEV_SUBSCRIBE_STATE, Store } from '../core/store'
```
dai-shi, line 5:
```suggestion
import { DEV_GET_ATOM_STATE, DEV_SUBSCRIBE_STATE } from '../core/store'
import type { Store } from '../core/store'
```

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,212 @@
...
+      devtools.current.shouldInit = false
+      return
+    }
+
+    if (!snapshot.size) {
+      return
+    }
+
+    if (isTimeTraveling.current) {
+      queueMicrotask(() => {
```
dai-shi, line 163: we never use queueMicrotask anywhere else. while it's technically fine, i'm concerned and curious why we don't need it with `useAtomDevtools`.

src/devtools/useAtomsDevtools.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  const state = {
+    values: serializedSnapshot,
+    dependencies: getDependencies(store, snapshot),
+  }
+
+  batchesQueue.push(state)
+  snapshotsQueue.push(snapshot)
+  const batchesLength = batchesQueue.length
+  setTimeout(() => {
```
dai-shi, line 196: This isn't what I expected. Why I wonder why `useEffect` doesn't batch... I might be missing something.

src/devtools/useAtomsDevtools.ts
Aslemammad: I tried different batching solutions, but only `queueMicrotask` worked since it's a little bit other than Promise.resolve().then & setTimeout

src/devtools/useAtomsDevtools.ts
Aslemammad: which part you mean it doesn't batch?

tests/devtools/useAtomsDevtools.test.tsx
dai-shi: @Aslemammad Do you have any other ideas to add some more tests?

src/devtools/useAtomsDevtools.ts
dai-shi: This is future work.

Aslemammad (commented):
We talked about this code, we can mark it as experimental in the release. so it's ready now!

dai-shi (approved):
finally, it looks good.

Aslemammad:
Finally!

PR #839, opened 2021-11-16
refactor(types): enable exactOptionalPropertyTypes

dai-shi:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,8 @@
     "test:ci": "jest",
     "test:dev": "jest --watch --no-coverage",
     "test:coverage:watch": "jest --watch",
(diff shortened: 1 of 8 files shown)


dai-shi:
@devanshj Curious if you know how to resolve the type error. Seems like it's coming from the xstate types.

devanshj:
Yep xstate does not have `--exactOptionalPropertyTypes` enabled in their codebase, hence they can emit these less strict types. You can solve this by turning on `--skipLibCheck` in this repo or sending a PR to xstate to fix these types. The fix would probably be rather trivial just removing `| undefined` in these three places (`onDone`, `exec` and `target`)

dai-shi:
Merging this. We will remove the patch when the new xstate package is released.

PR #844, opened 2021-11-22
fix(utils): Unwrap loadable/selectAtom promise types

Pinpickle:
#713 changed the type signature for atoms. The `data` field for a loadable ended up incorrectly typed with a promise, when it should be unwrapped.

A similar issue occurred with `selectAtom`

This adds a `ResolveType` (used in other files) to fix it.

### DIFF
--- a/src/utils/loadable.ts
+++ b/src/utils/loadable.ts
@@ -2,12 +2,14 @@ import { atom } from 'jotai'
 import type { Atom } from 'jotai'
 import { createMemoizeAtom } from './weakCache'
 
+type ResolveType<T> = T extends Promise<infer V> ? V : T
+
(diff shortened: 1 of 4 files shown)


dai-shi (approved):
ohhhhh, nice catch. thanks a lot!
something totally missed in #713. thanks for the test too.

dai-shi:
Maybe type assertion can be solved (eliminated) with the future version (with `Awaited`?)

Pinpickle:
> Maybe type assertion can be solved (eliminated) with the future version (with `Awaited`?)

I sure hope so! Although I don't know if it'll eliminate the assertion altogether... the code may need some refactoring so that TS knows the promise has been resolved.

PR #850, opened 2021-11-26
chore(test): Use SWC as Jest transformer

Thisen:
Use [@swc/jest](https://github.com/swc-project/jest) as Jest transformer, for faster test runs.

Locally I saw this:
- Before: ~10 secs
- With @swc/jest: ~6.6 secs

That's about 41% faster.

This is not much we gain currently, but as discussed between us maintainers, we want to focus more on tests, so this change will be better for us, as time goes.

NOTE: This is an experimental and a suggestion, I'm fine with this landing. 😄


tests/async.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { StrictMode, Suspense, useEffect, useRef } from 'react'
+import React, { StrictMode, Suspense, useEffect, useRef } from 'react'
```
dai-shi, line 1: It's our intention to prefer jsx-transform. If swc doesn't support it, can we somehow hack it without changing the code please?

tests/async.test.tsx
```diff
@@ -498,7 +498,7 @@ it('set promise atom value on write (#304)', async () => {
     set(
       countAtom,
       Promise.resolve(get(countAtom)).then(
-        (c) => new Promise((r) => setTimeout(() => r(c + 1), 500))
+        (c) => new Promise((r) => setTimeout(() => r(c + 1), 1))
```
dai-shi, line 501: not sure how this works. do they use the fake timer by default?

tests/async.test.tsx
```diff
@@ -767,9 +767,7 @@ it('non suspense async write self atom with setTimeout (#389)', async () => {
 })
 
 it('should override promise as atom value (#430)', async () => {
-  const countAtom = atom(
-    new Promise<number>((r) => setTimeout(() => r(-1), 3600 * 1000))
```
dai-shi, line 771: this means the promise should never resolve. sorry for the confusion.
let's make it `new Promise<number>)(() => {}) // infinite promise`.

dai-shi (changes requested):
we also need to make sure that all CI tests pass, fairly stably.

aulneau:
I use this in a number of my own projects, great add!

tests/async.test.tsx
Thisen: Can you elaborate?

tests/async.test.tsx
Thisen: Fixed :)

tests/async.test.tsx
dai-shi: In my experience with the previous one, this doesn't pass CI.
So, we would like to a) see if it passes, and b) know why.

Thisen: You need to it forward then :)

dai-shi (commented):
Alright, the code looks good.
We make this type of PRs for zustand and valtio too, before merging.
Would you be able to work on those two, or do you want me or someone else to work on them?

Thisen:
> Alright, the code looks good. We make this type of PRs for zustand and valtio too, before merging. Would you be able to work on those two, or do you want me or someone else to work on them?

Happy to do it for Valtio and Zustand aswell.

.github/workflows/test-multiple-versions.yml
```diff
@@ -65,6 +65,7 @@ jobs:
         run: |
           sed -i~ '1s/^/import React from "react";/' tests/*.tsx tests/*/*.tsx
           sed -i~ 's/automatic/classic/' babel.config.js
+          sed -i~ 's/automatic/classic/' .swcrc
```
dai-shi, line 68: 👏

dai-shi (approved):

dai-shi:
@Thisen I think it's because of this PR, but now jest error reporting doesn't show the correct error location (line number) in local development. This is pretty annoying. Hope you find a solution.

Thisen:
> @Thisen I think it's because of this PR, but now jest error reporting doesn't show the correct error location (line number) in local development. This is pretty annoying. Hope you find a solution.

I'll look into it.

tests/async.test.tsx
```diff
@@ -27,7 +27,7 @@ it('does not show async stale result', async () => {
     const [count, setCount] = useAtom(countAtom)
     const onClick = async () => {
       setCount((c) => c + 1)
-      await new Promise((r) => setTimeout(r, 10))
+      await new Promise((r) => setTimeout(r, 1))
```
dai-shi, line 30: @Thisen Unfortunately, this is causing the CI failure several times, so let me revert it.

PR #854, opened 2021-11-29
feat(core): experimental unstable versioned write for limited use case

dai-shi:
I've been working on this for a few months now. It's since #687. Let's see how it goes. I will do some experiments and some patch releases towards this.

PR #857, opened 2021-11-29
fix(utils): refactor loadable

dai-shi:
This is full refactor of the loadable util, introduced in #734. The previous implementation doesn't work with #854.

### DIFF
--- a/src/utils/loadable.ts
+++ b/src/utils/loadable.ts
@@ -11,32 +11,57 @@ type Loadable<Value> =
   | { state: 'hasError'; error: unknown }
   | { state: 'hasData'; data: ResolveType<Value> }
 
+const LOADING: Loadable<unknown> = { state: 'loading' }
+
 export function loadable<Value>(anAtom: Atom<Value>): Atom<Loadable<Value>> {
   return memoizeAtom(() => {
-    // TODO we should revisit this for a better solution than refAtom
-    const refAtom = atom(() => ({} as { prev?: Loadable<Value> }))
+    const loadableAtomCache = new WeakMap<
+      Promise<void>,
+      Atom<Loadable<Value>>
+    >()
 
-    const derivedAtom = atom((get): Loadable<Value> => {
-      const ref = get(refAtom)
-      let curr = ref.prev
+    const catchAtom = atom((get) => {
+      let promise: Promise<void>
       try {
-        const value = get(anAtom) as ResolveType<Value>
-        if (curr?.state !== 'hasData' || !Object.is(curr.data, value)) {
-          curr = { state: 'hasData', data: value }
-        }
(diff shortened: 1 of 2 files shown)


dai-shi:
@Pinpickle Would you be able to try this if it doesn't break your use case?

Pinpickle:
This looks good to me @dai-shi - if the tests pass then it works for my usecase!

I would suggest adding these tests (which are passing), just to make sure it's not getting tripped up by any async logic in the mix:

```ts
it('loadable immediately resolves sync values', async () => {
  const syncAtom = atom(5)
  const effectCallback = jest.fn()

  const { getByText } = render(
    <Provider>
      <LoadableComponent effectCallback={effectCallback} asyncAtom={syncAtom} />
    </Provider>
  )

  getByText('Data: 5')
  expect(effectCallback.mock.calls).not.toContain(
    expect.objectContaining({ state: 'loading' })
  )
  expect(effectCallback).toHaveBeenLastCalledWith({ state: 'hasData', data: 5 })
})

it('loadable can use resolved promises syncronously', async () => {
  const asyncAtom = atom(Promise.resolve(5))
  const effectCallback = jest.fn()

  const ResolveAtomComponent = () => {
    useAtomValue(syncAtom)

    return <div>Ready</div>
  }

  const { getByText, findByText, rerender } = render(
    <Provider>
      <Suspense fallback={null}>
        <ResolveAtomComponent />
      </Suspense>
    </Provider>
  )

  await findByText('Ready')

  rerender(
    <Provider>
      <LoadableComponent effectCallback={effectCallback} asyncAtom={asyncAtom} />
    </Provider>
  )
  getByText('Data: 5')

  expect(effectCallback.mock.calls).not.toContain(
    expect.objectContaining({ state: 'loading' })
  )
  expect(effectCallback).toHaveBeenLastCalledWith({ state: 'hasData', data: 5 })
})
```

It does require a slight change to the `LoadableComponent` at the bottom of the file:

```ts
interface LoadableComponentProps {
  asyncAtom: Atom<Promise<number> | Promise<string> | string | number>
  effectCallback?: (loadableValue: any) => void
}

const LoadableComponent = ({
  asyncAtom,
  effectCallback,
}: LoadableComponentProps) => {
  const value = useAtomValue(loadable(asyncAtom))

  useEffect(() => {
    if (effectCallback) {
      effectCallback(value)
    }
  }, [value, effectCallback])

  if (value.state === 'loading') {
    return <>Loading...</>
  }

  if (value.state === 'hasError') {
    return <>{String(value.error)}</>
  }

  // this is to ensure correct typing
  const data: number | string = value.data

  return <>Data: {data}</>
}
```

dai-shi:
Added the tests. Thanks!

PR #866, opened 2021-12-03
fix(core): refactor store with suspense promise

dai-shi:
This includes various refactors.

- interruptable promise -> suspense promise
- eliminate prepareNextAtomState
- simplify setting promise on write (this changes the behavior slightly)

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -159,7 +159,7 @@ export const createStore = (
 
   const setAtomReadError = <Value>(
     atom: Atom<Value>,
-    error: unknown,
+    error: ReadError,
     dependencies?: Set<AnyAtom>,
     suspensePromise?: SuspensePromise
   ): void => {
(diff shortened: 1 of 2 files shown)


dai-shi:
Oh no, CI fails with react 16.8.6. Can't figure out what's wrong. Wonder if anyone has ideas.

Thisen:
It looks like it gets stuck in `atomWithQuery.test.tsx`. Why, I don't know. However, a bunch of these are thrown when running the tests with `react@16.8.6` + `react-dom@16.8.6`:
```
It looks like you're using a version of react-dom that supports the "act" function, but not an awaitable version of "act" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning
```

I did a small commit that fixed minor typos, I hope that's ok.

dai-shi:
Thanks for the fix! The act warning was there from the beginning. Not sure if resolvable. I see it gets stuck even with `tests/async.test.tsx`.

dai-shi:
It turns out there was a bug in my refactor (I did the whole refactor again one by one with running tests in 16.8.6). What a relief. Thanks for looking into it.

PR #875, opened 2021-12-06
atomWithObservable options to allow enabled and initialData

wesleycole:
Allow an options object in `atomWithObservable` that provides the ability to enable/disable the atom and provide an initial data value. 

Currently, the enabled flag is working properly and initialData works as a writable atom, but it does not work as a readable atom only. There is a lot of cleanup that can be done with this around overloads and typing. Have been working on getting a working version first and then will clean up types.

https://github.com/pmndrs/jotai/issues/234

### DIFF
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -118,16 +118,17 @@ export function atomWithObservable<TData>(
     }
 
     const dataAtom = atom<TData | Promise<TData>>(
-      enabled &&
-        new Promise<TData>((resolve, reject) => {
-          settlePromise = (data, err) => {
-            if (err) {
-              reject(err)
-            } else {
-              resolve(data as TData)
+      enabled && initialData === undefined
+        ? new Promise<TData>((resolve, reject) => {
+            settlePromise = (data, err) => {
+              if (err) {
+                reject(err)
+              } else {
+                resolve(data as TData)
+              }
             }
-          }
-        })
+          })
+        : initialData
     )
     let setData: (data: TData | Promise<TData>) => void = () => {
       throw new Error('setting data without mount')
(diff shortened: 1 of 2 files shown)


dai-shi:
Just curious if `initialValue` should be reactive. There would be some use cases, but do you have any in mind?

Oh, this reminds me: I wonder if this is called `initialData` like atomWithQuery or `initialValue` like atomWithStorage. I think we use `initialValue` in general. atomWithQuery use `initialData` only because it's from react-query.

wesleycole:
> Just curious if initialValue should be reactive. There would be some use cases, but do you have any in mind?

Yea, probably should be reactive and should be able to be overwritten with the next value from the subscription (not implemented currently). 

What I'm attempting to do is initially query some data from an API asynchronously and seed the atomWithObservable atom with that initial data from the API. From there, I want to subscribe to all updates to that model in the DB and update another atom with the newly updated values from the subscription. 

```
export const initialDataAtom = atom(async (get) =>
  DataStore.query(Model, get(idAtom)),
);

const subscriptionAtom = atomWithObservable(
  () => DataStore.observe(Model, get(idAtom)),
  (get) => ({ initialData: get(initialDataAtom) })
)
```

> Oh, this reminds me: I wonder if this is called initialData like atomWithQuery or initialValue like atomWithStorage. I think we use initialValue in general. atomWithQuery use initialData only because it's from react-query.

Yep, I modeled it after atomWithQuery, but will update to initialValue.

Thank you for all of the great feedback! I'm probably in a little bit over my head with this PR ;)

wesleycole:
Also, I don't really have this use case anymore due to some changed business requirements with my app, so if this doesn't make sense to include this in Jotai, please feel free to close the PR. I think the initialValue can be useful, just not sure I know how to get this to the finish line.

dai-shi:
I like the idea of having `initialValue` to avoid initial suspense.
You can take time. But, I can take over if you like. Please let me know. Anyway, adding tests and testing is very helpful.

dai-shi:
Closing in favor of #1058.

PR #878, opened 2021-12-07
fix(core): Bail out of atom.write when atom update and current value are equal

### LINKED ISSUE #877: [core] dependency tracking fails with derived chains
Split from: https://github.com/pmndrs/jotai/issues/874#issuecomment-987338240

Repro with sync atoms
https://codesandbox.io/s/jotai-demo-forked-cucwg

Repro with async atoms
https://codesandbox.io/s/jotai-demo-forked-y58e3

Thisen:
Fixes #877.

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -426,8 +426,9 @@ export const createStore = (
           // NOTE technically possible but restricted as it may cause bugs
           throw new Error('atom not writable')
         }
-        const aState = readAtomState(a)
-        if ('v' in aState && Object.is(aState.v, v)) {
+        const aState = getAtomState(a)
+        // Bail out, if update value and current value are equal
+        if (aState && 'v' in aState && Object.is(aState.v, v)) {
           return
         }
         setAtomPromiseOrValue(a, v)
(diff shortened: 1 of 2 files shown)


src/core/store.ts
```diff
@@ -426,6 +426,10 @@ export const createStore = (
           // NOTE technically possible but restricted as it may cause bugs
           throw new Error('atom not writable')
         }
+        const aState = readAtomState(a)
```
dai-shi, line 429: Can this be `getAtomState`? readAtomState tries to do more work.

dai-shi (commented):
Great work!!!
So, setAtomPromiseOrValue bails out for non-promise values, but we are invoking invalidateDependents anyway, which causes this behavior. This is partly my intention, but overlooked the reported behavior.

src/core/store.ts
Thisen: Done 👍🏼

dai-shi (approved):
Thanks again for this! This is new to me and motivates me to investigate some more, because I feel like having some other bugs. I will try.

PR #879, opened 2021-12-08
docs(website): add documentation for atomWithHash options

notsidney:
Hi, while researching atomWithHash, I found the PR that added options (#758) but they were not documented on the site. I’ve added documentation for options.

### DIFF
--- a/docs/api/utils.mdx
+++ b/docs/api/utils.mdx
@@ -311,14 +311,34 @@ discarding changes made previously via the `set` function.
 ### Usage
 
 ```js
-atomWithHash(key, initialValue): PrimitiveAtom
+atomWithHash(key, initialValue, options): PrimitiveAtom
 ```
 
-This creats a new atom that is connected with URL hash.
+This creates a new atom that is connected with URL hash.
 The hash must be in the URLSearchParams format.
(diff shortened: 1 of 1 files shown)


notsidney:
Also, I read your comments saying you were thinking of removing atomWithHash altogether. Please don’t! I find it very useful and am now using it as part of [Rowy](https://github.com/rowyio/rowy)

dai-shi (approved):
Welcome to the jotai docs team!!
Thanks for working on it. Merging this.

dai-shi:
> Also, I read your comments saying you were thinking of removing atomWithHash altogether. 

At that point, it felt like `atomWithHash` is like a simple use case of atomWithStorage. But, since then, it have got improved/complicated, also supporting nextjs use cases. Worth keeping.
No immediate plans to change. In the future, we might consider moving to another package, if that makes sense (which might mean adding more features.)

dai-shi:
@notsidney Hey, this is a nice article! https://dev.to/notsidney/how-and-why-you-should-store-react-ui-state-in-the-url-34pi

notsidney:
@dai-shi Thanks! This actually reminded me, now that replaceState is documented, I can update the links in this article :)

PR #883, opened 2021-12-10
fix(tests): add useAtomDevtools test

Aslemammad:


dai-shi (approved):
Looks good! It's great to have a test for useAtomDevtools.

PR #887, opened 2021-12-13
doc: Make TOC link in Utils page kebab case

mauricedb:


dai-shi (approved):
Thanks for the fix. Confirmed the links work now in the preview.

PR #889, opened 2021-12-14
fix(type): Fix types on waitForAll to work with destructuring

russelldavis:
Example:
```typescript
const numberAtom = atom(0)
const stringAtom = atom('')
const [num, str] = useAtomValue(waitForAll([numberAtom, stringAtom]))
```

Previously, in the code above, both `num` and `str` would have the type `number | string | undefined`. With this fix, `num` has the type `number` and `str` has the type `string`.

### DIFF
--- a/src/utils/waitForAll.ts
+++ b/src/utils/waitForAll.ts
@@ -4,11 +4,15 @@ import { createMemoizeAtom } from './weakCache'
 
 const memoizeAtom = createMemoizeAtom()
 
-type ResolveType<T> = T extends Promise<infer V> ? V : T
+type ResolvePromise<T> = T extends Promise<infer V> ? V : T
+type ResolveAtom<T> = T extends Atom<infer V> ? V : T
+type ResolveType<T> = ResolvePromise<ResolveAtom<T>>
 
 export function waitForAll<
-  Values extends Record<string, unknown> | readonly unknown[]
->(atoms: { [K in keyof Values]: Atom<Values[K]> }): Atom<{
+  Values extends Record<string, Atom<unknown>> | readonly Atom<unknown>[]
+>(
+  atoms: Values
+): Atom<{
   [K in keyof Values]: ResolveType<Values[K]>
 }> {
   const createAtom = () => {
@@ -43,14 +47,18 @@ export function waitForAll<
 }
 
 const unwrapAtoms = <
-  Values extends Record<string, unknown> | unknown[]
->(atoms: { [K in keyof Values]: Atom<Values[K]> }): Atom<unknown>[] =>
+  Values extends Record<string, Atom<unknown>> | readonly Atom<unknown>[]
+>(
+  atoms: Values
+): Atom<unknown>[] =>
   Array.isArray(atoms)
     ? atoms
     : Object.getOwnPropertyNames(atoms).map((key) => atoms[key as keyof Values])
 
-const wrapResults = <Values extends Record<string, unknown> | unknown[]>(
-  atoms: { [K in keyof Values]: Atom<Values[K]> },
+const wrapResults = <
+  Values extends Record<string, Atom<unknown>> | readonly Atom<unknown>[]
+>(
+  atoms: Values,
   results: unknown[]
 ): unknown[] | Record<string, unknown> =>
   Array.isArray(atoms)
--- a/src/utils/waitForAll.ts
+++ b/src/utils/waitForAll.ts
@@ -9,11 +9,11 @@ type ResolveAtom<T> = T extends Atom<infer V> ? V : T
 type ResolveType<T> = ResolvePromise<ResolveAtom<T>>
 
 export function waitForAll<
-  Values extends Record<string, Atom<unknown>> | readonly Atom<unknown>[]
+  Atoms extends Record<string, Atom<unknown>> | readonly Atom<unknown>[]
 >(
-  atoms: Values
+  atoms: Atoms
 ): Atom<{
-  [K in keyof Values]: ResolveType<Values[K]>
+  [K in keyof Atoms]: ResolveType<Atoms[K]>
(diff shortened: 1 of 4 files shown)


dai-shi:
Oh, nice catch. I thought we had tests to detect it.

If you do this:
```js
const [num, str] = useAtomValue(waitForAll([numberAtom, stringAtom] as const))
```
It should work, right?

src/utils/waitForAll.ts
```diff
@@ -4,11 +4,15 @@ import { createMemoizeAtom } from './weakCache'
...
 
-type ResolveType<T> = T extends Promise<infer V> ? V : T
+type ResolvePromise<T> = T extends Promise<infer V> ? V : T
+type ResolveAtom<T> = T extends Atom<infer V> ? V : T
+type ResolveType<T> = ResolvePromise<ResolveAtom<T>>
 
 export function waitForAll<
-  Values extends Record<string, unknown> | readonly unknown[]
->(atoms: { [K in keyof Values]: Atom<Values[K]> }): Atom<{
+  Values extends Record<string, Atom<unknown>> | readonly Atom<unknown>[]
```
dai-shi, line 12:
```suggestion
  Atoms extends Record<string, Atom<unknown>> | readonly Atom<unknown>[]
```
Can we rename like this?

tests/types.test.tsx
```diff
@@ -1,8 +1,27 @@
...
 import type { SetAtom } from '../src/core/atom'
 
+it('waitForAll() should return the correct types', () => {
+  function Component() {
+    const numberAtom = atom(async () => 0)
+    const stringAtom = atom(async () => '')
+
+    const [number, string] = useAtomValue(waitForAll([numberAtom, stringAtom]))
+    expectType<number>(number)
+    expectType<string>(string)
```
dai-shi, line 14: Cool.

russelldavis:
Ah yep, it does already work with `as const`.

Note I just pushed a new version that updates the new test to use atoms with async getters, to make sure the types properly unwrap promises (they do).

tests/types.test.tsx
```diff
@@ -1,8 +1,27 @@
...
+
+    const { numberAtom: number2, stringAtom: string2 } = useAtomValue(
+      waitForAll({ numberAtom, stringAtom })
+    )
+    expectType<number>(number2)
+    expectType<string>(string2)
+  }
+  Component
+})
```
dai-shi, line 23: We are still not sure, but I feel like we should put this in `tests/utils`.
Can you make a new `tests/utils/waitForAll.types.text.tsx` file meanwhile?

dai-shi (commented):
Other than that, it looks good to me.

dai-shi:
> I thought we had tests to detect it.

https://github.com/pmndrs/jotai/blob/6ab8b75cd56186ca5445ec99eba29ccaaf6c1cb8/tests/utils/waitForAll.test.tsx#L70

https://github.com/pmndrs/jotai/blob/6ab8b75cd56186ca5445ec99eba29ccaaf6c1cb8/tests/utils/waitForAll.test.tsx#L129

I think we should remove `* 2` and `.toUpperCase()` if they are covered in the new tests. (and expect `getByText('num: 1, str: a')`)

src/utils/waitForAll.ts
russelldavis: 👍

tests/types.test.tsx
russelldavis: 👍

tests/types.test.tsx
```diff
@@ -1,6 +1,6 @@
 import { expectType } from 'ts-expect'
-import { atom, useAtom } from 'jotai'
 import type { Atom, PrimitiveAtom, WritableAtom } from 'jotai'
+import { atom, useAtom } from 'jotai'
```
dai-shi, line 3: let's revert this if it's not intentional.

russelldavis:
I've made all the suggested changes. Should be good to go. Thanks for the quick review, and the great library.

dai-shi (commented):
https://github.com/pmndrs/jotai/pull/889/files#diff-4180ba105f80d33e7b2603e6ea2e1805189c1049eabee392b3f20b44394f8d15R129
This is also covered by the new test, right?

tests/types.test.tsx
russelldavis: Whoops, good catch. Fixed.

russelldavis:
One thing to note: when using `useAtom` and only destructuring the first value, typescript still has the old behavior of assigning the union of all the types. Example:

```ts
const [[num, str]] = useAtom(waitForAll([numberAtom, stringAtom]))
```

There are several workarounds:

1) Using `as const`

```ts
const [[num, str]] = useAtom(waitForAll([numberAtom, stringAtom] as const))
```

2) Using a dummy var for the second value
```ts
const [[num, str], _] = useAtom(waitForAll([numberAtom, stringAtom]))
```

3) Using `useAtomValue`
```ts
const [num, str] = useAtomValue(waitForAll([numberAtom, stringAtom]))
```

Since (2) is semantically equivalent to the original code, this seems like a bug in Typescript.

dai-shi:
Thanks for the note. Interesting. I think it's fine.

---

https://github.com/russelldavis/jotai/blob/c25794729696bdad59c6ff41e6ac90cddebb5546/tests/utils/waitForAll.test.tsx#L129

If it's covered in waitForAll.types.test.tsx, this can be `return { num, str }` too?

russelldavis:
> I think it's fine.

Agreed -- it's the best we can do given typescript's behavior.

> If it's covered in waitForAll.types.test.tsx, this can be return { num, str } too?

Fixed.

dai-shi (approved):
Very nice! Thanks for your contribution!

PR #915, opened 2021-12-29
Comment store and useAtom internals

justjake:
- document the meaning of store production methods
- document reasoning in useAtom

Please review for correctness of these comments!

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -49,7 +49,29 @@ export type AtomState<Value = AnyAtomValue> = {
   d: ReadDependencies
 } & ({ e: ReadError } | { p: SuspensePromise } | { v: ResolveType<Value> })
 
-export type VersionObject = { p?: VersionObject } // "p"arent version
+/**
+ * Represents a version of a store. A version contains a state for all atoms.
+ *
+ * In concurrent React rendering, state can "branch" during transitions: the
+ * current state may continue to be rendered while a new state is being built
+ * concurrently. We key our atom state containers by this global version to
+ * represent the state for each diverging branch.
+ *
+ * While a new version is being built, we read atom previous state from the
+ * previous version.
+ */
+export type VersionObject = {
+  /**
+   * "p"arent version.
+   *
+   * Once a version is committed completely, the `p` property is deleted so the
+   * child version is independent, and the parent revision can be garbage
+   * collected.
+   *
+   * See [Provider] for more details on version data flow.
+   */
+  p?: VersionObject
+}
 
 type Listeners = Set<(version?: VersionObject) => void>
 type Dependents = Set<AnyAtom>
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -113,10 +113,11 @@ export const WRITE_ATOM = 'w'
  */
 export const COMMIT_ATOM = 'c'
 /**
- * Add a subscriber function to an atom. The atom will become "mounted" if this
- * is the first subscriber.
+ * Add a subscriber function to an atom. Returns a function that removes the
+ * subscriber.
  *
  * The subscriber is called in two cases:
+ *
  * - For writable atoms, the subscriber is called whenever the atom is directly
  *   changed by `atom.write`.
  * - For derived atoms, the subscriber is called whenever the atom is
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -481,7 +481,6 @@ export const createStore = (
       const atomState = getAtomState(version, atom)
       if (atomState) {
         // First, ensure that each atom we depend on is up to date.
-        //
         // Recursive calls to `readAtomState(version, a)` will recompute `a` if
         // it's out of date thus increment its revision number if it changes.
         atomState.d.forEach((_, a) => {
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -29,6 +29,7 @@ type ReadDependencies = Map<AnyAtom, Revision>
 
 /**
  * Immutable atom state, tracked for both mounted and unmounted atoms in a store.
+ * @private This is for internal use and not considered part of the public API.
  */
 export type AtomState<Value = AnyAtomValue> = {
   /**
@@ -38,7 +39,7 @@ export type AtomState<Value = AnyAtomValue> = {
   /**
    * Marks the revision of this atom when a transitive dependency was invalidated.
    * Mounted atoms are considered invalid when `r === i`. Unmounted atoms are
-   * always considered invalid.
+   * considered invalid until they are mounted again.
    */
   i?: InvalidatedRevision
   /**
@@ -50,7 +51,8 @@ export type AtomState<Value = AnyAtomValue> = {
 } & ({ e: ReadError } | { p: SuspensePromise } | { v: ResolveType<Value> })
 
 /**
- * Represents a version of a store. A version contains a state for all atoms.
+ * Represents a version of a store. A version contains a state for every atom
+ * read during the version's lifetime.
  *
  * In concurrent React rendering, state can "branch" during transitions: the
  * current state may continue to be rendered while a new state is being built
@@ -77,8 +79,11 @@ type Listeners = Set<(version?: VersionObject) => void>
 type Dependents = Set<AnyAtom>
 
 /**
- * State tracked for atoms with subscribers, or transitive dependencies of atoms
- * with subscribers. It is freed once an atom no longer has subscribers.
+ * State tracked for mounted atoms. An atom is considered "mounted" if it has a
+ * subscriber, or is a transitive dependency of another atom that has a
+ * subscriber.
+ *
+ * The mounted state of an atom is freed once it is no longer mounted.
  */
 type Mounted = {
   /** The list of subscriber functions. */
@@ -111,9 +116,13 @@ export const COMMIT_ATOM = 'c'
  * Add a subscriber function to an atom. The atom will become "mounted" if this
  * is the first subscriber.
  *
- * The subscriber is called whenever the atom is *invalidated* (i.e. when it's
- * possibly transitive dependencies change or become invalidated), **not** when
- * the actual Value of the atom changes. Derived atoms are only recomputed on read.
+ * The subscriber is called in two cases:
+ * - For writable atoms, the subscriber is called whenever the atom is directly
+ *   changed by `atom.write`.
+ * - For derived atoms, the subscriber is called whenever the atom is
+ *   *invalidated* (i.e. when it's possibly transitive dependencies change or
+ *   become invalidated), **not** when the actual Value of the atom changes.
+ *   Derived atoms are only recomputed on read.
  */
 export const SUBSCRIBE_ATOM = 's'
 /**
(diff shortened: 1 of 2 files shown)


src/core/store.ts
```diff
@@ -26,20 +26,65 @@ type Revision = number
 type InvalidatedRevision = number
 type ReadDependencies = Map<AnyAtom, Revision>
 
-// immutable atom state
+/**
+ * Immutable atom state, tracked for both mounted and unmounted atoms in a store.
```
dai-shi, line 30:
```suggestion
 * Immutable atom state, tracked for both mounted and unmounted atoms in a store.
 * This is for internal use and not for public API.
```

src/core/store.ts
```diff
@@ -26,20 +26,65 @@ type Revision = number
...
+ */
 export type AtomState<Value = unknown> = {
+  /**
+   * Counts number of times atom has actually changed or recomputed.
+   */
   r: Revision
+  /**
+   * Marks the revision of this atom when a transitive dependency was invalidated.
+   * Mounted atoms are considered invalid when `r === i`. Unmounted atoms are
+   * always considered invalid.
```
dai-shi, line 40: Unmounted atoms are not *always* considered invalid.
The validity will be checked when they are re-mounted.

src/core/store.ts
```diff
@@ -26,20 +26,65 @@ type Revision = number
...
+   * Maps from a dependency to the dependency's revision when it was last read.
+   * We can skip recomputation of an atom by comparing the ReadDependencies revision
+   * of each dependency to that dependencies's current revision.
+   */
   d: ReadDependencies
 } & ({ e: ReadError } | { p: SuspensePromise } | { v: ResolveType<Value> })
 
-export type VersionObject = { p?: VersionObject } // "p"arent version
+/**
+ * Represents a version of a store. A version contains a state for all atoms.
```
dai-shi, line 52: To be precise, it's not for *all* atoms. We only copy atom states on demand, so it will contain a subset of all atoms.

src/core/store.ts
```diff
@@ -26,20 +26,65 @@ type Revision = number
...
+ * While a new version is being built, we read atom previous state from the
+ * previous version.
+ */
+export type VersionObject = {
+  /**
+   * "p"arent version.
+   *
+   * Once a version is committed completely, the `p` property is deleted so the
+   * child version is independent, and the parent revision can be garbage
+   * collected.
```
dai-shi, line 68: Yes!

src/core/store.ts
```diff
@@ -26,20 +26,65 @@ type Revision = number
...
+   */
+  p?: VersionObject
+}
 
 type Listeners = Set<(version?: VersionObject) => void>
 type Dependents = Set<AnyAtom>
+
+/**
+ * State tracked for atoms with subscribers, or transitive dependencies of atoms
+ * with subscribers. It is freed once an atom no longer has subscribers.
```
dai-shi, line 80: There are mounted atoms without subscribers. They are only dependencies of other atoms.

src/core/store.ts
```diff
@@ -48,10 +93,31 @@ type StateListener = () => void
 type MountedAtoms = Set<AnyAtom>
 
 // store methods
```
dai-shi, line 95:
```suggestion
// store methods (not for public API)
```

src/core/store.ts
```diff
@@ -48,10 +93,31 @@ type StateListener = () => void
...
+/**
+ * Read an atom's hidden [AtomState]. Derived atom state may be recomputed if they
```
dai-shi, line 97: what's "hidden" mean?

src/core/store.ts
```diff
@@ -48,10 +93,31 @@ type StateListener = () => void
...
+ * are invalidated and any of their transitive dependencies have changed.
+ */
 export const READ_ATOM = 'r'
+/**
+ * Write an atom's Value.
```
dai-shi, line 102: To be precise, it invokes `atom.write` and it doesn't always write *this* atom's value.

src/core/store.ts
```diff
@@ -48,10 +93,31 @@ type StateListener = () => void
...
+ */
 export const WRITE_ATOM = 'w'
+/**
+ * Commit pending writes to an atom.
```
dai-shi, line 106: To be precise, **in the current implementation**, all pending writes will be committed. (not only to *this* atom)

src/core/store.ts
```diff
@@ -48,10 +93,31 @@ type StateListener = () => void
...
+ */
 export const COMMIT_ATOM = 'c'
+/**
+ * Add a subscriber function to an atom. The atom will become "mounted" if this
```
dai-shi, line 110: To be precise, an atom will be mounted if a) it has a subscriber, or b) another atom depend on it.

src/core/store.ts
```diff
@@ -48,10 +93,31 @@ type StateListener = () => void
...
+ * is the first subscriber.
+ *
+ * The subscriber is called whenever the atom is *invalidated* (i.e. when it's
```
dai-shi, line 113: To be precise, a subscriber is called when an atom is invalidated or the atom value is directly changed by `atom.write`.

src/core/useAtom.ts
```diff
@@ -59,7 +61,14 @@ export function useAtom<Value, Update, Result extends void | Promise<void>>(
...
   )
 
-  const [[version, value, atomFromUseReducer], dispatch] = useReducer<
+  // This reducer stores this hook's view of the atom's state (prev[1]).
+  //
+  // We know the component should re-render whenever it's view of the atom changes.
+  //
+  // If we did change tracking using a previous state stored outside of React,
+  // that state could be out of sync with React's concurrent rendering with
+  // `useTransition`.
```
dai-shi, line 70: I'm not sure if I understand what this note is trying to say.

src/core/useAtom.ts
```diff
@@ -86,12 +95,14 @@ export function useAtom<Value, Update, Result extends void | Promise<void>>(
...
 
   if (atomFromUseReducer !== atom) {
-    dispatch(undefined)
+    rerenderIfChanged(undefined)
   }
 
   useEffect(() => {
-    const unsubscribe = store[SUBSCRIBE_ATOM](atom, dispatch)
-    dispatch(undefined)
+    // Call `rerenderIfAtomStateChanged` whenever this atom is invalidated. Note
```
dai-shi, line 102:
```suggestion
    // Call `rerenderIfChanged` whenever this atom is invalidated. Note
```

btw, I wonder if there could be a better name than `rerenderIfChanged`. Seems good for now.

src/core/store.ts
```diff
@@ -26,20 +26,65 @@ type Revision = number
...
+ *
+ * While a new version is being built, we read atom previous state from the
+ * previous version.
+ */
+export type VersionObject = {
+  /**
+   * "p"arent version.
+   *
+   * Once a version is committed completely, the `p` property is deleted so the
+   * child version is independent, and the parent revision can be garbage
```
dai-shi, line 67:
```suggestion
   * child version is independent, and the parent version can be garbage
```

dai-shi (commented):
Wow, so surprising you have a good understanding of the internal behavior, even for versions.
Added some comments. It should be good to go once they are addressed.

src/core/useAtom.ts
justjake: I agree that the name could be better. I tried a few iterations (like `rerenderIfAtomStateChanged`), but I was also trying to avoid Prettier from reflowing the entire callback definition due to line length.

justjake:
@dai-shi Thank you for the review. I addressed you comments. Please take another look.

src/core/store.ts
dai-shi: @justjake Just to make sure if you are aware that a mounted atom may not have any listeners, are you?

src/core/store.ts
dai-shi: > Unmounted atoms are not always considered invalid.

This is still to be taken into account?

src/core/useAtom.ts
dai-shi: Yeah, good point about prettier.

dai-shi:
Thanks. Looking good. Please check unresolved comments. There are 6 of them.

src/core/store.ts
justjake: Yes. I think I fixed the comment.

src/core/useAtom.ts
justjake: I decided to remove this one, since I'm not sure either 😓

src/core/store.ts
justjake: I have changed the text to:

```
  * Marks the revision of this atom when a transitive dependency was invalidated.
   * Mounted atoms are considered invalid when `r === i`. Unmounted atoms are
   * considered invalid until they are mounted again.
```

Is that correct?

src/core/store.ts
justjake: I will change it to "internal". I mean, this state is not part of the public API, and our consumer does not need to consider it.

src/core/store.ts
dai-shi: > Unmounted atoms are considered invalid until they are mounted again.

Oh, I see the change now. I meant, unmounted atoms are not considered invalid, the invalidity will be checked after re-mount, and we then know if it's valid or invalid. So, unmounted atoms can still be valid. (Some of tests cover this behavior.)

src/core/store.ts
dai-shi: good. will you fix here or any reason not?

src/core/useAtom.ts
dai-shi: okay!

src/core/store.ts
dai-shi: gotcha.

src/core/store.ts
justjake: Yes!

src/core/store.ts
justjake: I see. I got the impression from this code:

https://github.com/justjake/jotai/blob/38c1d07616c564b5132af631f563b92f3d6e6526/src/core/store.ts#L485-L488

```
            if (!mountedMap.has(a)) {
              // not mounted
              readAtomState(version, a)
            } else {
```

because we always `readAtomState(version, a)` for unmounted atoms, but for mounted atoms, we only `readAtomState(version, a)` if 

```
              const aState = getAtomState(version, a)
              if (
                aState &&
                aState.r === aState.i // revision is invalidated
              ) {
                readAtomState(version, a)
              }
```

justjake: It seems like this case is hit when an unmounted atom is read for the first time -- we assume it's invalid?

dai-shi: > It seems like this case is hit when an unmounted atom is read for the first time -- we assume it's invalid?

Yeah, that sounds right. We may want to improve the `not mounted` comment.

dai-shi: Ah, I see. It's not mounted, so we need check the validity, thus invoking readAtomState.

justjake: Putting the invalidation algorithm into a comment is too difficult. I removed the note about unmounted atoms here entirely, and tried to add a few more comments to `readAtomState` instead.

justjake:
I think i got all the PR feedback I missed the first time!

src/core/store.ts
```diff
@@ -375,14 +477,22 @@ export const createStore = (
     force?: boolean
   ): AtomState<Value> => {
     if (!force) {
+      // See if we can skip recomputing this atom.
       const atomState = getAtomState(version, atom)
       if (atomState) {
+        // First, ensure that each atom we depend on is up to date.
+        //
```
dai-shi, line 484:
```suggestion
```
sorry, just my personal preference.

src/core/store.ts
```diff
@@ -393,6 +503,9 @@ export const createStore = (
             }
           }
         })
+
```
dai-shi, line 506:
```suggestion
```

src/core/store.ts
```diff
@@ -408,6 +521,8 @@ export const createStore = (
         }
       }
     }
+
```
dai-shi, line 524:
```suggestion
```

dai-shi (approved):
This looks great. Thanks tons for your work on this. This will help other developers reading the code.
Merging this and we can always fix or improve it.

PR #919, opened 2021-12-30
docs(advanced-recipes): atomWithCompare: atom with custom compare function to skip updates

justjake:
I use shallow equality semantics pervasively in my application, does it make sense to upstream a util atom creator that allows specifying a custom compare function? It's a simple layer on top of atomWithReducer, so maybe it's too trivial? Up to you.

Note that there's no way for us to "force" Jotai to update when areEqual(a, b) is false, but Object.is(a, b) is true without modifying the core. What do you think about adding `atom.areEqual(prev: Value, next: Value)` to the Atom config type? That's a rather invasive change, so let's start with a util that warns about the limitation.

Changes:

- Rename atomWithReducer arguments to use longer variable names. This makes it easier to understand `atomWithReducer` in IDEs that offer type hints.
- Create atomWithCompare util and a test.
- Fixed some spelling mistakes I saw along the way.

If the general idea is approved, I will add a simple .mdx doc page.


dai-shi:
Thanks for your work and suggestion.

> Note that there's no way for us to "force" Jotai to update when areEqual(a, b) is false, but Object.is(a, b) is true without modifying the core.

Yes, this is very intentional. Basically, we follow React useState behavior. The core is based on immutable state model, and will be so.
Someone requested before about avoiding updating even if Object.is(a, b) is false. We strongly rejected this idea for the same reason, and suggested the use of `selectAtom` for a workaround.
I think in your case, a workaround would be `{ ...b }` to force update.

> does it make sense to upstream a util atom creator that allows specifying a custom compare function?

In a sense, it makes sense that this is the inversion of `selectAtom`.
However, our principle is immutable state model, so in my suggestion, this falls into "Advanced Recipes" in docs, instead of a util.

> It's a simple layer on top of atomWithReducer

Yeah, that's also why having in "Advanced Recipes" makes sense.
In general, we don't provide a util that depends on other utils. It's the space for third-party libraries.

src/utils/atomWithReducer.ts
```diff
@@ -3,17 +3,17 @@ import type { WritableAtom } from 'jotai'
 
 export function atomWithReducer<Value, Action>(
   initialValue: Value,
-  reducer: (v: Value, a?: Action) => Value
+  reducer: (state: Value, action?: Action) => Value
```
dai-shi, line 6:
```suggestion
  reducer: (previousValue: Value, action?: Action) => Value
```
prefer "value" to "state".

justjake:
@dai-shi moved to "advanced recipes.

dai-shi (approved):
Looks good to me!

Issue #861, opened 2021-11-30
[RFC] unstable_createStore


dai-shi:
If #854 is successfully merged, we would like to consider exposing `unstable_createStore`.

```jsx
import { atom, Provider, unstable_createStore, unstable_readAtomValue } from 'jotai'

const countAtom = atom(0)
const store = unstable_createStore()

const App = () => (
  <Provider unstable_createStore={() => store}>
    ...
  </Provider>
)

// read atom value outside React
const value = unstable_readAtomValue(store, countAtom)
```

a-eid:
would it also be possible to set an atom value from outside react ?

dai-shi:
Yeah, that's the plan. Updated description.

a-eid:
@dai-shi that would be amazing.

loganvolkers:
This would be great!

It would let us use jotai with lit, haunted, and stencil via https://github.com/saasquatch/universal-hooks

langtube:
Is it possible to release an alpha version now?

dai-shi:
Please give us more time. It's still RFC. It's actually helpful if you can note why you need it. (And, maybe we can discuss current possible workaround.)

langtube:
Sorry, didn't mean to rush, just hope to use this feature soon.
In my case, I need to perform transient updates, similar to this one: https://github.com/pmndrs/jotai/discussions/610

dai-shi:
So, your commend is valuable, because this doesn't allow transient updates. `store.set()` will trigger re-render.

langtube:
Oh, it's a lucky thing I replied. I usually just read and don't reply because my English is not good.

Related Codes:
``` ts
  const offsetRef = useRef(useStore.getState().offsets[index]);
  useEffect(
    () =>
      useStore.subscribe(
        (state) => state.offsets[index],
        (offset: Offset) => {
          offsetRef.current = offset;

          animation.animateTo({
            transform: [{ translateX: offset.x }, { translateY: offset.y }],
          });
        },
      ),
    [],
  );
```

dai-shi:
jotai doesn't have a proper way for transient updates.
In some of utils, we do have a hack,
https://github.com/pmndrs/jotai/blob/685c8dc84e0ba7093af900a72d0bc14a37a09371/src/utils/selectAtom.ts#L15-L16
but it's not something to recommend, and hope to revisit this in the future.

[...]

PR #922, opened 2022-01-01
feat(core): unstable_createStore

dai-shi:
As discussed in #861, this adds `unstable_createStore` api and prop. This is an experimental feature and may change without notices.


src/core/store.ts
```diff
@@ -898,3 +898,22 @@ export const createStore = (
...
 export type Store = ReturnType<typeof createStore>
+
+export const createStoreForExport = (
+  initialValues?: Iterable<readonly [AnyAtom, AnyAtomValue]>
+) => {
+  const store = createStore(initialValues)
+  return {
+    get: <Value>(atom: Atom<Value>): ResolveType<Value> | undefined => {
+      const atomState = store[READ_ATOM](atom)
+      return 'v' in atomState ? atomState.v : undefined
```
justjake, line 909: This is an unfortunate erasure if `type Value = OtherThing | undefined` but the atom can also have a suspense promise. Perhaps the result of `get` can be a more interesting type like `{ value: Value } | undefined`?

justjake: And, what about the error case? Perhaps that should still throw, although I feel pretty strongly that `store.get(atom)` should never throw the suspense promise.

dai-shi: Yeah, this is unfortunate.
I think it should throw on error, and maybe suspense too.
We could have `has` method, hmm.

justjake: Does `has` cause recompute?

justjake: Re: throwing promise: I think outside of React library authors it’s pretty unusual to think about throwing promises, whereas throwing errors is common for most JS styles. Given that the public `store` intends to expose Jotai for use *outside* React, I think it might be inappropriate to throw the suspense promise since that’s a kind of react specific, framework internal practice.  But I am behind the times in terms of React best practices so maybe this is better known in the cutting edge JS community?

justjake: I understand the goal of Jotai is to be as close to React behavior as possible, but maybe this interface is the exception?

dai-shi: Hm, probably we can just return a promise.

dai-shi: but this isn't nice in terms of typing...

dai-shi: 281a617 would it be too complicated to explain?

dai-shi: > Does `has` cause recompute?

Good point, yeah...

justjake: https://github.com/pmndrs/jotai/commit/281a617d1247173fb0c4abf00b03ffbbf95f4cf5  This is very tricky to design!

the way I would write this in my personal, typescript only world would be to use a discriminating union result type that always forces the caller to consider the different types of result:

`type GetResult<Value> = { type: ‘value’, value: Resolve<Value> } | { type: ‘suspended’, promise: Promise<???> } | { type: ‘error’, error: unknown }`. 

But this style can be hard for JS-only consumers to understand or get right, and it’s not very natural in JS to return the error. So you could always throw the error, or return the two other variants. 

i am starting to think I am unqualified to understand the suspense cases though, I don’t know enough about suspense in React or how an atom ends up “suspended”.

dai-shi: From the typescript perspective, you can't know if an atom will suspend or not.
Do we want to check type=value, even if we never use async in our code?
So, after all, if we were to explain it to someone who don't know suspense (which is likely), the original `return undefined` idea would be easier (we throw on error). We would be simply explaining when the value is not ready, it returns `undefined`. It's like how `Map.get` is typed.

dai-shi: So, this: 4fd5afa

justjake: What about a second method with the result type, for callers that are working with async? You had as an option overload before, that works too. I do think that returning undefined is better than throwing promise.

justjake: Should the public interface be expressive enough to allow an alternative implementation of useAtom by consumers, including suspense, if they chose to have those semantics? It’s not important to me, though. Another thought exercise about the purpose of the public store API.

dai-shi: Yeah, that's the purpose of the RFC issue.
Exposing everything / having a second method would be all possible, but I don't think it's very good developer experience.

281a617 is good in a sense that `unstable_promise` is already introduced in getter in `atom.write`. It's still very unstable though.

a-eid:
@dai-shi I'm wondering how to patch up my node_modules package with this PR ?

justjake:
@a-eid is the component reading the atom via store.get, or by useAtom? Store.get is non-reactive. I’ve been testing a similar change in my Jotai fork with a very similar implementation and it seems to be working fine — the only difference is in how the provider works.

a-eid:
> @a-eid is the component reading the atom via store.get, or by useAtom? Store.get is non-reactive. I’ve been testing a similar change in my Jotai fork with a very similar implementation and it seems to be working fine — the only difference is in how the provider works.

`useAtom` and/or`useAtomValue`

dai-shi:
> I'm wondering how to patch up my node_modules package with this PR ?

See: https://github.com/pmndrs/jotai/pull/922#issuecomment-1003482782

Basically, find "Local Install Instructions" in: https://ci.codesandbox.io/status/pmndrs/jotai/pr/922

> doesn't seem that I'm able to pass the `store` to the provider components.

I haven't implemented this part yet. Will do shortly.

src/core/store.ts
```diff
@@ -898,3 +898,31 @@ export const createStore = (
 }
 
 export type Store = ReturnType<typeof createStore>
+
+export const createStoreForExport = (
```
dai-shi, line 902: I wonder if this should be named `createStoreForExporting`, `createStoreToExport` or something else.

justjake: We can rename the old function because it was private, right?

then we can rename Store -> InternalStore
createStore -> createInternalStore
And finally createStore can be the public name?

justjake: This should probably have an explicit return type so that changes to internal function return types won’t break the new public API

dai-shi: This name is still private and we can change it at any time.
I know what you mean, but let's not change the other part while this is unstable.

Typing is tricky. Let me know if you have trouble with it.

a-eid:
@dai-shi what happens if you're getting an async storage atom on startup with the store. 
would it return a promise ?


```ts
store.get("ASYNC_ATOM")
```

dai-shi:
@a-eid That's we had been discussing. Currently, it returns `undefined`, because outside React, it's not easy to understand Suspense behavior. What's your use case?

a-eid:
> @a-eid That's we had been discussing. Currently, it returns `undefined`, because outside React, it's not easy to understand Suspense behavior. What's your use case?

one use case that we can't get around is when running headlessJS task with react native.

a-eid:
I'm not sure how would `onMount` behave in case the `store.get` method is called before the atom is used ?

dai-shi:
> one use case that we can't get around is when running headlessJS task with react native.

hm, not sure i get the real issue. we can consider `store.asyncGet(atom)`. would that help?

> I'm not sure how would `onMount` behave in case the `store.get` method is called before the atom is used ?

good point. we wouldn't be able to finalize the behavior very soon, but `store.get` is not treated as "used". thus, not mounted and no `onMount`. #961 is related to make this behavior more consistent.

a-eid:
I think `store.asyncGet(atom)` would work.

dai-shi:
> currently to access an async atom outside react, we need to get it via `asyncStorage`, it would be nice to get it via the store.

If you "sub"scribe, you can "get" even the async value which is ready. I'm not sure if `asyncGet` is really appropriate for your case. Let's add it for experiments.

a-eid:
> > currently to access an async atom outside react, we need to get it via `asyncStorage`, it would be nice to get it via the store.
> 
> If you "sub"scribe, you can "get" even the async value which is ready. I'm not sure if `asyncGet` is really appropriate for your case. Let's add it for experiments.

"sub"scribe would not work unless the atom is mounted.

dai-shi:
☝️ `asyncGet` is here. Please experiment with it.

> "sub"scribe would not work unless the atom is mounted.

subscribe actually mounts the atom (and you need to unsubscribe to unmount it).

PR #924, opened 2022-01-02
Update license year

Awilum:
Update license year to 2022

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020-2021 Poimandres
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020-2021 Poimandres
+Copyright (c) 2020-2022 Poimandres
```
dai-shi, line 3:
```suggestion
Copyright (c) 2020 Poimandres
```
I don't think we need the range, so this is fine.
Anyway, merging this as it's harmless. It may look active.

PR #514, opened 2021-06-11
wip(splitAtom): item atom should be static

dai-shi:
~to close #510.~

Issue #510, opened 2021-06-09
Facing an error with splitAtom version of todos example


tell-y:
Hi,
I tried to use splitAtom() to learn it with the todos example.
Then I'm facing an error with a specific operation.

Representation code:
https://codesandbox.io/s/todos-splitatom-sample-ts-jh9mz

The specific operation:
https://user-images.githubusercontent.com/6883692/121356500-dd838600-c96b-11eb-9a09-b75caab9318c.mov

Maybe it's a bug so I just report It.
Thanks.

dai-shi:
> We need to handle removing items from upstream (arrAtom).

This is already handled. So, the bug seems to be caused by a subtle timing issue. @Aslemammad is investigating it.

PR #925, opened 2022-01-02
fix(utils): big refactor on splitAtom to handle stale atoms (no more warning, throwing only if unrecoverable)

### LINKED ISSUE #736: Possible bug with atom derived from variable sized splitted atom
The following warning appears in console if derivative was created from variable sized splitted atom:
```splitAtom: array index out of bounds, returning undefined warning in console.```

Here is sandbox https://codesandbox.io/s/client-splitatom-derivative-5uir4

To reproduce the issue just start clicking on the button and check out console.

### PR DESCRIPTION
**Pull request by Aslemammad**
fix #736

- [x] create a test to reproduce the reported codesandbox
- [x] fix the bug
- [x] minimize the test
- [x] resolve infinite loop in CI
- [x] refactored completely for not-idea-but-fairly-reasonable behavior
- [x] adjust tests to follow the new behavior

### DIFF
--- a/src/utils/splitAtom.ts
+++ b/src/utils/splitAtom.ts
@@ -9,6 +9,13 @@ import type {
 } from 'jotai'
 import { createMemoizeAtom } from './weakCache'
 
+const SafeWeakRef =
+  typeof WeakRef !== 'undefined'
+    ? WeakRef
+    : (class DummyWeakRef {
+        deref() {}
+      } as typeof WeakRef)
+
 const memoizeAtom = createMemoizeAtom()
 
 const isWritable = <Value, Update, Result extends void | Promise<void>>(
@@ -39,6 +46,7 @@ export function splitAtom<Item, Key>(
       type Mapping = {
         atomList: ItemAtom[]
         keyList: Key[]
+        prevRef?: WeakRef<Item[]>
       }
       const mappingCache = new WeakMap<Item[], Mapping>()
       const getMapping = (arr: Item[], prev?: Item[]) => {
@@ -62,23 +70,28 @@ export function splitAtom<Item, Key>(
             return
           }
           const read = (get: Getter) => {
-            const ref = get(refAtom)
-            const arr = get(arrAtom)
-            const mapping = getMapping(arr, ref.prev)
-            const index = mapping.keyList.indexOf(key) ?? -1
-            if (index < 0 || index >= arr.length) {
-              throw new Error('splitAtom: index out of bounds for read')
+            let prev = get(refAtom).prev
+            let arr = get(arrAtom) as Item[] | undefined
+            while (arr) {
+              const mapping = getMapping(arr, prev)
+              const index = mapping.keyList.indexOf(key) ?? -1
+              if (index < 0 || index >= arr.length) {
+                arr = prev
+                prev = mapping.prevRef?.deref()
+              } else {
+                return arr[index] as Item
+              }
             }
-            return arr[index] as Item
+            throw new Error('splitAtom: index out of bounds for read')
           }
           const write = (
             get: Getter,
             set: Setter,
             update: SetStateAction<Item>
           ) => {
-            const ref = get(refAtom)
+            const prev = get(refAtom).prev
             const arr = get(arrAtom)
-            const mapping = getMapping(arr, ref.prev)
+            const mapping = getMapping(arr, prev)
             const index = mapping.keyList.indexOf(key) ?? -1
             if (index < 0 || index >= arr.length) {
               throw new Error('splitAtom: index out of bounds for write')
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: tests/utils/splitAtom.test.tsx [hunk 1]
```diff
@@ -375,3 +375,50 @@ it('no error on wrong atom configs (fix 510)', async () => {
...
   expect(console.warn).toHaveBeenCalledTimes(1)
 })
+
+it('changing array of items (fix #736)', async () => {
+  const collectionAtom = atom<number[]>([])
+  const collectionAtomsAtom = splitAtom(collectionAtom)
+
+  const derivativeAtom = atom((get) => {
+    return get(collectionAtomsAtom).map(
+      (ca) => get(ca) + Math.round(Math.random() * 150)
```
dai-shi, line 385:
```suggestion
      (ca, index) => get(ca) + index
```
Can this be like this? Math.random() is a side effect, which is not generally allowed in the read function.

tests/utils/splitAtom.test.tsx
```diff
@@ -3,7 +3,7 @@ import type { ChangeEvent } from 'react'
 import { fireEvent, render, waitFor } from '@testing-library/react'
 import { atom, useAtom } from 'jotai'
 import type { Atom, PrimitiveAtom } from 'jotai'
-import { splitAtom } from 'jotai/utils'
+import { splitAtom, useAtomValue, useUpdateAtom } from 'jotai/utils'
```
dai-shi, line 6: I would like to simply use `useAtom` only, if that's not a problem to reproduce the test.

tests/utils/splitAtom.test.tsx
```diff
@@ -373,5 +373,64 @@ it('no error on wrong atom configs (fix 510)', async () => {
   fireEvent.click(evenCheckboxEl)
 
   expect(numbersEl.textContent).toBe('0')
-  expect(console.warn).toHaveBeenCalledTimes(1)
+  expect(console.warn).toHaveBeenCalledTimes(0)
```
dai-shi, line 376:
```suggestion
```
We should probably remove this entirely. It's basically a mistake, right?

tests/utils/splitAtom.test.tsx
```diff
@@ -373,5 +373,64 @@ it('no error on wrong atom configs (fix 510)', async () => {
...
+  const derivativeAtom = atom((get) =>
+    get(collectionAtomsAtom).map(
+      (ca) => get(ca) + Math.round(Math.random() * 150)
+    )
+  )
+
+  const numberAtom = atom(1)
+
+  const generateCollection = (number: number) =>
+    Array.from({ length: Math.round(Math.random() * 30) }, (_, i) => i + number)
```
dai-shi, line 392: Can we somehow make the test less complicated? Do we really need to create a random length of array?

tests/utils/splitAtom.test.tsx
```diff
@@ -373,5 +373,64 @@ it('no error on wrong atom configs (fix 510)', async () => {
...
+
+  const { getByText } = render(
+    <Provider>
+      <App />
+    </Provider>
+  )
+
+  fireEvent.click(getByText('+1'))
+
+  expect(console.warn).not.toHaveBeenCalled()
```
dai-shi, line 435: Is this the only "expect" this will fail? It would be nice if we check `undefined` somehow and remove the warn check.

dai-shi:
If we are sure about the previous test was wrong, this seems a right fix.
https://codesandbox.io/s/client-splitatom-derivative-forked-ljpgd

I will add some tasks in the description.

dai-shi:
So, I re-implemented completely to fix everything.
First of all, #510 case is something we don't want to support, but we shouldn't error if possible.
So, instead of warning possible out-of-bounds warning, we return the original value. This doesn't lie types. We only throw for serious errors.
This should also solve the warning issue with react-spring usage I saw with [my codesandbox](https://codesandbox.io/s/jotai-todos-focus-atom-ledxb).

The test we added in #514 is written to just check errors and possible type errors. It doesn't work correctly due to `useCachedAtoms` (it's a wrong usage) and the behavior is with this new splitAtom impl.

src/utils/splitAtom.ts
```diff
@@ -36,77 +36,101 @@ export function splitAtom<Item, Key>(
...
-              )
+            const itemRef = get(itemRefAtom)
+            const ref = get(refAtom)
+            const arr = get(arrAtom)
+            const mapping = getMapping(arr, ref.prev)
+            const index = mapping.keyList.indexOf(key) ?? -1
+            if (index < 0 || index >= arr.length) {
+              if ('prev' in itemRef) {
+                // returning a stale value to avoid errors for use cases such as react-spring
+                return itemRef.prev
```
Aslemammad, line 75: This PR also handles the useSpring use case, which needs a stale value. Amazing!

Aslemammad:
LGreatTM!

dai-shi (approved):
should be good.

PR #926, opened 2022-01-02
docs(guides): Composing atoms

### LINKED ISSUE #899: [docs] add "composing atoms" in "guides"
To show something like this:

```js
const baseWindowTitleAtom = atom('')
export const windowTitleAtom = atom(
  (get) => get(baseWindowTitleAtom) || window.title,
  (get, set, val) => {
    window.tite = val
    set(baseWindowTitleAtom, val)
  }
)
```

dai-shi:
close #899

### DIFF
--- a/docs/guides/composing-atoms.mdx
+++ b/docs/guides/composing-atoms.mdx
@@ -1,0 +1,206 @@
+---
+title: Composing atoms
+nav: 3.10
+---
+
+The `atom` function provided by library is very primitive,
+but it's also so flexible that you can combine multiple atoms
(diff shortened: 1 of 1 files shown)


docs/guides/composing-atoms.mdx
```diff
@@ -0,0 +1,206 @@
+---
+title: Composing atoms
+nav: 3.10
```
dai-shi, line 3: @sandren Looks like it's in the order of `3.1`, `3.10`, `3.2`...

PR #929, opened 2022-01-04
getting started section for the docs

Sajad-Sharhani:
Adding a new section to the docs
tell me your opinion and suggestions


docs/getting-started.mdx
```diff
@@ -0,0 +1,76 @@
+---
+title: Getting Started
+nav: 0.1
+---
```
dai-shi, line 4:
```suggestion
---

```

docs/getting-started.mdx
```diff
@@ -0,0 +1,76 @@
...
+Welcome to the Jotai Getting Started tutorial!
+This tutorial will introduce you to Jotai and teach you how to start using it correctly.
+
+For this tutorial, we assume that you're using Jotai with React. The example is based on [a Create-React-App folder structure](https://create-react-app.dev/docs/folder-structure) so the first thing you have to do is:
```
dai-shi, line 8: Do you think CRA is better than web based environment, like codesandbox or stackblitz?
I'm not sure if we should mention about the folder structure.

docs/getting-started.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+`npx create-react-app my-app`
+
+## Installation
+To install the latest stable version of Jotai, run the following command:
+
+`npm install jotai` or `yarn add jotai`
+
+## Provider
```
dai-shi, line 17: Jotai Provider is optional, and for getting started, we should teach the provider-less mode.

docs/getting-started.mdx
```diff
@@ -0,0 +1,76 @@
...
+    <Input />
+    <CharCount />
+    <Uppercase />
+  </Provider>
+)
+
+export default App
+```
+
+We'll implement the components mentioned above in the following section.
```
dai-shi, line 37: We should mention what we are going to develop in the following.

docs/getting-started.mdx
```diff
@@ -0,0 +1,76 @@
...
+## atom
+
+There's an exported function called `atom` that comes with jotai, which represents a piece of state. Atoms can be read from and written to from any component.
+
+To create an atom, all you need is to pass an initial value.
+
+```jsx
+const textAtom = atom('hello')
+const textLenAtom = atom((get) => get(textAtom).length)
+const uppercaseAtom = atom((get) => get(textAtom).toUpperCase())
```
dai-shi, line 48: These look a repetitive example to what we have in Introduction.
So, I guess we are still not sure and open for what we should teach in getting started.

docs/getting-started.mdx
```diff
@@ -0,0 +1,76 @@
...
+const CharCount = () => {
+  const [len] = useAtom(textLenAtom)
+  return <div>Length: {len}</div>
+}
+
+const Uppercase = () => {
+  const [uppercase] = useAtom(uppercaseAtom)
+  return <div>Uppercase: {uppercase}</div>
+}
+```
```
dai-shi, line 76: At last, we would want some wrapping out notes, and how we should proceed the following sections in the docs.

dai-shi (commented):
Thanks for working on this. Getting started section is something missing, and we should add it. I added some comments, but most notably, the goal of this section needs to be clarified.

docs/getting-started.mdx
Sajad-Sharhani: I think CRA is more common for explaining the concept and we can use codesandbox at the bottom as a demo for readers to interact with

dai-shi:
Let's pin @sandren who might be able to give us some advises.

docs/getting-started.mdx
Sajad-Sharhani: I chose this example from the showcase because it's really straightforward and easy to understand how to implement Jotai, and for getting started we should aim for simplicity

docs/getting-started.mdx
Sajad-Sharhani: I was thinking about that too.
do u think I should remove the Provider but also mention that u can add it too?

docs/getting-started.mdx
Sajad-Sharhani: I didn't understand

docs/getting-started.mdx
dai-shi: I would definitely remove it from the getting started example. That's one of the major reasons I developed the provider-less mode.
We can mention about Provider, and it would be nice if we can mention "when" we should use Provider. (I wonder if it's in the scope of getting started, or out of it.)

docs/getting-started.mdx
dai-shi: I didn't understand where something is "mentioned above".

So, it would be nice to tell briefly to the readers about what we will be developing in the getting started section, before diving into the code.

docs/getting-started.mdx
dai-shi: Let's wait for @sandren 's advice.

sandren:
> Adding a new section to the docs tell me your opinion and suggestions

Thanks for opening this PR! 🙂

I think Daishi's recommendations are a good start. Let's remove the `<Provider>` explanation and revise the example code to something unique compared to the home page and introduction if you don't mind. Also if the guide uses CRA instead of CSB it should explain each step of using it in more detail in case the reader is unfamiliar. Notes at the end concerning suggested next steps to continue learning would be really helpful as well.

docs/getting-started.mdx
sandren: I think it's best to only introduce concepts as they became necessary. The provider is not necessary for the getting started example so I think that section should be removed. 🙂

dai-shi:
Hi @sandren ,
@Sajad-Sharhani is pretty motivated to work on docs, do you think if you can give him a small task? Maybe, create some examples? Or some drafts for doc parts?

dai-shi:
Closing this for now.

PR #934, opened 2022-01-06
fix(utils): support RESET for atomWithStorage and atomWithHash

### LINKED ISSUE #930: AtomWithStorage doesn't have a 'remove' possibility
I can't figure out how to remove an item from asyncstorage with the AtomwithStorage.

Would it be possible to either have a remove argument we could pass in, or if you pass in null it could call remove internally?

I store user's auth token in async storage, I want to clear it on log out - can I achieve that with AtomWithStorage?

dai-shi:
close #930

- [x] code
- [x] tests
- [x] docs


dai-shi:
@notsidney Could you make sure this doesn't break your atomWithHash usage?

notsidney:
Hi @dai-shi, sorry for the late reply. Looks like it doesn’t affect our atomWithHash usage.

PR #936, opened 2022-01-07
selectAtom: replace refAtom with reducer-on-read

justjake:
I noticed a `// TODO` while reading selectorAtom.ts. I've tried this pattern which exploits a feature where a derived atom can get it's own previous state, and inspect the previous state and new information to produce the next state.  I call this pattern "reducer on read", but maybe it's weird or confusing to give it a name.

Here's a more complex example of reducer-on-read style implementing proxy-tracked atom dependencies: https://github.com/justjake/monorepo/blob/81ae6a0b62adf34764e17ff9baeb5ee731940809/packages/state/src/lib/proxyCompareAtom.ts#L74-L96

The downside of this style, as far as I can tell, is that it uses exception handling for flow control for the initial read state in a way that might not be considered part of the stable API. In fact, perhaps this is considered "an invalid derived atom"? None the less, this is a useful style to avoid weird mutation stuff like refAtom.

If we can work out a way to resolve the issues with this reducer-on-read style, it could also replace the other usages of refAtom, like in splitAtoms.

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -28,6 +28,7 @@ type Revision = number
 type InvalidatedRevision = number
 type ReadDependencies = Map<AnyAtom, Revision>
 const Initial = Symbol('initial')
+type SuspendedPreviousValue<Value> = ResolveType<Value> | typeof Initial
 
 /**
  * Immutable atom state, tracked for both mounted and unmounted atoms in a store.
@@ -51,7 +52,7 @@ export type AtomState<Value = AnyAtomValue> = {
   d: ReadDependencies
 } & (
   | { e: ReadError }
-  | { p: SuspensePromise; pv?: ResolveType<Value> | typeof Initial }
+  | { p: SuspensePromise; pv?: SuspendedPreviousValue<Value> }
   | { v: ResolveType<Value> }
 )
 
@@ -393,7 +394,7 @@ export const createStore = (
   const atomPreviousValue = <Value>(
     atom: Atom<Value>,
     atomState: AtomState<Value> | undefined
-  ): { pv: ResolveType<Value> | typeof Initial } | undefined => {
+  ): { pv: SuspendedPreviousValue<Value> } | undefined => {
     if (atomState && 'pv' in atomState) {
       return { pv: atomState.pv }
     }
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -5,7 +5,6 @@ import {
   isEqualSuspensePromise,
   isSuspensePromise,
   isSuspensePromiseAlreadyCancelled,
-  isSuspensePromiseOwner,
 } from './suspensePromise'
 import type { SuspensePromise } from './suspensePromise'
 
@@ -27,8 +26,6 @@ type ReadError = unknown
 type Revision = number
 type InvalidatedRevision = number
 type ReadDependencies = Map<AnyAtom, Revision>
-const Initial = Symbol('initial')
-type SuspendedPreviousValue<Value> = ResolveType<Value> | typeof Initial
 
 /**
  * Immutable atom state, tracked for both mounted and unmounted atoms in a store.
@@ -50,11 +47,7 @@ export type AtomState<Value = AnyAtomValue> = {
    * of each dependency to that dependencies's current revision.
    */
   d: ReadDependencies
-} & (
-  | { e: ReadError }
-  | { p: SuspensePromise; pv?: SuspendedPreviousValue<Value> }
-  | { v: ResolveType<Value> }
-)
+} & ({ e: ReadError } | { p: SuspensePromise } | { v: ResolveType<Value> })
 
 /**
  * Represents a version of a store. A version contains a state for every atom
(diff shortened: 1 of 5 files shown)


dai-shi:
Wow, this is interesting!
This is indeed an invalid usage (try-catch isn't good), _as well as_ refAtom is not technically valid. However, your approach seems promising. Let me play around with your idea later. (Actually, I'm currently working on `splitAtom`.)

dai-shi:
I added a commit that is more acceptable, but it doesn't pass all the tests. 😕

justjake:
Won’t relying on onMount cause the derived atom to “drop” its memoized value when unmounted? It seems okay in practice, but it would be cleaner to offer a better guarantee. 

It would help Jotai to support the reducer-on-read style since it allows implementing interesting semantics like custom memoization in an atom - both selectAtom and my proxyCompareAtom. 

Would it be too much of a hack to make get(thisAtom) return undefined instead of throwing an error? Would that be a breaking change?

dai-shi:
> Won’t relying on onMount cause the derived atom to “drop” its memoized value when unmounted? It seems okay in practice, but it would be cleaner to offer a better guarantee.

That sounds fair. Tried 5b57863, but this still causes tests to fail.

> It would help Jotai to support the reducer-on-read style since it allows implementing interesting semantics like custom memoization in an atom - both selectAtom and my proxyCompareAtom.

For broader usage, `atom.onUpdate` seems a proper solution. But, I'm not planning to add it because it's likely people would misuse.

> Would it be too much of a hack to make get(thisAtom) return undefined instead of throwing an error? Would that be a breaking change?

As I said, I like the idea. Let me think about it. I don't want to make it a public API yet. Maybe an `unstable_` flag.

justjake:
I like the new idea of combining `init` with a derived fn — definitely more satisfying and much faster than try/catch, feels more stable as well.

dai-shi:
> As I said, I like the idea. Let me think about it. I don't want to make it a public API yet. Maybe an `unstable_` flag.

I tried adding the flag locally, but this results in the same error. The problem is somewhere else.



> I like the new idea of combining `init` with a derived fn — definitely more satisfying and much faster than try/catch, feels more stable as well.

Okay, so let's go with it. Only the remaining issue for us is the failing tests. There must be a bug somewhere, can be in core. Would you please give it a look when you have a chance?

justjake:
Sure, I'll look.

justjake:
@dai-shi the issue is in core, and relates only to derived atoms that read themselves. The issue is that once the derived atom becomes suspended, whenever it reads itself, it throws the suspended promise again right away. See the latest commit (https://github.com/pmndrs/jotai/pull/936/commits/f1bcf5344531557d6d9899e451d8c22bae061b2d) for an illustration of the issue.

I haven't figured out a *good* fix. It seems like we could add an exception so that each suspense promise is tagged with an "owner" atom - the atom that the suspended promise will write to when resoled - and then if an atom reads itself, that read returns a suspended promise, and the suspended promise is owned by another atom, we instead return either the initial value (illogical, but easy to implement) or the previous value (logical, but requires more extensive changes). The second behavior makes sense to me, and can be explained as "the derived promise computation is suspended until all dependencies resolve, then proceeds with the previous value".

justjake:
Sorry, I rebased by accident.

justjake:
Hmm, retaining `'v'` while suspended seems like a bad idea since so much logic reasons about `'v' in state`. I'll move the previous value to a new property.

justjake:
Looks like somehow, this new logic breaks for an experimental version of React. I would appreciate your review of this idea before I spend more time debugging.

dai-shi:
Thanks for investigating. The CI error was the timing issue (btw, I want someone to help this. Hoped to use fake timers, but couldn't make it work.)

To be honest, I don't prefer the last idea. The elegance of your original idea was, we actually already had the previous value. The new one comes with too much burden. I don't think it's worth the complexity and the bundle size increase. The try-catch hack might seem better. (Now, I understand why try-catch works, when we throw, we don't store `dependencies`. By returning undefined, we update `dependencies`.)

On the other hand, if we have a real issue in core not related to this PR, we need to investigate it further. What do you think? Can the core issue persist without the reducer-on-read style? (We used to have the previous value in atom state while suspending, so I can consider reverting it if that is really necessary.)

dai-shi:
Sorry, reverted to what we had in 5b57863aec53910e9337ed3e40bebb42c2e165e8.

And, added the fix. Hope it's reasonable.

justjake:
I don't think the infinite suspense should be considered a bug since we already decide that a derived atom that reads itself is invalid, and it's impossible to create one without private API. I think the solution to store the zero state when suspended could be e a regression - selectors on async atoms with custom equality won't be stable anymore. Maybe ref atom remains the best we can do?

dai-shi:
Oh, so even the original idea has a regression.
Let me add some test cases for it, and then come back to this PR.
Please let me know if adding many commits to this PR by me is not comfortable, then I will create a new PR. We need a drastic change anyway.

dai-shi:
> We used to have the previous value in atom state while suspending, so I can consider reverting it if that is really necessary.

This would still be controversial. Unless React recommends it, I think clearing value while suspending makes sense, because suspense can last forever, and we want to avoid memory leaks. We may change this in the future, but not now.

> Maybe ref atom remains the best we can do?

Yeah, let's keep it for now. It's still not a recommended pattern outside the repo, because with versionedWrite, it will technically break. (A canonical solution would be some kind of versioned cache, but that'd be too complicated. Oh, yeah, another canonical solution is `onUpdate`, but not now.)

Closing this. Thanks so much for your suggestion and investigation. At least we learned a lot.

Issue #931, opened 2022-01-05
Improve debugging info in useAtomsDevtools


justjake:
The current `useAtomsDevtools` is a great start for debugging tools. I've been playing with it a bit and want to suggest a few features:

1. Better atom labeling. Currently, atoms are labeled sequentially as `atom1`, `atom2`, etc. Here's some ideas:
    - Automatically label atoms by capturing a stack trace during atom construction (in dev mode only), and adding a partial formatted stack trace to `atom.toString()`'s output. I've experimented a little with this, for example an atom I create as `const testAtom = implicitAtom(...)` has this `toString`: `WritableImplicitAtom 5 (created at Object.<anonymous> (/Users/jitl/src/jitl-monorepo/packages/state/src/lib/implicitAtom.test.ts:108:32))` 
    - A less general option is to encourage setting `debugName` or `toString` by adding a new optional final argument to `atom()` as `atom(read, write, options?: { debugName?: string, toString?: () => string })`; this would allow a nicer way to write atoms with labels in one line - `atom(initialValue, { debugName: 'foo atom' })` is easier to remember than `atom(initialValue); atom.debugName = 'foo atom'`.

2. Better action/mutation/change labeling. Currently, the only info about a change is when it occurred.
    Ideally, we could expose the following information about each new snapshot on the fake Redux action:
    - The place where the `setAtomState` function came from, eg the stack trace of the `useAtom` or `useSetAtom` hook that produced it. This allows the user to easily locate the React component that's updating Jotai's state.
    - The writable atoms (and/or their atom.toString()) involved in a state change. This could be complicated to gather for a derived writable atom that "fans out" by writing many other atoms, but we could at least capture the debug info of the top-level writable atom that started the write. That atom's identity is very analogous to the concept of "action type" in Redux.

3. Advanced: show *why* a change caused *what* React components to re-render. I call this “causality debugging”. This helps developers detect O(n) re-rendering due to change fan-out, which is a performance problem that can be very difficult to debug in large apps. This is the most complex idea, but I know it’s possible since I implemented this in my own framework. This would build on the stack trace capture work in 1 & 2. Here’s how it works:
    - Whenever derived atom A reads a dependency atom D, save the stack trace of that `get(D)` call alongside the revision number in the dependency map (or in a devtools only location). 
    - Whenever a writer atom W function writes an atom WD, save the stack trace of the `set(WD, …)` into a Map<W, [WD, Stack]> representing the action, stored in the devtools action log.
[...]

PR #943, opened 2022-01-09
useAtomsDevtools: Show all atoms updated by a write

justjake:
Draft PR to address one idea in https://github.com/pmndrs/jotai/issues/931.

The idea of this PR is to records all the atoms changed during an atom write into an event, and then uses that event data to enhance redux devtools.

I think I need to propagate the write event through all codepaths that call flushPending.

Here's a screenshot of the goal:

image

for this example component:

image

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -919,6 +919,11 @@ export const createStore = (
   }
 }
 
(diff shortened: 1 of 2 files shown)


dai-shi:
Closing this for now. Let's continue discussion in #931.

PR #944, opened 2022-01-09
Add "yarn watch" script to package.json

justjake:
This improves developer experience for eg iterating on devtools.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
   "scripts": {
     "prebuild": "shx rm -rf dist",
     "build": "concurrently 'yarn:build:*'",
-    "watch": "concurrently -n watch: 'yarn:build:* --watch'",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -109,6 +109,7 @@
   "scripts": {
     "prebuild": "shx rm -rf dist",
     "build": "concurrently 'yarn:build:*'",
+    "watch": "concurrently -n watch: 'yarn:build:* --watch'",
```
dai-shi, line 112: could this be `watch-build` or `build-watch`?

justjake: Both sound fine to me.

dai-shi: Let's make it `build-watch` then:
```suggestion
    "build-watch": "concurrently -n watch: 'yarn:build:* --watch'",
```

Issue #947, opened 2022-01-10
[Regression Bug] Updated atom value not propagated to component


xbaun:
I am using an atom (A) to control if an atom (B) should return the value of an atom (C) otherwise a default value is returned.  See the example below: https://codesandbox.io/s/dark-http-bsili?file=/src/App.tsx

```ts
const aAtom = atom<any>(false);
aAtom.debugLabel = "aAtom";
aAtom.onMount = (set) => {
    console.log("onMount: aAtom");
    setTimeout(() => {
        set(true);
    }, 1000);
};

const cAtom = atom<any>(1);
cAtom.debugLabel = "cAtom";
cAtom.onMount = (set) => {
    console.log("onMount: cAtom");
    set(2);
};

const bAtom = atom<any>((get) => {
    const a = get(aAtom);
    console.log("a=", a);
    if (a) {
        return get(cAtom);
    }
    return 1;
});
bAtom.debugLabel = "bAtom";

```

The problem is, that the changed value of atom (C) is never propagated to the invoking component through `useAtom`. The error occurs when there is no externally triggered rendering within the calling component, which re-evaluates the `useAtom` hook. I think, there is somewhere a `flushPending` missing. I tried to figure out what could be a possible fix, but I did not found any good solution. 

Affects version `1.5.0`.  Worked with ~~`1.4.2`~~ `1.4.9`.

----

BTW:  I think the `useEffect` call at https://github.com/pmndrs/jotai/blob/main/src/core/useAtom.ts#L103-L105 should pass its dependencies. If so, a forced rendering has no effect anymore on the bug above.

```ts
  useEffect(() => {
    store[COMMIT_ATOM](atom, version)
  }, [store, atom, version]);
```

dai-shi:
Thanks for reporting. I think we miss tests to cover this...

> Affects version `1.5.0`. Worked with `1.4.2`.

This almost indicates, this is a bug. Note: it works with `1.4.9` too.

xbaun:
I updated the version in the bug description.

smashercosmo:
Hm... I think I also stumbled upon this bug, but for me the last working version is 1.4.6

dai-shi:
Are you sure? That sounds like a different bug. Thanks for reporting anyway.
I'm currently working hard on issue.

smashercosmo:
[...]

PR #950, opened 2022-01-11
fix(useAtom): issue-947 propagate atom revision changes correctly

xbaun:
Track revision changes in `useAtom` to ensure that the latest atom value is returned.

Issue #947

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -295,9 +295,9 @@ export const createStore = (
     version: VersionObject | undefined,
     prevReadDependencies: ReadDependencies = new Map(),
     dependencies?: Set<AnyAtom>
-  ): ReadDependencies => {
+  ): { d: ReadDependencies; changed: boolean } => {
     if (!dependencies) {
-      return prevReadDependencies
+      return { d: prevReadDependencies, changed: false }
     }
     const readDependencies: ReadDependencies = new Map()
     let changed = false
@@ -309,9 +309,9 @@ export const createStore = (
       }
     })
     if (prevReadDependencies.size === readDependencies.size && !changed) {
-      return prevReadDependencies
+      return { d: prevReadDependencies, changed }
     }
-    return readDependencies
+    return { d: readDependencies, changed }
   }
 
   const setAtomValue = <Value>(
@@ -335,15 +335,23 @@ export const createStore = (
         cancelSuspensePromise(atomState.p)
       }
     }
+
+    const { d, changed: dependenciesChanged } = createReadDependencies(
+      version,
+      atomState?.d,
+      dependencies
+    )
+
     const nextAtomState: AtomState<Value> = {
       v: value,
       r: atomState?.r || 0,
-      d: createReadDependencies(version, atomState?.d, dependencies),
+      d,
     }
     if (
       !atomState ||
       !('v' in atomState) || // new value, or
-      !Object.is(atomState.v, value) // different value
+      !Object.is(atomState.v, value) || // different value,
+      dependenciesChanged
     ) {
       ++nextAtomState.r // increment revision
       if (nextAtomState.d.has(atom)) {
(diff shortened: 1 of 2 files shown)


src/core/useAtom.ts
```diff
@@ -54,39 +54,49 @@ export function useAtom<Value, Update, Result extends void | Promise<void>>(
...
+        VersionObject | undefined
+      >,
+      undefined
+    >(
+      useCallback(
+        (prev, nextVersion) => {
+          const [nextValue, nextRevision] = getAtomValue(nextVersion)
+          if (
+            Object.is(prev[1], nextValue) &&
+            Object.is(prev[2], nextRevision) &&
```
dai-shi, line 83: This technically means, when we have the same value, but a different revision, this hook triggers _unnecessary_ re-render. We should avoid it and should make a test case to catch it if possible. (guessing, it may or may not happen with suspense.)

dai-shi (changes requested):
Oh, you solved it? Nice!
You might feel it too, but I'm not super happy with this ceremony in useAtom.ts.
I feel like we should solve this in store.ts.

Would you like to continue investigating it? We need a test case for this too.

src/core/useAtom.ts
xbaun: > This technically means, when we have the same value, but a different revision ...

That's true, but was exactly my case where the problem emerged: An atom that conditionally read another atom which should invoke `onMount` and start a subscription, but had initially the same default value.

> ... this hook triggers unnecessary re-render

 Because the same value was returned as in the previous render cycle, the effect was not getting re-rendered, therefore `commitAtom` not fired and `flushPending` not invoked.

With the current implementation the effect firing `commitAtom` should be a good place to call it.
I guess, as an alternative it could be also fired directly in the `rerenderIfChanged` dispatcher when the revision changed. This would save a render cycle.

```ts
      useCallback(
        (prev, nextVersion) => {
          const [nextValue, nextRevision] = getAtomValue(nextVersion)

          if (
            Object.is(prev[1], nextValue) &&
            Object.is(prev[2], nextRevision) &&
            prev[3] === atom
          ) {
            return prev // bail out
          }

          if (!Object.is(prev[2], nextRevision)) {
             setTimeout(() => store[COMMIT_ATOM](atom, version))
          }
          
          return [nextVersion, nextValue, nextRevision, atom]
        },
        [getAtomValue, atom]
      ),
```
https://github.com/pmndrs/jotai/blob/84a0d685b2f774b684673f488897a9f525e56e3d/src/core/useAtom.ts#L78-L91

dai-shi: Please forgive me, I didn't looked your issue #947 closely yet. You filed this PR so quickly. 😅 

I think if we suspend and resolve with the same state, not useEffect will fire. Or maybe not?

Can you do some experiments without jotai only with React useState, and see if useEffect fires with the said situation?

In any case, this can be an issue on jotai end. If useEffect doesn't fire, and we don't want to do side effects in dispatch phase, so the only solution would be requiring `Provider` to fire effects. Okay, I feel like I need to sit down and take some time to think about it. If you can create some test cases, that would be helpful.

dai-shi: Wait, this is not related to Suspense? My misunderstanding...

src/core/store.ts
```diff
@@ -335,15 +335,23 @@ export const createStore = (
...
       r: atomState?.r || 0,
-      d: createReadDependencies(version, atomState?.d, dependencies),
+      d,
     }
     if (
       !atomState ||
       !('v' in atomState) || // new value, or
-      !Object.is(atomState.v, value) // different value
+      !Object.is(atomState.v, value) || // different value,
+      dependenciesChanged
```
xbaun, line 354: Updates the revision if the dependencies have changed. That's necessary to track when the dependent atom has the same initial value as the accessed one.

dai-shi:
I ended up with another hacky way in #960. But, it feels comfortable without modifying useAtom.ts and without changing the semantics of revision.

closing this. this pr helped to understand the issue.

PR #952, opened 2022-01-11
fix(babel): Require that atom is in module scope for react refresh

### LINKED ISSUE #891: React Refresh Babel plugin causes unexpected behaviour in atoms returned from functions
I was pulling my hair out when trying to get `atomWithStorage` to work (with Expo's SecureStore in a React Native context).
The issue I was having was that no matter which atom I updated, all `atomWithStorage` atoms would receive the same value – that of the first atom created.
The occurrence started at a (seemingly) random time at first, but after creating a completely blank project I was able to boil it down to being the fault of the React Refresh Babel plugin (which of course needed the Metro cache to be cleared before jumping into effect, which was why it popped up at a random time – after a reboot of my laptop 😵‍💫).

Disabling `plugin-react-refresh` or `jotai/babel/preset` in the babel config and then restarting Metro bundler with `--reset-cache` fixes the issue.

[Reproducible example (using plugin-react-refresh)](https://codesandbox.io/s/react-with-custom-babel-config-forked-311h6?file=/src/App.js)

[Working example (using plugin-debug-label)](
https://codesandbox.io/s/react-with-custom-babel-config-forked-28cbf?file=/src/App.js)

Click the buttons in the examples above to demonstrate the behavior and look at the console to see the output of `useAtomsSnapshot()`.

In essence, updating the state of any atom that is returned from a function (countAtom2 or countAtom3 in the example) will result in both receiving the same state (and debugLabel).

Thisen:
Fixes #891.

For simplicity, we require that the atom is in module scope for the cache to work. If we wanted to support the case in #891, we need need to depend on the call order (like hooks), but I think that's too complex for current use cases. If it gets requested, we can investigate what we can do.

### DIFF
--- a/src/babel/plugin-react-refresh.ts
+++ b/src/babel/plugin-react-refresh.ts
@@ -48,13 +48,16 @@ export default function reactRefreshPlugin({
         }
       },
       VariableDeclarator(nodePath, state) {
+        console.log(nodePath.parentPath.parentPath?.type)
         if (
           t.isIdentifier(nodePath.node.id) &&
           t.isCallExpression(nodePath.node.init) &&
           isAtom(t, nodePath.node.init.callee) &&
           // Make sure atom declaration is in module scope
-          nodePath.parentPath.parentPath?.isProgram()
+          (nodePath.parentPath.parentPath?.isProgram() ||
+            nodePath.parentPath.parentPath?.isExportNamedDeclaration())
         ) {
+          console.log('yay!')
           const filename = state.filename || 'unknown'
           const atomKey = `${filename}/${nodePath.node.id.name}`
 
--- a/src/babel/plugin-react-refresh.ts
+++ b/src/babel/plugin-react-refresh.ts
@@ -48,7 +48,6 @@ export default function reactRefreshPlugin({
         }
       },
       VariableDeclarator(nodePath, state) {
-        console.log(nodePath.parentPath.parentPath?.type)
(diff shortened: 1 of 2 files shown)


src/babel/plugin-react-refresh.ts
```diff
@@ -48,11 +48,16 @@ export default function reactRefreshPlugin({
         }
       },
       VariableDeclarator(nodePath, state) {
+        console.log(nodePath.parentPath.parentPath?.type)
```
dai-shi, line 51: remove this?
```suggestion
```

src/babel/plugin-react-refresh.ts
```diff
@@ -48,11 +48,16 @@ export default function reactRefreshPlugin({
...
         if (
           t.isIdentifier(nodePath.node.id) &&
           t.isCallExpression(nodePath.node.init) &&
-          isAtom(t, nodePath.node.init.callee)
+          isAtom(t, nodePath.node.init.callee) &&
+          // Make sure atom declaration is in module scope
+          (nodePath.parentPath.parentPath?.isProgram() ||
+            nodePath.parentPath.parentPath?.isExportNamedDeclaration())
         ) {
+          console.log('yay!')
```
dai-shi, line 60: yay!
```suggestion
```

dai-shi (approved):
Sounds okay to me.

PR #958, opened 2022-01-12
feat(docs): debugging

Aslemammad:


dai-shi (approved):
LGTM

Some notes:
- Should we mention about babel plugin for `debugLabel`? At least, a link would be nice.
- Should we clarify that `Provider` is required for React Devtools? Provider-less mode can't be used.

dai-shi:
Let's merge this.
@sandren If you have some comments, feel free to leave it here or open a discussion.

PR #960, opened 2022-01-13
fix(core): handle an edge case of changing dependencies without changing atom value

### LINKED ISSUE #947: [Regression Bug] Updated atom value not propagated to component

xbaun:
fix #947 

(edit by @dai-shi) this is done by a hacky way. wish to have a nicer way.

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -349,6 +349,17 @@ export const createStore = (
       if (nextAtomState.d.has(atom)) {
         nextAtomState.d = new Map(nextAtomState.d).set(atom, nextAtomState.r)
       }
+    } else if (
+      nextAtomState.d !== atomState.d &&
+      (nextAtomState.d.size !== atomState.d.size ||
+        !Array.from(nextAtomState.d.keys()).every((a) => atomState.d.has(a)))
+    ) {
+      // value is not changed, but dependencies are changed
+      // we should schdule a flush in async
+      // FIXME any better way? https://github.com/pmndrs/jotai/issues/947
+      Promise.resolve().then(() => {
+        flushPending(version)
+      })
     }
     setAtomState(version, atom, nextAtomState)
     return nextAtomState
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -479,8 +479,9 @@ export const createStore = (
       // See if we can skip recomputing this atom.
       const atomState = getAtomState(version, atom)
       if (atomState) {
-        // First, if we have suspending promise
+        // First, check if we already have suspending promise
         if (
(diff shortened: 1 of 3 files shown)


src/core/store.ts
```diff
@@ -423,13 +422,12 @@ export const createStore = (
...
-              if (
-                isSuspensePromise(e) &&
-                isSuspensePromiseAlreadyCancelled(e)
-              ) {
+              if (isSuspensePromise(e)) {
                 // schedule another read later
                 // FIXME not 100% confident with this code
-                e.then(() => readAtomState(version, atom, true))
+                e.then(() => {
+                  readAtomState(version, atom, true)
```
Aslemammad, line 429: What's the difference here? Like readAtomState won't pass the returned value to the next `then`?

dai-shi: Sorry for confusion. This doesn't mean anything special. Just making it clear that it returns void.

dai-shi:
Okay, seems like I was solving a different/general problem. Let's separate that fix in a separate PR and come back to this.

PR #966, opened 2022-01-15
docs(integrations): atomWithImmer: countAtom var

### LINKED ISSUE #962: [docs] wrong variable used in doc, https://jotai.org/docs/integrations/immer#atom-with-immer
It seems that, the `setCountAtom` should be corrected to `countAtom`,  or line 27 should add something else ? I am not sure.

Doc site: https://jotai.org/docs/integrations/immer#atom-with-immer 
Source:  https://github.com/pmndrs/jotai/blame/49c7b813cceac5eab4d208f7f710c598bc834bd9/docs/integrations/immer.mdx#L34

lgh06:
fix #962

### DIFF
--- a/docs/integrations/immer.mdx
+++ b/docs/integrations/immer.mdx
@@ -31,7 +31,7 @@ const Counter = () => {
 }
 
 const Controls = () => {
-  const [, setCount] = useAtom(setCountAtom)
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Thanks for the fix!

PR #968, opened 2022-01-15
feat(build): use __DEV__ to control development mode

dai-shi:
- [x] rollup config
- [x] replace code with `__DEV__`
- [x] test config
- [x] fix failing tests
- [x] fix codesandbox ci
- [x] test matrix for NODE_ENV in CI


src/core/store.ts
```diff
@@ -804,17 +786,19 @@ export const createStore = (
...
-    pending.forEach(([atom, prevAtomState]) => {
-      const atomState = getAtomState(undefined, atom)
-      if (atomState && atomState.d !== prevAtomState?.d) {
-        mountDependencies(atom, atomState, prevAtomState?.d)
-      }
-      const mounted = mountedMap.get(atom)
-      mounted?.l.forEach((listener) => listener())
-    })
-    if (typeof process === 'object' && process.env.NODE_ENV !== 'production') {
+    while (pendingMap.size) {
```
justjake, line 789: This looks like a good correctness improvement possibly - my state management framework has the same queue draining logic. But, is it needed for the build changes?

tests/devtools/useAtomDevtools.test.tsx
```diff
@@ -98,22 +96,8 @@ describe('If there is no extension installed...', () => {
...
-    const consoleWarn = jest.spyOn(console, 'warn')
-
-    render(
-      <Provider>
-        <Counter />
-      </Provider>
-    )
-    expect(consoleWarn).not.toBeCalled()
-
-    consoleWarn.mockRestore()
```
justjake, line 115: Thoughts: it could be worth setting up a test matrix variant to test with the different build outputs. Or at least running jest with `__DEV__=false` as well — we should cover production behavior!

package.json
```diff
@@ -177,6 +177,9 @@
         "@swc/jest"
       ]
     },
+    "globals": {
+      "__DEV__": true
```
justjake, line 181: Does jest read this package.json attribute?

justjake (commented):
Changes look good but I’m surprised about the build failure.

But, new memory leak!? Could it be the new pendingMap dequeue loop?

I’m most curious to see how we get the right build selected by consumers’ bundling process. I’ll have to check my installed react packages to see how their build output works with NPM and webpack.

rollup.config.js
```diff
@@ -21,9 +23,9 @@ function getBabelOptions(targets) {
   }
 }
 
-function getEsbuild(target) {
+function getEsbuild(target, env = 'development') {
```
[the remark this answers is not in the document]
dai-shi: It's for browser. L124.

src/core/store.ts
dai-shi: Oooooooops, I was accidentally including my wip code in #950. That's why tests failed. Will revert them.

tests/devtools/useAtomDevtools.test.tsx
dai-shi: Let me see if we can setup the test matrix in the CI.

package.json
dai-shi: If we remove this, we see lots of `ReferenceError: __DEV__ is not defined` errors.

dai-shi:
> Changes look good but I’m surprised about the build failure.
> 
> But, new memory leak!? Could it be the new pendingMap dequeue loop?

Sorry, it was my mistake. I mixed my wip code in #950.
It was such a mess, so force pushed.

But, it still fails. How come...

> I’m most curious to see how we get the right build selected by consumers’ bundling process. I’ll have to check my installed react packages to see how their build output works with NPM and webpack.

If you use bundlers, it won't change much. `__DEV__` is simply replaced and we don't run dead code elimination here. (It would be nice if we can, but I don't know how to do it for ESM.)

---
In summary:

### CJS for webpack & create-react-app (v4/v5) & nodejs

`__DEV__` is replaced with `process.env.NODE_ENV!=="production"`

### ESM for vite & nodejs

`__DEV__` is replaced with `import.meta.env!=="production"`

You would need to do `import.meta.env='production'` in nodejs. (Does it work?)

### UMD for browser

- *.development.js with `__DEV__=true`
- *.production.js with `__DEV__=false` and minification

### SystemJS for browser

- *.development.js with `__DEV__=true`
- *.production.js with `__DEV__=false` and minification

.eslintrc.json
```diff
@@ -5,6 +5,9 @@
     "node": true,
     "es6": true
   },
+  "globals": {
+    "__DEV__": "readonly"
```
dai-shi, line 9: I hoped this fixes GitHub Actions / lint errors. But, no? Why?

dai-shi: it was my bad.

justjake:
We do run dead code elimination (terser) for systemJS and UND, right? In those cases you are inlining a static true or false. I don’t use those build types, but it should give consumers a reasonable idea of the bundle size of Jotai in production.

tests/devtools/useAtomsDevtools.test.tsx
```diff
@@ -115,26 +118,8 @@ describe('If there is no extension installed...', () => {
       'Please install/enable Redux devtools extension'
     )
 
-    process.env.NODE_ENV = originalNodeEnv
     console.warn = originalConsoleWarn
   })
-
-  it('does not warn if not in dev env', () => {
```
dai-shi, line 122: We can't use this hook in non-DEV env anyway. So, let's remove this case.

dai-shi:
Hmm, how come codesandbox builds fail?

dai-shi:
Seems like it was a temporary failure.

Aslemammad (approved):

PR #980, opened 2022-01-19
Add documentation for react-native

justinhandley:
Based on recent conversation I've written out documentation on the functions I used to get AtomWithStorage working effectively on React Native.

### DIFF
--- a/docs/guides/react-native.mdx
+++ b/docs/guides/react-native.mdx
@@ -1,0 +1,49 @@
+---
+title: React Native
+description: Using Jotai in React Native
+nav: 3.5
+---
+
+Jotai atoms can be used in React Native applications with very few changes except for the use of persistence.
+
+Jotai has an [atomWithStorage function in the utils bundle](../api/utils.mdx#atom-with-storage) for persistance that supports persisting state in `sessionStorage`, `localStorage`, `AsyncStorage`, or the URL hash.
+
+Because react-native only supports AsyncStorage, you need to tell Jotai to use AsyncStorage instead of the default.
+
+Then, depending on what you are storing, you may need to take additional steps.  By default, AsyncStorage stores strings.  So, if you are simply storing a string, just passing AsyncStorage in is enough.
+
+## A simple pattern with AsyncStorage
+
+```js
+export const strAtom = atomWithStorage('string', undefined, AsyncStorage)
+```
+
+If you want to store objects and booleans, however, there are a series of helper functions that make this easier.
+
+## Storing non-string items in AsyncStorage
+
+First, create a storage item using the createJSONStorage function.  This automatically stringifies and parses objects before writing to AsyncStorage.
+
+```js
+export const storage = createJSONStorage(() => AsyncStorage)
+```
+
+Now you can pass an object into the store just by passing storage as the storage type.
+
+```js
+export const storage = createJSONStorage(() => AsyncStorage)
+export const objectAtom = atomWithStorage('object', undefined, storage)
+```
+
+## Storing booleans in AsyncStorage
+
+To store a boolean, there is an additional helper function atomWithToggleAndStorage.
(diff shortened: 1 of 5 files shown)


docs/guides/react-native.mdx
```diff
@@ -0,0 +1,49 @@
...
+Jotai has an [atomWithStorage function in the utils bundle](../api/utils.mdx#atom-with-storage) for persistance that supports persisting state in `sessionStorage`, `localStorage`, `AsyncStorage`, or the URL hash.
+
+Because react-native only supports AsyncStorage, you need to tell Jotai to use AsyncStorage instead of the default.
+
+Then, depending on what you are storing, you may need to take additional steps.  By default, AsyncStorage stores strings.  So, if you are simply storing a string, just passing AsyncStorage in is enough.
+
+## A simple pattern with AsyncStorage
+
+```js
+export const strAtom = atomWithStorage('string', undefined, AsyncStorage)
```
dai-shi, line 18: Ah, I never knew you can do this. We don't do this with `localStorage` because that will break SSR. But, RN doesn't have SSR, it might be okay.

docs/guides/react-native.mdx
```diff
@@ -0,0 +1,49 @@
...
+```
+
+If you want to store objects and booleans, however, there are a series of helper functions that make this easier.
+
+## Storing non-string items in AsyncStorage
+
+First, create a storage item using the createJSONStorage function.  This automatically stringifies and parses objects before writing to AsyncStorage.
+
+```js
+export const storage = createJSONStorage(() => AsyncStorage)
```
dai-shi, line 28: Do we need to export it?
```suggestion
const storage = createJSONStorage(() => AsyncStorage)
```

docs/guides/react-native.mdx
```diff
@@ -0,0 +1,49 @@
...
+```
+
+Now you can pass an object into the store just by passing storage as the storage type.
+
+```js
+export const storage = createJSONStorage(() => AsyncStorage)
```
dai-shi, line 34:
```suggestion
const storage = createJSONStorage(() => AsyncStorage)
```

docs/guides/react-native.mdx
```diff
@@ -0,0 +1,49 @@
...
+export const objectAtom = atomWithStorage('object', undefined, storage)
+```
+
+## Storing booleans in AsyncStorage
+
+To store a boolean, there is an additional helper function atomWithToggleAndStorage.
+
+```js
+export const storage = createJSONStorage(() => AsyncStorage)
```
dai-shi, line 43:
```suggestion
const storage = createJSONStorage(() => AsyncStorage)
```

docs/guides/react-native.mdx
```diff
@@ -0,0 +1,49 @@
...
+export const booleanAtom = atomWithToggleAndStorage('boolean', false, storage)
```
dai-shi, line 44: This atom helper is only described in an advanced recipe. I'm a little bit concerned to have it here. At least we need a link to the recipe.

dai-shi (commented):
Thanks for the new guide! Very nice to see docs improving.
Added some minor comments.

docs/guides/react-native.mdx
justinhandley: I have removed exports throughout

docs/guides/react-native.mdx
justinhandley: I have left this in for now but have added a link to the recipe.  I added some text throughout so you may want to review the new commit again, and I've updated to meet your new numbering.

dai-shi (approved):
Thanks for your work!
Let's merge this. We may want to tweak something or the entire guide section in the future, but it looks good as of now.
Nice contribution. We need more contributors on docs!

PR #981, opened 2022-01-21
fix(query): Add QueryKey types to atomWithQuery

Pinpickle:
Type inference of the query key in `atomWithQuery` wasn't working correctly, so usage like this:

```ts
atomWithQuery(() => ({
  queryKey: ['pages', { page: 2 }] as const,
  queryFn: async ({ queryKey: [, { page }] }) => {
    return (await fetch(`pages/${page}`)).json()
  },
}))
```

Would result in a type error.

This adds the correct types by utilising react-query's fifth type argument, `TQueryKey`.

### DIFF
--- a/src/query/atomWithInfiniteQuery.ts
+++ b/src/query/atomWithInfiniteQuery.ts
@@ -24,18 +24,33 @@ export type AtomWithInfiniteQueryOptions<
   TQueryFnData,
   TError,
   TData,
-  TQueryData
-> = InfiniteQueryObserverOptions<TQueryFnData, TError, TData, TQueryData> & {
-  queryKey: QueryKey
-}
+  TQueryData,
+  TQueryKey extends QueryKey
+> = Omit<
+  InfiniteQueryObserverOptions<
+    TQueryFnData,
+    TError,
+    TData,
+    TQueryData,
+    TQueryKey
+  >,
+  'queryKey'
+> & { queryKey: TQueryKey }
 
 export type AtomWithInfiniteQueryOptionsWithEnabled<
   TQueryFnData,
   TError,
   TData,
-  TQueryData
+  TQueryData,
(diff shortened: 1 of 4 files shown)


dai-shi (approved):
Thanks. It looks good to me. Wow, this is fairly complicated.

aulneau (approved):
wow this is great! thanks so much, approved!

Pinpickle:
I just discovered a weird typing bug which happens when you add the "get" argument to the getter function. I've changed the tests to fail here, but I haven't found a fix yet.

Please hold off merging until I work it out!

Pinpickle:
I _think_ I've uncovered a TS bug. Tracking here: https://github.com/microsoft/TypeScript/issues/47547

Pinpickle:
@dai-shi it looks like a fix for this is landing in TS 4.7: https://github.com/microsoft/TypeScript/pull/48538

I think older versions of TS will continue working with `atomWithQuery` as they do now, even with these new types.

dai-shi:
@Pinpickle You can try 4.7 beta in this draft PR. Once 4.7 is released, please change it to 4.7 and mark this "ready for review"

Pinpickle:
@dai-shi @TwistedMinda thanks for the reminder!

I rebased the branch, which updated to TS 4.7.4, which caused a weird bug that I can't quite get to the bottom of.

The type of `TQueryKey` is now being correctly inferred:

image

Even `queryKey` is being typed correctly before renaming

image

However, the variable itself is being marked as `unknown[] | [string]`:

image

This smells like another TS bug (which may have yet been fixed), but I'm not sure.

I don't know when I'll get a chance to have a proper look at this. If someone else wants to try, be my guest!

dai-shi:
I think this is covered by #1126 in v1.7.7.

PR #989, opened 2022-01-29
feat(core): Move useAtomValue and useSetAtom to core

Thisen:
This change renames `useUpdateAtom` to `useSetAtom` and moves `useAtomValue` and `useSetAtom` to core.
The hooks are still exported from `jotai/utils` with original names for backwards compatibility.

- [X] Impl
- [X] Tests


Thisen:
@dai-shi @sandren I have yet to change the docs, because I'm unsure how we want to tackle it. 
How should leave it for another PR?

If not, where do we want to place the sections about `useAtomValue` and `useSetAtom` inside the docs? There's no obvious place for it. 

I think we should leave a section in `utils` about them, that says we're deprecated and should be imported from `jotai`.

Let me know your toughts.

src/utils/useAtomValue.ts
```diff
@@ -1,7 +1 @@
-import { useAtom } from 'jotai'
-import type { Atom } from 'jotai'
-import type { Scope } from '../core/atom'
-
-export function useAtomValue<Value>(anAtom: Atom<Value>, scope?: Scope) {
-  return useAtom(anAtom, scope)[0]
-}
+export { useAtomValue } from 'jotai'
```
dai-shi, line 1: We can do this in `src/utils.ts`. What do you think?

src/core/useAtomValue.ts
```diff
@@ -0,0 +1,7 @@
+import { useAtom } from 'jotai'
+import type { Atom } from 'jotai'
+import type { Scope } from '../core/atom'
+
+export function useAtomValue<Value>(anAtom: Atom<Value>, scope?: Scope) {
+  return useAtom(anAtom, scope)[0]
```
dai-shi, line 6: Let's move the implementation here.
In the end, our `useAtom` will be something like this (written without types to explain simply):
```js
export function useAtom(anAtom, scope) {
  return [useAtomValue(anAtom, scope), useSetAtom(anAtom, scope)]
}
```
The whole point of this change is to avoid duplicated code.

dai-shi (commented):
Cool. Added some comments. Surprised with the number of files changed, and found they are tests. 😅

dai-shi:
> @dai-shi @sandren I have yet to change the docs, because I'm unsure how we want to tackle it. How should leave it for another PR?

It can be another PR, but we would like to do it soon. I'm not merging this PR soon, so until then we can work on docs in this PR.

So, my idea is to modify this page https://jotai.org/docs/basics/primitives.
Split into two sections and describe the core APIs:

>    Basic APIs: atom, useAtom
>    Additional APIs: Provider, useAtomValue, useSetAtom

It's important to show that first thing to learn is atom/useAtom, and only reach to the additional APIs, when necessary.

Also, modify the API page: https://jotai.org/docs/api/core
Maybe, for additional APIs, we don't describe details in the Primitives page, but just an introduction and link to the Core API page.

As for Utils, we don't deprecate useAtomValue/useUpdateAtom in `jotai/utils` for now. So, in docs (https://jotai.org/docs/utils/use-atom-value, https://jotai.org/docs/utils/use-update-atom), just say we have the same api in core and add links.

src/core/useAtom.ts
```diff
@@ -1,22 +1,6 @@
...
-  useEffect,
-  useReducer,
-} from 'react'
-import type { Reducer } from 'react'
 import type { Atom, Scope, SetAtom, WritableAtom } from './atom'
-import { getScopeContext } from './contexts'
-import { COMMIT_ATOM, READ_ATOM, SUBSCRIBE_ATOM, WRITE_ATOM } from './store'
-import type { VersionObject } from './store'
-
-type ResolveType<T> = T extends Promise<infer V> ? V : T
```
dai-shi, line 14: Let's keep this. It's a bit awkward, but we don't want to export this internal type (yet).

src/core/useAtom.ts
```diff
@@ -1,22 +1,6 @@
...
-
-const isWritable = <Value, Update, Result extends void | Promise<void>>(
-  atom: Atom<Value> | WritableAtom<Value, Update, Result>
-): atom is WritableAtom<Value, Update, Result> =>
-  !!(atom as WritableAtom<Value, Update, Result>).write
+import { ResolveType, useAtomValue } from './useAtomValue'
```
dai-shi, line 2:
```suggestion
import { useAtomValue } from './useAtomValue'
```

src/core/useAtomValue.ts
```diff
@@ -0,0 +1,87 @@
...
+  useEffect,
+  useReducer,
+} from 'react'
+import type { Reducer } from 'react'
+import type { Atom, Scope } from './atom'
+import { getScopeContext } from './contexts'
+import { COMMIT_ATOM, READ_ATOM, SUBSCRIBE_ATOM } from './store'
+import type { VersionObject } from './store'
+
+export type ResolveType<T> = T extends Promise<infer V> ? V : T
```
dai-shi, line 14:
```suggestion
type ResolveType<T> = T extends Promise<infer V> ? V : T
```

src/core/useAtom.ts
```diff
@@ -38,85 +22,8 @@ export function useAtom<Value, Update, Result extends void | Promise<void>>(
...
-    },
-    [store, versionedWrite, atom]
-  )
-
-  useDebugValue(value)
-  return [value, setAtom]
+  return [
+    useAtomValue(atom, scope),
+    useSetAtom(atom as WritableAtom<Value, Update, Result>, scope),
+  ]
```
dai-shi, line 28: This is what I expected, but now after reviewing this commit, I think checking `isWritable` here seems better.
```suggestion
  return [
    useAtomValue(atom, scope),
    isWritable(atom) ? useSetAtom(atom, scope) : () => {
      throw new Error('not writable atom')
    },
  ]
```

dai-shi (commented):
Sorry for nits, but important. Please also check the conflicts.

dai-shi:
> Attempted import error: 'useSetAtom' is not exported from 'jotai' (imported as 'useSetAtom').

☝️ This is an error from Vercel deploy log. Seems like a tough one? cc @sandren

src/core/useAtom.ts
dai-shi: Oh my bad, this breaks the rules of hooks. Void.

src/core/useSetAtom.ts
```diff
@@ -0,0 +1,32 @@
+import { useCallback, useContext } from 'react'
+import { SECRET_INTERNAL_getScopeContext as getScopeContext } from 'jotai'
+import type { WritableAtom } from 'jotai'
+import type { Atom, Scope, SetAtom } from '../core/atom'
+import { WRITE_ATOM } from '../core/store'
+import type { VersionObject } from '../core/store'
```
loganvolkers, line 6: I believe that things in `core` should not import anything `from 'jotai'`

```suggestion
import { getScopeContext } from './contexts'
import type { Atom, Scope, SetAtom, WritableAtom } from './atom'
import { WRITE_ATOM } from './store'
import type { VersionObject } from './store'
```

website/src/components/docs.js
```diff
@@ -1,12 +1,12 @@
 import cx from 'classnames'
 import { Link, graphql, useStaticQuery } from 'gatsby'
-import { useUpdateAtom } from 'jotai/utils'
+import { useSetAtom } from 'jotai'
```
dai-shi, line 3: Please don't change this. We can only change this after releasing the new version.

dai-shi (changes requested):

dai-shi (approved):
Cool. This should be ready.
Let's work on docs in a separate PR.

PR #996, opened 2022-02-04
fix(readme,website): update assets and fix search bug

sandren:
This adds the Jotai mascot to the project README file and consolidates all of Jotai's brand assets into the `./img` directory for others to easily download, but elsewhere consumes the same files from Google Cloud Storage for improved caching. 🙂

It also fixes a bug on the website where an empty query would still display search results.


dai-shi (approved):
LGTM!

PR #1001, opened 2022-02-08
Add atomWithListeners

elliotwaite:
I think I need to fix some types since I'm getting type errors in the [demo](https://codesandbox.io/s/jotai-atomwithlisteners-hbymp?file=/src/App.tsx), but I wasn't sure what to change (still learning TypeScript). Was hoping you could make a suggestion.

I also changed my previous code to remove a listener from:
```
setListeners((prev) => prev.filter((c) => c !== callback))
```
... to:
```
setListeners((prev) => {
  const index = prev.indexOf(callback)
  return [...prev.slice(0, index), ...prev.slice(index + 1)]
})
```
... to allow the same callback to be registered multiple times.

And I added a check to see if the value actually changed:
```
const prevVal = get(baseAtom)
set(baseAtom, arg)
const newVal = get(baseAtom)
if (newVal !== prevVal) {
  get(listenersAtom).forEach((callback) => callback(get, set, newVal))
}
 ```
However, this changes the behavior from the callbacks being called each time the atom is set to only being called if its value actually changed. Let me know if you think the previous behavior was more correct.

Also, in certain places I wasn't sure if I should use the term `callback` or `listener`, for example, should this:
```
get(listenersAtom).forEach((callback) => callback(get, set, newVal))
```
... be this?:
```
get(listenersAtom).forEach((listener) => listener(get, set, newVal))
```

I also called the hook `useListener`, but let me know if think the hook should be called something else.

### CODE REVIEW

**Comment by elliotwaite** [comment]
Also, let me know if you have any suggestions for how I should change the text that describes what the atom does.

**Comment by dai-shi** [comment]
Thanks for working on this.

> I think I need to fix some types since I'm getting type errors in the [demo](https://codesandbox.io/s/jotai-atomwithlisteners-hbymp?file=/src/App.tsx), but I wasn't sure what to change (still learning TypeScript). Was hoping you could make a suggestion.

Add `as const`.

image


> I also changed my previous code to remove a listener from:
> 
> ```
> setListeners((prev) => prev.filter((c) => c !== callback))
> ```
> 
> ... to:
> 
> ```
> setListeners((prev) => {
>   const index = prev.indexOf(callback)
>   return [...prev.slice(0, index), ...prev.slice(index + 1)]
> })
> ```
> 
> ... to allow the same callback to be registered multiple times.

Sounds and looks good.

> And I added a check to see if the value actually changed:
> 
> ```
> const prevVal = get(baseAtom)
> set(baseAtom, arg)
> const newVal = get(baseAtom)
> if (newVal !== prevVal) {
>   get(listenersAtom).forEach((callback) => callback(get, set, newVal))
> }
> ```
> 
> However, this changes the behavior from the callbacks being called each time the atom is set to only being called if its value actually changed. Let me know if you think the previous behavior was more correct.

Yeah, I wouldn't add the check. How about passing `prevVal` in callback so that it can be done on the user's end.
```js
  callback(get, set, newVal, prevVal)
```

> Also, in certain places I wasn't sure if I should use the term `callback` or `listener`,

I'm not sure either... sometimes mixed.

> I also called the hook `useListener`, but let me know if think the hook should be called something else.

It's only about the internal name, right? It looks good.

elliotwaite:
> Add `as const`.

Ah, thanks. I also changed where I previously had `Value | undefined` to just `Value` and made the `initialValue` required instead of optional, as that fixed another issue.

> Yeah, I wouldn't add the check. How about passing prevVal in callback so that it can be done on the user's end.

Okay, I changed it to do that. I also changed some of the description text to mention that the callback is called every time the atom's value is set instead of every time its value is changed.

> I'm not sure either... sometimes mixed.

Okay, I kept it is as `callback` in this case since that seemed easier to understand to me.

> It's only about the internal name, right?

Yeah, the `useListener` name is only for the internal hook, and its style is just reflected in the name given to the variable it's assigned to externally in the demo code, `useCountListener`.

Let me know if there is anything else I should change.

dai-shi (approved):
Looks great! Thanks for your contribution!

elliotwaite:
Nice. Glad I could contribute.

PR #1004, opened 2022-02-09
feat(utils/atomWithStorage): Add subscribe to storage events in defaultStorage

### LINKED ISSUE #882: Synchronize atomWithStorage across tabs with "storage" event
I've noticed that using `atomWithStorage` util, across two tabs of the same origin, with the same key, does not synchronize the values.

Synchronization would be relatively simple to implement by listening to the [storage event](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) on the window and updating the atom's value if the localStorage changes at the same key.

Of course, we'd need to take care that two tabs do not notify one another in a loop with these events (the recipient tab of the event should only update the in-memory state and not write to localStorage).

What do you think? I'd be happy to take a stab at implementation.

Thisen:
Adds `subscribe` to `defaultStorage` for `atomWithStorage`.
The `subscribe` listens to `storage` events.

Closes #882.

### DIFF
--- a/src/utils/atomWithStorage.ts
+++ b/src/utils/atomWithStorage.ts
@@ -75,6 +75,17 @@ export function createJSONStorage<Value>(
     setItem: (key, newValue) =>
       getStringStorage().setItem(key, JSON.stringify(newValue)),
     removeItem: (key) => getStringStorage().removeItem?.(key), // TODO remove optional chaining
+    subscribe: (key, callback) => {
+      const storageEventCallback = (e: StorageEvent) => {
+        if (e.key === key && e.newValue) {
+          callback(JSON.parse(e.newValue))
+        }
+      }
+      window.addEventListener('storage', storageEventCallback)
+      return () => {
+        window.removeEventListener('storage', storageEventCallback)
+      }
+    },
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/atomWithStorage.ts
```diff
@@ -75,6 +75,17 @@ export function createJSONStorage<Value>(
...
+      const storageEventCallback = (e: StorageEvent) => {
+        if (e.key === key && e.newValue) {
+          callback(JSON.parse(e.newValue))
+        }
+      }
+      window.addEventListener('storage', storageEventCallback)
+      return () => {
+        window.removeEventListener('storage', storageEventCallback)
+      }
+    },
```
dai-shi, line 88: `createJSONStorage` is an exported function and used with other storages like AsyncStorage in RN.

We need to apply this logic only for `defaultStorage`.

dai-shi (changes requested):
Thanks for working on this.

- It looks good, but we need to do it only for `defaultStorage`.
- I know it's impossible to write tests for this. Correct?
- Please create a codesandbox (using PR's build) demo and add it in the docs.

dai-shi (approved):
Looks good! Thanks.

dai-shi:
Working good:

https://user-images.githubusercontent.com/490574/153555914-b5446d9e-84e7-4039-8474-a3e5ae3ac2b5.mov

Issue #1012, opened 2022-02-15
AtomFamily type isn't exported


MasonD:
Currently as a user, explicitly typing the return value of atomFamily is awkward, because the AtomFamily type isn't exported, so it has to be redefined by hand. Any reason not to export this type?

dai-shi:
Hi, thanks for opening it up.
I expect, generally, you don't need to type atomFamily explicitly.
What is your use case look like?

(Not only for jotai, but we prefer to avoid exporting types unless necessary. It's to try avoiding breaking changes in types.)

MasonD:
I'm building an AtomFamily dynamically within a component and placing it in react context. I would need the AtomFamily type to easily type the context. Basically whenever you want to type a variable that isn't immediately set as an AtomFamily. Function arguments, etc.

dai-shi:
Do you mean you are storing atomFamily instance in useRef or something?
I assume, you don't consume all atomFamily methods, so create your own atomFamily instance type might work.

```ts
type MyAtomType = PrimitiveAtom<number>
type MyAtomFamily = {
  (param: string): MyAtomType
  // add others if you use
}

const MyContext = createContext<MyAtomFamily | null>(null)
```

---
If you would like to extract `MyAtomFamily` type, the easiest way would be to define a create function in advance: https://tsplay.dev/WvVonW
```ts
import { atom } from "jotai";
import { atomFamily } from "jotai/utils";

const createMyAtomFamily = () => atomFamily((s: string) => atom(Number(s)));
type MyAtomFamily = ReturnType<typeof createMyAtomFamily>;
```

MasonD:
Indeed, I maintain referential identity with useMemo. The atoms are built using environment (graphql client, etc) that aren’t accessible before react mounts. Currently I’m just typing the context as (a: Value) => WritableAtom<…>.

Just feels like, to me, since the return type of atomFamily is already part of the api contract, it should be exposed in a more convenient manner. Creating my own atomfamily instance seems counterintuitive, since I would effectively just be copy pasting code and manually keeping it in sync if there are any changes to the atomFamily return. 

Anyway, I’ve said my two cents. I’m perfectly capable of working around it not being exported myself, it just felt like the nontrivial return type of a function should be exported. Feel free to close the issue if that’s against this project’s ideal.

dai-shi:
Yeah, I know what you mean. I still think it's less common case.
[...]

PR #1013, opened 2022-02-15
chore(types): refactor imports and add comments

dai-shi:
Just minor refactoring.
I noticed some of types are exported but not considered as public api, so commented. (inspired by #1012)


Thisen (approved):
👍🏼

PR #1016, opened 2022-02-18
Fix `atomWithObservable` is not resubscribed after remount

11bit:
`atomWithObservable` seems to be not resubscribing to observable when a component is being hidden and then shown again. 

There is a test reproducing it and also a [codesandbox](https://codesandbox.io/s/wizardly-resonance-28rvue?file=/src/App.js): atom value updates when clicking on a document, but if you hide/show component, it doesn't update anymore.

### DIFF
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -97,7 +97,10 @@ export function atomWithObservable<TData>(
       if (!subscription) {
         subscription = observable.subscribe(dataListener, errorListener)
       }
-      return () => subscription?.unsubscribe()
+      return () => {
+        subscription?.unsubscribe()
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Nice catch!!!
Thanks for the fix and the test!

dai-shi:
@11bit Thanks for your contribution. If you are interested, would you like to contribute more? We need more contributors for atomWithObservable. See #875 for example.

11bit:
@dai-shi sure. will have a look at the next week

11bit:
@dai-shi I made PR #1058 that adds `initialData` (api is similar to #875 but different implementation). Please have a look

PR #1018, opened 2022-02-19
Replace useAtom w/ useSetAtom for write only

devuxer:
Updated the "write only atoms" example in the README to use the `useSetAtom` instead of the `useAtom` hook.

I decided to keep the original code commented out with explanation why it's not longer recommended. Maintainers might want to delete this altogether if they think it clutters up the docs.

I did not scour the whole docs to change any other outdated examples, but I can do that if the maintainers feel my initial edit is on the right track.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -136,7 +136,8 @@ Just do not define a read function.
 const multiplyCountAtom = atom(null, (get, set, by) => set(countAtom, get(countAtom) * by))
 
 function Controls() {
-  const [, multiply] = useAtom(multiplyCountAtom)
+  // causes unnecessary re-render: const [, multiply] = useAtom(multiplyCountAtom)
+  const nultiply = useSetAtom(multiplyCountAtom)
   return <button onClick={() => multiply(3)}>triple</button>
 ```
 


dai-shi:
So, this seems controversial, and we haven't finished on the docs yet, but my view is:
- Basic functions: `atom` and `useAtom`
- Additional functions: `Provider`, `useAtomValue`, and `useSetAtom`

(they are all exported from `'jotai'`)

I would like to educate people to use `atom` and `useAtom` for the first time.
And then, when necessary, they can reach out additional functions, allowing more use cases.
So, readme.md only shows basic functions. docs website will explain additional functions.

Hope you understand it.

readme.md
```diff
@@ -136,7 +136,8 @@ Just do not define a read function.
 const multiplyCountAtom = atom(null, (get, set, by) => set(countAtom, get(countAtom) * by))
 
 function Controls() {
-  const [, multiply] = useAtom(multiplyCountAtom)
+  // causes unnecessary re-render: const [, multiply] = useAtom(multiplyCountAtom)
```
dai-shi, line 139: No, it doesn't cause unnecessary re-renders.
We will keep using the original pattern for the basic usage.

readme.md
```diff
@@ -136,7 +136,8 @@ Just do not define a read function.
...
+  const nultiply = useSetAtom(multiplyCountAtom)
```
dai-shi, line 140: Please note that `useSetAtom` from `jotai` is nothing new. We have `useUpdateAtom` from `jotai/utils` from the beginning with the same functionality and with the same mind.

dai-shi:
Closing this. Let's continue the discussion in #1019.

devuxer:
Hi @dai-shi ,

You wrote...

> Hope you understand it.

Yes. I kept the pull request very minimal in case I was on the track, so no problem. Thank you for your response.

I do feel confused, though. When it is advisable to use `useSetAtom` vs. `useAtom`. Is `useSetAtom` premature optimization? Do you use it only if you're experience slow performance?

The comment in this example is the reason I thought `useSetAtom` was the "preferred" hook:

image

dai-shi:
Happy to answer your questions (and we should include it in docs eventually).
There are two scenarios:

For write-only atoms, the atom value never changes, so `useAtom` is fine. `useSetAtom` is  premature optimization in this scenario.

For primitive(-like) atoms, if you do `const [, setValue] = useAtom(valueAtom)` it can cause unnecessary re-renders. So, `useSetAtom` (or `useUpdateAtom`) helps to avoid extra re-renders. Or, always create write-only atoms in such cases. I prefer write-only atoms as a pattern.

devuxer:
That makes sense, thanks. I will go with your advice to create write-only atoms. That seems cleaner.

PR #1030, opened 2022-02-28
Try using function toString for map key

### LINKED ISSUE #1025: Problems with React Fast Refresh
Hi, while it works well in most cases, we experience some bugs.

1. Editing functions out of react tree aren't updated. For example, update the logic of your read-only atom and issue a state update.
2. The referential integrity of the atom store gets broken. I receive no longer state updates. I need to reload to fix it.

All issues happen when using react fast refresh with https://github.com/gsoft-inc/craco. We use the jotai babel preset as documented on the website!

Thank you and keep up the good work!

Thisen:
Implementation is an experiment. Lets see.

Closes #1025.

### DIFF
--- a/src/babel/plugin-react-refresh.ts
+++ b/src/babel/plugin-react-refresh.ts
@@ -19,9 +19,9 @@ export default function reactRefreshPlugin({
             cache: new Map(),
             get(name, inst) { 
               if (this.cache.has(name)) {
-                return this.cache.get(name)
+                return this.cache.get(name + inst.toString())
               }
-              this.cache.set(name, inst)
+              this.cache.set(name + inst.toString(), inst)
               return inst
             },
           }`)()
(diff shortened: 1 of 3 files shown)


dai-shi (commented):
the change looks okay.

dai-shi:
@Thisen I was trying to test this, but before going into it, found the original codesandbox example doesn't work.
https://codesandbox.io/s/priceless-benz-bgf3b in https://jotai.org/docs/api/babel

Do you know what causes it?

Thisen:
I can check it out, but since we can't support the issue fully, I think we should just close. WDYT @dai-shi ?

dai-shi:
I still don't understand the real issue in #1025. Can you help with [the comment](https://github.com/pmndrs/jotai/issues/1025#issuecomment-1062362840)?

On the other hand, this PR seems reasonable and valuable.
Do you think if you can create a codesandbox example that fails without this PR, and will be fixed by this PR?

Thisen:
Closing due to somewhat hacky solution. Can re-open if we decide otherwise.

PR #1032, opened 2022-02-28
[docs] Add base example

### LINKED ISSUE #1031: [Docs] Create a base example
There are some examples with only core functions, without `codesandbox` for developers to try. Such as [Use Reset Atom](https://jotai.org/docs/utils/use-reset-atom) or [Resetting default values](https://jotai.org/docs/utils/atom-with-default#resetting-default-values)
So I think it's a good idea to add an base example, developers only need to copy core function and paste, then can enjoy playing around the code.
Currently, I think we can add it at the top of `utils`.

Suggestions are welcome!

LuckyPigeon:
fix #1031

### DIFF
--- a/docs/api/utils.mdx
+++ b/docs/api/utils.mdx
@@ -5,6 +5,7 @@ nav: 2.02
 ---
 
 This is an overview over atom creators/hooks utilities that can found under `jotai/utils`. Each utility is listed below with a link to their description/API.
+You can play around the examples with this [base example](https://codesandbox.io/s/red-wind-eptiu0)
 
 ## Overview
 


dai-shi (commented):
Hmm, I'm not sure if this helps anything meaningful. It's not harmful though.
@sandren What do you think?

sandren:
> Hmm, I'm not sure if this helps anything meaningful. It's not harmful though. @sandren What do you think?

@dai-shi I think something like a "base" sandbox may be helpful for developing more complex examples, but I don't think it's helpful to link to this sandbox from the docs.

PR #1040, opened 2022-03-04
[docs] Update debugging link

LuckyPigeon:
hyper links in [Frozen Atom](https://jotai.org/docs/guides/debugging#frozen-atoms) is too old, needs some update.


dai-shi (approved):
Thanks so much for the fix!!

PR #1047, opened 2022-03-06
update jotai to v1.6.0  and correct example to make it work fine

tsingson:
update example to make its work correct.

### DIFF
--- a/examples/hacker_news/src/App.tsx
+++ b/examples/hacker_news/src/App.tsx
@@ -2,29 +2,33 @@ import { Suspense } from 'react'
 import { a, useSpring } from '@react-spring/web'
 import Parser from 'html-react-parser'
 import { Provider, atom, useAtom, useSetAtom } from 'jotai'
+import { atomWithQuery } from 'jotai/query'
 
-type PostData = {
-  by: string
-  descendants?: number
-  id: number
-  kids?: number[]
-  parent: number
-  score?: number
-  text?: string
-  time: number
-  title?: string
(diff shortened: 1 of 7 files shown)


examples/hacker_news/src/App.tsx
```diff
@@ -2,29 +2,33 @@ import { Suspense } from 'react'
 import { a, useSpring } from '@react-spring/web'
 import Parser from 'html-react-parser'
 import { Provider, atom, useAtom, useSetAtom } from 'jotai'
+import { atomWithQuery } from 'jotai/query'
```
dai-shi, line 5: We don't want to change this example with RQ. If you need such a example, please create a new one.

dai-shi (changes requested):
Thanks for working on this. A quick first review.

dai-shi:
> to make its work correct.

what was the incorrect example, previously?

TwistedMinda:
@tsingson Hey tsingson, do you think you will be able to complete this? 

I can take care of updating to last version & fixing the problem, but do you remember what was not working in the first place?

The [Example CodeSandBox](https://codesandbox.io/s/github/pmndrs/jotai/tree/main/examples/hacker_news?file=/src/App.tsx) seems to be working just fine

TwistedMinda:
@dai-shi I guess we can close it and we'll rework examples if there is a bug found or we make a major update. Correct me if I'm wrong.
Most important is that the example is working fine.

OP can still comment here if he has an issue

PR #1049, opened 2022-03-07
fix(devtools): better types with official devtools extension types

Aslemammad:
This PR tries to bring the native `window.__REDUX_DEVTOOLS_EXTENSION__` types from redux devtools and then avoids some custom types which may be incorrect in some cases!

### DIFF
--- a/src/devtools/types.ts
+++ b/src/devtools/types.ts
@@ -1,6 +1,6 @@
 import type {} from '@redux-devtools/extension'
 
-export type Message = {
+type Message = {
   type: string
   payload?: any
   state?: any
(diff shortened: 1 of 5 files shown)


src/devtools/useAtomDevtools.ts
```diff
@@ -2,46 +2,15 @@ import { useEffect, useRef } from 'react'
...
   name?: string,
   scope?: Scope
 ): void {
-  let extension: Extension | undefined
-  try {
-    extension = (window as any).__REDUX_DEVTOOLS_EXTENSION__ as Extension
-  } catch {
-    // ignored
-  }
+  let extension = window?.__REDUX_DEVTOOLS_EXTENSION__
```
dai-shi, line 12: Do we need `?` optional chaining?
Also, please keep try-catch, which is important.

src/devtools/types.ts
```diff
@@ -0,0 +1,20 @@
...
+
+interface Action<T = any> {
+  type: T
+}
+
+export interface ConnectResponse {
+  init: (state: unknown) => void
+  send: (action: Action<unknown>, state: unknown) => void
+  subscribe?: (
+    listener: (message: any) => void // FIXME no-any
```
dai-shi, line 17: Can this be `Message` instead of `any`?

dai-shi (changes requested):
Please also add description.

dai-shi (approved):
This should be good to go.

PR #1058, opened 2022-03-12
fix(utils): Initial data for atomWithObservable 

11bit:
Reimplementation of initial data option for `atomWIthObservable` inspired by #875 + some refactoring and simplification of `atomWithObservable` code

### DIFF
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -72,6 +72,8 @@ export function atomWithObservable<TData>(
     }
     let subscription: Subscription | null = null
 
+    // FIXME this implementation is not fully compatible with concurrent rendering.
+    // we need to deal with the case `onMount` is not invoked after the atom is initialized.
     dataAtom.onMount = (update) => {
       setData = update
       if (!subscription) {
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -29,43 +29,61 @@ type ObservableLike<T> = {
 
 type SubjectLike<T> = ObservableLike<T> & Observer<T>
 
+type InitialDataFunction<T> = () => T | undefined
+
+export type AtomWithObservableOptions<TData> = {
+  initialData?: TData | InitialDataFunction<TData>
+}
+export type CreateObservableOptions<TData> = {
+  initialData?: TData
+}
+
 export function atomWithObservable<TData>(
-  createObservable: (get: Getter) => SubjectLike<TData>
+  createObservable: (get: Getter) => SubjectLike<TData>,
+  options?: AtomWithObservableOptions<TData>
 ): WritableAtom<TData, TData>
 
 export function atomWithObservable<TData>(
-  createObservable: (get: Getter) => ObservableLike<TData>
+  createObservable: (get: Getter) => ObservableLike<TData>,
+  options?: AtomWithObservableOptions<TData>
 ): Atom<TData>
 
 export function atomWithObservable<TData>(
-  createObservable: (get: Getter) => ObservableLike<TData> | SubjectLike<TData>
+  createObservable: (get: Getter) => ObservableLike<TData> | SubjectLike<TData>,
+  options?: AtomWithObservableOptions<TData>
 ) {
   const observableResultAtom = atom((get) => {
     let settlePromise: ((data: TData | null, err?: unknown) => void) | null =
       null
+    let initialized = false
     let observable = createObservable(get)
     const itself = observable[Symbol.observable]?.()
     if (itself) {
       observable = itself
     }
-    const dataAtom = atom<TData | Promise<TData>>(
-      new Promise<TData>((resolve, reject) => {
-        settlePromise = (data, err) => {
-          if (err) {
-            reject(err)
-          } else {
-            resolve(data as TData)
-          }
-        }
-      })
+    const initialData = getInitialData(options)
+
+    const dataAtom = atom(
+      initialData === undefined
+        ? new Promise<TData>((resolve, reject) => {
+            settlePromise = (data, err) => {
+              if (err) {
+                reject(err)
+              } else {
+                resolve(data as TData)
+              }
+            }
+          })
+        : initialData
     )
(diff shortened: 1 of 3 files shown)


src/utils/atomWithObservable.ts
```diff
@@ -29,69 +29,52 @@ type ObservableLike<T> = {
 
 type SubjectLike<T> = ObservableLike<T> & Observer<T>
 
+type InitialDataFunction<T> = () => T | undefined
+
+export type AtomWithObservableOptions<TData> = {
```
dai-shi, line 34: I know it's a bit controversial, but we prefer not exporting types.

```suggestion
type AtomWithObservableOptions<TData> = {
```

src/utils/atomWithObservable.ts
```diff
@@ -29,69 +29,52 @@ type ObservableLike<T> = {
...
+  initialData?: TData | InitialDataFunction<TData>
+}
+export type CreateObservableOptions<TData> = {
+  initialData?: TData
+}
```
dai-shi, line 39: I don't think we need this. Leftover?

```suggestion
```

src/utils/atomWithObservable.ts
```diff
@@ -29,69 +29,52 @@ type ObservableLike<T> = {
 
 type SubjectLike<T> = ObservableLike<T> & Observer<T>
 
+type InitialDataFunction<T> = () => T | undefined
+
+export type AtomWithObservableOptions<TData> = {
+  initialData?: TData | InitialDataFunction<TData>
+}
+export type CreateObservableOptions<TData> = {
+  initialData?: TData
```
dai-shi, line 38: Can we name it `initialValue`?
```suggestion
  initialValue?: TData
```

src/utils/atomWithObservable.ts
```diff
@@ -120,3 +104,32 @@ export function atomWithObservable<TData>(
...
+
+function getInitialData<TData>(options: AtomWithObservableOptions<TData>) {
+  const initialData = options.initialData
+  return initialData instanceof Function ? initialData() : initialData
+}
+
+function firstValueFrom<T>(source: ObservableLike<T>): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    let resolved = false
+    const subscription = source.subscribe({
```
dai-shi, line 116: Although this looks cleaner, I think there's a risk of memory leaks.
Such case is, a) we create an atom, b) before getting the first value, c) atom can be unmounted.
We need to unsubscribe this subscription on `onMount` cleanup.

dai-shi (changes requested):
Hey, nice to see you work on this!! Please see comments.

src/utils/atomWithObservable.ts
11bit: @dai-shi hmm, I thought that the suspended component is not mounted before the promise resolves. So, we don't actually have an option to unsubscribe in any case or do I miss something?  

To illustrate, here is an atom that suspends and `onMount` is never called https://codesandbox.io/s/modern-glade-rutmog?file=/src/App.js:0-413

src/utils/atomWithObservable.ts
11bit: sure

src/utils/atomWithObservable.ts
11bit: yes, accidentally copied that from the previous pr. Thanks for noticing

src/utils/atomWithObservable.ts
11bit: Renamed everywhere to initialValue

src/utils/atomWithObservable.ts
11bit: BTW I've already changed other things. Should I do anything else here?

dai-shi: Let's add a comment like this:
```js
// FIXME this implementation is not fully compatible with concurrent rendering.
// we need to deal with the case `onMount` is not invoked after the atom is initialized.

11bit: @dai-shi I've added the comment. Please have a look

dai-shi:
https://github.com/pmndrs/jotai/pull/1058#discussion_r834925047
After reading my comment again, I notice I misunderstood something.
I think I confused with several issues.

Issue 1: If we subscribe in render (= `read` function body), there might not be a way to unsubscribe, leading to memory leaks.
Issue 2: Even if we unsubscribe in the callback, it may not fire.
Issue 3: Ideally, we should just subscribe once, if it mounts soon.

Issue 1 is important, but we unsubscribe in the callback (both before this PR and this PR), so it's not a big issue if the callback is invoked soon.
But, Issue 2 is still present for both. We would like to fix it but the known workaround is only the setTimeout hack done in `jotai/query` (internally in RQ) and `jotai/urql`. 
Issue 3 is something introduced in this PR. It's only the performance issue, and if `subscribe` is lightweight, it's not a big issue.

So, if Issue 3 is something we care, we would like to revert the change. What do you think? @11bit 

The `FIXME` comment should be about Issue 2. Let me try clarifying more.

dai-shi (changes requested):
Wait, there is more important issue.
Issue 4: If the value is changed twice the second value is dropped.
I guess it's the regression with this PR.

So, let's revert it, or come up with a better solution.

dai-shi:
> Issue 4: If the value is changed twice the second value is dropped.

Hmm, it's not solved even with the original code without this PR.
My bad again. I'll wait for comments anyway, hoping we have a better solution.

dai-shi (approved):
The biggest concern is Issue 4, but it's not a regression of this PR.
Issue 3 is the one introduced in this PR, but only a performance concern.

So, at least, it's mergeable.

11bit:
So, you mean that current implementation can miss values received between first value and befoe `onMount`, right?

I think solution might be to subscribe only once and do the following:
1. First value => resolve promise
2. Store everything that is receiverd before onMount somewhere
3. onMount => set the last received value from step 2 and then work as usual

PR #1076, opened 2022-03-30
feat(devtools): enabled option useAtomsDevtools

Aslemammad:
This PR tries to achieve what we achieved in https://github.com/pmndrs/zustand/pull/880 so we can add an option called `enabled` for better testing/dev env friendly errors/warnings!

This PR is for useAtomsDevtools, so we can make reviewing easier. after this, we'll try achieving this for other dev tools.


src/devtools/useAtomsDevtools.ts
```diff
@@ -52,11 +52,16 @@ const getDevtoolsState = (atomsSnapshot: AtomsSnapshot) => {
...
 }
 
-export function useAtomsDevtools(name: string, scope?: Scope) {
-  const ScopeContext = getScopeContext(scope)
+interface DevtoolsOptions {
+  scope?: Scope
+  enabled?: boolean
+}
+
+export function useAtomsDevtools(name: string, options?: DevtoolsOptions) {
```
dai-shi, line 60: Please make it backward compatible.

```ts
export function useAtomsDevtools(name: string, options?: DevtoolsOptions)
/**
 * @deprecated Please use object options
 */
export function useAtomsDevtools(name: string, options?: Scope)
export function useAtomsDevtools(name: string, options?: DevtoolsOptions | Scope) {
  if (typeof options !== 'undefined' && typeof options !== 'object') {
    console.warn('DEPRECATED...')
    options = { scope: options }
```

Check https://github.com/pmndrs/valtio/pull/400 for reference.

src/devtools/useAtomsDevtools.ts
```diff
@@ -52,11 +52,16 @@ const getDevtoolsState = (atomsSnapshot: AtomsSnapshot) => {
...
+  const ScopeContext = getScopeContext(options?.scope)
```
dai-shi, line 61: Maybe we want to destructure in advance, so that code looks cleaner.
```ts
  const { enabled, scope } = options || {}
```

src/devtools/useAtomsDevtools.ts
```diff
@@ -52,11 +52,16 @@ const getDevtoolsState = (atomsSnapshot: AtomsSnapshot) => {
...
   const { s: store, w: versionedWrite } = useContext(ScopeContext)
 
-  if (!store[DEV_SUBSCRIBE_STATE]) {
+  if (!store[DEV_SUBSCRIBE_STATE] && options?.enabled && __DEV__) {
```
dai-shi, line 64: I don't think we need this.
Ohhh! we can't enable this in PRD anyway! That was my confusion...

So, it should be something like this. Open for suggestions.
```js
if (enabled === false) {
  return
}
if (!store[DEV_SUBSCRIBE_STATE]) {
}
```

src/devtools/useAtomsDevtools.ts
```diff
@@ -117,18 +122,21 @@ export function useAtomsDevtools(name: string, scope?: Scope) {
   let extension: typeof window['__REDUX_DEVTOOLS_EXTENSION__']
 
   try {
-    extension = window.__REDUX_DEVTOOLS_EXTENSION__
+    extension =
+      ((options?.enabled ? options.enabled : __DEV__) &&
```
dai-shi, line 126: This isn't what we want. It should be `typeof enabled === 'boolean' ? enabled : __DEV__`, to deal with `enabled=false`.


Hmm, it would be much cleaner if we do:
```js
  enabled ?? __DEV__
```
(we should do this in zustand and valtio too.)

src/devtools/useAtomsDevtools.ts
```diff
@@ -117,18 +122,21 @@ export function useAtomsDevtools(name: string, scope?: Scope) {
...
-    if (__DEV__ && typeof window !== 'undefined') {
+    if (__DEV__ && options?.enabled) {
       console.warn('Please install/enable Redux devtools extension')
     }
   }
 
-  if (!store[DEV_SUBSCRIBE_STATE]) {
+  if (!store[DEV_SUBSCRIBE_STATE] && options?.enabled && __DEV__) {
     throw new Error('useAtomsSnapshot can only be used in dev mode.')
   }
```
dai-shi, line 141: Why do we have this again? It seems like a leftover. `useAtomsSnapshot` 👀

dai-shi (changes requested):
first review

src/devtools/useAtomsDevtools.ts
Aslemammad: not putting undefined causes type problems!

src/devtools/useAtomsDevtools.ts
Aslemammad: This makes problems for us! make tests failing!

src/devtools/useAtomsDevtools.ts
```diff
@@ -52,11 +53,29 @@ const getDevtoolsState = (atomsSnapshot: AtomsSnapshot) => {
...
+) {
+  if (typeof options !== 'undefined' && typeof options !== 'object') {
+    options = { scope: options }
+  }
+  const { enabled, scope } = options || {}
   const ScopeContext = getScopeContext(scope)
   const { s: store, w: versionedWrite } = useContext(ScopeContext)
 
-  if (!store[DEV_SUBSCRIBE_STATE]) {
+  if (!store[DEV_SUBSCRIBE_STATE] && enabled && __DEV__) {
```
dai-shi, line 78: I still want the early return.
Which test fails? I mean it should only fail if `enabled === false`, right? It should be a new test.

```suggestion
  if (enabled === false) {
    return
  }
  if (!store[DEV_SUBSCRIBE_STATE]) {
```

dai-shi (changes requested):
second one!

src/devtools/useAtomsDevtools.ts
dai-shi: I completely forgot about this being a hook. We can't early return. That's really my bad. So sorry to have your time wasted. Added a commit: dc485b7

src/devtools/useAtomsDevtools.ts
```diff
@@ -114,83 +152,68 @@ export function useAtomsDevtools(name: string, scope?: Scope) {
...
-    throw new Error('useAtomsSnapshot can only be used in dev mode.')
-  }
-
   const isTimeTraveling = useRef(false)
   const isRecording = useRef(true)
   const devtools = useRef<
-    ReturnType<NonNullable<typeof extension>['connect']> & {
+    ReturnType<
+      NonNullable<typeof window['__REDUX_DEVTOOLS_EXTENSION__']>['connect']
+    > & {
```
dai-shi, line 160: Is this change necessary?

dai-shi (approved):

PR #1083, opened 2022-04-04
docs: Minor language change for more accurate reading

pietgit:
Hi, I added a small improvement to the docs to make the below sentence a little more accurate (Objects and Arrays aren't primitive types in JS). 

from

"First create a primitive atom.
An atom represents a piece of state. All you need is to specify an initial value, which can be a primitive value like a string, number, object or array. You can create as many primitive atoms as you like."

to

"First create a primitive atom.
An atom represents a piece of state. All you need is to specify an initial
value, which can be a **primitive value like a string or number, or an object or
array.** You can create as many primitive atoms as you like."

More info: I'm new to Jotai, and are going to use it for a project upgrade, and found myself wondering about the usage of the term 'primitive' here.

### DIFF
--- a/docs/introduction.mdx
+++ b/docs/introduction.mdx
@@ -20,9 +20,9 @@ You can try live demos here:
 
 ### First create a primitive atom
 
-An atom represents a piece of state. All you need is to specify an initial
-value, which can be a primitive value like a string or number, or an object or
-array. You can create as many primitive atoms as you like.
+A piece of state in Jotai is represented by an atom. All you need is to specify
+an initial value, be it a primitive type like a number, string, or more complex
+structures like arrays and objects. You can create as many primitive atoms as you like.
 
 ```jsx
 import { atom } from 'jotai'
--- a/docs/introduction.mdx
+++ b/docs/introduction.mdx
@@ -21,7 +21,7 @@ You can try live demos here:
 ### First create a primitive atom
 
 An atom represents a piece of state. All you need is to specify an initial
-value, which can be a primitive value like a string, number, object or
+value, which can be a primitive value like a string or number, or an object or
 array. You can create as many primitive atoms as you like.
 
 ```jsx


AjaxSolutions:
How do you like this description?

"A piece of state in Jotai is represented by an atom. An atom accepts an initial value, be it a primitive type like a number, string, or more complex structures like arrays and objects."

https://dmitripavlutin.com/react-jotai-state-management/

pietgit:
> 

Yes, that's perfect. I've added the following into a new commit:

"A piece of state in Jotai is represented by an atom. All you need is to specify
an initial value, be it a primitive type like a number, string, or more complex
structures like arrays and objects. You can create as many primitive atoms as you like."

sandren:
@pietgit  Thank you for making the language more accurate. 🙂

Though perhaps it would be even easier to understand our meaning of the term 'primitive atom' if we remove the reference to 'primitive types' altogether since it isn't necessary for the explanation of atoms?

AjaxSolutions:
@sandren Jotai is defined as "Primitive and flexible state management for React"

What does primitive mean?  Does it mean that it's low-level?

https://jotai.org/

dai-shi:
> @sandren Jotai is defined as "Primitive and flexible state management for React"
> 
> What does primitive mean? Does it mean that it's low-level?

It's me who noted "primitive and flexible". I wanted to emphasize it's considered as building blocks.
atom+useAtom is a replacement for useState(+useContext). I would say useState is low-level as well as atom+useAtom.
jotai/utils and other jotai/* are all developed on jotai core public api. so, I think it's very flexible.

dai-shi (changes requested):
Would love to see this change. https://github.com/pmndrs/jotai/pull/1083#issuecomment-1089199467

dai-shi:
@pietgit are you still around?

TwistedMinda:
@pietgit If u're not available I can take care of this, we can close this PR and i'll create one from this. Seems like a small change that we can do fast, no need to let it stale

PR #1088, opened 2022-04-08
fix(utils): undefined bug with no entry in storage

### LINKED ISSUE #1086: atomWithStorage seems to be broken with 1.6.2
I updated Jotai to 1.6.2 just now and I noticed this seems to have broken code in my current project which uses atomWithStorage.

I can't give you a minimal reproduction now (might later). All I can say now that I have several useAtoms() of atomWithStorage in different React components and suddenly at least one started to return the local storage value of a completely different key. 

I reverted to 1.6.1 and everything is working fine.

dai-shi:
close #1086 

This was a silly mistake, but can't be noticed without a test.

### DIFF
--- a/src/utils/atomWithStorage.ts
+++ b/src/utils/atomWithStorage.ts
@@ -50,8 +50,8 @@ export function createJSONStorage<Value>(
       const parse = (str: string | null) => {
         str = str || ''
         if (lastStr !== str) {
-          lastStr = str
           lastValue = JSON.parse(str)
+          lastStr = str
         }
         return lastValue
       }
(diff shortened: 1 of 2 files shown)


Moshyfawn:
Thank you for your work!

Do you have an ETA for the next release that will include this fix? I'm not familiar with the Jotai release cycle yet. Not trying to rush or anything ;)

dai-shi:
very soon.
I put a date in the milestone, so you can expect when I'm planning: https://github.com/pmndrs/jotai/milestone/20

Moshyfawn:
I appreciate it, thx!

PR #1091, opened 2022-04-11
feat(core,utils): abortable atom

dai-shi:
ref: https://twitter.com/dai_shi/status/1513173028674617350

core doesn't depend on `AbortController`, only the one in utils does.

- [x] impl
- [x] tests
- [x] docs with codesandbox example


src/utils/abortableAtom.ts
```diff
@@ -0,0 +1,45 @@
...
+
+// read-only derived atom
+export function abortableAtom<Value>(read: Read<Value>): Atom<Value>
+
+export function abortableAtom<
+  Value,
+  Update,
+  Result extends void | Promise<void>
+>(read: Read<Value>, write?: Write<Update, Result>) {
+  const { signal, abort } = new AbortController()
```
incepter, line 37: I am not a zustand user, so may be my comment will be irrelevant 🙏


AbortController only works with fetch and axios and may be some other libraries.

Cancellations in general affect any asynchronous behavior: browser api (timeout, interval), Promises, async/await .. I suggest supporting a mechanism that allows cancellations of all kinds, and not instantiating an abort controller at each run.

In a library I am preparing, I solved this by allowing the user's function to register any amount of abort callbacks, that are invoked either by user action or by effect cleanup.


```
function myStateProducer(props) {
  props.onAbort(() => clearTimeout(id));
  props.onAbort(() => clearInterval(id));
  props.onAbort((reason) => controller.abort(reason));
}
```

This allows the user to easily implement cancellations of all kinds rather that relying on AbortController  which isn't supported in legacy browsers.

How can we cancel atoms ? is there any `abort` function provided ?

src/utils/abortableAtom.ts
```diff
@@ -0,0 +1,45 @@
...
+  return atom((get) => {
+    const promise = read(get, { signal })
+    if (promise instanceof Promise) {
```
incepter, line 40: For better support, I guess we should support thenable, not only promises.

ie:
`typeof promise === 'function' and typeof promise.then === 'function'`

Is this promise catched somewhere in case of rejection?

incepter:
I may help with cancellations, but i ll need first to know how the repo and internals of zustand work.

If there is any document describing them somehwere ? thanks in advance.

src/utils/abortableAtom.ts
dai-shi: As far as I understand, AbortController is a general subscriber, and has nothing to do with fetch and others. We can use it for anything. setTimeout/clearTimeout can work too. (DX-wise, not sure if it's ideal) It would be nice to have such usage in docs.

src/utils/abortableAtom.ts
dai-shi: We heavily depend on `instanceof Promise` everywhere. So, that means general thenable is not supported by jotai in the first place.

> Is this promise catched somewhere in case of rejection?

What does this mean? It is cached in store.

dai-shi:
I'm sure you already visited https://jotai.org. That covers mostly. Some comments in source code may help, but it will still require the understanding of public API first.

src/utils/abortableAtom.ts
dai-shi: My plan was to use `signal.aborted`. It's not super nice, though. Maybe there would be better solutions. What we keep in mind is we don't often directly use promises. In most cases we just use `async` functions.

I'm not sure if we are on the same page. Feel free to try your own idea. It can be AC based, or independent. We shouldn't need to change jotai core in either case.

(I feel like I'm missing some cases like complex dependency graph...

PR #1096, opened 2022-04-14
chore(docs): waitForAll with objects

dai-shi:
Even I forgot about it...

### DIFF
--- a/docs/utils/wait-for-all.mdx
+++ b/docs/utils/wait-for-all.mdx
@@ -55,13 +55,9 @@ const catsAtom = atom(async (get) => {
   return await response.json()
 })
 
-const animalsAtom = atom((get) => {
-  return get(
-    waitForAll({
-      dogs: dogsAtom,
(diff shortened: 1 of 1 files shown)


docs/utils/wait-for-all.mdx
```diff
@@ -55,13 +55,9 @@ const catsAtom = atom(async (get) => {
...
-const animalsAtom = atom((get) => {
-  return get(
-    waitForAll({
-      dogs: dogsAtom,
-      cats: catsAtom,
-    })
-  )
+const animalsAtom = waitForAll({
+  dogs: dogsAtom,
+  cats: catsAtom,
```
dai-shi, line 60: @Thisen Was there any intention to show the atom wrapper usage? If so, we probably want more meaningful example.

Thisen: I can't remember tbh 😅

PR #1101, opened 2022-04-15
fix(utils): make useAtomCallback return sync if possible

### LINKED ISSUE #1100: Imperatively & synchronously read atoms (`useAtomCallback` only supports `Promise`s)
I am in one of those weird edge cases where I am forced to be un-Reacty as I am using the Leaflet map library which requires a lot of imperative code. Thus I have something like the following:

```typescript
const markersLayer = useRef<L.MarkerClusterGroup>();
useQuery("markersKey", () => fetchMarkers(), {
    onSuccess: (data) => {
      if (!markersLayer.current) {
        markersLayer.current = L.markerClusterGroup({
          iconCreateFunction: (cluster) => {
            const highlightedMarkerIds = valueFromMyAtom;
            // ... create markers based on highlight
```

Unfortunately, the `iconCreateFunction` is triggered later and `valueFromMyAtom` is stale. It seems that I need an imperative getter function to put there instead.
`useAtomCallback` is perhaps what I need, but it only supports asynchronous reading which Leaflet doesn't accept. I investigated implementing a synchronous version but couldn't seem to work it out.

For now I have the following workaround but it feels like awful code smell:

```typescript
export const useAtomGetter = () => {
  const getter = useRef<Getter | null>(null);
  const derived = useMemo(
    () =>
      atom((get) => {
        getter.current = get;
      }),
    []
  );
  useAtomValue(derived);
  return getter.current;
};

//in my component
const atomGetter = useAtomGetter();

//...as above
const highlightedMarkerIds = atomGetter(myAtom);
```

Love the library so far! <3

dai-shi:
close #1100

### DIFF
--- a/src/utils/useAtomCallback.ts
+++ b/src/utils/useAtomCallback.ts
@@ -6,21 +6,25 @@ import type { Scope } from '../core/atom'
 type WriteGetter = Parameters<WritableAtom<unknown, unknown>['write']>[0]
 
 type Callback<Result, Arg> = undefined extends Arg
-  ? (arg?: Arg) => Promise<Result>
-  : (arg: Arg) => Promise<Result>
+  ? (arg?: Arg) => Result
+  : (arg: Arg) => Result
 
 export function useAtomCallback<Result, Arg>(
   callback: (get: WriteGetter, set: Setter, arg: Arg) => Promise<Result>,
   scope?: Scope
-): Callback<Result, Arg>
+): Callback<Promise<Result>, Arg>
 
 export function useAtomCallback<Result, Arg>(
   callback: (get: WriteGetter, set: Setter, arg: Arg) => Result,
   scope?: Scope
-): Callback<Result, Arg>
+): Callback<Result | Promise<Result>, Arg>
 
 export function useAtomCallback<Result, Arg>(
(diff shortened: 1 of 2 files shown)


JLarky:
I only recently updated to this version and I had a typescript error in one of my helper function :) just in case anyone would need this. Here are two examples.

If you want to read the value of jotai atom so that callback always returns the promise:
```ts
export function useAsyncAtomValue<V>(atom: WritableAtom<V, any>) {                                                                                           
  return useAtomCallback(React.useCallback(async get => get(atom), [atom]));
}
```

If you want to read the value of jotai atom synchronously:
```ts
export function useGetAtomValue<V>(atom: WritableAtom<V, any>) {
  const callback = useAtomCallback(React.useCallback(get => get(atom), [atom]));
  return React.useCallback(() => {
    const syncValue = callback();
    if (syncValue instanceof Promise) {
      throw new Error("useGetAtomValue should not be used with async atoms");
    }
    return syncValue;
  }, [callback]);
}
```

And example usage:
```ts
const xAtom = atom('');

function useX() {
  const value = useGetAtomValue(xAtom);
  const promise = useAsyncAtomValue(xAtom);
  setTimeout(async () => {
    value().toUpperCase();
    (await promise()).toUpperCase();
  }, 1000);
}
```

Issue #982, opened 2022-01-23
Feature request for atomWithBroadcast util


thachp:
A  way to share atoms between browser tabs and frames, similar to atomWithStorage.

See https://github.com/pubkey/broadcast-channel 
and  https://hookstate.js.org/docs/extensions-broadcasted.

```javascript
import { useAtom } from 'jotai'
import { atomWithBroadcast } from 'jotai/utils'

const darkModeAtom = atomWithBroadcast ('darkMode', false)

const Page = () => {
  const [darkMode, setDarkMode] = useAtom(darkModeAtom)

  return (
    <>
      <h1>Welcome to {darkMode ? 'dark' : 'light'} mode!</h1>
      <button onClick={() => setDarkMode(!darkMode)}>toggle theme</button>
    </>
  )
}
```

dai-shi:
This works without localStorage (or any other storage), right?
Then, we can't implement on top of atomWithStorage. It would be a new util.

I wonder if this should be in `jotai/utils` or in a third-party library. Maybe, an advanced recipe would fit well based on `broadcast-channel` (instead of native method).

---
Implementation-wise, I also wonder how we could pull the initial value. 🤔

dai-shi:
Would anyone be interested in creating an advanced recipe for a start?

dryrainbow:
Hello, I'm interested in implementing it, or just in writing an example in documentation with the recipe for using library with broadcast-channel

dai-shi:
@dryrainbow Why won't you try creating a codesandbox example and an advanced recipe in docs? Happy to review.

dai-shi:
Note that we already have a plan to add sync feature in for localStorage #882. For IndexDB, we could do the same with `atomWithStorage`.
What's not possible with `atomWithStorage` is pure BroadcastChannel without storage.

dai-shi:
Now #1004 is merged and released, what's remaining is `atomWithBroadcast` (with native method or boradcast-channel library) to be in "advanced-recipes" in docs. Would anyone be interested?

dai-shi:
wip #1111

dai-shi:
#1111 is merged.

PR #1111, opened 2022-04-20
docs(advanced-recipes): atomWithBroadcast util

Sajad-Sharhani:
A Util to share atoms between browser tabs and frames, similar to atomWithStorage but with the initialization limitation.


dai-shi (changes requested):
As we discussed and noted in #982, our plan is to add it as an advanced recipe. Please modify `./docs/advanced-recipes/atom-creators.mdx`.

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,61 @@ function EvenCounter() {
...
+      if (!update.fromOnMessage) {
+        channel.postMessage(update.value);
+      }
+      set(baseAtom, update.value);
+    }
+  );
+  broadcastAtom.onMount = (setAtom) => {
+    channel.onmessage = (event) => {
+      setAtom({ fromOnMessage: true, value: event.data });
+    };
```
dai-shi, line 327: we should clean up something. or is it garbage collectable?

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,61 @@ function EvenCounter() {
...
+    (get) => get(baseAtom),
+    (get, set, update) => {
+      if (!update.fromOnMessage) {
+        channel.postMessage(update.value);
+      }
+      set(baseAtom, update.value);
+    }
+  );
+  broadcastAtom.onMount = (setAtom) => {
+    channel.onmessage = (event) => {
```
dai-shi, line 325: Isn't this overwriting event listener if we use multiple providers?

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,61 @@ function EvenCounter() {
...
+
+  return (
+    <div>
+      {count}
+      <button onClick={() => setCount(count + 1)}>+1</button>
+    </div>
+  );
+};
+
+```
```
dai-shi, line 351: Would you please add a codesandbox example? That will be helpful very much.

dai-shi (commented):
here's the first review!

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,61 @@ function EvenCounter() {
...
+  const returnedAtom = atom<Value, Value>(
+    (get) => get(broadcastAtom),
+    (get, set, update) => {
+      set(broadcastAtom, { fromOnMessage: false, value: update });
+    }
+  );
+  return returnedAtom;
+}
+
+const atomBroad = atomWithBroadcast("count", 0);
```
Aslemammad, line 338: Let's follow the general naming of jotai's atoms! like something more meaningful + `atom` in the end. like `broadcastAtom`

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,61 @@ function EvenCounter() {
...
+```ts
+import { atom } from "jotai";
+
+export function atomWithBroadcast<Value>(key: string, initialValue: Value) {
+  const channel = new BroadcastChannel(key);
+
+
+  const baseAtom = atom(initialValue);
+
+  const broadcastAtom = atom<Value, { fromOnMessage: boolean; value: Value }>(
```
Aslemammad, line 315: if you'd like to keep fromOnMessage, let's change the name to `isEvent`!

Aslemammad:
@Sajad-Sharhani Congratulations on this PR! it's exciting!

docs/advanced-recipes/atom-creators.mdx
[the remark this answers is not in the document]
dai-shi: I guess you don't understand what I mean.

Please try something like this:
```js
const baseAtom = atom(initialValue);
baseAtom.onMount = (setAtom) => {
  ...
});

const broacastAtom = atom(
  (get) => get(baseAtom);
  (get, set, update) => {
      set(baseAtom, update);
      channel.postMessage(get(baseAtom));
  }
);
```

This will support function update too.

docs/advanced-recipes/atom-creators.mdx
dai-shi: We would need to pair onMount and onUnmount.
At least, you should only `.close()` when `listerns` is empty.
You can try having two atoms in the example and one of them behind show/hide toggle.

I'm not very familiar with BroadcastChannel. Can we create multiple channels in a single thread?
```js
const c1 = new BroadcastChannel('foo');
const c2 = new BroadcastChannel('foo');
```

If it's too hard to solve, we could get `channel` from function arguments.

dai-shi (changes requested):
second one. let's keep going!

docs/advanced-recipes/atom-creators.mdx
Sajad-Sharhani: solved it in the new commit.
tell me what do u think?

Aslemammad:
@Sajad-Sharhani Any updates on this PR?

docs/advanced-recipes/atom-creators.mdx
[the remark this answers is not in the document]
dai-shi: Oh, okay. I think I missed the fact that `returnedAtom` writes with `fromOnMessage: false`.

docs/advanced-recipes/atom-creators.mdx
dai-shi: on second thought, I don't think this needs to be in `onMount`. Can you move it right after L312?

docs/advanced-recipes/atom-creators.mdx
dai-shi: Oh, I missed that you added `channel.close` in clean up. If we go with this, we need to create `channel` in onMount. Otherwise, it will work unexpectedly. Basically, onMount and cleanup should be paired. Same applies to `useEffect`.

dai-shi (changes requested):
One last thing needs to be adjusted.
So, the question is when do you want to create `new BroadcastChannel()`?
1. when you define an atom, or
2. first time when the atom is used.

Sajad-Sharhani:
> One last thing needs to be adjusted. So, the question is when do you want to create `new BroadcastChannel()`?
> 
> 1. when you define an atom, or
> 2. first time when the atom is used.

When u define an atom.
Because I guess it's possible to call BroadcastChannel multiple times.

dai-shi:
> When u define an atom.

In that case, let's remove `channel.close();` from onMount. It's not the right place.
The limitation is we can't close it. (because we can't "undefine" the atom.)

> Because I guess it's possible to call BroadcastChannel multiple times.

But, you can only create it with the first use. (i.e. listeners.length === 1)
If it works, I prefer this. `new BroadcastChannel` in onMount, and close() on unmount.

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,76 @@ function EvenCounter() {
...
+
+This can be useful when you want the state to interact with each other without
+the use of localStorage and that by just using The Broadcast Channel API allows
+basic communication between browsing contexts (that is, windows, tabs, frames, 
+create a component or iframes) and workers on the same origin. According to the MDN documentation receiving a message in initialization is not supported in broadcast and if we want to support that we may need to add extra stuff to atomWithBroadcast (like local storage).
+
+```tsx
+import { atom } from "jotai";
+
+const channelCache = new Map<string, BroadcastChannel>();
```
dai-shi, line 309: My apologies if my comment was confusing.
We don't need to deal with two `atomWithBroadcast` invocation with the same `key`.
```suggestion
```

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,76 @@ function EvenCounter() {
...
+export function atomWithBroadcast<Value>(key: string, initialValue: Value) {
+ let channel: BroadcastChannel;
+  if (channelCache.has(key)) {
+    channel = channelCache.get(key)!;
+  } else {
+    channel = new BroadcastChannel(key);
+    channelCache.set(key, channel);
+  }
```
dai-shi, line 317:
```suggestion
  const channel = new BroadcastChannel(key);
```

dai-shi (commented):
Thanks for your work. Let's partially revert it.

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,69 @@ function EvenCounter() {
...
+    }
+  );
+  broadcastAtom.onMount = (setAtom) => {
+    const listener = (event: MessageEvent<any>) => {
+      setAtom({ isEvent: true, value: event.data });
+    };
+    listeners.add(listener);
+    channel.onmessage = (event) => {
+      listeners.forEach((l) => l(event));
+    };
```
dai-shi, line 331:
```suggestion
```

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -291,3 +291,69 @@ function EvenCounter() {
...
+basic communication between browsing contexts (that is, windows, tabs, frames, 
+create a component or iframes) and workers on the same origin. According to the MDN documentation receiving a message in initialization is not supported in broadcast and if we want to support that we may need to add extra stuff to atomWithBroadcast (like local storage).
+
+```tsx
+import { atom } from "jotai";
+
+export function atomWithBroadcast<Value>(key: string, initialValue: Value) {
+  const channel = new BroadcastChannel(key);
+  const baseAtom = atom(initialValue);
+  const listeners = new Set<(event: MessageEvent<any>) => void>();
```
dai-shi, line 312:
```suggestion
  const baseAtom = atom(initialValue);
  const listeners = new Set<(event: MessageEvent<any>) => void>();
  const channel = new BroadcastChannel(key);
  channel.onmessage = (event) => {
    listeners.forEach((l) => l(event));
  };
```

dai-shi (approved):
It has a few limitations,
- atom values will not be in sync, if they are mounted at different times (and an event is received in-between).
- there's no way to close the created channel.

but, this is an advanced recipe in docs, which is valuable more in readability.
Let's merge this.

Issue #911, opened 2021-12-28
improve useAtomsSnapshot / useGotoAtomsSnapshot


dai-shi:
Found some issues while developing #837:
- [x] Ignore snapshots with invalidated atoms
- [x] Avoid extra re-renders
- [x] Maybe we should somehow return `dependents` too. (Provider's useDebugState does it)
- [x] Support versionedWrite in useGotoAtomsSnapshot (See useAtomsDevtools.ts)

We would need to write tests too.

Thisen:
Good idea!

Thiesl137:
I'm going through the code-base of Atomic this week. My time has opened up this month to get the product features using snapshots to the chrome store. I'll post here if I have any feedback on improvement.

dai-shi:
@Thiesl137 That would be great! Please also keep #918 posted for any updates or even no updates.

Aslemammad:
@Thiesl137 Let's close this, but your help would be so appreciated, so feel free to re-open it!

dai-shi:
@Aslemammad Actually, I have some ideas written in the description to improve, without having @Thiesl137 feedback.
So, let's do them. Would you like to work on this?

Aslemammad:
@dai-shi I'd be happy, but it'd be good if you good them somewhere in my priority list! so I can work on it soon!

dai-shi:
> Avoid extra re-renders

Currently, `useAtomsSnapshot` has this:
https://github.com/pmndrs/jotai/blob/0682db6e94efe38d4f4d36ff0a0e4ef8b004aad7/src/devtools/useAtomsSnapshot.ts#L38-L41
which can cause extra re-renders when nothing changes.

`useAtomsDevtools` has a guard for it:
https://github.com/pmndrs/jotai/blob/0682db6e94efe38d4f4d36ff0a0e4ef8b004aad7/src/devtools/useAtomsDevtools.ts#L128-L137
see "bail out" condition.

PR #1116, opened 2022-04-22
feat(breaking-change): return dependants with useAtomsSnapshot (v2)

Aslemammad:
This issue address this in #911 
> Maybe we should somehow return dependents too. (Provider's useDebugState does it)

### DIFF
--- a/src/devtools/useAtomsSnapshot.ts
+++ b/src/devtools/useAtomsSnapshot.ts
@@ -6,9 +6,8 @@ import {
   DEV_GET_MOUNTED,
   DEV_GET_MOUNTED_ATOMS,
   DEV_SUBSCRIBE_STATE,
-  Mounted,
 } from '../core/store'
-import type { AtomState, Store } from '../core/store'
+import type { AtomState, Mounted, Store } from '../core/store'
 
 type AtomsValues = Map<Atom<unknown>, unknown>
 type AtomsDependants = Map<Atom<unknown>, Set<Atom<unknown>>>
--- a/src/devtools/useAtomsSnapshot.ts
+++ b/src/devtools/useAtomsSnapshot.ts
@@ -3,12 +3,16 @@ import { SECRET_INTERNAL_getScopeContext as getScopeContext } from 'jotai'
 import type { Atom, Scope } from '../core/atom'
 import {
   DEV_GET_ATOM_STATE,
+  DEV_GET_MOUNTED,
   DEV_GET_MOUNTED_ATOMS,
   DEV_SUBSCRIBE_STATE,
+  Mounted,
 } from '../core/store'
 import type { AtomState, Store } from '../core/store'
 
-type AtomsSnapshot = Map<Atom<unknown>, unknown>
+type AtomsValues = Map<Atom<unknown>, unknown>
+type AtomsDependants = Map<Atom<unknown>, Set<Atom<unknown>>>
+type AtomsSnapshot = readonly [AtomsValues, AtomsDependants]
 
 const createAtomsSnapshot = (
   store: Store,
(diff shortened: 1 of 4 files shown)


src/core/store.ts
```diff
@@ -84,7 +84,7 @@ type Dependents = Set<AnyAtom>
  *
  * The mounted state of an atom is freed once it is no longer mounted.
  */
-type Mounted = {
+export type Mounted = {
```
dai-shi, line 87: Please don't export internal types.

src/devtools/useAtomsSnapshot.ts
```diff
@@ -3,12 +3,15 @@ import { SECRET_INTERNAL_getScopeContext as getScopeContext } from 'jotai'
...
   DEV_GET_MOUNTED_ATOMS,
   DEV_SUBSCRIBE_STATE,
 } from '../core/store'
-import type { AtomState, Store } from '../core/store'
+import type { AtomState, Mounted, Store } from '../core/store'
 
-type AtomsSnapshot = Map<Atom<unknown>, unknown>
+type AtomsValues = Map<Atom<unknown>, unknown>
+type AtomsDependants = Map<Atom<unknown>, Set<Atom<unknown>>>
+type AtomsSnapshot = readonly [AtomsValues, AtomsDependants]
```
dai-shi, line 14: So, we are changing our terminology. We will need to follow it in `useGotoAtomsSnapshot`.

src/devtools/useAtomsSnapshot.ts
```diff
@@ -18,7 +21,12 @@ const createAtomsSnapshot = (
     const atomState = store[DEV_GET_ATOM_STATE]?.(atom) ?? ({} as AtomState)
     return [atom, 'v' in atomState ? atomState.v : undefined]
   })
-  return new Map(tuples)
+  const dependants = atoms.map<[Atom<unknown>, Set<Atom<unknown>>]>((atom) => {
+    const mounted = store[DEV_GET_MOUNTED]?.(atom) ?? ({} as Mounted)
+    return [atom, mounted.t]
```
dai-shi, line 26: I guess you can simply do this?
```suggestion
    const mounted = store[DEV_GET_MOUNTED]?.(atom)
    return [atom, mounted?.t]
```

tests/devtools/useGoToAtomsSnapshot.test.tsx
```diff
@@ -20,7 +20,7 @@ it('[DEV-ONLY] useGotoAtomsSnapshot should modify atoms snapshot', async () => {
   }
 
   const UpdateSnapshot = () => {
-    const snapshot = useAtomsSnapshot()
+    const [snapshot] = useAtomsSnapshot()
```
dai-shi, line 23: So, I mean we don't change this, but change useGotoAtomsSnapshot(). Luckily, for this one, we can make it backward compatible with `Array.isArray()`.

dai-shi (changes requested):
Looks like a good first step. Added some comments.

src/devtools/useGotoAtomsSnapshot.ts
```diff
@@ -13,9 +22,17 @@ export function useGotoAtomsSnapshot(scope?: Scope) {
   }
 
   return useCallback(
-    (values: Iterable<readonly [Atom<unknown>, unknown]>) => {
-      store[RESTORE_ATOMS](values)
+    (values: Iterable<readonly [Atom<unknown>, unknown]> | AtomsSnapshot) => {
+      if (isIterable(values)) {
```
dai-shi, line 26: Oh, I see. We used to support Iterable not just Map. Seems fair.

src/devtools/useGotoAtomsSnapshot.ts
```diff
@@ -13,9 +22,17 @@ export function useGotoAtomsSnapshot(scope?: Scope) {
...
+        store[RESTORE_ATOMS](values)
```
dai-shi, line 27: Let's add a deprecation message.
```suggestion
        if (__DEV__) {
          console.warn('snapshot as iterable is no longer supported. use an object instead.')
        }
        store[RESTORE_ATOMS](values)
```

dai-shi (commented):
Looks good. one change request.

dai-shi (approved):
LGTM
Note to anyone: This is a PR to a branch. We will create another PR later to main.

Issue #1114, opened 2022-04-21
Problems with derived atom which has a promise as its initial value


riemonyamada:
I played around with the [codesandbox](https://codesandbox.io/s/nervous-fermi-j89jhw) in this issue https://github.com/pmndrs/jotai/issues/1054#issuecomment-1093566184 to understand the behavior when the atom has `Promise` as its initial value.

Then, I have noticed that this pull request https://github.com/pmndrs/jotai/pull/1089 does not solve the problem in many cases, and that even in cases where it appears to solve the problem, there is a another problem.

### Cases where the problem is not resolved

This is just one atom added to the first codesandbox. Now the `atomWithSuspense` no longer works.

https://codesandbox.io/s/vigorous-roman-45dbx5?file=/src/App.js

After digging deeper into the issue, I realized that this behavior was the expected behavior.

### Why this codesandbox is the expected behavior

The reason why `atomWithSuspense` does not work when the initial value is a infinite `Promise` is because it is not registered in the `Dependents` of `nullableAtom`.

Normally, a derived atom is registered in `Dependents` of dependencies at `store[SUBSCRIBE_ATOM]` in `useAtomValue`, but if initial value is a `Promise`, it is not executed because `throw Promise` occurs before that. So, `atomWithSuspense` will not registered in the `Dependents` of `nullableAtom` and will not work.

### Why does the first codesandbox still work

Basically, the only chance for a derived atom to be registered in dependencies' `Dependents` is when it is mounted for the first time. However, there is one more place where this can happen. That is `flushPending`.

Inside `flushPending`, atoms registered in the `pendingMap` are registered in dependencies' `Dependents` under certain conditions. Note that atoms that are not mounted are also registered in the `pendingMap`. Atoms are registered in the `pendingMap` when they are first obtained, so `atomWithSuspense`, which is not mounted, is also registered in the `pendingMap`.

The conditions under which atoms are registered as `Dependents` are below:

- The dependencies of the atom must be changed from the last time
- The dependencies must have been mounted

For this to work, `nullableAtom` must be mounted before the first `flushPending`, as in the first codesandbox. In other words, `nullableAtom` must be the first atom to be mounted.

### What is a another problem in cases where it works

[...]

Issue #1054, opened 2022-03-09
Derived atom getting stuck in suspense when initial value is a Promise that does not resolve


carlos-marchal:
We are starting to use jotai for state management on an existing codebase. One pattern we currently use for data loading is to have a state be null when loading, and hold the data when loaded. So we check if `state === null` to display loading components, and display the data itself (`state`) if the state is not null.

With jotai, we want to start using suspense, so what we do in these cases is as follows:
1. Use an atom to store the original nullable state
1. Derive a read only atom from the step 1 atom, where we return the data if not null, or a promise that never resolves otherwise (`new Promise(() => {})`). 

This makes any component reading the value of atom 2 suspend while the value of the atom 1 is null, and render the data when there atom 1 is not null. So far so good. But the issue appears when the initial state for atom 1 is null, then even after changing it to some other value any component rendering atom 2 remains suspended, and the read function of the atom 2 is not executed.

This sandbox shows the issue: https://codesandbox.io/s/keen-snowflake-cy88ek?file=/src/App.js. Try changing the `INITIAL_VALUE` between a number or null, and see the difference in behavior when fetching new values.

I suspect there must be something that changes with how suspend and jotai interact, depending on whether the component that used the atom has been able to render at least once or is suspended before the first render. I don't really know what the proper expected behavior would be here, as I am not an expert on suspense or jotai, I'm afraid I can't provide more information.

dai-shi:
The infinite promise `new Promise(() => {})` is really hard to deal with. We do support some cases (with tricky implementation), but this one is not covered.
It would be nice if this can be supported, but I'm not sure if it's possible/how easy at this point.

On the other hand, I don't think this pattern will be common (or recommended) as the use case of Suspense. We don't know yet. Will see.

carlos-marchal:
Sounds good, thanks for the response. I've been taking a look at jotai's implementation of SuspensePromise, and even though it's quite elegant, I can see how you'd run into dangerous edge cases.

As I said, I'm not an expert on jotai or suspense, so maybe I'm taking the wrong approach. But can I ask, in your opinion, how you would handle a case such as mine? Refactoring the whole fetching library would be just too much work right now, and I can't find any other incremental approach to handle my use case with suspense.

dai-shi:
[...]

PR #1118, opened 2022-04-24
fix(utils): resolve infinite loop with loadable of derived async atom in some cases

riemonyamada:
fix https://github.com/pmndrs/jotai/issues/1114


dai-shi (approved):
Thanks for your investigation!
So, this reverts #1089 which was a wrong fix, and we claim #1054 is an expected behavior.

Issue #1068, opened 2022-03-20
[atomWithQuery | atomWithInfiniteQuery]: return all react query state.


sondh0127:
Currently query atom only return the {data} from useQuery hooks.
It would be better to return all other states as well. 
![image](https://user-images.githubusercontent.com/62163604/159172388-32fc3f28-5423-4926-8220-24d9bb9122c4.png)
https://react-query.tanstack.com/reference/useQuery

dai-shi:
Hi, thanks for the suggestion.
The point of `jotai/query` is to make RQ as jotai atoms, and jotai atoms are based on Suspense for async operations.
That means, we shouldn't return loading/pending states with properties, and not for errors which are handled with ErrorBoundary. Also, actions are for `write` function, and we can `refetch`.
That will leave only a few properties out of listed ones, and I would rather like to hear some real use cases, because some of them are already covered by jotai methods.

Hope it makes sense!

sondh0127:
Thank you for your explanation. @dai-shi 
I'm facing a real use-case here with atomWithInfiniteQuery, that I need to get `hasNextPage` to display loadmore button.
```ts
infiniteMetaData.onMount = (update) => {
	  setInfiniteData = update
	  if (options.enabled !== false) {
	    return observer.subscribe(listener2)
	  }
}
```
I tried to add second atom and listen to the same observer of infiniteData does. and return it from wrapper atom.

```ts
const queryAtom = atom<
    { data: InfiniteData<TData | TQueryData> | undefined } & InfiniteMetaData,
    AtomWithInfiniteQueryAction<TQueryFnData>
  >(
    (get) => {
      const { dataAtom, infiniteMetaData } = get(queryDataAtom)
      return { data: get(dataAtom), ...get(infiniteMetaData) }
    },
    (_get, set, action) => set(queryDataAtom, action) // delegate action
  )
```
It's work but I still got some trouble with nextjs ssr. 
I would like to ask the author (you) for an official solution

sondh0127:
> Thank you for your explanation. @dai-shi I'm facing a real use-case here with atomWithInfiniteQuery, that I need to get `hasNextPage` to display loadmore button.
> 
> ```ts
> infiniteMetaData.onMount = (update) => {
> 	  setInfiniteData = update
> 	  if (options.enabled !== false) {
> 	    return observer.subscribe(listener2)
> 	  }
> }
> ```
> 
[...]

PR #1126, opened 2022-04-28
feat(query): migrate to tanstack/query-core v4 & make QueryKey's type generic

liaoliao666:
BREAKING CHANGES:
- It depends on `@tanstack/query-core`.
- atomWithInfiniteQuery's refetch action has `payload`.


src/query/atomWithInfiniteQuery.ts
```diff
@@ -14,28 +14,46 @@ import { queryClientAtom } from './queryClientAtom'
 import { CreateQueryOptions, GetQueryClient } from './types'
 
 export type AtomWithInfiniteQueryAction<TQueryFnData> =
-  | ({ type: 'refetch' } & Partial<
-      RefetchOptions & RefetchQueryFilters<TQueryFnData>
-    >)
+  | {
+      type: 'refetch'
+      payload: Partial<RefetchOptions & RefetchQueryFilters<TQueryFnData>>
```
dai-shi, line 19: This is breaking change. We want to avoid breaking changes unless they are super necessary.
Unless `type` is conflicting, we shouldn't need to change this.

dai-shi (changes requested):
Thanks for the suggestion!
This is the first comment. Will have another look after the fix.

src/query/atomWithInfiniteQuery.ts
liaoliao666: its already conflicted with QueryTypeFilter of ReactQuery V4, maybe u can see this link for the detail
https://github.com/tannerlinsley/react-query/blob/fdbc002006d7d8c3e236da71cb0c23cf394c1835/src/core/utils.ts#L19

dai-shi: Oh, I wasn't aware that. That's too bad.

src/query/atomWithInfiniteQuery.ts
```diff
@@ -192,8 +233,8 @@ export function atomWithInfiniteQuery<
       const { observer } = get(queryDataAtom)
```
dai-shi, line 233: On second thought, as [useInfiniteQuery](https://react-query.tanstack.com/reference/useInfiniteQuery) doesn't support customizing options anyway. Maybe, we don't need to specify it.
How about using the same `options`?
```suggestion
      const { observer, options } = get(queryDataAtom)
```

Or, we don't even need to pass `options` to `overserver.refetch()`?

src/query/atomWithInfiniteQuery.ts
```diff
@@ -166,7 +200,14 @@ export function atomWithInfiniteQuery<
         }
       }
 
-      const defaultedOptions = queryClient.defaultQueryObserverOptions(options)
+      const defaultedOptions = queryClient.defaultQueryOptions(
+        options
+      ) as unknown as InfiniteQueryObserverOptions<
```
dai-shi, line 205: How would types break without this assertion?

src/query/atomWithInfiniteQuery.ts
liaoliao666: I'm not very similar with this `useInfiniteQuery`, u could deal this with ur  thought

src/query/atomWithInfiniteQuery.ts
dai-shi: Thanks. Why was it working previously? Do you think it's the issue of RQ?

src/query/atomWithInfiniteQuery.ts
dai-shi: @aulneau Hey, I wonder if we can completely remove options for `type: "refetch"`. What do you think? Are there any use cases?

aulneau: hmm i'll need to dig into v4 changes of react-query a bit, but I think it could be possible to remove these and instead build out that atom we mentioned in a different thread. the atom would contain all the status/actions associated with a given query atom, rather than combining them. see here: https://github.com/pmndrs/jotai/issues/1068#issuecomment-1101737345

src/query/atomWithInfiniteQuery.ts
liaoliao666: Because i passed QueryKey to InfiniteQueryObserverOptions, so that returns a type that contains QueryKey, but InfiniteQueryObserver's constructor that can not accept a options that contains QueryKey

src/query/atomWithInfiniteQuery.ts
dai-shi: This is slightly different from #1068. Here, it's our action that we defined. It's not about the returned status. I just found `useInfiniteQuery` from RQ doesn't even have `refetch` action.

src/query/atomWithInfiniteQuery.ts
dai-shi: > can not accept a options that contains QueryKey

That sounds unlikely unless it a tricky type. Can you then use `Omit<>` to remove `QueryKey`? I really hope to remove `as unknown`.

src/query/atomWithInfiniteQuery.ts
liaoliao666: yep, u r right.

src/query/atomWithInfiniteQuery.ts
liaoliao666: I tried, but its impossible. maybe u can wait for this issue to be merged and published then remove this trick
https://github.com/tannerlinsley/react-query/pull/3571

dai-shi:
@aulneau is busy??

src/query/atomWithInfiniteQuery.ts
```diff
@@ -5,37 +5,49 @@ import type {
...
-  RefetchOptions,
-  RefetchQueryFilters,
 } from 'react-query'
 import { atom } from 'jotai'
 import type { WritableAtom } from 'jotai'
 import { queryClientAtom } from './queryClientAtom'
 import { CreateQueryOptions, GetQueryClient } from './types'
 
-export type AtomWithInfiniteQueryAction<TQueryFnData> =
-  | ({ type: 'refetch' } & Partial<
```
dai-shi, line 17: I started think that we need `refetch` action to recover from errors. Can't we keep it?

tests/query/atomWithInfiniteQuery.test.tsx
```diff
@@ -216,76 +216,3 @@ it('infinite query with enabled 2', async () => {
   fireEvent.click(getByText('set enabled'))
   await findByText('slug: hello-world')
 })
-
-// adapted from https://github.com/tannerlinsley/react-query/commit/f9b23fcae9c5d45e3985df4519dd8f78a9fa364e#diff-121ad879f17e2b996ac2c01b4250996c79ffdb6b7efcb5f1ddf719ac00546d14R597
-it('should be able to refetch only specific pages when refetchPages is provided', async () => {
```
dai-shi, line 221: Why can't we keep this test?

dai-shi (commented):
My second review while waiting a response from @aulneau .

tests/query/atomWithInfiniteQuery.test.tsx
liaoliao666: > Why can't we keep this test?

https://github.com/pmndrs/jotai/pull/1126#discussion_r861380333
u should see this reply u propose😂, so i remove the test of `refetch`.
but oddly i found that `useInfiniteQuery` still returns `refetch` method after my check just now
so we should still keep this?

dai-shi: 😿 Yeah, sorry, I change my mind...
I start to think about #1199.
So, this PR should be minimal to move forward.
I feel like we need some kind of `refetch`, which can be different from what we have now, but it's out of the scope of this PR anyway.

package.json
```diff
@@ -241,7 +241,7 @@
     "prettier": "^2.6.2",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-query": "^3.39.1",
+    "react-query": "^4.0.0-beta.23",
```
dai-shi, line 244: It would probably be a good idea (although we can't test with v3 anymore.)
I'm working on #1239 now and will add more tests, so once it's ready, we'll be able to run the new tests.

liaoliao666: you're replied so soon, are u live in github haha

liaoliao666: https://github.com/pmndrs/jotai/pull/1239
This is amazing about the enhancement, Thanks for ur work. I'll gladly see how do u solve the problem of error handling, cuz i tried but failed😓

src/query/atomWithQuery.ts
```diff
@@ -150,7 +177,7 @@ export function atomWithQuery<
           setData(result.data)
         }
       }
-      const defaultedOptions = queryClient.defaultQueryObserverOptions(options)
+      const defaultedOptions = queryClient.defaultQueryOptions(options)
```
dai-shi, line 180: Is this supported in RQ v3 too?

liaoliao666: yep, also supported with RQ v3

dai-shi: Thanks. Alright, I think I should include this change in #1239 (or something different), so that we can run tests with RQ v3.

dai-shi:
@liaoliao666 Hi, #1239 is merged. Can you make this to base on it?
I just noticed RQ v4 is released 15 hours ago.

dai-shi:
I should have been clearer.
@liaoliao666 Can you merge main and make the diff minimal?

package.json
```diff
@@ -241,7 +241,7 @@
     "prettier": "^2.6.2",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-query": "^3.39.1",
+    "react-query": "^4.0.0",
```
dai-shi, line 244: Actually, we should use `@tanstack/query-core`.
Are you available to work on it?
(Anyway, I'll cut a release before that, with RQ v3.)

liaoliao666: Also migrated to `@tanstack/query-core`.
I haven't focus on this library lately. I thought it was still `react-query` and It's my fault😂

src/query/atomWithInfiniteQuery.ts
```diff
@@ -1,45 +1,55 @@
...
 import { CreateQueryOptions, GetQueryClient } from './types'
 
-type AtomWithInfiniteQueryAction<TQueryFnData> =
-  | ({ type: 'refetch' } & Partial<
-      RefetchOptions & RefetchQueryFilters<TQueryFnData>
-    >)
+export type AtomWithInfiniteQueryAction<TQueryFnData> =
+  | {
+      type: 'refetch'
+      payload: Partial<RefetchOptions & RefetchQueryFilters<TQueryFnData>>
```
dai-shi, line 18: This is a breaking change. Is it intentional?

dai-shi (commented):
The diff looks clean. Thanks.

dai-shi:
We also need to change the docs.

src/query/atomWithInfiniteQuery.ts
liaoliao666: yep. we have discussed this a few years ago

dai-shi: It's been a while... 😅

src/query/atomWithInfiniteQuery.ts
```diff
@@ -168,8 +194,8 @@ export function atomWithInfiniteQuery<
             state.unsubscribe?.()
             state.unsubscribe = null
           }
-          const { type: _type, ...rest } = action
-          observer.refetch(rest).then((result) => {
+          const { type: _type, payload } = action
+          observer.refetch(payload).then((result) => {
```
dai-shi, line 198:
```suggestion
          observer.refetch(action.payload).then((result) => {
```

dai-shi (changes requested):
The code looks good. (I do worry about too complicated generic types though.)

The remaining is to make the doc to follow the change.

dai-shi (approved):

PR #1130, opened 2022-04-28
fix(urql): atomWithQuery: improve isOperationResultWithData to avoid `undefined` in case of error

svilen-ivanov:
Despite the runtime check (`isOperationResultWithData`), there is a key `data` in the result, but the value is `undefined` in case of error. Let's retain optional `data` definition from `urql` `OperationResult`

Screenshot_20220429_001040

### DIFF
--- a/src/urql/atomWithQuery.ts
+++ b/src/urql/atomWithQuery.ts
@@ -18,11 +18,14 @@ type AtomWithQueryAction = {
 type OperationResultWithData<Data, Variables> = OperationResult<
   Data,
   Variables
->
+> & {
+  data: Data
+}
 
 const isOperationResultWithData = <Data, Variables>(
   result: OperationResult<Data, Variables>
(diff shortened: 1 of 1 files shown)


dai-shi:
Hi, thanks for the suggestion.
Can we fix the issue with `isOperationResultWithData`?
Something like this:
```js
const isOperationResultWithData = <Data, Variables>(
  result: OperationResult<Data, Variables>
): result is OperationResultWithData<Data, Variables> => 'data' in result && !result.error
```

We'd like to be strict about atom types.

dai-shi:
From the type definition, `Data` can be `undefined` and that's why I used `'data' in result`. Does URQL require `Data` to be an object?

If `'data' in result && !result.error` works, let's do it. I wonder if it's possible to write a test.

dai-shi (approved):
I hope we could add a test for this, but the code looks good as is.
If you can write a test, please open a follow-up PR.

PR #1139, opened 2022-05-04
fix(core): improve re-render behavior with useTransition

### LINKED ISSUE #1125: (react 18) Extra commit with useTransition
https://codesandbox.io/s/bold-antonelli-04z766?file=/src/App.js

Initial log:
```
commit {isLoading: false, resolvedThree: 3}
```

After clicking the button:
```
commit {isLoading: true, resolvedThree: 3}
commit {isLoading: true, resolvedThree: 4}
commit {isLoading: false, resolvedThree: 4}
```

The second commit with `{isLoading: true, resolvedThree: 4}` should not exist, and it's how `useState` behaves.

dai-shi:
close #1125

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -427,7 +427,6 @@ export const createStore = (
         promiseOrValue
           .then((value: Awaited<Value>) => {
             setAtomValue(version, atom, value, dependencies, suspensePromise)
-            flushPending(version)
           })
           .catch((e) => {
             if (e instanceof Promise) {
@@ -441,7 +440,6 @@ export const createStore = (
               return e
             }
             setAtomReadError(version, atom, e, dependencies, suspensePromise)
-            flushPending(version)
           })
       )
       return setAtomSuspensePromise(
--- a/tests/transition.test.tsx
+++ b/tests/transition.test.tsx
@@ -7,48 +7,53 @@ const Provider = getTestProvider()
 
 jest.mock('../src/core/useDebugState.ts')
 
-it('no extra commit with useTransition (#1125)', async () => {
-  const countAtom = atom(0)
-  const delayedAtom = atom(async (get) => {
-    await new Promise((r) => setTimeout(r, 100))
-    return get(countAtom)
-  })
+const describeWithUseTransition =
+  typeof useTransition === 'function' ? describe : describe.skip
 
-  const commited: { pending: boolean; delayed: number }[] = []
-
-  const Counter = () => {
-    const setCount = useSetAtom(countAtom)
-    const delayed = useAtomValue(delayedAtom)
-    const [pending, startTransition] = useTransition()
-    useEffect(() => {
-      commited.push({ pending, delayed })
+describeWithUseTransition('useTransition', () => {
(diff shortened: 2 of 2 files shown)


dai-shi:
@riemonyamada Hi, do you happen to be interested in this issue?

riemonyamada:
Hi @dai-shi. I'm not very familiar with React 18, but I'll try it next week.

riemonyamada:
The cause of this problem is that `flushPending` is triggered from within `suspensePromise`. Updating state in an asynchronous context is reflected immediately because it is not recognized as triggered from within `startTransition`.

Removing [`flushPending`](https://github.com/pmndrs/jotai/blob/17de495d04b28353b3d3b774d40d7510920e8afb/src/core/store.ts#L430) from `suspensePromise` solves this problem.

I don't think this change to cause any other problems, because the resolution of `suspensePromise` triggers a rerender anyway, but do you have any concerns?

dai-shi:
@riemonyamada Thanks for the investigation.
I guess there was a reason we have `flushPending` there, but not sure.
Let's see how tests pass/fail with removing it.

dai-shi:
Although I had a occasional error locally, CI passes without problems.

```
 FAIL  tests/dependency.test.tsx
  ● works a primitive atom and a dependent async atom

    Unable to find an element with the text: count: 3, doubled: 6. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

    Ignored nodes: comments, <script />, <style />
    <body>
      <div>
        <div
          style="display: none;"
        >
          count: 
          2
          , doubled: 
          4
        </div>
        <button
          style="display: none;"
        >
          button
        </button>
        loading
      </div>
    </body>

      85 |   fireEvent.click(getByText('button'))
      86 |   await findByText('loading')
    > 87 |   await findByText('count: 3, doubled: 6')
         |         ^
      88 | })
      89 |
      90 | it('should keep an atom value even if unmounted', async () => {

      at waitForWrapper (node_modules/@testing-library/dom/dist/wait-for.js:187:27)
      at node_modules/@testing-library/dom/dist/query-helpers.js:101:33
      at findByText (tests/dependency.test.tsx:87:9)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:63:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:294:22)
      at Generator.next (node_modules/regenerator-runtime/runtime.js:119:21)
      at asyncGeneratorStep (tests/dependency.test.tsx:19:28)
      at _next (tests/dependency.test.tsx:37:17)


Test Suites: 1 failed, 49 passed, 50 total
Tests:       1 failed, 10 skipped, 274 passed, 285 total
Snapshots:   19 passed, 19 total
Time:        11.918 s
```

dai-shi (commented):
The change looks okay. I hope our tests cover well. 🤞 
Thanks @riemonyamada !

dai-shi:
https://codesandbox.io/s/boring-torvalds-cn7lxn
confirmed the original sandbox is also fixed.

Issue #1112, opened 2022-04-21
Infinite loop when Suspense contains a Provider with initialValues set


macarie:
I played around with Jotai to create a (custom) translation library.

We need to show some app components using a different locale, so I thought we could do this by defining a `scope` as `useAtom`'s second argument and providing the locale through `Provider`'s `initialValues` and `scope`.

I have two `atom`s: `localeAtom` and `translationsAtom`.

`translationsAtom` is async and does a `get(localeAtom)`, then checks if the translations for that `locale` are already cached. If they are, it returns them; otherwise, it imports them, sets the cache, and returns them.

Everything works fine. I can change the `localeAtom`'s value, and the `translationsAtom`'s value changes.

The problem arises when I use the `Provider` component and set `initialValues`; at this point, `translationsAtom` enters an infinite loop.

I've set up a [barebones example here](https://codesandbox.io/s/jotai-loop-async-atom-provider-pcnnh1?file=/src/index.tsx), without the custom hook implementation and the use of `scope` for simplicity's sake. I've left some logs in there to see where it falls apart. To reproduce the issue is enough to uncomment the lines where the `Provider` is declared (`index.tsx:13` and `index.tsx:15` if the link doesn't open the correct file).

dai-shi:
The example is too long, but I guess the problem is `<Suspense>` is not inside `<Provider>`. Hope it helps.

macarie:
Oh, yeah, that was the problem!

I've just found the paragraph about that in the docs. I think I've missed it before.

Thanks for the help, and sorry for having wasted your time 😅

Do you think it would be possible to detect this scenario and add a warn/error to the console with a link to the docs, like React does for some common issues? Having an infinite loop doesn't feel right, and honestly, moving the `Suspense` component inside the `Provider` didn't even cross my mind, but this might be because `Suspense` is kinda new.

dai-shi:
Infinite loop detection would generally be hard. Maybe possible with some heuristics. If anyone has a not-too-complex idea, please let us know.

What we should do at least is to improve the docs to explicitly tell (or even warn) the usage of `Suspense`. Do you have any suggestion?

macarie:
I've tried debugging this issue because I thought it was weird that it doesn't work only with `initialValues`.

I've found out that what causes the infinite loop is this line of code: https://github.com/pmndrs/jotai/blob/3f7d555355bba3d8743915a8b7f8fda4e0eb0991/src/core/Provider.ts#L46

[...]

PR #1142, opened 2022-05-05
docs: warn about usage of `Provider` inside `Suspense`

macarie:
As discussed in #1112, I've made the warning about using a `Provider` inside a `Suspense` a bit more noticeable.

Initially, I wanted to create a section with a heading and the paragraph, but I think the yellow box (blockquote) inside the "Suspense" paragraph makes more sense.

Do you think this is okay? For example, should I write what could happen if the `Provider` contains no `Suspense`?

### DIFF
--- a/docs/basics/async.mdx
+++ b/docs/basics/async.mdx
@@ -12,7 +12,7 @@ Async support is first class in jotai. It fully leverages React Suspense.
 
 To use async atoms, you need to wrap your component tree with `<Suspense>`.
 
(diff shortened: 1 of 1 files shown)


dai-shi:
> Do you think this is okay? For example, should I write what could happen if the `Provider` contains no `Suspense`?

Thanks for the suggestion. Yeah, actually, a small note about what could happen seems good to add.

I can merge this as is, if you want.

dai-shi (approved):
Thanks! Let's merge this.

PR #1157, opened 2022-05-14
fix(utils): atomWithObservable: fix ReferenceError with BehaviourSubject (#1156)

### LINKED ISSUE #1156: [atomWithObservable] Uncaught ReferenceError: Cannot access 'subscription' before initialization
If I'm using `atomWithObservable` with the `BehaviourSubject` from rjxs, I get the following error:

image

Minimal reproducible example:

```typescript
import { BehaviorSubject } from "rxjs";
import { atomWithObservable, useAtomValue } from "jotai/utils";

const number$ = new BehaviorSubject(1);
const numberAtom = atomWithObservable(() => number$);

  const number = useAtomValue(numberAtom);
  return <div>{number}</div>;
}
```

Based on my analysis, the error is caused here: https://github.com/pmndrs/jotai/blob/main/src/utils/atomWithObservable.ts#L124-L131

The `subscription` is accessed, before the variable is declared.

To fix the issue, we should declare the variable beforehand:

```typescript
let subscription: Subscription | undefined = undefined;
subscription = source.subscribe({
  next: (value) => {
    resolve(value)
    resolved = true
    if (subscription) {
      subscription.unsubscribe()
    }
  },
  error: reject,
  complete: () => {
    reject()
  },
})
```

I can create an PR to fix the issue.

TobiasWalle:
Fixes #1156

### DIFF
--- a/.swcrc
+++ b/.swcrc
@@ -1,6 +1,6 @@
 {
   "jsc": {
-    "target": "es5",
+    "target": "es2015",
     "parser": {
       "syntax": "typescript",
       "tsx": true
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -121,12 +121,12 @@ function getInitialValue<TData>(options: AtomWithObservableOptions<TData>) {
 function firstValueFrom<T>(source: ObservableLike<T>): Promise<T> {
   return new Promise<T>((resolve, reject) => {
     let resolved = false
-    let subscription: Subscription | undefined = undefined
-    subscription = source.subscribe({
+    let isSync = true
+    const subscription = source.subscribe({
       next: (value) => {
         resolve(value)
         resolved = true
-        if (subscription) {
+        if (!isSync) {
           subscription.unsubscribe()
         }
       },
@@ -135,6 +135,7 @@ function firstValueFrom<T>(source: ObservableLike<T>): Promise<T> {
         reject()
       },
     })
+    isSync = false
 
     if (resolved) {
       // If subscription was resolved synchronously
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -121,7 +121,8 @@ function getInitialValue<TData>(options: AtomWithObservableOptions<TData>) {
 function firstValueFrom<T>(source: ObservableLike<T>): Promise<T> {
   return new Promise<T>((resolve, reject) => {
     let resolved = false
-    const subscription = source.subscribe({
+    let subscription: Subscription | undefined = undefined
+    subscription = source.subscribe({
       next: (value) => {
         resolve(value)
         resolved = true
--- a/tests/utils/atomWithObservable.test.tsx
+++ b/tests/utils/atomWithObservable.test.tsx
@@ -1,6 +1,6 @@
 import { ReactElement, Suspense, useState } from 'react'
 import { act, fireEvent, render } from '@testing-library/react'
-import { Observable, Subject } from 'rxjs'
+import { BehaviorSubject, Observable, Subject } from 'rxjs'
 import { useAtom } from 'jotai'
 import { atomWithObservable } from 'jotai/utils'
 import { getTestProvider } from '../testUtils'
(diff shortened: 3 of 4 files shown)


dai-shi (changes requested):
Thanks for your investigation! The change looks good if it solves the issue.
Can you add a test in `tests/utils/atomWithObservable.test.tsx`?

TobiasWalle:
Sure!

TobiasWalle:
@dai-shi I added the test.

To reproduce the behaviour, I needed to change the target version in swc from "es5" to "es2015". Otherwise the `const` variables are probably transpiled to `var` and the issue doesn't occur.

I hope that's okay.

dai-shi:
Okay, I'll make a change.

.swcrc
```diff
@@ -1,6 +1,6 @@
 {
   "jsc": {
-    "target": "es5",
+    "target": "es2015",
```
dai-shi, line 3: Seems like this breaks other tests in atomWithObservable.test.tsx.

src/utils/atomWithObservable.ts
```diff
@@ -121,11 +121,12 @@ function getInitialValue<TData>(options: AtomWithObservableOptions<TData>) {
 function firstValueFrom<T>(source: ObservableLike<T>): Promise<T> {
   return new Promise<T>((resolve, reject) => {
     let resolved = false
+    let isSync = true
```
dai-shi, line 124: this is just to make it more explicit and more consistent with other code.

tests/utils/atomWithObservable.test.tsx
```diff
@@ -41,7 +41,7 @@ it('writable count state', async () => {
     // is this usual to delay the subscription?
     setTimeout(() => {
       observable.subscribe(subject)
-    }, 500)
+    }, 100)
```
dai-shi, line 44: This test is still unstable. This might not affect anything, but just hoping.

dai-shi (approved):
Looks good to me.

@TobiasWalle please check the changes.

src/utils/atomWithObservable.ts
TobiasWalle: LGTM

tests/utils/atomWithObservable.test.tsx
TobiasWalle: I analysed the issue. The problem is, that the `onMount` subscription is created __after__ the dispatch of the button. My logs (I hope it makes sense):

image

I can look into this issue this evening.

dai-shi: Thanks for taking a look. Please work on a separate PR as I will merge this some time eventually.

btw, contributions on `atomWithObservable` are very welcome, because currently no contributors on it are active except for you. Please keep it going.

TobiasWalle: Sounds good to me!

.swcrc
dai-shi: Later noticed this change breaks some CI tests. Worked around: b996394

TobiasWalle:
I added a small workaround to fix the flaky test:
https://github.com/pmndrs/jotai/pull/1157/commits/fab0fa953a5676632b0f8fc87809763775c26258

As discussed, I will create a separate PR to fix the real underlying issue.

If it is okay for you the MR can be merged now.

tests/utils/atomWithObservable.test.tsx
```diff
@@ -37,11 +37,11 @@ it('writable count state', async () => {
     const observable = new Observable<number>((subscriber) => {
       subscriber.next(1)
     })
-    const subject = new Subject<number>()
+    const subject = new ReplaySubject<number>()
     // is this usual to delay the subscription?
```
dai-shi, line 41: By changing from `Subject` to `ReplaySubject`, this comment is no longer valid or is it still?

TobiasWalle: I'm not sure. If the `setTimeout` is removed, it should still work, but then the value is immediately available on the first subscription instead of after a small delay, which changes the test case a little bit. I would then behave similar to a `BehaviourSubject`.

dai-shi: Okay, I'm asking a little differently. This test is probably what I made without knowing much about observable usages. So, the question is if this reflects a typical usage, or at least something we should cover. Adding more tests is also welcome.

In any case, we can work on a new PR.

TobiasWalle: A more typical usage would be the following:
```typescript
const observableAtom = atomWithObservable(() => {
  const subject = new ReplaySubject<number>()
  subject.next(1);
  return subject
})
```

> In any case, we can work on a new PR.

+1. I'm happy to help you with that. I worked a lot with Angular in the past, so I have some experience with RXJS.

PR #1165, opened 2022-05-16
Implement atomWithSuspense util

eolme:


docs/utils/atom-with-suspense.mdx
```diff
@@ -0,0 +1,12 @@
...
+---
+
+## Usage
+
+```ts
+atomWithSuspense<Value>(): WritableAtom<Value | Promise<Value>, SetStateAction<Value>
+```
+
+This creates a new atom that suspend by default.
+When the value is set for the first time it will act like a normal atom.
```
dai-shi, line 12: We cover this case with this pattern:
```js
const anAtom = atom(new Promise<Value>(() => {}))
```

I think this is probably a doc issue. Feel free to open a new PR for improving docs.

Having that said, providing a shorthand with an explicit function name can be a good idea. This can fall into my 3rd-party package: https://github.com/jotai-labs/jotai-suspense Would you be interested in contributing there? I would maybe call it `atomWithPending`.

src/utils/atomWithSuspense.ts
```diff
@@ -0,0 +1,26 @@
+import { atom } from 'jotai'
+import type { SetStateAction } from 'jotai'
+
+export function atomWithSuspense<Value>() {
+  const SUSPENSE = Symbol()
+  let ready: (value: Value) => void
```
dai-shi, line 6: Implementation-wise, this has an issue. This doesn't work with multi provider use cases. We can't have a module state outside atoms.

dai-shi (commented):
If I understand it correctly, this is something we don't merge, so I will close this immediately.
But, please feel free to add comments (and I can be misunderstanding something.)

src/utils/atomWithSuspense.ts
eolme: Oh, really, then I don't think I've delved enough into internal implementation.

docs/utils/atom-with-suspense.mdx
dai-shi: @eolme If you are willing to contribute, my suggestions are

a) in this `jotai` repo, open a PR to improve docs to illustrate this infinite pending usage. A codesandbox example would be nice to be included.
b) in the other `jotai-suspense` repo, open an issue or PR to add `atomWithPending` which should be as simple as follows, but not tested. Also an example would be nice. (you are generally welcome to contribute in the `jotai-suspense` repo, as it's almost empty. 😄 Especially, README.)

```js
// in JS
const atomWithPending = () => atom(new Promise(() => {}))
```

PR #1167, opened 2022-05-17
fix(docs): common usage for async storage

TwistedMinda:
I'm not sure why the `atomWithStorage` util doesn't show the storage option.
I also feel like introducing the `storage` option too early would be less comprehensible.

But it feels misleading to show `atomWithStorage`'s code with 2 parameters and then use it with 3. Maybe I should make it clearer by adding the util import statement?

`import { atomWithStorage } from "jotai/utils"`

### DIFF
--- a/docs/utils/atom-with-storage.mdx
+++ b/docs/utils/atom-with-storage.mdx
@@ -33,7 +33,7 @@ The `atomWithStorage` function creates an atom with a value persisted in `localS
 **storage** (optional): an object with:
 - `getItem` and `setItem` methods for storing/retreiving persisted state; defaults to using `localStorage` for storage/retreival and `JSON.stringify()`/`JSON.parse()` for serialization/deserialization.
 - Optionally, the storage has a `subscribe` property, which can be used to synchronize storage. The default `localStorage` handles `storage` events for cross-tab synchronization.
-- Optionally, the storage has a `delayInit` (`boolean`) property, which can be used to tell jotai whether use Suspense
+- Optionally, the storage has a `delayInit` (`boolean`) property, which can be used to tell jotai whether to use Suspense
   - `delayInit: true` will not suspend when first reading the value of your atom
   - `delayInit: false` (default) will suspend when first reading the value of your atom
 


dai-shi:
Thanks for working on docs!

> I'm not sure why the `atomWithStorage` util doesn't show the storage option.

What do you mean?

dai-shi:
> But that got me confused for a second.

The fact is that `guides/persistence.mdx` is written before `atomWithStorage` was included and it's not updated since then.

So, we can revisit the entire content in persistence.mdx.

> Which is not bad actually, to allow the reader to understand what the utils does inside!

I like it too. It tells it's less magical.

> So what do we do from now?

I can merge it as is, and you can do some follow-up work if you want.
Or, I can wait if you want some more work in this PR.

I'm pretty sure you find huge room for improving docs. Happy to have you on board.

> Do we wait for people to upvote the PR ?

For docs improvements, we don't do it. Unless there are obvious mistakes, it goes through pretty quickly. I may ask other contributors for review if someone is more appropriate.

> It's my first actual contribution to a project ever :D

Cool!

TwistedMinda:
@dai-shi I was thinking the same about the overall repartition.
We currently talk about persistence in:

- "React-native" main menu
- "Persistence" main menu
- "atomWithStorage" dedicated page

Would probably feel better to have just:
- "Persistence": describe how it works, basic usage and common use case + provide a link towards the `atomWithStorage` utils page
- "atomWithStorage utils page": show the dummy implementation, describe the different options and possible caveat, things to think about like "what happens when the schema of a given stored atom changes", etc...

I'm gonna continue working on this branch then. I let you decide to merge the current state now or not :p 
I think I can propose a better organisation for this topic indeed.

I'll ping you when I have a commit to show

dai-shi:
> "React Native" could be hidden until required for very specific implementation details

While it might be true, but I would like to keep it. I would at least tell that RN is in the scope. I mean, it would be nice if we can write some guide for RN users.

> I'll ping you when I have a commit to show

👍

TwistedMinda:
@dai-shi  

>  It would at least tell that RN is in the scope. I mean, it would be nice if we can write some guide for RN users.

Sure thing, rick. I just meant to take out the persistence out of there. Don't see anything to talk about tho.

"Just works on React-Native too"

TwistedMinda:
@dai-shi Done. Here's how I thought the whole thing:

- Persistence : teaches you about how the storage works, sync & async (no mention to react-native specifically)
- atomWithStorage util: added React-Native specific implementation instructions
- React-Native: cleaned to briefly link to persitence, added brief showcase of what jotai can do for your performance (yes that's really a thing for mobile developers :p)

docs/guides/persistence.mdx
```diff
@@ -73,6 +73,68 @@ const atomWithAsyncStorage = (key, initialValue) => {
...
 ```
 
+### Usage in Async actions
+
+When using an asynchronous storage, the reading of the created atoms are asynchronous too. That comes with some limitations. Before `useAtom` has been called at least once, the value is not set.
+This can cause an issue, for example, if you want to retrieve the value of your stored atom in an async action (write-only atom).
+
+Let's use the [atomWithStorage function in the utils bundle](../api/utils.mdx#atom-with-storage) for the examples, so that we have more options available.
+
+```typescript
```
dai-shi, line 83: If you are using `any` anyway, can we just use`js` here and remove typing?

docs/guides/persistence.mdx
```diff
@@ -73,6 +73,68 @@ const atomWithAsyncStorage = (key, initialValue) => {
...
+const storage = createJSONStorage(() => AsyncStorage)
+const userId = atomWithStorage('user-id-key', null, storage)
+const userInfo = atom({})
+const fetchUserInfo = atom(null, async (_get, set, payload: any) => {
+    const id = _get(userId) // This throws an error if "userId" is not loaded yet
```
dai-shi, line 88: Please just use `get`. `_get` is a convention for "unused" value.

docs/guides/persistence.mdx
```diff
@@ -73,6 +73,68 @@ const atomWithAsyncStorage = (key, initialValue) => {
...
+
+We want to make sure the action will never fail. As always with asynchronous flow in jotai, we have 2 options: with or without **Suspense**.
+
+#### With Suspense
+
+You may want to preload atoms at root level of your app directly:
+
+```typescript
+const Preloader = () => {
+    useAtomValue(userId) // Trigger the "onMount" function that will load the data from the store
```
dai-shi, line 102:
```suggestion
  useAtomValue(userId) // Trigger the "onMount" function that will load the data from the store
```

Please use two spaces. (I think I should somehow recover prettier config for mdx files.)

dai-shi (commented):
Thanks for your work. Overall looks good. Can you address some styling issues?

docs/guides/persistence.mdx
TwistedMinda: @dai-shi Oh, right, makes sense! I was wondering. Is this mentioned somewhere?

docs/guides/persistence.mdx
TwistedMinda: @dai-shi Yep sure, typing is not a concern here, the example is already complicated. enough

docs/guides/persistence.mdx
TwistedMinda: @dai-shi Oh sure, weird, i usually use 2 spaces. My bad

TwistedMinda:
@dai-shi Comments fixed, waiting your next review ;)

docs/guides/persistence.mdx
dai-shi: For example, TypeScript `noUnusedParameters` option will skip `_foo`.
https://stackoverflow.com/questions/41685054/clarification-on-typescripts-nounusedparameters-compiler-option

dai-shi (approved):
It looks okay for me. Let's keep improving if anything is left.

PR #1168, opened 2022-05-17
fix(docs): added info about infinite loop on useAtom

TwistedMinda:
@dai-shi I know you wanted to talk about `selectAtom` at this point, but I'm not sure why it's related, as it will still cause an infinite loop if used directly in the `useAtom`

I tested this:

```typescript
const [value] = useAtom(selectAtom(atom(0), (val) => val))
```

On the first render it does okay, but as soon as there is another render triggered by something else, this will cause an infinite loop.
I may be missing something about this selectAtom, for me it was just a fancy tool for redux lovers :p

### DIFF
--- a/docs/basics/primitives.mdx
+++ b/docs/basics/primitives.mdx
@@ -71,6 +71,7 @@ const Component = ({ value }) => {
 }
 ```
 
+
 ## useAtom
 
 The `useAtom` hook is to read an atom value in the state.
@@ -94,6 +95,16 @@ const [value, updateValue] = useAtom(anAtom)
 The `updateValue` takes just one argument, which will be passed
 to the third argument of the write function of the atom.
 The behavior depends on how the write function is implemented.
+
+**Note:** as mentioned in the *atom* section, you have to take care of handling the reference of your atom, otherwise it may enter an infinite loop
+```typescript
+const safeAtom = atom(0)
+const Component = () => {
+  const [value] = useAtom(atom(0)) // Will cause an infinite loop
+  const [safeValue] = useAtom(safeAtmo) // is fine
+  const [alsoSafeValue] = useAtom(React.useMemo(0, [])) // is also fine
+}
+```
 
 ## Provider
 
--- a/docs/basics/primitives.mdx
+++ b/docs/basics/primitives.mdx
@@ -98,11 +98,11 @@ The behavior depends on how the write function is implemented.
 
 **Note:** as mentioned in the *atom* section, you have to take care of handling the reference of your atom, otherwise it may enter an infinite loop
 ```typescript
-const safeAtom = atom(0)
+const stableAtom = atom(0)
 const Component = () => {
-  const [value] = useAtom(atom(0)) // Will cause an infinite loop
-  const [safeValue] = useAtom(safeAtmo) // is fine
-  const [alsoSafeValue] = useAtom(React.useMemo(0, [])) // is also fine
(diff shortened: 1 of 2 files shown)


docs/basics/primitives.mdx
```diff
@@ -95,6 +96,16 @@ The `updateValue` takes just one argument, which will be passed
 to the third argument of the write function of the atom.
 The behavior depends on how the write function is implemented.
 
+**Note:** as mentioned in the *atom* section, you have to take care of handling the reference of your atom, otherwise it may enter an infinite loop
+```typescript
+const safeAtom = atom(0)
+const Component = () => {
+  const [value] = useAtom(atom(0)) // Will cause an infinite loop
+  const [safeValue] = useAtom(safeAtmo) // is fine
```
dai-shi, line 104: A typo. I would call it `stableAtom`. How does it sound?

docs/basics/primitives.mdx
```diff
@@ -95,6 +96,16 @@ The `updateValue` takes just one argument, which will be passed
...
+  const [alsoSafeValue] = useAtom(React.useMemo(0, [])) // is also fine
```
dai-shi, line 105: You mean `useMemo(() => atom(0), [])` ?
This is fine, but useMemo doesn't have a semantic guarantee.
So, what we should show as a usage with useMemo is a derived atom.
```js
useMemo(() => atom((get) => get(countAtom) * 2), [])
```

dai-shi (commented):
Thanks for working on this. added a few comments.

dai-shi:
> I know you wanted to talk about `selectAtom` at this point, but I'm not sure why it's related, as it will still cause an infinite loop if used directly in the `useAtom`

We can skip `selectAtom` here. Or, maybe we should add a note in the selectAtom doc.

> I tested this:
> 
> ```ts
> const [value] = useAtom(selectAtom(atom(0), (val) => val))
> ```

This doesn't work because both an atom and a selector function are not stable.

```js
// Have a stable atom
const baseAtom = atom(0)

// Solution 1
  const [value] = useAtom(useMemo(() => selectAtom(baseAtom, (v) => v), []))

// Solution 2
  const [value] = useAtom(selectAtom(baseAtom, useCallback((v) => v, [])))

// Solution 3
// Define selector outside
const selector = (v) => v
  const [value] = useAtom(selectAtom(baseAtom, selector))
```

`useAtom` basically requires "stable" atoms.
`selectAtom` is clever enough to create a stable atom **if** both base atom and selector are stable.

docs/basics/primitives.mdx
TwistedMinda: @dai-shi sounds better

docs/basics/primitives.mdx
TwistedMinda: @dai-shi Yep forgot the inner function of useMemo :p Good point on semantic guarantee. Shouldn't we just show an example with useRef then ?

Okay your example has the merit of showing that it may make sense to create atoms on the fly. Myself never even thought of the idea. Nice !

TwistedMinda:
@dai-shi Alright we should let the selectAtom part for another PR. I'm not at ease with it :p

TwistedMinda:
@dai-shi Fixed typos & added stable references explanations for selectAtom, in the [latest commit](https://github.com/pmndrs/jotai/pull/1168/commits/e46254d29a90839155ceae0c9f8bc925ef6c6a05) if you may check ;)

dai-shi (approved):
Looks good. Thanks for your contribution!

PR #1169, opened 2022-05-17
docs: add atomWithPending atom creator

eolme:
See: https://github.com/pmndrs/jotai/pull/1165

### DIFF
--- a/docs/advanced-recipes/atom-creators.mdx
+++ b/docs/advanced-recipes/atom-creators.mdx
@@ -291,3 +291,61 @@ function EvenCounter() {
   return <>Count was set to an even number {evenCount} times.</>
 }
 ```
+
+## atomWithPending
+
+> `atomWithPending` creates an atom that suspend by default.
+> When the value is set for the first time it will act like a normal atom.
+
+This can be useful when you need to delegate asynchronous data, for example,
+if some business logic cant be promisified.
+
+```ts
+import { atom } from 'jotai'
+
+export function atomWithPending<Value>() {
+  return atom(new Promise(() => {}) as unknown as Value)
+}
+```
+
+Here's how you'd use it to make an pending atom:
+
+```ts
+import { atomWithPending } from 'XXX'
+
+const workAtom = atomWithPending<string>()
+```
+
+In components:
+
+```ts
+import { useSetAtom, useAtomValue } from 'jotai'
+
+const WorkController = () => {
+  const done = useSetAtom(workAtom)
+
+  return (
+    <>
+      <button onClick={() => done('success')}>success</button>
+      <button onClick={() => done('fail')}>fail</button>
+    </>
+  )
+}
+
+const WorkStatus = () => {
+  const status = useAtomValue(workAtom)
+
+  return <>Work is done with {status}.</>
+}
+
+const Work = () => {
+  return (
+    <>
+      <WorkController />
+      <Suspense fallback="Work in progress...">
+        <WorkStatus />
+      </Suspense>
+    </>
+  )
+}
+```


dai-shi:
@eolme Thanks for working on improving docs.

Hm, I was expecting to add it in a guide or somewhere, instead of adding it as an advanced recipe.

What I mean is, in JS without types, it's just an easy pattern;
```js
const anAtom = atom(new Promise(() => {}));
```
and we don't even need to create an atom creator function.

What do you think?

Maybe creating a new file `docs/guides/suspense.mdx`?

dai-shi:
> By the way this is very much linked to async storage, being first async then normal, maybe it could be a good transition in the docs ?

good point. but, i'm not sure how it would be in the docs.

> I'm voting "yes" for the "atomWithPending" creator, as we're here for minimalism, and returning an empty promise doesn't feel enough minimalist to me :p Maybe i'm too hyped by it and how grateful it goes with Suspense to see that there may not be enough use cases for this one-liner to deserve an official creator.

okay, two votes vs. one from me... I may accept having both "a guide" and "an atom creator".

@eolme and I had a discussion in [another thread](https://github.com/pmndrs/jotai/pull/1165#discussion_r874700753), and we will have such an atom creator in `jotai-labs/jotai-suspense`. https://github.com/jotai-labs/jotai-suspense/pull/1

dai-shi:
> i'm not sure I understand the concept of this second library.

For the official `jotai` library, we want to make it small and concise.
For `atomWithPending`, I don't prefer adding it in `jotai/utils`, because it's more like a pattern like "atoms-in-atom", which is already supported by nature. So, my first priority is to educate the `atom(new Promise(() => {}))` usage in docs. In addition to it, we can show it as an atom creator in advanced recipes. (Like other atom creators, which we decided not to have in `jotai/utils`.)
`jotai-labs` is a new activity of mine and ours, which we think something out of the scope of the official `jotai` repo (yet). We have some functions in jotai-labs repos, that could have been in advanced recipes. Some other repos there include experimental ones.

TwistedMinda:
@dai-shi I feel you.

I like the idea of:
- "patterns/recipes": just some examples to see what atoms are capable of, introducing the common usages
- "utils": you wouldn't like to recode it yourself, just use it :D 

And indeed `atomWithPending` enters the first category. You changed my mind!

> We have some functions in jotai-labs repos, that could have been in advanced recipes. Some other repos there include experimental ones.

Maybe "jotai-experiments" or "jotai-labs-experimental" would be clearer to understand your long term goals, in my opinion. 

But I get it now, the idea is to have:
- a very small core package
- a very small list of utils for things that you're 100% sure developers will need when using jotai
- then there may be another lib to embrace suspense
- maybe another to embrace loadables

....potentially anything that could become a bit more opinionated/specific, but too big for someone that just wants to use jotai for a basic project. Main objective being to be lightweight, unopinionated and never become a big messy engine. I feel ya

TwistedMinda:
@eolme @dai-shi Up on this guys.
After the discussion, I would propose adding a new sub-section "Async read once" in the [Basic's Async section](https://jotai.org/docs/basics/async):

- Talk about `atomWithPending` and how cool it is
  - with Suspense
  - without Suspense
  - preload pattern examplpe
- Introduce another `atomWithPending`-like by linking to Persistence page "You may note that this pattern is also used in the Persistence (view atomWithAsyncStorage)"

It would mean that the "Common Usage for async actions" I added in the Persistence would be in the Async sections which makes lot more sense.
Should I open a new PR to make my proposal ?

dai-shi:
@TwistedMinda 

> "Async read once"

Actually, it's not just "once".
You can set a new promise on `write`. And, it suspends! This is actually a cool technique to start async function early and still use `Suspense`.

`read` is invoked in render (in useReducer) which is delayed, `write` is invoked in sync (so usually it's in callback).

I think we need a new discussion to dig into it?

> Should I open a new PR to make my proposal ?

Yeah, that sounds great.

eolme:
Ok, then I'm closing this one.

PR #1170, opened 2022-05-17
fix(utils): fix memory leaks and other issues in atomWithObservable

TobiasWalle:
@dai-shi As discussed this PR builds on https://github.com/pmndrs/jotai/pull/1157

- I added more tests for some of the edge cases
- I fixed some known and unknown issues that are covered by the new tests

This is the test result with the __current__ implementation:
image

This is the test result __new__ implementation:
image

Fixed issues:
- Emitted values between the initialValue and onMount are no longer ignored
- Falsy initial values are no longer ignored (`0`, `false`, `''`, etc.)
- If the observable is an `BehaviourSubject` or already has a value, `Suspense` is no longer required.
- If the observable never emits and no initial value is provided, the subscription is cancelled after 10 seconds.  This value can be configured with the `timeout` option
    - I implemented this solution, based on your recommendation in https://github.com/pmndrs/jotai/pull/1058#discussion_r825990484. Unfortunately I also didn't find a better solution.
    - I choose a timeout of ten seconds as a default, instead of one second, because it is not uncommon for requests, to take more than one second.
    - We might want to log a warning during development if the `timeout` gets applied. This way the user knows that something went wrong and can fix the issue (What are your thoughts on that?)
    - __This is actually a breaking change.__ Because it will break observables that take more than ten seconds to emit their first value without manually setting or disabling the timeout. If this is not okay for you, we can remove the default for now and add it separately in the next major release.

I'm looking forward to your thoughts and feedback!

### DIFF
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -105,6 +105,16 @@ export function atomWithObservable<TData>(
     if (options?.initialValue !== undefined) {
       initialValue = getInitialValue(options)
     } else {
+      // FIXME
+      // There is the potential for memory leaks in this implementation.
+      //
+      // If the observable doesn't emit an initial value before the component that uses the atom gets destroyed,
+      // the onMount function never gets called and therefore the subscription never gets cleaned up.
+      //
+      // Unfortunately, currently there is no good way to prevent this issue (as of 2022-05-23).
+      // Timeouts may lead to an endless loading state, if the subscription get's cleaned up too quickly.
+      //
+      // Discussion: https://github.com/pmndrs/jotai/pull/1170
       subscription = observable.subscribe(dataListener, errorListener)
       initialValue = initialEmittedValue
     }
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -33,6 +33,7 @@ type InitialValueFunction<T> = () => T | undefined
 
 type AtomWithObservableOptions<TData> = {
   initialValue?: TData | InitialValueFunction<TData>
+  timeout?: number | false
 }
 
 export function atomWithObservable<TData>(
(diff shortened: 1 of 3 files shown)


dai-shi:
Hi, thanks for working on it. I'll have a look later.
Meanwhile, I just merged #1157. So, you can update it here.

TobiasWalle:
> Hi, thanks for working on it. I'll have a look later. Meanwhile, I just merged #1157. So, you can update it here.

I updated the MR.

To my earlier point:
> We might want to log a warning during development if the timeout gets applied. This way the user knows that something went wrong and can fix the issue (What are your thoughts on that?)

I don't think that this would be a good idea anymore. I think in some scenarios the timeout just cannot be avoided and in these cases warning logs would be pretty annoying.

TobiasWalle:
I see that a test is failing. I will look into later (probably tomorrow).

dai-shi:
> If the observable never emits and no initial value is provided, the subscription is cancelled after 10 seconds.

You seem to be misunderstanding something, at least from my understanding. I would expect to clear timeout on mount. So, 1 second is mostly enough (or we don't know for sure anyways). It's not about first emitting.

TobiasWalle:
@dai-shi My understanding was, if the first value doesn't emit the "loading" indicator from the suspense is always shown.

This way the `onMount` effect gets never called, as the component that uses `useAtom` is never rendered (It just throws the promise). If the component now gets removed before the first render/first emit, it will always leak the subscription, independent from the time that has passed.

This test should reproduce the behaviour:
https://github.com/pmndrs/jotai/blob/dfb6262cf6a5d46a598b583d66448f27aa7f192c/tests/utils/atomWithObservable.test.tsx#L246-L277

This is always throws:
https://github.com/pmndrs/jotai/blob/main/src/core/useAtomValue.ts#L32

So this never gets called (which calls onMount):
https://github.com/pmndrs/jotai/blob/main/src/core/useAtomValue.ts#L76

But I will test my assumptions again and create a code sandbox to reproduce the memory leak in a real app. Maybe this was a misunderstanding on my side.

dai-shi:
> if the first value doesn't emit the "loading" indicator from the suspense is always shown.

Oh, yeah. You are right. Hm, what was I confused with..
It's different from atomWithSubscription in jotai/urql.

Okay, in such a case, I think the default behavior is no timeout, pending infinitely.
`timeout` option is just optional (so that developers can avoid memory leaks if they want).

TobiasWalle:
> > if the first value doesn't emit the "loading" indicator from the suspense is always shown.
> 
> Oh, yeah. You are right. Hm, what was I confused with.. It's different from atomWithSubscription in jotai/urql.
> 
> Okay, in such a case, I think the default behavior is no timeout, pending infinitely. `timeout` option is just optional (so that developers can avoid memory leaks if they want).

Okay, I will update the default accordingly 👍

TobiasWalle:
@dai-shi I just figured out, that the subscription only get's leaked if the `Provider` is destroyed before the the first value is emitted. 

As this is a pretty uncommon edge case and the timeout can introduce other bugs, I decided to just completely remove it. This also makes the code a little less complicated.

We could avoid the leak completely if the `Provider` would run a callback on all atoms, once it unmounts. But as this is a very uncommon issue, I wouldn't give this any priority at the moment.

I also fixed the tests. So this MR should be ready.

dai-shi:
> that the subscription only get's leaked if the `Provider` is destroyed before the the first value is emitted.

This is only true in React 17. React 18 can invoke render multiple time before comitting for various reasons. jotai atom's `read` function is same as the render function. So, technically it has to be "pure". https://gist.github.com/sebmarkbage/75f0838967cd003cd7f9ab938eb1958f#what-does-pure-mean-in-react

See more here: https://github.com/facebook/react/issues/15317


> I decided to just completely remove it.

That said, it's fine for now, but I'd like to leave some comments about the memory leak risk.

src/utils/atomWithObservable.ts
```diff
@@ -56,24 +56,68 @@ export function atomWithObservable<TData>(
...
-    const dataAtom = atom(
-      options?.initialValue
-        ? getInitialValue(options)
-        : firstValueFrom(observable)
-    )
-    let setData: (data: TData | Promise<TData>) => void = () => {
-      throw new Error('setting data without mount')
+    // To differentiate beetwen no value was emitted and `undefined` was emitted,
+    // this symbol is used
+    const NotEmitted = Symbol()
```
dai-shi, line 61: I used `EMPTY` in `atomWithDefault`.

src/utils/atomWithObservable.ts
```diff
@@ -56,24 +56,68 @@ export function atomWithObservable<TData>(
...
+    type NotEmitted = typeof NotEmitted
```
dai-shi, line 62: This is confusing. Just use `typeof NotEmitted` below without aliasing.

src/utils/atomWithObservable.ts
```diff
@@ -56,24 +56,68 @@ export function atomWithObservable<TData>(
...
+    let initialValue:
+      | TData
+      | Promise<TData>
+      | InitialValueFunction<TData>
+      | undefined
+    if (options?.initialValue !== undefined) {
+      initialValue = getInitialValue(options)
+    } else {
+      subscription = observable.subscribe(dataListener, errorListener)
+      isSync = false
```
dai-shi, line 111: Let's move this out of the conditional block.

dai-shi (commented):
Some minor comments.

TobiasWalle:
> That said, it's fine for now, but I'd like to leave some comments about the memory leak risk.

Should I add the comments in the code or in the documentation?

dai-shi:
> Should I add the comments in the code or in the documentation?

In the code, please.

TobiasWalle:
> > Should I add the comments in the code or in the documentation?
> 
> In the code, please.

I added the comment.

dai-shi (approved):
Looks good. Thanks for your contribution.
I'll merge it when we are ready to cut a new release.

PR #1176, opened 2022-05-21
docs(async): Rework async education

TwistedMinda:
I need to re-read myself a few times more but I guess the main idea is here.
I would also like to test if the loadable API could be used inside actions to also explain how preloading can be done without Suspense.

Anyways, what do you think?

EDIT: What did I do to break the prettier check? :/


dai-shi:
Thanks for the suggestion.

One request: I want to limit the discussion in "BASICS" to the core module `jotai`. If we need to talk about `loadable` from `jotai/utils`, "GUIDES" is a better place. How about moving "Async" section to the first item of "GUIDES"?

> What did I do to break the prettier check? :/

`yarn run prettier` will fix it. I recently added it, which had once included and been broken for a while. There are pros and cons, but appreciate for understanding.

docs/basics/async.mdx
```diff
@@ -80,6 +110,57 @@ const Component = () => {
 
 > There's no way to know as of now if an atom suspends because of `read` or `write`.
```
dai-shi, line 111: We can remove this. It's a note for pre-1.4. `write` will not suspend. Exception: if you `write` a promise object, it will suspend, but "you know it".

TwistedMinda: @dai-shi That part was so misleading me that I didn't even dare to touch it.
So you confirm me that:
```js

cont [_, setNum] = useAtom(writeAsync) is never gonna suspend
```

only the below will
```js
cont [_, setNum] = useAtom(writeAsync) is never gonna suspend

React.useEffect(() => {
   setTimeout(() => { setNum() }, 1000) // This triggers the suspend
}, [])
```

Are we good? 

> if you write a promise object, it will suspend, but "you know it".

I'm not sure I understood that part

> How about moving "Async" section to the first item of "GUIDES"?

Arf I feel bad about it. I would have really liked for people to consider it basics. But I feel your pain of attacking utils straight in the basics... In that case, I would propose a re-order of the menu to put Async in top of the scene.

What do you think?
Just a proposition, we can fallback to just moving Async if you want, for now ;)

```
BASICS
- Concepts
- Primitives

API
- Core
- Utils

GUIDES
- Async
- Resettable <== Too simply to be a guide, should only be an util/recipe, no?
- Persistence
- Atoms in atom
- Composing atoms

EXPERIENCE
- TypeScript
- Debugging
- Devtools
- Babel
- Testing
- Core Internals

BRIDGES
- Next.js
- React Native
- Vite

SHOWCASE:
- Comparison
- Showcase
```

TwistedMinda: @dai-shi I'm not sure how to document the loadable util correctly, tho.
Is there anything more that I should be aware of, more of the simple usage:

```jsx
const loadableAtom = loadable(asyncAtom)
// Does not need to be wrapped by a <React.Suspense> element
const Component = () => {
  const value = useAtom(loadableAtom)
  if (value.error) return <Text>Error</Text>
  if (value.state === 'pending') {
    // 'pending' | 'hasData'
    return <Text>Loading...</Text>
  }
  console.log(value.data) // Results of the Promise
  return <Text>Value: {value.data}</Text>
}
```

If it's good, should I add it as the last util ? #20 ?

docs/utils/loadable.mdx
```diff
@@ -0,0 +1,32 @@
+---
+title: loadable
+---
+
+For some reason, you may not want to profit from the React's `Suspense` core handling of jotai, and prevent your app from suspending when using async atoms.
```
Pinpickle, line 5: This comes off as a little negative to me ("not want to profit"), I think there are lots of valid use-cases for `loadable`. 

You could say something like

```
If you don't want async atoms to suspend or throw to an error boundary (for example, for finer-grained control of loading and error logic), you can use the `loadable` util. It returns a value with one of three states: `loading`, `hasData` and `hasError`.
```

Pinpickle (commented):
@TwistedMinda I wanted to say thanks for the docs that I never got to writing for the past 6 months! I added a little nit, but it being documented is far better than the situation we're in at the moment.

docs/utils/loadable.mdx
TwistedMinda: @Pinpickle Sounds way better!

docs/basics/async.mdx
dai-shi:
```js
cont [_, setNum] = useAtom(writeAsync) is never gonna suspend

React.useEffect(() => {
   setTimeout(() => { setNum() }, 1000) // This triggers the suspend
}, [])
```
It will not trigger suspend.

dai-shi:
> Arf I feel bad about it. I would have really liked for people to consider it basics.

If you want to keep async section in basics, I would require that to be only about `jotai`, so no `loadable`. We can still more about async in guides. I actually like it better.

dai-shi:
> Just a proposition, we can fallback to just moving Async if you want, for now ;)

I want @sandren 's comment on it.

TwistedMinda:
@dai-shi 
> If you want to keep async section in basics, I would require that to be only about jotai, so no loadable. We can still more about async in guides. I actually like it better.

Hm I feel stuck now. Cutting the async guide in half seems like a bad idea to me. But moving it out of Basics also seem bad to me.
I prefer your solution of moving it to Guides instead of considering it basic, if you we have to choose

dai-shi:
> If you want to cut in 2 parts, what would you approach in each?

BASICS: only about `atom(async(get) => ...)` and the fact that we need `<Suspense>` higher in the tree, but within `<Provider>`.
GUIDES: async write, loadable, promise as value, infinite pending

I don't mean I'm very confident on this. So, I'd like to hear your opinions. But, I strongly feel we should have "BASICS" to explain only about `jotai`, no `jotai/utils`. If it's too confusing, "BASICS" can be only about "sync".

TwistedMinda:
@dai-shi Well even after meditating I still strongly believe we shouldn't cut the Async part, as its the "most-complicated" and should not be confusing in any way!

So I moved it to guides as you recommended. But really had to move away the show-case that was preventing a good transition of basics to the first guide. A demo is not a basic, right? I moved it at the end in a Demo sub-section.

Did you review the content of the async page? Do you think it's ok for now?
I'm not experienced with async atoms enough to documentate them better, so maybe just a polish and we can let someone take over?
I actually only use the simple patterns :p

dai-shi:
> So I moved it to guides as you recommended.

Fair enough. I'll think about it once again.

> Did you review the content of the async page?

Sorry, not yet. Please give me some time as this is such an important change/improvement.

sandren:
@dai-shi I think moving the async doc to 'guides' is a good idea as it allows us to cover more than what is available in Jotai core. 🙂

docs/guides/persistence.mdx
```diff
@@ -75,6 +75,8 @@ const atomWithAsyncStorage = (key, initialValue) => {
 }
 ```
 
+Don't forget to check out the [Async documentation](../basics/async) for more details on how to use async atoms.
```
dai-shi, line 78:
```suggestion
Don't forget to check out the [Async documentation](./async) for more details on how to use async atoms.
```

readme.md
```diff
@@ -206,3 +203,6 @@ yarn add jotai react
 - Advanced Recipes
   - [Large Objects](./docs/advanced-recipes/large-objects.mdx)
   - [Atom Creators](./docs/advanced-recipes/atom-creators.mdx)
+- Demo
+  - [Comparison](./docs/basics/comparison.mdx)
+  - [Showcase](./docs/basics/showcase.mdx)
```
dai-shi, line 208: @sandren what do you think about moving them down here?
I expected many people are interested in "comparison" and "showcase" before diving into detailed docs.

docs/guides/async.mdx
```diff
@@ -0,0 +1,188 @@
...
+Jotai is inherently leveraging `Suspense` to handle asynchronous flows.
+
+```jsx
+const ComponentUsingAsyncAtoms = () => {
+  const [num] = useAtom(asyncAtom)
+  // here `num` is always `number` even though asyncAtom returns a Promise
+}
+const App = () => {
+  return (
+    <React.Suspense fallback={/* What to show while suspended */}>
```
dai-shi, line 32:
```suggestion
    <Suspense fallback={/* What to show while suspended */}>
```
Let's prefer shorter one. I don't think this confuse people too much.

docs/guides/async.mdx
```diff
@@ -0,0 +1,188 @@
...
+**Note**: An atom becomes async not only if the atom read function is async,
+but also if one or more of its dependencies are async.
+
+```js
+const anotherAtom = atom((get) => get(asyncAtom) / 2)
+// even though this atom doesn't return a promise,
+// it's a read async atom because `asyncAtom` is async.
+```
+
+**Note**: You cannot get the value of an async atom in a write-atom (wether its async or not) before its value has been set
```
dai-shi, line 50: This only applies to derived atoms. primitive atoms such as `atom(1)` should be fine.

dai-shi (commented):
Thanks for working on this! Looks very good. Added some comments.

docs/guides/async.mdx
TwistedMinda: @dai-shi I'm not sure I understand your statement.

To me, the error appears as soon as its an `async atom`, not `derived atom`
I was trying to talk about `async` atoms (`atom(1)` is not async).

Maybe there's something I'm missing?
Do you think I should remove this part?

readme.md
TwistedMinda: @dai-shi We can move the Demo Sub-section to the top of basics, I wouldn't mind at all

docs/guides/async.mdx
dai-shi: Oh, you are right. I was confused with "async or not" part.
```suggestion
**Note**: You cannot get the value of an async atom in a write-atom (whether the write function is async or not) before its value has been resolved
```

readme.md
dai-shi: In this PR, let's move them between Basics and Guides.
@sandren will probably re-organize anyway in follow-up PRs.

dai-shi: done: 5d3d203

dai-shi (approved):
Let's merge this. Thanks so much for your contribution!

PR #1181, opened 2022-05-24
fix(utils): does not throw with an empty list to waitForAll

### LINKED ISSUE #1177: `waitForAll` throws with an empty list
This code:

```
useAtom(waitForAll([]));
```

Throws with the following error:

```
TypeError: Invalid value used as weak map key
    at WeakMap.set (<anonymous>)
    at setWeakCacheItem (https://zwq75z.csb.app/node_modules/jotai/utils.js:245:13)
    at memoizeAtom (https://zwq75z.csb.app/node_modules/jotai/utils.js:269:5)
    at waitForAll 
```

From what I can tell, this is due to the `memoizeAtom` logic:

https://github.com/pmndrs/jotai/blob/88f4edd32a2a3b0cde743a1fc26ae88cd7a56aa0/src/utils/weakCache.ts#L28-L41

^ this loop will be run at least once, even if the array is empty. This leads to a `weakMap.set(undefined, value)` call, which throws the error.

## Reproduction

https://codesandbox.io/s/react-typescript-forked-zwq75z?file=/src/App.tsx

dai-shi:
close #1177


dai-shi:
Fixed: https://codesandbox.io/s/react-typescript-forked-cgsl00

dai-shi:
The change causes jest errors with react 17. Can anybody know how to workaround this?

PR #1197, opened 2022-06-01
feat(website): add table of contents

sandren:
This adds support for a new Table of Contents component for use in the Introduction doc.


docs/utils/use-atom-value.mdx
```diff
@@ -1,5 +1,6 @@
 ---
 title: useAtomValue
+status: deprecated
```
dai-shi, line 3: `useAtomsValue` and `useUpdateAtom` from `jotai/utils` are not really deprecated in code *yet*.

dai-shi (approved):
LGTM

docs/utils/use-atom-value.mdx
TwistedMinda: @dai-shi Sorry, but what do you mean by "yet"?

You want to "force" usage of useAtom ?
`const [val] = useAtom(...)`

I liked the useAtomValue pattern a lot?

On the other hand, why not state `useUpdateAtom` as deprecated now? As it's [already mentioned in the documentation](https://jotai.org/docs/utils/use-update-atom) to use `useSetAtom` instead, for newer versions.

Issue #1052, opened 2022-03-09
[testing] Can we seed atoms with a value for tests?


hellos3b:
I've looked at the [Testing Documentation](https://github.com/pmndrs/jotai/blob/main/docs/guides/testing.mdx) and completely understand the idea of testing components and not implementation. But I think more often than not we will want to test component  functionality when it has some state

For example in the Counter sample from the docs, imagine if the counter maxes out at 99. Is there a way to inject a `countAtom` that's set to 99, so you can test that clicking the button doesn't change the text?

In a more realistic example, testing a filtered todo list component who's todos are stored in a centralized atom that's initialized from a separate component. We'd need to prefill that atom with a list of todos in order to see that toggling filters are applied correctly, otherwise the component would just be an empty list.

Are there any utils or recommended practices to fill some atoms with data before rendering components?

dai-shi:
Provider's `initialValues` prop is designed for such test cases.
The testing guide docs should include that use case.

Aslemammad:
Hey @Sajad-Sharhani, This may be a good take for you! It'd give you an overview of `initialValues` in providers

TwistedMinda:
Closing this in favor of the PR.

PR #1201, opened 2022-06-02
docs(testing): initialValues & React-Native demo

TwistedMinda:
Hello,

wanted to address [this issue](https://github.com/pmndrs/jotai/issues/1052) that states it's not obvious, when reading the Testing docs, that we can use `Provider` for initia values while testing.

I also added a bonus example with React-Native to be fair to "@testing-library/react", providing his brother "@testing-library/react-native"

@dai-shi What do you think?
Also, how would you use `renderHooks` to setup initial values ? Just call the setter(s) in the first lines ?


dai-shi (commented):
It looks good to me. Can anyone else review?

Thisen (approved):
LGTM 👍🏼

PR #1202, opened 2022-06-02
docs(primitives): clarify what can contain an atom

TwistedMinda:
This is the follow-up of [this PR](https://github.com/pmndrs/jotai/pull/1083) that I believe is way too old compared to the importance.

Wanted to prevent it from staling ;) 

EDIT: Maybe it's not the best place @sandren ? I know u've been working on Introduction recently. Maybe you know better where to put this or even if it's still needed. Very good job by the way, the Table of contents is great!

### DIFF
--- a/docs/basics/primitives.mdx
+++ b/docs/basics/primitives.mdx
@@ -3,7 +3,7 @@ title: Primitives
 nav: 1.02
 ---
 
-An atom in jotai represents a piece of state. It can hold any value ranging from string, boolean or number to complex structures like arrays, objects, functions and so on.
+An atom in jotai represents a piece of state. It can hold any value ranging from string, boolean or number to complex structures like arrays, objects, functions and so on. You can consider them the building blocks of your app state.
 
 Unlike useState in React, atoms are not tied to specific components.
 Let's see how to define and use atoms.
--- a/docs/basics/primitives.mdx
+++ b/docs/basics/primitives.mdx
@@ -60,7 +60,7 @@ For async behavior, please refer to the [async](../guides/async.mdx) doc.
 `set` in the write function is to write atom value.
 It will invoke the write function of the target atom.
 
-Atom configs can be created anywhere, but referential equality is important.
+**Note**: Atom configs can be created anywhere, but referential equality is important.
 They can be created dynamically too.
 To create an atom in render function, `useMemo` or `useRef` is required to get a stable reference. If in doubt about using `useMemo` or `useRef` for memoization, use `useMemo`.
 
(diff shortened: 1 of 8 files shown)


docs/basics/primitives.mdx
```diff
@@ -3,8 +3,8 @@ title: Primitives
 nav: 1.02
 ---
 
-State in Jotai is a set of atoms.
-An atom is a piece of state.
+An atom in jotai represents a piece of state. It can hold any value ranging from string, boolean or number to complex structures like arrays, objects, functions and so on. You can consider them the building blocks of your app state.
```
dai-shi, line 6: A gotcha is that we cant store a function directly in an atom. We need to wrap it with an object.

TwistedMinda: @dai-shi Yeah I was thinking about it too. But I still feel like it's important to mention it, even if it may be a bit confusing at first.

The key is that I have to add somewhere that you can store anything, maybe mentioning more examples like Blob or Base64 or you name it... BUT, in the case of a function, you just need to store it inside an objet.
Let me find the good spot

TwistedMinda: @dai-shi Last commit contains my proposition

sandren: @TwistedMinda  Also if you don't mind, could you capitalize Jotai for consistency with the rest of the docs. 🙂

docs/basics/primitives.mdx
```diff
@@ -112,6 +112,16 @@ const Component = () => {
...
+**Note**: We mentioned that our atoms can contain any kind of data, it can be a string, Blob, Observer, anything really. There is just one exception. Because derived atoms are defined using a function, jotai will not understand if we pass it a function that isn't exactly a pure getter.
+So what you can do is simply wrap your function in an object.
+
+```js
+const doublerAtom = atom({ callback: (n) => n * 2 })
+// Usage
+const [doubler] = useAtom(doublerAtom)
+const doubledValue = doubler.callback(50) // Will compute to 100
+```
+
```
sandren, line 124: @TwistedMinda Thank you. This demo helps clarify. 🙂

docs/basics/primitives.mdx
TwistedMinda: Added my new proposition :)

sandren:
@TwistedMinda Thanks for cleaning up the other docs as well. It looks good to me! 🙂

TwistedMinda:
@sandren Np! Was my mistake in the first place tho :p 
@dai-shi All good for you?

dai-shi (approved):
Looks good. Thanks for your work.

Issue #1150, opened 2022-05-10
State resets with React Fast Refresh


alan-nf:
Expected: when React Fast Refresh updates UI on save, jotai state should not be lost. E.g. a primitive counter atom like `const countAtom = atom(0)` should maintain it's value across fast refreshes.
Actual: all jotai global state resets when React Fast Refresh updates the UI.

Codesandbox: https://codesandbox.io/s/jotai-react-refresh-bug-807nez?file=/src/Jotai.tsx
Local demo: https://github.com/alan-nf/jotai-react-refresh-issue

Instructions:
1. Click the +1 button to increment the counter
2. Edit the text in the `Copy` component to say something else. The counter will be reset back to 0.

NOTE: related discussion about React Fast Refresh being broken in https://github.com/pmndrs/jotai/issues/1025. I cannot reproduce things being totally broken and I'm just seeing this reset issue.

dai-shi:
@Thisen Can you take a look?

Thisen:
I'll take a look!

Thisen:
@alan-nf, the Codesandbox doesn't seem to be using the `jotai/babel/plugin-react-refresh`. Am I blind? 😅

alan-nf:
No you're not blind, heh. It's using Create React App which uses react refresh under the covers

from https://javascript.plainenglish.io/react-fast-refresh-the-new-react-hot-reloader-652c6645548c:

> Starting from 4.x, Create React App enables React Fast Refresh by default. If your project is built off of Create React App, all you need to do is upgrade.

more specifically, it's using `@pmmmwh/react-refresh-webpack-plugin@^0.5.3`

This is where it's hooked up: https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/config/webpack.config.js#L35

alan-nf:
And, if you grab [the demo that can run locally](https://github.com/alan-nf/jotai-react-refresh-issue), you can save a file edit and then see the network requests related to the react refresh update:

![image](https://user-images.githubusercontent.com/100322190/167920423-960d42b4-260e-46e7-b7e7-3971fe9ad860.png)

alan-nf:
I noticed that the hot update includes the atom declaration, presumably because it's in the same file as the component I'm editing. If I move that component to a separate file and edit it there, the jotai state stops resetting. That seems like totally reasonable behavior, and therefore isn't really reproducing the issue I'm seeing.

[...]

PR #1203, opened 2022-06-02
docs(fast refresh): introduce dx issue

TwistedMinda:
Following-up to [issue #1025](https://github.com/pmndrs/jotai/issues/1025)

Quick reminder: goal is to have some kind of "official statement" of the status with Fast Refresh:

- current workaround: as long as you don't modify a file that contains the atom declaration, it will not reset the state
- our feeling: DX not the best, forcing to split atoms to avoid state reset, but we are aware/working on it
- can be very hard to understand why it got reset in some situations
- we're very much open to support/contribution on this
- (edit) can fix instantly with babel plugin!

### DIFF
--- a/docs/guides/fast-refresh.mdx
+++ b/docs/guides/fast-refresh.mdx
@@ -5,11 +5,11 @@ nav: 3.12
 ---
 
 In order to deliver the best Developer Experience (DX), we're working on being compatible with Fast Refresh from the ground up.
+Currently, by default, editing a file will result in resetting the state of all the declared atoms inside it.
 
-### Developer experience issue
+### Splitting workaround
 
-Currently, editing a file will result in resetting the state of all the declared atoms inside it.
-In order to best avoid resetting, you can split your atoms in seperate files. This way, the state is reset according to more granular file updates.
+In order to best avoid resetting, you could split your atoms in seperate files. This way, the state is reset according to more granular file updates.
 
 ```ts
 // file: atoms/atom1.ts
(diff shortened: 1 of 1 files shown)


dai-shi:
I think you miss `jotai/babel/plugin-react-refresh`.

With the plugin, fast refersh should work for globally defined atoms even if they are in the same file.

TwistedMinda:
@dai-shi No, unfortunately I do have `jotai/babel/plugin-react-refresh`.
Maybe I did bad at setting it up?
I have followed the docs by updating my babel.config.js & restarting my metro bundler: 
```json
{
  "plugins": ["jotai/babel/plugin-react-refresh"]
}
```
Do I miss anything?

dai-shi:
> @dai-shi No, unfortunately I do have `jotai/babel/plugin-react-refresh`. Maybe I did bad at setting it up? I have followed the docs by updating my babel.config.js & restarting my metro bundler:

Then, it's not a doc issue, but a real issue for the plugin.

> Do I miss anything? Note: I'm using React-Native, does that change anything?

Can you try checking the behavior with codesandbox first?
We are not yet sure where the problem is. #1025 is an issue we wanted to hear details.

Meanwhile, let's hold this PR as it may not be accurate.

TwistedMinda:
@dai-shi Okay I think it was the good old mistake:
I had 
```ts
{
  "plugins": [
     ["jotai/babel/plugin-react-refresh"]
  ]
}
```
instead of
```ts
{
  "plugins": ["jotai/babel/plugin-react-refresh"]
}
```

Works perfectly now. I'm confused now, what is the problem with FastRefresh then ? And should we still do a FastRefresh page to at least mention that the plugin has to be used in order to avoid a reset ?

alan-nf:
@dai-shi if you look at my original repro in https://github.com/pmndrs/jotai/issues/1150 you can see that the jotai atom gets reset if it's included in the bundles that are hot-updated. If that's not intended behavior I can re-open https://github.com/pmndrs/jotai/issues/1150. Can you let me know?

(And before anyone asks, the code sandbox is using create-react-app which uses react refresh under the covers. See [this post](https://github.com/pmndrs/jotai/issues/1150#issuecomment-1124152252) for more info)

Thisen:
Closing this, as it has been resolved in 1.7.3.

Issue #1175, opened 2022-05-20
Extra re-renders with selectable list


moroshko:
Given a list of selectable items, when an item's selection is toggled, I'd like only that item to be re-rendered.
However, I see that ALL items are getting re-rendered.
What am I doing wrong?

Use case: Select items to perform bulk action of the selected item ids.

[CodeSandbox](https://codesandbox.io/s/jotai-select-list-forked-nljpsh)

dai-shi:
https://codesandbox.io/s/jotai-select-list-forked-p4zzwk

Nothing is wrong. React runs render function and finds that atom values are the same, and stop rendering (no commit). This behavior is basically the same as `useReducer`.

TwistedMinda:
@moroshko I think your algorithm could be improved if you really need to focus on performances.

Here is a schema of your pattern:
- create map of <string, atom>
- for each item, create a derived atom on the map, for a given id
- everytime you update one toggle in your list
  - this will re-render all your items
  - as dai-shi mentioned it's not a big deal because there will be no commits by react (but it's still not ideal)

Here's how you couold avoid the extra re-renders:

- for each list item create a read-write atom<boolean> like you did, **except not depending on the map atom**
- of course, you will need to fill up the initial value for each row at the first render
```tsx
const selectedAtom = atom<Record<string, boolean>>({});
const Item = (props) => {
  const { item, initialSelection } 
  const reference = React.useRef(atom(props.initialSelection))
  const [toggle, setToggle] = useAtom(reference.current)
  // ...
  return (null)
}
const List = (props) => {
  const renderItem = (item) => {
    const value = props.initialSelection[item.id]
    return (
       <Item item={item} defaultValue={value} />
    )
  }
  return items.map(renderItem)
}
const Component = () => {
  const initialValues = useAtomValue(selectedAtom)
  // Stop listening changes after first mount
  const MemoizedList = React.useMemo(() => <List initialValues={initialValues} />, [])
  return (
    <MemoizedList />
  )
}
```

That way, only the toggle item is the only one re-rendered.

TwistedMinda:
@dai-shi I think we can close it as it's all intended behaviour. I even tried to provide a work-around if needs be to avoid extra-renders (can't remember if I tested it, tho, but looks more like a react problem than jotai problem anyway)

OP could still add a comment to re-open if need more info.

dai-shi:
Yeah.
Would it be nice if we could describe such a behavior is expected in docs somewhere?

[...]

PR #1206, opened 2022-06-03
docs(primitives): Educate idempotent components

TwistedMinda:
Following-up [this issue](https://github.com/pmndrs/jotai/issues/1175) to clarify in the docs that extra-rerenders even if atoms value didn't change is an expected behaviour


dai-shi (approved):
Looks good! Thanks!

Issue #1190, opened 2022-05-29
`useHydrateAtoms` w/ NextJS sometimes persists previous value even on new page loads


arcoraven:
I'm seeing a strange and intermittently reproducible issue with useHydrateAtoms and NextJS.

I have a page `/item/:item_id` where `getServerSideProps()` queries the item from DB and renders a page with that data. I added a `console.log(item)` in `getServerSideProps()` and in the React component that uses this atom.

Steps to reproduce:
- Navigate to `/item/1`
  - Server-side console logs `item 1`
  - Client-side console logs `item 1`
- Open a new tab, navigate to `/item/2`
  - Server-side console logs `item 1`
  - Client-side console logs `item 2`
  - NextJS throws a client-side warning:
    - > Warning: Prop `alt` did not match. Server: "Item 1" Client: "Item 2"

**Result:** Some elements on the page render item 2's data, and some (like images) render item 1's data.

Code (simplified for readability):
```typescript
const myItemAtom = atom<MyItem | undefined>(undefined)

function MyItemPage({ item }: { item: MyItem }) {
  useHydrateAtoms([[myItemAtom, item]]);
  return (
    <>
      <ItemMetadata />
    </>
  )
}
export default MyItemPage;

export const getServerSideProps({ query }) {
  const item = // retrieve item with ID `query.item_id`
  console.log(item.id)
  return { props: { item } }
}

// in another file
export function ItemMetadata() {
  const item = useAtomValue(myItemAtom)
  console.log(item.id)
  return (
    <>
      <Text>{item.name}</Text>
      <Image src={item.image_url} alt={item.name} />
    </>
  )
}
```

So on subsequent page loads (even on a _new tab_) NextJS seems to be loading an old/existing value for the atom instead of the newest value. I understand that hydration is meant to be done once and does not change with re-renders, but is there some caching or storage happening that causes the value to persist across sessions?

Note: Setting the atom in a `useEffect` instead of using `useHydrateAtoms` fixes the consistency issue, but it's less elegant and has an initial delay where the values are not properly set.

Thanks in advance for the help!

dai-shi:
Just some quick questions / comments without looking in detail:
- Do you use `Provider`?
- `useHydrateAtoms` can only set values if atoms are not initialized. (If `useEffect` works, it sounds the case of initialized atoms.)
- If you use Provider, and clean up by remounting, `useHydrateAtoms` would work as expected.

[...]

PR #1207, opened 2022-06-03
docs(nextjs): educate provider

TwistedMinda:
Follow-up of [this issue](https://github.com/pmndrs/jotai/issues/1190), where it was unclear that NextJS should use `Provider` while we see no use-case without it.

@dai-shi Please review carefully, I'm not sure what I'm talking about too well

### DIFF
--- a/docs/guides/nextjs.mdx
+++ b/docs/guides/nextjs.mdx
@@ -44,6 +44,10 @@ const postData = atom((get) => {
 })
 ```
 
+## Provider
+
+Because of the nature of SSR, your app can have more instances existing in JS memory in the same time. You need to wrap your app inside a `Provider` (view more details [in the Core section](../api/core.mdx)) so that each instance has its own state and will not interfere with previous values from a common Provider (provider-less mode).
+
 ## Examples
 
 ### Clock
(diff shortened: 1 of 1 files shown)


docs/guides/nextjs.mdx
```diff
@@ -44,6 +44,10 @@ const postData = atom((get) => {
 })
 ```
 
+## Provider
+
+Because of the nature of SSR, your app can have more instances existing in JS memory in the same time. You need to wrap your app inside a `Provider` (view more details [in the Core section](../api/core.mdx)) so that each instance has its own state and will not interfere with previous values from a common Provider (provider-less mode).
```
dai-shi, line 49:
```suggestion
Because of the nature of SSR, your app can have more instances existing in JS memory in the same time. You need to wrap your app inside a `Provider` (view more details [in the Core section](../api/core.mdx)) so that each instance has its own state and will not interfere with previous values from a default store (provider-less mode).
```

to make a little more precise. Though, we don't often use the term "store" in docs yet.

dai-shi (approved):
Thanks for the suggestion. Let's merge this with a small change.

Issue #1041, opened 2022-03-04
[docs] refactor docs


LuckyPigeon:
Docs current have different typesetting with different pages. It will increase a lot of reading experience if we set a typesetting, 
I think the typesetting of [freezeAtom](https://jotai.org/docs/utils/freeze-atom) would be nice, we can start by refactor `utils` typesetting.

dai-shi:
@sandren is working on refactoring docs. Let's see what we can do after some progress.

sandren:
> @sandren Need some help on this?

@TwistedMinda I'm currently working on writing a new introduction doc and reorganizing the utils section. If you would like to continue copyediting some of the other docs, that would be really helpful. Thanks! 🙂

TwistedMinda:
@sandren No prob'

TwistedMinda:
@sandren Just for you to know, we're gonna add `loadable` util very soon to the utils page. Maybe that changes something in your rework of utils page :p

TwistedMinda:
@sandren What's the state of this issue? The issue here is mostly about code blocks typings.
Now that you did set up the Table of contents, I think it can be a good time to clear that up.

- when there is a component
   - if typescript matters in the example : use "tsx"
   - if typescript doesn't really matter: use "jsx"
- when there is no component
  - if typescript matters in the example: use "ts"
  - if typescript matters in the example: use "js"

I understood the issue correctly, right?
I can make a deep check of all pages if you confirm ;)

dai-shi:
Is the OP about providing code snippet in TS?
Your pattern seem to be about how to write markdowns, which is the same pattern of mine.

TwistedMinda:
@dai-shi OP "typechecking" means the codesandbox at the end of the doc in freezeAtom? I'm confused

TwistedMinda:
@sandren @dai-shi Let's break down the possibilities:
a) Make sure all markdown are using the correct typing depending on the context
b) Make sure all markdown are "tsx"
c) Make sure all pages contain a CodeSandBox at the end to demonstrate

What do you think we have to do for this issue?
Seems like a simple issue but we just need to understand each other :p!

dai-shi:
> we just need to understand each other

well said.

While I don't understand what OP suggests, I think we should do a) for this issue, and close it once done.

TwistedMinda:
@dai-shi Good. I'll do it then

TwistedMinda:
Closing it in favor of the PR.

[...]

PR #1211, opened 2022-06-07
docs(markdowns): clean-up

TwistedMinda:
Following-up [this issue](https://github.com/pmndrs/jotai/issues/1041) we decided to clean all markdowns, a lot were incorrect and some were overkill.

Pattern followed to clean:

- when there is a component & the component needs to return jsx
  - if typescript matters in the example : use "tsx"
  - if typescript doesn't really matter: use "jsx"
- when there is no component or the component doesn't need to return jsx
  - if typescript matters in the example: use "ts"
  - if typescript doesn't really matter: use "js"

### DIFF
--- a/docs/advanced-recipes/atom-creators.mdx
+++ b/docs/advanced-recipes/atom-creators.mdx
@@ -30,7 +30,7 @@ The setter function can have an optional argument to force a particular state, s
 
 Here is how it's used.
 
-```ts
+```js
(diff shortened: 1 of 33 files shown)


dai-shi (approved):
Thanks. I like to see this fixed.
We no longer use `React.FC` in our lib code. Docs should follow it too.

PR #1212, opened 2022-06-07
docs(mental-modal): new section

TwistedMinda:
The idea here is to create a page that would entertain new comers to embrace the mental model of Jotai. Ideally with:
- how to forget everything and start from scratch
- examples of patterns that were hard before jotai
- examples of patterns that were boring before jotai
- that it's just the beginning (maybe introduce atom-with-* future public repositories ?)


dai-shi:
> I think anyone reading the docs carefully should understand that jotai replaces useContext.

Okay then.

> in a Comparison doc with, a) useContext

That sounds great.

TwistedMinda:
@dai-shi Okay I've added your comparison to useContext to the Comparison page.

But I also noticed that the "Concept" page is just a comparison to recoil & useContext, so I moved it to Comparison too.
For the concept page, I instead added the content of my new "mental model" doc, that I still need to write. What do you think ?

dai-shi:
> But it feels weird that we have no "Sync" guide. Without "primitives", it's directly "Async" before we even mentioned "useAtom".

We need "Getting Started" at first. Hope @sandren can work on it.

dai-shi:
Hi @TwistedMinda !
So sorry for the pause. Are you still around?
If so, I think we should move this and #1208 forward, even if it's partially complete.
We will then re-organize things.

dai-shi:
There is room for improvement, but instead of leaving the PR open, let's merge it and improve it. So sorry for the delay.

PR #1242, opened 2022-06-20
fix: copy built-in awaited type

Newbie012:
It fixes a bug when trying to use the atom in conjunction with generics. Given the following code:

```typescript
const useX = <$StageName extends string>(params: {
    atom: WritableAtom<$StageName, SetStateAction<$StageName>, void>;
    stages: $StageName[];
  }) => {
    const [currentStage, setCurrentStage] = useAtom(params.atom);
    const stageNames = Object.keys(params.stages) as $StageName[];
    const currentStageNumber = stageNames.indexOf(currentStage) + 1;
}
```

I get the following error:

```
Argument of type 'Awaited<$StageName>' is not assignable to parameter of type '$StageName'.
  'Awaited<$StageName>' is assignable to the constraint of type '$StageName', but '$StageName' could be instantiated with a different subtype of constraint 'string'.
    Type 'unknown' is not assignable to type '$StageName'.
      '$StageName' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.ts(2345)
```

Using the built-in `Awaited` type fixes this issue.

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -1,4 +1,10 @@
-type Awaited<T> = T extends Promise<infer V> ? Awaited<V> : T
+type Awaited<T> =
+    T extends null | undefined ? T : // special case for `null | undefined` when not in `--strictNullChecks` mode
+        T extends object & { then(onfulfilled: infer F): any } ? // `await` only unwraps object types with a callable `then`. Non-object types are not unwrapped
+            F extends ((value: infer V, ...args: any) => any) ? // if the argument to `then` is callable, extracts the first argument
+                Awaited<V> : // recursively unwrap the value
+                never : // the argument to `then` was not callable
+        T; // non-object or non-thenable
 
 interface Getter {
   <Value>(atom: Atom<Value | Promise<Value>>): Value
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -8,7 +8,13 @@ import {
 } from './suspensePromise'
 import type { SuspensePromise } from './suspensePromise'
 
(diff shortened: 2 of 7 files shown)


dai-shi:
Thanks for reporting. I wasn't aware the issue with generics.
While your solution makes sense to fix the issue, it doesn't feel very ideal. We want to use the built-in one, right?
Would you be interested in contributing to https://github.com/sandersn/downlevel-dts ?

dai-shi:
> Although, I was wondering why did you redeclare the type in each module instead of exporting it once?

It's not only about this type, but I try not to export private types (public types are defined in `./src/index.ts`), because we will change private types without notice, which breaks if people accidentally using private types (if it's exported).

Newbie012:
Interesting!
Anyway, not sure why I got so confused last night. I've opened a PR to downlevel-dts. Placing it here for a follow-up

 https://github.com/sandersn/downlevel-dts/pull/79

Newbie012:
@dai-shi I haven't got a response for the PR I made on downlevel-dts. In short, I've stumbled across [this issue](https://github.com/sandersn/downlevel-dts/pull/79#issuecomment-1162949598) which I believe would be a problem merging this PR.

For now, I believe we have two choices - keep the current state and open an a new bug issue, or merge it until further development. Honestly, I'm not happy with either of these options 🙁.

dai-shi:
> keep the current state and open an a new bug issue

Let's do this. Closing this PR for now.

For your case, I think module augmentation would work.
Check this workaround: https://tsplay.dev/mLqX2W

PR #1244, opened 2022-06-21
fix(core): improve versionedWrite logic

dai-shi:
following up #1154, it should somehow be improved before fixing #1239 tests with versionedWrite.

- [x] investigating behavior
- [x] fix logic and pass tests


.github/workflows/test-multiple-versions.yml
```diff
@@ -47,6 +47,8 @@ jobs:
         mode: [NORMAL, PROVIDER_LESS, VERSIONED_WRITE]
         exclude:
           - { react: 16.8.6, mode: VERSIONED_WRITE }
+          - { react: 16.9.0, mode: VERSIONED_WRITE }
+          - { react: 17.0.0, mode: VERSIONED_WRITE }
```
dai-shi, line 51: The new implementation depends on the new useReducer behavior in React 18, so pre-18 versions are no longer supported.

dai-shi:
Finally, the tests are fairly stable more than ever.

PR #1254, opened 2022-06-27
Handle ESM imports for @babel/template

### LINKED ISSUE #1251: `templateBuilder is not a function` in plugin-react-refresh.mjs
When using `'jotai/babel/plugin-react-refresh'` w/ vite it throws `templateBuilder is not a function` at `/node_modules/jotai/esm/babel/plugin-react-refresh.mjs`.

According to [@babel/template #13719](https://github.com/babel/babel/issues/13719#issuecomment-910252622), when using native mjs, it requires
```
import _template from "@babel/template";
const templateBuilder = _template.default;
```
instead of normal
```
import templateBuilder from "@babel/template";
```

Thisen:
This PR (hopefully) fixes ESM imports for `@babel/template`, when using `vite`.

Fixes #1251.

### DIFF
--- a/src/babel/plugin-debug-label.ts
+++ b/src/babel/plugin-debug-label.ts
@@ -1,7 +1,9 @@
 import path from 'path'
 import babel, { PluginObj } from '@babel/core'
-import templateBuilder from '@babel/template'
+import _templateBuilder from '@babel/template'
 import { isAtom } from './utils'
+
+const templateBuilder = (_templateBuilder as any).default || _templateBuilder
 
 export default function debugLabelPlugin({
   types: t,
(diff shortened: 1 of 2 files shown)


dai-shi:
> Getting the same error.

Not sure either, but the error message is different. 

> [plugin:vite:react-babel] /home/projects/vitejs-vite-5z9vhf/src/main.tsx: _babel_template.default is not a function

Thisen:
Should we merge and try?

dai-shi:
No, we don't need to merge for a try. Doesn't csb build work? One can install it locally.

Thisen:
@dai-shi I tried out this PR version with https://github.com/alan-nf/jotai-vite-react-refresh and it works 💪🏼 

Ready to merge!

dai-shi (approved):
The change looks okay.

PR #1260, opened 2022-07-01
fix(types): can not export atoms

u3u:
```ts
// ts-error:
// Exported variable 'itmeSizeAtom' has or is using name 'WithInitialValue'
// from external module but cannot be named.ts(4023)
export const itmeSizeAtom = atom(16)
```

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -37,10 +37,6 @@ type Write<Update, Result extends void | Promise<void>> = (
   update: Update
 ) => Result
 
-type WithInitialValue<Value> = {
-  init: Value
-}
-
 export type Scope = symbol | string | number
 
 // Not exported for public API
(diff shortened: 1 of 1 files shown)


dai-shi:
Oh no, I didn't realize it. This is a regression with #1234.

We don't want to export the internal types and I have another possible fix.
Would you create an issue for tracking? I can do that, otherwise.

u3u:
Okay, I've created #1261

src/core/atom.ts
```diff
@@ -37,7 +37,7 @@ type Write<Update, Result extends void | Promise<void>> = (
   update: Update
 ) => Result
 
-interface WithInitialValue<Value> {
+export interface WithInitialValue<Value> {
```
dai-shi, line 40: I think it should work with type alias as it used to be so:
```suggestion
type WithInitialValue<Value> = {
```
(but, eslint yells at this...)

dai-shi (approved):

PR #1273, opened 2022-07-06
feat: versionedWrite in useGotoAtomsSnapshot

Aslemammad:
This addresses another task of #911 which tries to bring the versionedWrite ability to useGotoAtomsSnapshot (as it was in useAtomsDevtools).

### DIFF
--- a/src/devtools/useGotoAtomsSnapshot.ts
+++ b/src/devtools/useGotoAtomsSnapshot.ts
@@ -21,26 +21,28 @@ export function useGotoAtomsSnapshot(scope?: Scope) {
   }
 
   return useCallback(
-    (values: Iterable<readonly [Atom<unknown>, unknown]> | AtomsSnapshot) => {
-      const writeStore = (value: any) => {
+    (snapshot: Iterable<readonly [Atom<unknown>, unknown]> | AtomsSnapshot) => {
+      const restoreAtoms = (
+        values: Iterable<readonly [Atom<unknown>, unknown]>
+      ) => {
         if (versionedWrite) {
           versionedWrite((version) => {
-            store[RESTORE_ATOMS](value, version)
+            store[RESTORE_ATOMS](values, version)
           })
         } else {
-          store[RESTORE_ATOMS](value)
+          store[RESTORE_ATOMS](values)
         }
       }
-      if (isIterable(values)) {
+      if (isIterable(snapshot)) {
         if (__DEV__) {
           console.warn(
             'snapshot as iterable is deprecated. use an object instead.'
           )
         }
-        writeStore(values)
+        restoreAtoms(snapshot)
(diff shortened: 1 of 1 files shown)


src/devtools/useGotoAtomsSnapshot.ts
```diff
@@ -14,27 +14,35 @@ type AtomsSnapshot = Readonly<{
...
     (values: Iterable<readonly [Atom<unknown>, unknown]> | AtomsSnapshot) => {
+      const writeStore = (value: any) => {
+        if (versionedWrite) {
+          versionedWrite((version) => {
+            store[RESTORE_ATOMS](value, version)
+          })
+        } else {
+          store[RESTORE_ATOMS](value)
+        }
+      }
```
dai-shi, line 33:
```suggestion
      const restoreAtoms = () => {
        if (versionedWrite) {
          versionedWrite((version) => {
            store[RESTORE_ATOMS](values, version)
          })
        } else {
          store[RESTORE_ATOMS](values)
        }
      }
```

Should be good without `any`?

dai-shi (commented):
Thanks for working on it.
(tbh, I'm not sure if `versionedWrite` makes sense in real use cases, but this is more consistent and why i requested.)

src/devtools/useGotoAtomsSnapshot.ts
```diff
@@ -14,27 +14,35 @@ type AtomsSnapshot = Readonly<{
...
           console.warn(
             'snapshot as iterable is deprecated. use an object instead.'
           )
         }
-        store[RESTORE_ATOMS](values)
+        writeStore(values)
         return
       }
-      store[RESTORE_ATOMS](values.values)
+      writeStore(values.values)
```
Aslemammad, line 43: Look at this!

src/devtools/useGotoAtomsSnapshot.ts
dai-shi: Oh, I missed it. Then, I'd do this.

```ts
  return useCallback(
    (snapshot: AtomsSnapshot | Iterable<readonly [Atom<unknown>, unknown]>) => {
      const restoreAtoms = (values: Iterable<readonly [Atom<unknown>, unknown]>) => {
        if (versionedWrite) {
          versionedWrite((version) => {
            store[RESTORE_ATOMS](values, version)
          })
        } else {
          store[RESTORE_ATOMS](values)
        }
      }
      if (isIterable(snapshot)) {
        if (__DEV__) {
          console.warn(
            'snapshot as iterable is deprecated. use an object instead.'
          )
        }
        restoreAtoms(snapshot)
        return
      }
      restoreAtoms(snapshot.values)
    },
    [store, versionedWrite]
  )
```

PR #1281, opened 2022-07-14
fix(xstate): Allow restarting a machine in a global atom

### LINKED ISSUE #1280: [jotai/xstate] Allow restarting a machine in a global atom
Basically there is no way to restart a machine that has been set to an atom in global scope. When it reaches its final state it would not accept any events.

jahglow:
The XState machine can be started with a default state and a custom passed state. This PR handles only the default one (as is the atom - doesn't handle hydration).

It is done by sending a jotai `RESET` in `send`:
```ts
import { RESET } from 'jotai/utils'
// ...

const [current, send] = useAtom(myMachineAtom);

// example usage of resetting machine by leaving a route in react-navigation:
  useFocusEffect(
    useCallback(
      () =>
        navigation.addListener('beforeRemove', () => {
          send(RESET);
        }),
      [navigation, send],
    ),
  );

```
Typescript will not complain as `typeof RESET` is added to the types of events that `send` my accept

### DIFF
--- a/src/xstate/atomWithMachine.ts
+++ b/src/xstate/atomWithMachine.ts
@@ -160,6 +160,10 @@ export function atomWithMachine<
       if (event === RESET) {
         service.stop()
         set(cachedMachineAtom, null)
+        set(machineAtom, null)
+        set(machineStateAtom, () => {
+          return
+        })
         service.start()
       } else {
         service.send(event)
--- a/src/xstate/atomWithMachine.ts
+++ b/src/xstate/atomWithMachine.ts
@@ -13,6 +13,7 @@ import type {
 } from 'xstate'
 import type { Atom, Getter, WritableAtom } from 'jotai'
 import { atom } from 'jotai'
+import { RESET } from 'jotai/utils'
 
 export interface MachineAtomOptions<TContext, TEvent extends EventObject> {
   /**
@@ -56,7 +57,7 @@ export function atomWithMachine<
   getOptions?: Options<TMachine> | ((get: Getter) => Options<TMachine>)
 ): WritableAtom<
   StateFrom<TMachine>,
-  MaybeParam<Prop<TInterpreter, 'send'>>,
+  MaybeParam<Prop<TInterpreter, 'send'>> | typeof RESET,
   void
 > {
   const cachedMachineAtom = atom<{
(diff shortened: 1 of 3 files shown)


tests/xstate/atomWithMachine.test.tsx
```diff
@@ -47,4 +64,10 @@ it('toggle machine', async () => {
 
   fireEvent.click(getByText('Active! Click to deactivate'))
   await findByText('Click to activate')
+
+  fireEvent.click(getByText('Go to final state'))
+  await findByText('Not actionable')
+
+  fireEvent.click(getByText('Restart machine'))
+  await waitFor(() => expect(findByText('Click to activate')).toBeDefined())
```
dai-shi, line 72: cool. seems working.

can you please create a new spec, instead of modifying the existing one?
```js
it('resettable machine', () => {
```

jahglow: okay.

jahglow:
It turns out that stopping machine and starting it was actually resetting its context to the initial one, while the context stayed intact which was a bug and i fixed it in the other commit. In case of the `RESET` which solves the problem of restarting global atoms - they do not need unsubscription like being stopped (disposed with the app/page). The atoms with scope Providers will be disposed on unmount of the said provider and thus do not their context be reset.

jahglow:
And the test that fails here is not related to my changes. is it flaky?

src/xstate/atomWithMachine.ts
```diff
@@ -154,9 +156,21 @@ export function atomWithMachine<
...
+      set,
+      event: Parameters<AnyInterpreter['send']>[0] | typeof RESTART
+    ) => {
       const { service } = get(machineAtom)
-      service.send(event)
+      if (event === RESTART) {
+        service.stop()
+        set(cachedMachineAtom, null)
+        set(machineAtom, null)
+        set(machineStateAtom, () => void 0)
```
dai-shi, line 169: This doesn't seem correct.
we can't stop machine on unmount.

A possible fix would be to make L133 `service.stop()` to
```js
  const { service } = get(machineAtom)
  service.stop()
```
and change here:
```suggestion
        const { service: newService } = get(machineAtom)
        newService.onTransition((nextState: any) => {
          set(cachedMachineStateAtom, nextState)
        })
        newService.start()
```
and remove L170.

I'm not so sure...

jahglow: the proposal LGTM

jahglow:
> The code looks to me.
Does everything work as expected for you?
Could you create a small codesandbox example?
Also, can you add some description about this new feature in docs, with the codesandbox example?

Yes it does work as expected for me, actually. I use this fix in react-native and it proves to work as expected.

for the sandbox do you mind if its the same example as in the test? It illustrates the usage very well. I might think of something else though but I don't have much time on my hands, the deadline for the project is in this sprint.

I'll update the docs tomorrow, let it hang for a while until I get my hands on em

dai-shi:
> for the sandbox do you mind if its the same example as in the test? It illustrates the usage very well.

the same example in the test is fine. it's still helpful to have it as a codesandbox and put it in docs.

thanks for your help!

dai-shi:
Let me merge this soon.

@jahglow Please work on the docs and the example in a new PR.

PR #1293, opened 2022-07-20
docs(advanced-recipes): initialize atom on render

austinwoon:
Docs for initialising atom on render.

Had a use-case for creating a reusable component which relies on shared atom state in my project. The root component had to take in different props to initialise atom state.

Before Jotai, we would initialise values with with React Context by passing down the values as props to the parent Context Provider. 

Read this discussion here: https://github.com/pmndrs/jotai/discussions/581 and was able to achieve the same outcome with jotai.

Noted that this advanced-recipe wasn't created so took the opportunity to do a small quick contribution!

Please let me know if the example is and phrasing of the docs are clear. Thank you!


dai-shi:
Thanks for working on docs (we need more contributor on docs).

Per my vague/undocumented/unstable definition of "advanced recipes", this is not an advanced recipe. Can you put it in "guides"?

dai-shi:
I just noticed that in my comment, I said "recipes".

> I'd encourage someone to write this in docs / recipes / initialize-atom-on-render or something.

https://github.com/pmndrs/jotai/discussions/581#discussioncomment-972361

dai-shi (approved):
This looks a good start. Thanks for your contribution.

PR #1301, opened 2022-07-22
refactor(utils): use prev property instead of refAtom

eolme:
Instead of refAtom we can use prev property safely.

TODO: write tests

### DIFF
--- a/src/utils/selectAtom.ts
+++ b/src/utils/selectAtom.ts
@@ -4,6 +4,8 @@ import { createMemoizeAtom } from './weakCache'
 
 type Awaited<T> = T extends Promise<infer V> ? Awaited<V> : T
 
+type AtomWithPrev<Value, Prev = Value> = Atom<Value> & { prev?: Prev }
+
 const memoizeAtom = createMemoizeAtom()
 
 export function selectAtom<Value, Slice>(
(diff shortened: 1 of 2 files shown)


dai-shi:
Unfortunately, this doesn't work with multiple provider use cases.

src/core/store.ts
```diff
@@ -667,8 +672,7 @@ export const createStore = (
     ) => {
       let promiseOrVoid: void | Promise<void>
       if ((a as AnyWritableAtom) === atom) {
-        if (!hasInitialValue(a)) {
-          // NOTE technically possible but restricted as it may cause bugs
+        if (!isWritable(a)) {
```
dai-shi, line 675: This is not correct. We only allow setting value to non-derived atom.

dai-shi:
You might be interested in our past trial: #936

dai-shi:
To elaborate, your idea can be implemented not with `atom.init` but with `atom.prev`, without touching core. (Or, even with `WeakMap<atom, prev>`). It should then pass tests, because we don't have multiple provider use case tests.

dai-shi:
Thanks for trying. We should eventually add some tests for multiple provider use cases to make this change fail.

Let me close this to avoid confusion (this is not a right approach.)

PR #1311, opened 2022-07-26
breaking(devtools): renew useAtomsSnapshot and useGotoAtomsSnapshot

### LINKED ISSUE #911: improve useAtomsSnapshot / useGotoAtomsSnapshot

dai-shi:
Close #911

What's breaking: Previously `useAtomsSnapshot` returns a single `Map`, and now it returns `{ values: Map, dependents: Map }`.

This PR includes:
- [x] #1116 
- [x] #1273 
- [x] #1286


dai-shi:
This should be good. Let's merge.

@Aslemammad As you notice, we have common code in useAtomsDevtools and useAtomsSnapshot. We may want to try refactoring to use `useAtomsSnapshot` inside `useAtomsDevtools`. What do you think? Let's tackle it later.

Issue #957, opened 2022-01-12
Prevent suspending atomWithQuery when keepPreviousData is true


alexski-perry:
I think this has been brought up before in #374, but it seems that setting `keepPreviousData` to true on `atomWithQuery` doesn't prevent suspending, which I believe should be the intended behaviour. 

For instance in [this example](https://codesandbox.io/s/atomwithquery-keeppreviousdata-yowuv), whenever the query key changes, the list jumps to and from the loading state. I don't think this should be the case with `keepPreviousData` set to true.

dai-shi:
Sounds reasonable. It can suspend at the first time, right?

Hope @aulneau comes up with a solution.

alexski-perry:
Yes it should suspend the first time, unless `initialData` is set or initial data is supplied using `queryClient.setQueryData` or `queryClient.prefetchQuery`.

TwistedMinda:
@alexski-perry It feels like your example works just great!
It's actually an impressive demo of how lovely react-query is :p 

Simple workflow that shows it works:
- type "red" (it will load)
- erase everything
- type "red" again (it will not load)

and just to be sure, repeat with another color.
Maybe I missed something?

TwistedMinda:
@alexski-perry Oh, yes indeed, my bad! Thanks for the explanation
I don't know how I came up with that interpretation. Because what I was describing (your 1st example's behaviour) is actually the exact behaviour of just "suspend: true"

Thanks for the demo!

PR #1313, opened 2022-07-27
fix(query): prevent creating QueryObserver multiple times

Aslemammad:
Resolves: #957

This issue tries to track the query options through `queryObserver.setOptions`, so @tanstack/query-core features like `keepPreviousData` work properly!


src/query/atomWithQuery.ts
```diff
@@ -156,7 +170,6 @@ export function atomWithQuery<
             }
           }, 1000)
           resolve(result)
-          resolve = null
```
dai-shi, line 159: This seems bad. (Interesting all tests pass... 🤔

src/query/atomWithQuery.ts
```diff
@@ -123,8 +130,15 @@ export function atomWithQuery<
         typeof createQuery === 'function' ? createQuery(get) : createQuery
 
       const defaultedOptions = queryClient.defaultQueryOptions(options)
-      const observer = new QueryObserver(queryClient, defaultedOptions)
-      const initialResult = observer.getCurrentResult()
+
+      if (observer) {
```
dai-shi, line 134: This doesn't work if `queryClient` is changed, right?

src/query/atomWithQuery.ts
```diff
@@ -95,6 +94,14 @@ export function atomWithQuery<
     isMounted: boolean
     unsubscribe: (() => void) | null
   }
+
+  let observer: QueryObserver<
```
dai-shi, line 98: this is going to be a module scope variable, which breaks the use case with multiple `<Provider>`s.

src/query/atomWithQuery.ts
Aslemammad: I cannot say yes or no! but in useQuery, it tracks the queryClient!

src/query/atomWithQuery.ts
Aslemammad: Isn't refAtom the best solution for that now?

src/query/atomWithQuery.ts
Aslemammad: Removing it will make the tests fail!

Aslemammad: in versionedWrite

src/query/atomWithQuery.ts
dai-shi: It's better, but not best. I still believe there's a way without refAtom.

src/query/atomWithQuery.ts
dai-shi: hmm, there might be something wrong in other places..

src/query/atomWithQuery.ts
```diff
@@ -90,29 +90,39 @@ export function atomWithQuery<
...
+    TError,
+    TData,
+    TQueryData,
+    TQueryKey
+  >
   type Result = QueryObserverResult<TData, TError>
-  type State = {
-    isMounted: boolean
-    unsubscribe: (() => void) | null
-  }
```
dai-shi, line 97: Good that I was able to remove this. Didn't like this.

src/query/atomWithQuery.ts
dai-shi: We don't have a test to catch if we remove `resolve = null`.
It's a test to update result from queryClient, I suppose.

dai-shi (approved):
Should be good to go.

Aslemammad (commented):
LGTM!

PR #1314, opened 2022-07-27
docs(core): clarify `useSetAtom`

dwelle:
A quick change as discussed on chat to clarify that `useSetAtom()` is useful in general as a performance improvement.

### DIFF
--- a/docs/api/core.mdx
+++ b/docs/api/core.mdx
@@ -266,7 +266,7 @@ export default function App() {
 
 In case you need to update a value of an atom without reading it, you can use `useSetAtom()`.
 
-This is especially useful when the performance is a concern, as the `const [, setValue] = useAtom(valueAtom)` will cause unnecessary rerenders on each `valueAtom` update. This can be a problem in case of large components when the atom value changes often.
+This is especially useful when the performance is a concern, as the `const [, setValue] = useAtom(valueAtom)` will cause unnecessary rerenders on each `valueAtom` update.
 
 ## useAtomValue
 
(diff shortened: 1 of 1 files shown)


docs/api/core.mdx
```diff
@@ -264,11 +264,9 @@ export default function App() {
 }
 ```
 
-In case you create a write only atom where the value never changes you can use the `useSetAtom` hook.
-`useSetAtom` is premature optimization in this scenario.
+In case you need to update a value of an atom without reading it, you can use `useSetAtom()`.
 
-For primitive values if you use `const [, setValue] = useAtom(valueAtom)` it can cause unnecessary re-renders,
-so `useSetAtom` helps to avoid those extra re-renders.
+This is especially useful when the performance is a concern, as the `const [, setValue] = useAtom(valueAtom)` will cause unnecessary rerenders on each `valueAtom` update. This can be a problem in case of large components when the atom value changes often.
```
dai-shi, line 269: My concern is this could discourage too much about the usage of `const [, dispatch] = useAtom(actionAtom)` which is kind of recommended in readme.md. This is a valid usage with write-only atoms and it doesn't cause unnecessary re-renders. (It's better to use `useSetAtom`, but it can be premature optimization.)

Any idea to mitigate it?

dai-shi: @dwelle

dai-shi:
```suggestion
This is especially useful when the performance is a concern, as the `const [, setValue] = useAtom(valueAtom)` will cause unnecessary rerenders on each `valueAtom` update.
```
@sandren Can you review wordings? ☝️

dwelle:
hey @dai-shi — sorry for completely missing this! Inbox zero is sadly something I have not seen in a while 😭

PR #1338, opened 2022-08-09
docs: atomWithDebounce atomCreator

austinwoon:
Docs for new `atomWithDebounce` atom creator as discussed here: https://github.com/pmndrs/jotai/discussions/553#discussioncomment-3354600

### DIFF
--- a/docs/advanced-recipes/atom-creators.mdx
+++ b/docs/advanced-recipes/atom-creators.mdx
@@ -414,4 +414,8 @@ export function atomWithDebounce<Value>(
 
 ### Example Usage
 
+The sandbox link below shows how we would use a derived atom to fetch state based on the value of `debouncedValue`.
+
+When typing a pokemon's name in `<SearchInput>`, we do not fire a fetch on every letter, but only after `delayMilliseconds` has passed since the last text input.
+
 [CodeSandbox](https://codesandbox.io/s/withered-tdd-cjrz2y?file=/src/App.tsx)
--- a/docs/advanced-recipes/atom-creators.mdx
+++ b/docs/advanced-recipes/atom-creators.mdx
@@ -364,7 +364,7 @@ const ListOfThings = () => {
 This util is useful for text search inputs, where you would like to call **functions in derived atoms only once** after waiting for a duration, instead of firing an action on every keystroke.
 
 ```tsx
-import { atom, Getter, PrimitiveAtom, Setter } from 'jotai'
+import { atom } from 'jotai'
 import { SetStateAction } from 'react'
 
 export default function atomWithDebounce<T>(
@@ -372,29 +372,19 @@ export default function atomWithDebounce<T>(
   delayMilliseconds = 500,
   shouldDebounceOnReset = false
 ) {
-  const prevTimeoutAtom = atom<ReturnType<typeof setTimeout> | null>(null)
+  const prevTimeoutAtom = atom<ReturnType<typeof setTimeout> | undefined>(
+    undefined
+  )
 
   // DO NOT EXPORT currentValueAtom as using this atom to set state can cause
   // inconsistent state between currentValueAtom and debouncedValueAtom
   const _currentValueAtom = atom(initialValue)
   const isDebouncingAtom = atom(false)
 
-  const clearTimeoutId = (
-    get: Getter,
-    set: Setter,
-    prevTimeoutAtom: PrimitiveAtom<ReturnType<typeof setTimeout> | null>
-  ) => {
-    const prevTimeoutId = get(prevTimeoutAtom)
-    if (prevTimeoutId) {
-      clearTimeout(prevTimeoutId)
-      set(prevTimeoutAtom, null)
-    }
-  }
-
   const debouncedValueAtom = atom(
     initialValue,
     (get, set, update: SetStateAction<T>) => {
-      clearTimeoutId(get, set, prevTimeoutAtom)
+      clearTimeout(get(prevTimeoutAtom))
 
       const prevValue = get(_currentValueAtom)
       const nextValue =
@@ -430,7 +420,7 @@ export default function atomWithDebounce<T>(
 
   // exported atom setter to clear timeout if needed
   const clearTimeoutAtom = atom(null, (get, set, _arg) => {
-    clearTimeoutId(get, set, prevTimeoutAtom)
+    clearTimeout(get(prevTimeoutAtom))
     set(isDebouncingAtom, false)
   })
 
--- a/docs/advanced-recipes/atom-creators.mdx
+++ b/docs/advanced-recipes/atom-creators.mdx
@@ -355,19 +355,11 @@ const ListOfThings = () => {
 }
 ```
 
-<CodeSandbox id="ugkzm0" />
-
 ## atomWithDebounce
 
 > `atomWithDebounce` helps with creating an atom where state set should be debounced.
 
 This util is useful for text search inputs, where you would like to call **functions in derived atoms only once** after waiting for a duration, instead of firing an action on every keystroke.
-
-### Caveat
-
-Please note that this atom has different objectives from concurrent features in React 18 such as `useTransition` and `useDeferredValue` whose main aim is to prevent blocking of interaction with the page for expensive updates.
-
-For more info, please read this github discussion https://github.com/reactwg/react-18/discussions/41 under How is it different from setTimeout?
 
 ```tsx
 import { atom, Getter, PrimitiveAtom, Setter } from 'jotai'
@@ -450,6 +442,12 @@ export default function atomWithDebounce<T>(
 }
 ```
 
+### Caveat
+
+Please note that this atom has different objectives from concurrent features in React 18 such as `useTransition` and `useDeferredValue` whose main aim is to prevent blocking of interaction with the page for expensive updates.
+
+For more info, please read this github discussion https://github.com/reactwg/react-18/discussions/41 under the section titled **"How is it different from setTimeout?"**
+
 ### Example Usage
 
 The sandbox link below shows how we would use a derived atom to fetch state based on the value of `debouncedValueAtom`.
@@ -458,4 +456,4 @@ When typing a pokemon's name in `<SearchInput>`, we do not send a get request on
 
 This reduces the number of backend requests to the server.
 
-[CodeSandbox](https://codesandbox.io/s/withered-tdd-cjrz2y?file=/src/App.tsx)
+<CodeSandbox id="cjrz2y" />
--- a/docs/advanced-recipes/atom-creators.mdx
+++ b/docs/advanced-recipes/atom-creators.mdx
@@ -361,52 +361,90 @@ const ListOfThings = () => {
 
 > `atomWithDebounce` helps with creating an atom where state set should be debounced.
 
-This can be especially useful for text search inputs, where you would like to fire an action only after the user has stopped typing for a delay, instead of firing an action on every keystroke.
+This util is useful for text search inputs, where you would like to call **functions in derived atoms only once** after waiting for a duration, instead of firing an action on every keystroke.
+
+### Caveat
+
+Please note that this atom has different objectives from concurrent features in React 18 such as `useTransition` and `useDeferredValue` whose main aim is to prevent blocking of interaction with the page for expensive updates.
+
+For more info, please read this github discussion https://github.com/reactwg/react-18/discussions/41 under How is it different from setTimeout?
 
 ```tsx
-export function atomWithDebounce<Value>(
-  initialValue: Value,
+import { atom, Getter, PrimitiveAtom, Setter } from 'jotai'
+import { SetStateAction } from 'react'
+
+export default function atomWithDebounce<T>(
+  initialValue: T,
   delayMilliseconds = 500,
   shouldDebounceOnReset = false
 ) {
-  let timeout: NodeJS.Timeout
+  const prevTimeoutAtom = atom<ReturnType<typeof setTimeout> | null>(null)
 
   // DO NOT EXPORT currentValueAtom as using this atom to set state can cause
   // inconsistent state between currentValueAtom and debouncedValueAtom
-  const currentValueAtom = atom(initialValue)
+  const _currentValueAtom = atom(initialValue)
+  const isDebouncingAtom = atom(false)
+
+  const clearTimeoutId = (
+    get: Getter,
+    set: Setter,
+    prevTimeoutAtom: PrimitiveAtom<ReturnType<typeof setTimeout> | null>
+  ) => {
+    const prevTimeoutId = get(prevTimeoutAtom)
+    if (prevTimeoutId) {
+      clearTimeout(prevTimeoutId)
+      set(prevTimeoutAtom, null)
+    }
+  }
 
   const debouncedValueAtom = atom(
     initialValue,
-    (get, set, update: SetStateAction<Value>) => {
-      clearTimeout(timeout)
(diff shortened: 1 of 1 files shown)


AjaxSolutions:
How do I cancel debounce cycle?

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -356,3 +356,66 @@ const ListOfThings = () => {
...
+
+This can be especially useful for text search inputs, where you would like to fire an action only after the user has stopped typing for a delay, instead of firing an action on every keystroke.
+
+```tsx
+export function atomWithDebounce<Value>(
+  initialValue: Value,
+  delayMilliseconds = 500,
+  shouldDebounceOnReset = false
+) {
+  let timeout: NodeJS.Timeout
```
dai-shi, line 372: Let's remove this for now as this shared variable is not correct for atom creators.

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -356,3 +356,66 @@ const ListOfThings = () => {
...
+      }, delayMilliseconds)
+    }
+  )
+
+  const isDebouncingAtom = atom(
+    (get) => get(currentValueAtom) !== get(debouncedValueAtom)
+  )
+
+  return {
+    readOnlyCurrentValueAtom: atom((get) => get(currentValueAtom)),
```
dai-shi, line 408: Can we just call it `valueAtom`?
```suggestion
    valueAtom: atom((get) => get(currentValueAtom)),
```

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -356,3 +356,66 @@ const ListOfThings = () => {
...
+    debouncedValueAtom,
+  }
+}
+```
+
+### Example Usage
+
+The sandbox link below shows how we would use a derived atom to fetch state based on the value of `debouncedValue`.
+
+When typing a pokemon's name in `<SearchInput>`, we do not fire a fetch on every letter, but only after `delayMilliseconds` has passed since the last text input.
```
dai-shi, line 419: We should explore `useDefferedValue` and `useTransition` for this kind of use cases. It should be far better. This example gives me a little hesitation to merge because it essentially guides to wrong direction.

Do you have another example?

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -356,3 +356,66 @@ const ListOfThings = () => {
...
+
+[CodeSandbox](https://codesandbox.io/s/withered-tdd-cjrz2y?file=/src/App.tsx)
```
dai-shi, line 421: I got an error...
image

dai-shi (changes requested):
Please see comments.

docs/advanced-recipes/atom-creators.mdx
austinwoon: the api only works for valid pokemon names, did not catch errors as it was just a demo haha 😆 try `ditto` or `charmander`

docs/advanced-recipes/atom-creators.mdx
austinwoon: May i clarify if `useDefferedValue` and `useTransition` are custom hooks to get atom setters for debouncing the value and setting `isDebouncing`?

docs/advanced-recipes/atom-creators.mdx
austinwoon: Continuing from where we left off, could we set `timeout` to be an atom that stores the `timeoutId` then? Would this solve the issue?

austinwoon:
> How do I cancel debounce cycle?

At the moment there is no way to "cancel" a debounce cycle. If you need the current value that is updated, you can do so by obtaining `readOnlyCurrentValueAtom` (about to be renamed `currentValueAtom`) which is returned from `atomWithDebounce` creator

docs/advanced-recipes/atom-creators.mdx
dai-shi: Do you mean this?
```ts
const timeoutAtom = atom<NodeJS.Timeout | null>(null)
```
Yea, it should solve the issue.

docs/advanced-recipes/atom-creators.mdx
dai-shi: It's different from debouncing. It will allow data fetching in a lower priority task, so we can keep typing in the text field. The benefit is we don't need to specify the `delay` number. React will automatically adjust it depending on the cpu availability.
This is still a new practice in React, but essentially this will replace the debouncing use case.

docs/advanced-recipes/atom-creators.mdx
dai-shi: Is it too hard to fix? Would be nice even if it's just a demo. Or, it's even nice because it's a *demo*.

docs/advanced-recipes/atom-creators.mdx
dai-shi: So, I think we can merge this PR, with a note about new practice in React 18. And, we should work on a new guide about the practice. Hope you can help on it too.

docs/advanced-recipes/atom-creators.mdx
austinwoon: Did a hacky fix in sandbox here: https://codesandbox.io/s/withered-tdd-cjrz2y?file=/src/atoms.ts

Ideally should be done with `ErrorBoundary` but codesandbox seems to not work with ErrorBoundary well so had a hacky solution just for demo.

austinwoon:
> How do I cancel debounce cycle?

Have not updated code in the repo, but please look at sandbox here. I have created a new atom setter `clearTimeoutAtom` to suit this purpose, though i do not see any obvious use-case for it other than it being a nice to have. Could you perhaps share when you would need to trigger a cancel a cycle manually?

docs/advanced-recipes/atom-creators.mdx
austinwoon: Did amendments here: https://codesandbox.io/s/withered-tdd-cjrz2y?file=/src/atomWithDebounce.ts (will update the PR code at a later date after https://github.com/pmndrs/jotai/pull/1338#discussion_r942555474 pending discussion has been addressed)

I still do not see when collisions can occur as I thought the `window` object would guarantee that there will be no possible occurrence in the browser. If you have the time, could you perhaps share an instance where this might occur for my own learning?

dai-shi: https://codesandbox.io/s/beautiful-shannon-gy153b?file=/src/App.tsx

It's not easy to see from the UI (because `isDebouncingAtom` is an atom correctly), but it behaves incorrectly.

docs/advanced-recipes/atom-creators.mdx
dai-shi: I think your use of `useEffect` and `useDefferedValue` is not correct. But, I'm not sure if this use case can be solved with `useDefferedValue`.
I tried with `useTransition`, but still doesn't work as expected. https://codesandbox.io/s/brave-albattani-ldj47c

Hope I can try it again later.

docs/advanced-recipes/atom-creators.mdx
austinwoon: I see! just forked and tried it myself. You're right that it behaves incorrectly. Thanks for doing up this sandbox! TIL.

docs/advanced-recipes/atom-creators.mdx
austinwoon: Thanks! Good example, I was unfamiliar with that `transition` api as well.

We still don't have control over the number of times the expensive calculation should be done, hence there is some benefit in having a specified timeout delay. The only con would be that this delay sets a minimum threshold before the calculations can happen and is also arbitrarily set.

dai-shi: > Hope I can try it again later.

https://codesandbox.io/s/dawn-forest-2wrz9y

Okay, this ☝️ is what I got. We can't stop the big sleep because JS is single threaded. What works is something like this.

Another trial with `useDefferedValue`: https://codesandbox.io/s/gracious-wind-vp54iy?file=/src/_app.js Hmm, not working. I might be misunderstanding something.



> I think there is still some good use-cases for `atomWithDebounce` in react 18.

Yes, overall I agree with it. It's trade-off between automatic and manual. Both are valid.

I'm still interested in solutions with useDefferedValue and useTransition, but this PR is just fine without them.

AjaxSolutions:
See lodash [debounce](https://www.geeksforgeeks.org/lodash-_-debounce-method/)

They use an option named `wait`, it reads better imo than `delayMilliseconds`.  The [useDebounce](https://github.com/xnimorz/use-debounce) hook was mimicked after lodash.

The author of this hook was the first one IIRC who responded to Dan Abramov's [question](https://twitter.com/dan_abramov/status/1060729512227467264?lang=en).

It's up to you if you want to use `useDebounce` or `lodash` as a reference.

docs/advanced-recipes/atom-creators.mdx
austinwoon: Thanks! Tested your example vs the performance with debounce, and read [this](https://github.com/reactwg/react-18/discussions/41) explanation.

I felt that this author's explanation was the best at exemplifying the cons of `atomWithDebounce`. Using transitions makes the ui responsive, and does not block interaction with the page.

> Another important difference is that a large screen update inside a setTimeout will still lock up the page, just later after the timeout. If the user is still typing or interacting with the page when the timeout fires, they will still be blocked from interacting with the page. But state updates marked with startTransition are interruptible, so they won’t lock up the page. They lets the browser handle events in small gaps between rendering different components. If the user input changes, React won’t have to keep rendering something that the user is no longer interested in.

Will rephrase the doc to explicitly state that the use case for `atomWithDebounce` would be to limit number of function calls derived atoms from the debounced value will make for updates that frequently happen.

austinwoon: [5ac4625](https://github.com/pmndrs/jotai/pull/1338/commits/5ac462556b0871b64c44e839203d902f8ffd98ed)

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -355,4 +355,105 @@ const ListOfThings = () => {
 }
 ```
 
-<CodeSandbox id="ugkzm0" />
```
dai-shi, line 358: We should not delete this, right?

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -355,4 +355,105 @@ const ListOfThings = () => {
...
+    get: Getter,
+    set: Setter,
+    prevTimeoutAtom: PrimitiveAtom<ReturnType<typeof setTimeout> | null>
+  ) => {
+    const prevTimeoutId = get(prevTimeoutAtom)
+    if (prevTimeoutId) {
+      clearTimeout(prevTimeoutId)
+      set(prevTimeoutAtom, null)
+    }
+  }
```
dai-shi, line 390: This is just fine, but we prefer write-only atom pattern.

```js
const clearTimeoutAtom = atom(null, (get, set) => {
  const prevTimeout = get(prevTimeoutAtom)
  if (prevTimeout) {
    clearTimeout(prevTimeout)
    set(prevTimeoutAtom, null)
  }
})
```

In JS, `clearTimeout(null)` is no-op. So, we could simply do:
```ts
clearTimeout(get(prevTimeoutAtom) as ...)
```
But, in this recipe, the write-only atom above would be good.

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -355,4 +355,105 @@ const ListOfThings = () => {
...
+      // set previous timeout atom in case it needs to get cleared
+      set(prevTimeoutAtom, nextTimeoutId)
+    }
+  )
+
+  // exported atom setter to clear timeout if needed
+  const clearTimeoutAtom = atom(null, (get, set, _arg) => {
+    clearTimeoutId(get, set, prevTimeoutAtom)
+    set(isDebouncingAtom, false)
+    set(_currentValueAtom, get(debouncedValueAtom))
```
dai-shi, line 433: Why do you have this?
If we clear timeout, don't we want the old value, or do we want the new value?
Either way is fine for a recipe, though. Curious what would be expected in general.

dai-shi (changes requested):

docs/advanced-recipes/atom-creators.mdx
austinwoon: Ah, my purpose of making this function was so that it could be used in the write-only atom `clearTimeoutAtom` in line `432`, and also in the setter of `debouncedValueAtom`. Also to improve readability in the `debouncedValueAtom` setter. It was not meant to be exported.

docs/advanced-recipes/atom-creators.mdx
austinwoon: > If we clear timeout, don't we want the old value, or do we want the new value?

I misinterpreted AjaxSolution's comment as also to clear to the old value. Saw `useDebounce` implementation, and they simply just cancelled the debounce status and cleared the timeout. 

For now i will copy that implementation, and leave it to the users who read the docs to change the implementation to suit their usecase.

austinwoon: The initial idea I had was to set it to the value before the debounce state was updated, so that "clearing" the debounce cycle would also mean resetting state to what it was before.

austinwoon: [214c7aa](https://github.com/pmndrs/jotai/pull/1338/commits/214c7aa68d5ba263a8aa19b42802461383ad3f65)

docs/advanced-recipes/atom-creators.mdx
austinwoon: Sorry, was cutting the snippet for my code sandbox section and forgot to add it back. Amended here: [f5e72d3](https://github.com/pmndrs/jotai/pull/1338/commits/f5e72d36ee9820191a02e1ec40c774c55795ed37)

docs/advanced-recipes/atom-creators.mdx
austinwoon: Amended my typings to undefined, and removed this utility function so that the signature could be cleaner. Using `undefined` instead of `null` avoids the need for any typecasting or conditional checks to make the ts linter happy.

v

dai-shi: I didn't know `clearTimeout(undefined)` isn't complained by tsc.
https://tsplay.dev/wQ8rjW

docs/advanced-recipes/atom-creators.mdx
```diff
@@ -356,3 +356,95 @@ const ListOfThings = () => {
...
+
+## atomWithDebounce
+
+> `atomWithDebounce` helps with creating an atom where state set should be debounced.
+
+This util is useful for text search inputs, where you would like to call **functions in derived atoms only once** after waiting for a duration, instead of firing an action on every keystroke.
+
+```tsx
+import { atom } from 'jotai'
+import { SetStateAction } from 'react'
```
dai-shi, line 368: jotai exports the one, which seems cleaner?
```suggestion
import type { SetStateAction } from 'jotai'
```

dai-shi (approved):

docs/advanced-recipes/atom-creators.mdx
austinwoon: Ah, wrong import. 

Amended to

`import { atom, SetStateAction } from "jotai";`

PR #1346, opened 2022-08-13
feat(utils): support RESET with update function for atomWithStorage/atomWithHash/atomWithReset

### LINKED ISSUE #1335: `SetStateAction<T>` should accept `T | typeof RESET`
When using `atomWithHash` or `atomWithReset` it would be nice if the state setter function would also accept the `RESET` symbol as return type.

## Problem

Suppose I want to encode an `isOpen` flag into the url hash but also remove it entirely from the URL when it's false.

```tsx
import { atomWithHash, RESET } from 'jotai/utils'
import React, { useCallback, useMemo } from 'react'

export const useUrlHashToggle = (key: string) => {
  const isOpenAtom = useMemo(() => atomWithHash(key, false), [key])
  const [isOpen, setIsOpen] = useAtom(isOpenAtom)

  const toggle = useCallback(
    () => setIsOpen(isOpen ? RESET : true),
    [isOpen, setIsOpen]
  )

  return { isOpen, toggle }
}

const Component: React.FC = () => {
  const { isOpen, toggle } = useUrlHashToggle('key')

  const expensiveComponents = useMemo(
    () => <button onClick={toggle}>{/* expensive stuff */}</button>,
    [toggle]
  )

  return (
    <div>
      {expensiveComponents}
      {isOpen && <div>{/* other stuff */}</div>}
    </div>
  )
}
```

As the `toggle` function depends on `isOpen` it will change reference when `isOpen` is toggled. This way `expensiveComponents` will have to rerender on state change although they don't necessarily depend on `isOpen`.

## Desired Usage

It would be great if it were possible to return `RESET` in a state setter function:

```ts
export const useUrlHashToggle = (key: string) => {
  const isOpenAtom = useMemo(() => atomWithHash(key, false), [key])
  const [isOpen, setIsOpen] = useAtom(isOpenAtom)

  const toggle = useCallback(
    () => setIsOpen((prevIsOpen) => prevIsOpen ? RESET : true),
    [setIsOpen]
  )

  return { isOpen, toggle }
}
```

Now the `toggle` function would not change reference.

austinwoon:
## Context
Closes #1335 

## Approach

- Created new type `StorageSetStateAction` which allows for the State setter's function to return `typeof RESET`
- Added conditional check to detect if returned `newValue` is `typeof RESET` after calling callback. If it is, run steps to reset storage

## Other Approaches considered
- Update `SetStateAction` callback union type from `type SetStateAction<Value> = Value | ((prev: Value) => Value)` to `type SetStateAction<Value> = Value | ((prev: Value) => Value) | typeof RESET` but this was cause affect all atom state setters. It seems only `atomWithStorage` / `atomWithHash` would benefit from this type, so I decided to create a type specifically for this util atom. 
    - If possible, maybe we can derive types from `SetStateAction`, but I am relatively new to typescript and do not know of a way to do this
    - An approach I tried with the mindset of reusing `SetStateAction`: https://github.com/austinwoon/jotai/pull/1 
        - There may be no good reason for this as `SetStateAction` signature should be stable, but this ensures if we do change `SetStateAction` signature in the future, `atomWithStorage` will catch lint errors if any.

## Tests
- Added to `atomWithStorage` tests by checking if conditional reset based on previous value works and that storage key clears as expected.
- Added to `atomWithHash` test by checking if element toggles as expected, and that url hash clears as expected

### DIFF
--- a/docs/utils/atom-with-storage.mdx
+++ b/docs/utils/atom-with-storage.mdx
@@ -53,11 +53,7 @@ The suggested workaround for this issue is to only render the content dependent 
 For the case you want to delete an item from storage,
 the atom created with `atomWithStorage` accepts the `RESET` symbol on write.
 
-See the following example for RESET usage:
-
-- The input element exemplifies how you could use a previous value to conditionally trigger a reset based on previous value.
-  - In the example, `text` will be reset to `hello` if previous atom value === 'reset'
-- The button resets text to 'hello' when clicked.
+See the following example for the usage:
 
 ```jsx
 import { useAtom } from 'jotai'
@@ -70,13 +66,32 @@ const TextBox = () => {
 
   return (
     <>
-      <input
-        value={text}
-        onChange={(e) =>
-          setText((prev) => (prev === 'reset' ? RESET : e.target.value))
-        }
-      />
+      <input value={text} onChange={(e) => setText(e.target.value)} />
       <button onClick={() => setText(RESET)}>Reset (to 'hello')</button>
+    </>
+  )
+}
+```
+
+If needed, you can also do conditional resets based on previous value.
+
+This can be particularly useful if you wish to clear keys in localStorage if previous values meet a condition.
+
+Below exemplifies this usage that clears the `visible` key whenever the previous value is `true`.
+
+```jsx
+import { useAtom } from 'jotai'
+import { atomWithStorage, RESET } from 'jotai/utils'
+
+const isVisibleAtom = atomWithStorage('visible', false)
+
+const TextBox = () => {
+  const [isVisible, setIsVisible] = useAtom(isVisibleAtom)
+
+  return (
+    <>
+      { isVisible && <h1>Header is visible!</h1> }
+      <button onClick={() => setIsVisible((prev) => prev ? RESET : true))}>Toggle visible</button>
     </>
   )
 }
--- a/docs/utils/atom-with-storage.mdx
+++ b/docs/utils/atom-with-storage.mdx
@@ -55,8 +55,9 @@ the atom created with `atomWithStorage` accepts the `RESET` symbol on write.
 
 See the following example for RESET usage:
 
-- The button resets text to 'hello'
-- The second input element exemplifies how you could use a previous value to conditionally trigger a reset
+- The input element exemplifies how you could use a previous value to conditionally trigger a reset based on previous value.
+  - In the example, `text` will be reset to `hello` if previous atom value === 'reset'
+- The button resets text to 'hello' when clicked.
 
 ```jsx
 import { useAtom } from 'jotai'
@@ -69,19 +70,13 @@ const TextBox = () => {
 
   return (
     <>
-      <input value={text} onChange={(e) => setText(e.target.value)} />
-      <button onClick={() => setText(RESET)}>Reset (to 'hello')</button>
-
-      {/* 
-        trigger a reset based on the previous value. If input text is reset, we
-        will reset value to 'hello' 
-      */}
       <input
         value={text}
         onChange={(e) =>
           setText((prev) => (prev === 'reset' ? RESET : e.target.value))
         }
       />
+      <button onClick={() => setText(RESET)}>Reset (to 'hello')</button>
     </>
   )
 }
--- a/docs/utils/atom-with-storage.mdx
+++ b/docs/utils/atom-with-storage.mdx
@@ -68,6 +68,12 @@ const TextBox = () => {
     <>
       <input value={text} onChange={(e) => setText(e.target.value)} />
       <button onClick={() => setText(RESET)}>Reset (to 'hello')</button>
+      <button
+        onClick={(e) =>
+          setText((prev) => (prev === 'reset' ? RESET : e.target.value))
+        }>
+        Reset value if input text === reset
+      </button>
     </>
   )
 }
--- a/docs/utils/atom-with-storage.mdx
+++ b/docs/utils/atom-with-storage.mdx
@@ -53,7 +53,10 @@ The suggested workaround for this issue is to only render the content dependent 
 For the case you want to delete an item from storage,
 the atom created with `atomWithStorage` accepts the `RESET` symbol on write.
 
-See the following example for the usage:
+See the following example for RESET usage:
+
+- The button resets text to 'hello'
+- The second input element exemplifies how you could use a previous value to conditionally trigger a reset
 
 ```jsx
 import { useAtom } from 'jotai'
@@ -68,6 +71,11 @@ const TextBox = () => {
     <>
       <input value={text} onChange={(e) => setText(e.target.value)} />
       <button onClick={() => setText(RESET)}>Reset (to 'hello')</button>
+
+      {/* 
+        trigger a reset based on the previous value. If input text is reset, we
+        will reset value to 'hello' 
+      */}
       <input
         value={text}
         onChange={(e) =>
--- a/docs/utils/atom-with-storage.mdx
+++ b/docs/utils/atom-with-storage.mdx
@@ -68,8 +68,8 @@ const TextBox = () => {
     <>
       <input value={text} onChange={(e) => setText(e.target.value)} />
       <button onClick={() => setText(RESET)}>Reset (to 'hello')</button>
-      <button
-        onClick={(e) =>
+      <input value={text}
+        onChange={(e) =>
           setText((prev) => (prev === 'reset' ? RESET : e.target.value))
         }>
         Reset value if input text === reset
--- a/docs/utils/atom-with-storage.mdx
+++ b/docs/utils/atom-with-storage.mdx
@@ -68,12 +68,12 @@ const TextBox = () => {
     <>
       <input value={text} onChange={(e) => setText(e.target.value)} />
       <button onClick={() => setText(RESET)}>Reset (to 'hello')</button>
-      <input value={text}
+      <input
+        value={text}
         onChange={(e) =>
           setText((prev) => (prev === 'reset' ? RESET : e.target.value))
-        }>
-        Reset value if input text === reset
-      </button>
+        }
+      />
     </>
   )
 }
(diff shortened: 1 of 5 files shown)


src/utils/atomWithStorage.ts
```diff
@@ -1,9 +1,13 @@
 import { atom } from 'jotai'
-import type { SetStateAction, WritableAtom } from 'jotai'
+import type { WritableAtom } from 'jotai'
 import { RESET } from './constants'
 
 type Unsubscribe = () => void
 
+type SetStateActionFunc<Value> = (prev: Value) => Value | typeof RESET
+
+type StorageSetStateAction<Value> = Value | SetStateActionFunc<Value>
```
dai-shi, line 9: How about this?
```suggestion
type SetStateActionWithReset<Value> = Value | typeof RESET | ((prev: Value) => Value | typeof RESET)
```
Just a preference.

src/utils/atomWithStorage.ts
```diff
@@ -148,15 +156,22 @@ export function atomWithStorage<Value>(
 
   const anAtom = atom(
     (get) => get(baseAtom),
-    (get, set, update: SetStateAction<Value> | typeof RESET) => {
+    (get, set, update: StorageSetStateAction<Value> | typeof RESET) => {
       if (update === RESET) {
         set(baseAtom, initialValue)
         return storage.removeItem(key)
       }
```
dai-shi, line 163: We no longer need this.
```suggestion
```

src/utils/atomWithStorage.ts
```diff
@@ -148,15 +156,22 @@ export function atomWithStorage<Value>(
...
+
       const newValue =
         typeof update === 'function'
-          ? (update as (prev: Value) => Value)(get(baseAtom))
+          ? (update as SetStateActionFunc<Value>)(get(baseAtom))
```
dai-shi, line 167: With the above change, this would become:
```suggestion
          ? (update as (prev: Value) => Value | typeof RESET)(get(baseAtom))
```

src/utils/atomWithStorage.ts
dai-shi: Yeah, it should be better as we can remove most of `typeof RESET` below.

dai-shi (commented):
Thanks for working on this! Let's do some tweaks.

src/utils/atomWithStorage.ts
austinwoon: Initially did the function type to help with func typecasting, but since this can't be reused anywhere I will stick with the repo conventions!

1aa8917

src/utils/atomWithStorage.ts
austinwoon: 1aa8917

austinwoon:
> Thanks for working on this! Let's do some tweaks.
> 
> And then can you work on `atomWithReset` too after the 2nd review?

Done here! https://github.com/pmndrs/jotai/pull/1346/commits/9030ed5b8f2105e96945b4402abd668c1ab8cae2

src/utils/atomWithStorage.ts
```diff
@@ -84,28 +89,32 @@ export function atomWithStorage<Value>(
   key: string,
   initialValue: Value,
   storage: AsyncStorage<Value> & { delayInit: true }
-): WritableAtom<Value, SetStateAction<Value> | typeof RESET, Promise<void>>
+): WritableAtom<
+  Value,
+  SetStateActionWithReset<Value> | typeof RESET,
```
dai-shi, line 94: We don't need `typeof REST`, right?
```suggestion
  SetStateActionWithReset<Value>,
```

src/utils/atomWithStorage.ts
```diff
@@ -84,28 +89,32 @@ export function atomWithStorage<Value>(
...
+>
 
 export function atomWithStorage<Value>(
   key: string,
   initialValue: Value,
   storage: AsyncStorage<Value>
 ): WritableAtom<
   Promise<Value>,
-  SetStateAction<Value> | typeof RESET,
+  SetStateActionWithReset<Value> | typeof RESET,
```
dai-shi, line 104:
```suggestion
  SetStateActionWithReset<Value>,
```

src/utils/atomWithStorage.ts
```diff
@@ -84,28 +89,32 @@ export function atomWithStorage<Value>(
...
   Promise<void>
 >
 
 export function atomWithStorage<Value>(
   key: string,
   initialValue: Value,
   storage: SyncStorage<Value>
-): WritableAtom<Value, SetStateAction<Value> | typeof RESET>
+): WritableAtom<Value, SetStateActionWithReset<Value> | typeof RESET>
```
dai-shi, line 112:
```suggestion
): WritableAtom<Value, SetStateActionWithReset<Value>>
```

src/utils/atomWithStorage.ts
```diff
@@ -84,28 +89,32 @@ export function atomWithStorage<Value>(
...
 
 export function atomWithStorage<Value>(
   key: string,
   initialValue: Value
-): WritableAtom<Value, SetStateAction<Value> | typeof RESET>
+): WritableAtom<Value, SetStateActionWithReset<Value> | typeof RESET>
```
dai-shi, line 117:
```suggestion
): WritableAtom<Value, SetStateActionWithReset<Value>>
```

src/utils/atomWithStorage.ts
```diff
@@ -177,7 +188,7 @@ export function atomWithHash<Value>(
     replaceState?: boolean
     subscribe?: (callback: () => void) => () => void
   }
-): WritableAtom<Value, SetStateAction<Value> | typeof RESET> {
+): WritableAtom<Value, SetStateActionWithReset<Value> | typeof RESET> {
```
dai-shi, line 191:
```suggestion
): WritableAtom<Value, SetStateActionWithReset<Value>> {
```

dai-shi (commented):
Just reviewed `atomWithStorage.ts`.

src/utils/atomWithStorage.ts
austinwoon: ah, i just pushed out a commit to clean this up! https://github.com/pmndrs/jotai/pull/1346/commits/3ab023acb932910476889065816ba442cce7f6d2

docs/utils/atom-with-storage.mdx
```diff
@@ -68,6 +68,12 @@ const TextBox = () => {
     <>
       <input value={text} onChange={(e) => setText(e.target.value)} />
       <button onClick={() => setText(RESET)}>Reset (to 'hello')</button>
+      <button
+        onClick={(e) =>
+          setText((prev) => (prev === 'reset' ? RESET : e.target.value))
```
dai-shi, line 73: Hmm? I don't think this example is very convincing.

src/utils/atomWithReset.ts
```diff
@@ -1,20 +1,26 @@
...
-      set(
-        anAtom,
-        typeof update === 'function'
-          ? (update as (prev: Value) => Value)(get(anAtom))
-          : update
-      )
+      return
     }
+
+    set(anAtom, nextValue)
```
dai-shi, line 23: Can we do this?
```ts
set(anAtom, nextValue === RESET ? initialValue : nextValue)
```

dai-shi (commented):
thanks. here are some new comments.

docs/utils/atom-with-storage.mdx
austinwoon: Should I add an explanation? or a separate example?

src/utils/atomWithReset.ts
austinwoon: https://github.com/pmndrs/jotai/pull/1346/commits/3ab023acb932910476889065816ba442cce7f6d2

docs/utils/atom-with-storage.mdx
austinwoon: Ah, I was using the wrong. HTML element type. Adjusted here, let me know if this is an appropriate example

[ca565fb](https://github.com/pmndrs/jotai/pull/1346/commits/ca565fb01414b4e030c82bb9f04f37b940cbe4c8)

dai-shi: I mean, when does `prev` become `'reset'`? My interpretation is you type "reset" and type another key, it will be reset. It's weird or not realistic usage.
If you want to type "reset" to make it reset, this should work:
```js
setText(e.target.value === 'prev' ? RESET : e.target.value)
```
It's still not realistic, though.

src/utils/atomWithStorage.ts
```diff
@@ -148,15 +149,17 @@ export function atomWithStorage<Value>(
 
   const anAtom = atom(
     (get) => get(baseAtom),
-    (get, set, update: SetStateAction<Value> | typeof RESET) => {
-      if (update === RESET) {
-        set(baseAtom, initialValue)
-        return storage.removeItem(key)
-      }
+    (get, set, update: SetStateActionWithReset<Value>) => {
       const newValue =
```
dai-shi, line 153: It's not your fault, but can we rename it to `nextValue` for consistency?

dai-shi (commented):
Other than that, it looks good!

docs/utils/atom-with-storage.mdx
austinwoon: Agreed. Its a weird example, I've added a new example that might be more appropriate here: https://github.com/pmndrs/jotai/pull/1346/commits/3059bcfc8fcba3469c7e3d6f614d4f825ab38f27

docs/utils/atom-with-storage.mdx
```diff
@@ -73,6 +73,30 @@ const TextBox = () => {
...
+
+const isVisibleAtom = atomWithStorage('visible', false)
+
+const TextBox = () => {
+  const [isVisible, setIsVisible] = useAtom(isVisibleAtom)
+
+  return (
+    <>
+      { isVisible && <h1>Header is visible!</h1> }
+      <button onClick={() => setIsVisible((prev) => prev ? RESET : true))}>Toggle visible</button>
```
dai-shi, line 94: Cool. Yeah, this is something impossible without this PR.

dai-shi (approved):
LGTM

I will merge this sometime next week.

docs/utils/atom-with-storage.mdx
austinwoon: nice! thanks for the quick reviews

PR #1350, opened 2022-08-16
fix(urql): remove ad-hoc error handling

lukechu10:
Throwing an error makes it much harder to handle the error in application code. Returning the error instead in the atom value is much more ergonomic.

This also makes the behavior more consistent as some other places simply ignore the error.

### DIFF
--- a/src/urql/atomWithQuery.ts
+++ b/src/urql/atomWithQuery.ts
@@ -97,9 +97,9 @@ export function atomWithQuery<Data, Variables extends object>(
         return
       }
       // TODO error handling
-      if (!isOperationResultWithData(result)) {
-        throw new Error('result does not have data')
-      }
+      // if (!isOperationResultWithData(result)) {
+      //   throw new Error('result does not have data')
+      // }
       if (resolve) {
         resolve(result)
         resolve = null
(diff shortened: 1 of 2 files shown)


dai-shi (commented):
This is not a error handling but a type guard.
Do you actually see this error? Please provide a minimal reproduction if possible. I might misunderstand something.

Slightly related to the topic, I will be working on a proper error handling (marked as `TODO`) some time soon. (as `jotai/query` is already done, we just need to follow it.)

dai-shi:
Closing this. Please open a new issue with a reproduction.

lukechu10:
> Do you actually see this error? Please provide a minimal reproduction if possible. I might misunderstand something.

Yes, I do see the error. When an error occurs with `atomWithQuery`, the suspense never resolves, presumably because of the thrown exception. Commenting out this code makes the suspense resolve with the error. I'll try to create a minimal reproduction and update you as soon as possible when I have the time.

dai-shi:
Thanks. Here are steps we follow/prefer:
1. open a new issue with a minimal reproduction with codesandbox
2. create a failing test spec
3. fix it

I can work on 2 and 3, but your contribution would also be welcome for 2 and/or 3, in addition to 1.

lukechu10:
I created issues #1358 and #1359 with minimal reproductions.

PR #1360, opened 2022-08-21
refactor(devtools): use useAtomsSnapshot in useAtomsDevtools

### LINKED ISSUE #1356: devtools: using useAtomsSnapshot in useAtomsDevtools
From @dai-shi:

> As you notice, we have common code in useAtomsDevtools and useAtomsSnapshot. We may want to try refactoring to use useAtomsSnapshot inside useAtomsDevtools. What do you think? Let's tackle it later.

I guess this would be possible, although we haven't investigated it properly!

Aslemammad:
Resolves #1356 

Since we have so common code between useAtomsDevtools and useAtomsSnapshot, we try to use the latter in the former. 

Note: This an exception, we don't usually use utils in themselves !

### DIFF
--- a/src/devtools/useAtomsDevtools.ts
+++ b/src/devtools/useAtomsDevtools.ts
@@ -1,13 +1,7 @@
-import { useCallback, useContext, useEffect, useRef, useState } from 'react'
+import { useContext, useEffect, useRef } from 'react'
 import { SECRET_INTERNAL_getScopeContext as getScopeContext } from 'jotai'
 import type { Atom, Scope } from '../core/atom'
-import {
-  DEV_GET_ATOM_STATE,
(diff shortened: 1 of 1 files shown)


src/devtools/useAtomsDevtools.ts
```diff
@@ -97,78 +75,12 @@ export function useAtomsDevtools(
     throw new Error('useAtomsDevtools can only be used in dev mode.')
   }
```
dai-shi, line 76: Let's remove this too.

dai-shi (approved):
wow, looks so clean!

dai-shi:
> Note: This an exception, we don't usually use utils in themselves !

Let's leave this in the source code comment.

Issue #949, opened 2022-01-11
[Next.js @ jotai 1.5.1] `atomWithStorage` : localStorage is not defined 


comingAlive:
```
import { useAtom } from "jotai";
import { atomWithStorage } from "jotai/utils";

const counterAtom = atomWithStorage("counter", 0);

export default function HomePage() {
  const [c] = useAtom(counterAtom);
  return <div>{c}</div>;
}
```

<=1.5.0 working as expected.

dai-shi:
Oops! Will make a fix soon. Thanks for reporting.

dai-shi:
Can you try #951 to see if it really fixes the issue?
https://ci.codesandbox.io/status/pmndrs/jotai/pr/951
☝️ See "Local Install Instructions"

comingAlive:
It does. Amazing library btw :) Thank you for your work.

dai-shi:
https://github.com/pmndrs/jotai/releases/tag/v1.5.2 is released with the fix.

PR #1363, opened 2022-08-23
fix(utils): avoid catching storage error in atomWithStorage

### LINKED ISSUE #1347: (atomWithStorage): Exceptions from Storage providers should not be silenced
At the moment if anything throws inside StorageSync the error will be swallowed and not even logged in console.
This behavior is very unfriendly.

```ts
const storage = createJSONStorage(() => localStorage);
const prefsAtom = atomWithStorage(
  "prefs",
  { x: 1 },
  {
    ...storage,
    getItem: (key) => {
      console.log("loading", key);
      throw new Error("Cannot load prefs");
      return storage.getItem(key);
    }
  }
);
```

https://codesandbox.io/s/jotai-useatomcallback-async-error-forked-zr737n?file=/src/App.tsx:187-470

I had a bug where zod couldn't parse data but took me a while to figure out because of suppressed exceptions.

IMHO exceptions should be at the very least logged with `console.error` if not made fatal.

dai-shi:
close #1347 

I ended up with introducing a symbol and removing big try-catch completely.
(while I originally thought about throwing a symbol or something, returning a symbol should be cleaner.)

I had to fix a test for #949.
Also, did a tweak for `.subscribe` #1004, to work with non `defaultStorage`. (There was an issue/discussion which I don't remember.)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -13,6 +13,7 @@ export { splitAtom } from './utils/splitAtom'
 export { atomWithDefault } from './utils/atomWithDefault'
 export { waitForAll } from './utils/waitForAll'
 export {
+  NO_STORAGE_VALUE,
   atomWithStorage,
   atomWithHash,
   createJSONStorage,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -13,7 +13,7 @@ export { splitAtom } from './utils/splitAtom'
 export { atomWithDefault } from './utils/atomWithDefault'
 export { waitForAll } from './utils/waitForAll'
 export {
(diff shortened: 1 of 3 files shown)


joshmkennedy:
I'm unfamiliar with symbols. What is the benefit of Symbols as opposed to a strings?

dai-shi:
A string would work too. A symbol is unique, so no conflicting risks.
I'm actually fine with a string in this case, because exporting symbol isn't very nice. But, if some other libs happen to use the same string, it would catch it unexpectedly.

PR #1375, opened 2022-08-29
fix(utils/atomWithStorage): Prevent createJSONStorage from adding subscribe method when window.addEventListener is not defined

### LINKED ISSUE #1373: React Native (Expo) atomWithStorage error: "window.addEventListener is not a function"
I'm running into an issue using the `atomWithStorage` util with React Native. Here's the relevant code I'm trying with it:

Creating the atom:
```ts
import { atomWithStorage, createJSONStorage, loadable } from "jotai/utils";
import AsyncStorage from "@react-native-async-storage/async-storage";

const storage = createJSONStorage<boolean>(() => AsyncStorage);
export const walkThroughStatusAtom = atomWithStorage(
  "@walkThroughStatus",
  false,
  storage
);
```

Using the atom:
```ts
const [walkThroughStatus, setWalkThroughStatus] = useAtom(walkThroughStatusAtom);
```

Running this, I get the following error:
```
TypeError: window.addEventListener is not a function. (In 'window.addEventListener('storage', storageEventCallback)', 'window.addEventListener' is undefined)
```

It would appear this is happening at the following line: https://github.com/pmndrs/jotai/blob/main/src/utils/atomWithStorage.ts#L86

There's a check that window is not undefined [here](https://github.com/pmndrs/jotai/blob/main/src/utils/atomWithStorage.ts#L79), but at least when running the app with Expo, window is indeed defined, so the check fails to prevent the `window.addEventListener` from being called. I'm new to React Native, so I'm not sure if window is a global variable once the app is actually built and running on a native environment. This may just be an incompatibility with Expo, but with Expo being a common tool for React Native, it might be something to look into. Perhaps the check could be changed to:

```ts
if (typeof window !== 'undefined' && typeof window.addEventListener === 'function')
```

mkossoris:
## Related Issues

Fixes #1373

## Summary

Added a condition to the if statement that already existed that, in addition to checking whether the global `window` object exists, also checks that `window.addEventListener` exists. Also added a simple test to prevent regression in the future.

I tested this in the project I first discovered the bug using yalc (since React Native doesn't support symlinks), and, other than a weird type issue where the `value` in `[value] = useAtom(atom)` shows as unknown, it doesn't throw an error now. I doubt my one-line change could have actually messed with the typings, but wanted to call it out for you to verify as the expert. 

Note: The beforeAll function in the added test looks a bit odd with the `;` in front but was added by prettier, probably to prevent the parentheses around `(window as any)` from accidentally being used to invoke a function. In order to set `addEventListener` as `undefined`, I had to cast `window` as `any`, which requires those parentheses. Alternatively, I could do something like the following if preferred:

```ts
const windowAny = window as any
windowAny.addEventListener = undefined
```

## Check List

### DIFF
--- a/src/utils/atomWithStorage.ts
+++ b/src/utils/atomWithStorage.ts
@@ -76,7 +76,7 @@ export function createJSONStorage<Value>(
       getStringStorage()?.setItem(key, JSON.stringify(newValue)),
     removeItem: (key) => getStringStorage()?.removeItem(key),
   }
-  if (typeof window !== 'undefined') {
+  if (typeof window !== 'undefined' && typeof window.addEventListener === 'function') {
     storage.subscribe = (key, callback) => {
       const storageEventCallback = (e: StorageEvent) => {
         if (e.key === key && e.newValue) {
--- a/src/utils/atomWithStorage.ts
+++ b/src/utils/atomWithStorage.ts
@@ -76,7 +76,10 @@ export function createJSONStorage<Value>(
       getStringStorage()?.setItem(key, JSON.stringify(newValue)),
     removeItem: (key) => getStringStorage()?.removeItem(key),
   }
-  if (typeof window !== 'undefined' && typeof window.addEventListener === 'function') {
+  if (
+    typeof window !== 'undefined' &&
+    typeof window.addEventListener === 'function'
+  ) {
     storage.subscribe = (key, callback) => {
       const storageEventCallback = (e: StorageEvent) => {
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
This looks great!
Don't worry about prettier putting `;` in front. It's how it works.

dai-shi:
> the `value` in `[value] = useAtom(atom)` shows as unknown, it doesn't throw an error now.

Maybe it was using old TS? Can't tell for sure without a reproduction.

PR #1399, opened 2022-09-05
fix(types): patch awaited for ts < 4.5

### LINKED ISSUE #1398: jotai 1.6.4+ does not support typescript 4.0
What happens: `useAtom` returns the value with type `any`. 
Why: `declare type Awaited<T> = T extends Promise<infer V> ? Awaited<V> : T;` this recursive declaration from [useAtom](https://unpkg.com/browse/jotai@1.8.2/core/useAtom.d.ts) is not supported in ts 4.0
What I have tested:
I have ts 4.0 project and I noticed that it stopped building with jotai 1.7 and 1.8 but works in 1.6
I also tested jotai 1.8 on different versions using typescript playground and found that it works in 4.1 and up and doesn't work (i.e. returns `any`) in 4.0 and down

TS playground link: https://www.typescriptlang.org/play?ts=4.0.5#code/JYWwDg9gTgLgBAbzgQxhEAaOBXAzgUwEE0Q4BfOAMynTgHIArCGZYOgbgChOBjCAO1zwaQ-FADCyHgAt8cALwoSACjp0AlF14ChOAsXQBGaACYFcZeoUA+RJzhw+g+AG0AblkoBdc3iIqRGDFJGXx1ezgAekiHWIA9AH4ItwA6NAAZCAB3YOQCS04yTiA

dai-shi:
close #1398

We used have our custom `Awaited`, but it doesn't work with TS < 4.0.
This PR eliminates the custom one, and uses the built-in one available since TS >= 4.5.
For older versions, we patch with a limited (non-recursive) `Awaited` type.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "main": "./index.js",
   "types": "./index.d.ts",
   "typesVersions": {
-    "<4.1": {
+    "<4.5": {
       "esm/*": [
         "ts3.4/*"
       ],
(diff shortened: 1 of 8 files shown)


dai-shi:
I'm thinking about using built-in Awaited for ts >=4.5 and patched one otherwise,
which should fix #1242 @Newbie012.

PR #1402, opened 2022-09-06
feat: atomWithQueryStatus

Aslemammad:
wip: #1068

### DIFF
--- a/src/query/atomWithQuery.ts
+++ b/src/query/atomWithQuery.ts
@@ -6,8 +6,9 @@ import type {
 } from '@tanstack/query-core'
 import { atom } from 'jotai'
 import type { PrimitiveAtom, WritableAtom } from 'jotai'
-import { queryClientAtom } from './queryClientAtom'
+import { defaultGetQueryClient, queryClientAtom } from './queryClientAtom'
 import type { CreateQueryOptions, GetQueryClient } from './types'
+import { queryObserverFamily } from './queryObserver'
 
 type AtomWithQueryAction = {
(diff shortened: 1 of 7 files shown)


src/query/atomWithQuery.ts
```diff
@@ -88,7 +89,7 @@ export function atomWithQuery<
   createQuery: CreateQueryOptions<
     AtomWithQueryOptions<TQueryFnData, TError, TData, TQueryData, TQueryKey>
   >,
-  getQueryClient: GetQueryClient = (get) => get(queryClientAtom)
+  getQueryClient: GetQueryClient = defaultGetQueryClient
```
dai-shi, line 92: We don't want this change, if this is required, then it probably means `<Provider>` doesn't work and it's breaking change.

src/query/atomWithQuery.ts
```diff
@@ -132,8 +133,10 @@ export function atomWithQuery<
       const options =
         typeof createQuery === 'function' ? createQuery(get) : createQuery
       const defaultedOptions = queryClient.defaultQueryOptions(options)
-      const observer = get(observerAtom)
-      observer.destroy()
+      const observer = get(
+        queryObserverFamily([queryDataAtom, options.queryKey, getQueryClient])
+      )
+      // observer.destroy()
```
dai-shi, line 139: This seems really unexpected. We can't get status only with the key? In that case, we need to re-ideate from the ground.

dai-shi:
closing in favor of #1435.

PR #1420, opened 2022-09-15
fix: incorrect atom invalidation logic leads to stale atom values (#1412)

### LINKED ISSUE #1412: New atom invalidation logic is incorrect
Hi, we recently upgraded Jotai in our app from 1.7.0 to 1.8.2. This was a great performance improvement for our app, but we've found a subtle bug that sometimes causes `useAtomValue` to return old/stale values. We're using React 17.0.2 without Suspense and can reproduce it on both our web app and React Native app.

In particular, we have many usages of `useAtomValue` for the same derived atom, and some instances return the correct, current value while other instances are "stuck" returning the old, stale value. 

Unfortunately I don’t have a simple reproduction, as our application is quite complex with a large tree of primitive and derived atoms, and we can only reproduce this when performing some complex operations that update many of those atoms.

However, I’ve noticed that when commenting out these lines: https://github.com/pmndrs/jotai/blob/fc96b2c72a057cb00afef0dfbe2ca30cbb7b23c4/src/core/store.ts#L866-L875 we get correct behavior, but we also get extra re-renders, and of course Jotai’s ‘no extra rerenders after commit with derived atoms (#1213)’ optimization test fails.

I’ll keep digging to see if I can identify the cause or a simple reproduction case, but wanted to raise this issue in case the cause is obvious to one of the maintainers.

Thanks!

samkline:
## Related Issues

Fixes #1412.

## Summary

This fixes the issue we were seeing with stale atom values from `useAtomValue` (#1412).

The general approach is to compare the previous and current revision to decide whether we can bail out flushing the atom to listeners. If the atom is invalidated, we first have to read it to get the current revision.

Unfortunately, I still don't have a simple reproduction, so I'm not sure how to write a new test that covers this functionality. All of Jotai's existing tests in `yarn test` pass.

This is my first time working in Jotai's codebase. @dai-shi , what do you think of this solution? Are there any edge cases I'm not thinking about? And now that we have a solution, can you think of a way to trigger the behavior in a test?

## Check List

### DIFF
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -387,6 +387,11 @@ export const createStore = (
         // newer async read is running, not updating
         return atomState
       }
+      if ('e' in atomState && Object.is(atomState.e, error)) {
+        // the same error, not updating
+        // avoids infinite loop in atomWithObservable's "writable count state with error" test
+        return atomState
+      }
       if ('p' in atomState) {
         cancelSuspensePromise(atomState.p)
       }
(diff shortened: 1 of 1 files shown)


src/core/store.ts
```diff
@@ -387,6 +387,11 @@ export const createStore = (
         // newer async read is running, not updating
         return atomState
       }
+      if ('e' in atomState && Object.is(atomState.e, error)) {
```
dai-shi, line 390: This makes sense. Nice.

src/core/store.ts
```diff
@@ -859,15 +864,15 @@ export const createStore = (
       const pending = Array.from(pendingMap)
       pendingMap.clear()
       pending.forEach(([atom, prevAtomState]) => {
-        const atomState = getAtomState(undefined, atom)
+        let atomState = getAtomState(undefined, atom)
+        if (!atomState || !atomState.y) {
+          atomState = readAtomState(undefined, atom)
```
dai-shi, line 869: This seems really bad. We should never call readAtomState in flushPending.

dai-shi (changes requested):
Thanks for working on it!

We really need to have a test to reproduce this issue. We'll be modifying core code quite drastically in the future, and without a test, we can't keep it.

dai-shi:
Closing this for now.
We are interested in the issue, and the first thing needed is to reproduce it in tests.
Let's continue discussion in #1412.

PR #1426, opened 2022-09-19
refactor(utils): atomWithObservable

### LINKED ISSUE #1404: [atomWithObservable] support error handling
Having a better understanding with

- https://github.com/pmndrs/jotai/pull/1239 
- https://github.com/pmndrs/jotai/pull/1381 

I would like to revisit the implementation of `atomWithObservable`. Hope it will be a good building block.

dai-shi:
close #1404 

error handling was already supported, so what this PR does are:
- add more tests with error handling.
- better error handling code eliminating Promise.reject().
- refactor the entire code with our recent experience with jotai/query and jotai/urql.
- options type is slightly changed to avoid confusion.

### DIFF
--- a/src/utils/atomWithObservable.ts
+++ b/src/utils/atomWithObservable.ts
@@ -1,6 +1,9 @@
 import { atom } from 'jotai'
 import type { Atom, Getter, WritableAtom } from 'jotai'
 
+type Timeout = ReturnType<typeof setTimeout>
+type AnyError = unknown
+
 declare global {
   interface SymbolConstructor {
     readonly observable: symbol
@@ -13,7 +16,7 @@ type Subscription = {
 
 type Observer<T> = {
   next: (value: T) => void
-  error: (error: unknown) => void
+  error: (error: AnyError) => void
   complete: () => void
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/atomWithObservable.ts
```diff
@@ -13,41 +16,40 @@ type Subscription = {
...
 
 type SubjectLike<T> = ObservableLike<T> & Observer<T>
 
-type InitialValueFunction<T> = () => T | undefined
-
-type AtomWithObservableOptions<TData> = {
-  initialValue?: TData | InitialValueFunction<TData>
+type Options<Data> = {
+  initialValue?: Data | (() => Data)
+  timeout?: number
```
dai-shi, line 37: This new option is still an undocumented one. It only exists to avoid memory leaks with concurrent rendering. we are not sure what's the best practice. Hmm, maybe we should add `unstable_` prefix.

src/utils/atomWithObservable.ts
```diff
@@ -56,113 +58,98 @@ export function atomWithObservable<TData>(
...
+    let timer: Timeout | undefined
+    const isNotMounted = () => !setResult
+    const start = () => {
+      if (subscription) {
+        clearTimeout(timer)
+        subscription.unsubscribe()
+      }
+      subscription = observable.subscribe(
+        (d) => listener({ d }),
+        (e) => listener({ e })
```
dai-shi, line 95: Seems like this style is deprecated in rxjs. https://rxjs.dev/deprecations/subscribe-arguments

PR #1435, opened 2022-09-20
feat(query): new implementation with peer dep

### LINKED ISSUE #309: [jotai/query] atomWithMutation
follow up #248

It would be nice if we can add the mutation support.

### LINKED ISSUE #1068: [atomWithQuery | atomWithInfiniteQuery]: return all react query state.

### LINKED ISSUE #1418: [jotai/query] atomWithQuery gets stuck in react strict mode
https://codesandbox.io/s/jotai-query-stuck-in-strict-8ffrtn?file=/src/App.tsx

If `<React.StrictMode>` is enabled then the query is stuck showing placeholder data.

Also, in bigger application I have this problem even without placeholderData - there the query is forever stuck in loading state with no observers attached. Seems to happen when there are many derived atoms. Hopefully it's the same root cause since I haven't been able to create smaller test case without placeholderData.

### LINKED ISSUE #1454: [jotai/query] atomWithQuery gets stuck when request takes too long
https://codesandbox.io/s/long-request-does-not-render-wnpber

When doing a promise inside a `atomWithQuery` that takes longer than 1 second to resolve the atom doesn't trigger a render and gets stuck on the fallback component provided to `<Suspense>`. Methods like `onSettled` doesn't trigger.

When jotai is locked to version `1.8.3` it works and renders when the promise resolves but when updating to `1.8.4` it gets stuck.

dai-shi:
This is a big refactor on tanstack query integration. the goals are:
- more maintainable code, thinner wrapper, using jotai/utils
- full, non suspense api #1068, #1424 
- better versioning https://github.com/pmndrs/jotai/discussions/1312#discussioncomment-3329805

---

- [x] compatible atomWithQuery
- [x] compatible atomWithInfiniteQuery
- [x] new docs (incl. status api and mutation api)

close #1418
close #1068
close #309
close #1454


dai-shi:
```
 FAIL  tests/query/atomWithQuery.test.tsx (8.468 s)
  ✓ query basic test (127 ms)
  ✓ query basic test with object instead of function (109 ms)
  ✕ query refetch (1255 ms)
  ✕ query loading (1254 ms)
  ✕ query loading 2 (1271 ms)
  ✕ query no-loading with keepPreviousData (257 ms)
  ✕ query with enabled (1011 ms)
  ✕ query with enabled 2 (243 ms)
  ✓ query with enabled (#500) (369 ms)
  ✓ query with initialData test (10 ms)
  ✓ query dependency test (351 ms)
  ✓ query expected QueryCache test (117 ms)
  error handling
    ✓ can catch error in error boundary (136 ms)
    ✕ can recover from error (1373 ms)
```

src/query/atomWithQuery.ts
```diff
@@ -88,119 +82,24 @@ export function atomWithQuery<
...
-        if (unsubscribe) {
-          unsubscribe()
-          // FIXME why does this fail?
-          // unsubscribe = null
-        }
-      }
-    }
-    return { resultAtom, makePending, startQuery }
+  const getOptions = (get: Getter) => ({
+    staleTime: 500,
```
dai-shi, line 86: This is to avoid failing tests with strict effect. this or `retryOnMount: false` should help. otherwise, it will fetch twice in DEV mode, which is actually fine and expected.

src/query/atomWithQuery.ts
```diff
@@ -88,119 +82,24 @@ export function atomWithQuery<
...
+    ...(typeof createQuery === 'function' ? createQuery(get) : createQuery),
   })
-
-  const queryAtom = atom(
+  const [dataAtom] = atomsWithTanstackQuery(getOptions, getQueryClient)
```
dai-shi, line 89: It's unfortunate that we can't simply return `dataAtom`. We need a small wrapper to pass our current tests.

tests/query/atomWithQuery.test.tsx
```diff
@@ -576,7 +575,7 @@ it('query dependency test', async () => {
 
   await new Promise((r) => setTimeout(r, 100))
   fireEvent.click(getByText('increment'))
-  await findByText('loading')
+  // await findByText('loading')
```
dai-shi, line 578: it no longer suspends on refetch by default.

tests/query/atomWithQuery.test.tsx
```diff
@@ -354,7 +353,7 @@ it('query with enabled', async () => {
   expect(mockFetch).toHaveBeenCalledTimes(1)
 })
 
-it('query with enabled 2', async () => {
+it.skip('query with enabled 2', async () => {
```
dai-shi, line 356: passing this test would require many hacky code. so my conclusion for now is that this is a wrong test.

dai-shi (commented):
I think this is reasonably good, even though it introduces some behavioral changes.

dai-shi:
Yay, tests are all passing.

EduardoLopes:
I'm already using jotai-tanstack-query in a app, and i noticed that when more than one componente is using the same queryAtom and we change the query key using another atom, we got a warning from react about a _bad setState()_. 

Here's the full error:

```
Warning: Cannot update a component (`UserData`) while rendering a different component (`UserData`). To locate the bad setState() call inside `UserData`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
    at UserData (https://q06r83.csb.app/src/App.tsx:45:39)
    at Suspense
    at App
```

Here's the reproduction on codesandbox: https://codesandbox.io/s/beautiful-volhard-q06r83?file=/src/App.tsx:120-140
Just click in the prev/next button, and the warning will be shown in the console

dai-shi:
Oh, this is very interesting. 
https://codesandbox.io/s/vigorous-chandrasekhar-2t9v85?file=/src/App.tsx
It doesn't happen with the current `jotai/query`. So, it's a bug in `jotai-tanstack-query` (or `atomWithObservable`).

Thanks for the reproduction.

dai-shi:
oops, the fix broke the tests. it was too naive.

dai-shi:
@EduardoLopes It should be fixed.

src/query/atomWithQuery.ts
```diff
@@ -1,16 +1,10 @@
-import { QueryClient, QueryObserver } from '@tanstack/query-core'
-import type {
-  QueryKey,
-  QueryObserverOptions,
-  QueryObserverResult,
-} from '@tanstack/query-core'
+import type { QueryKey, QueryObserverOptions } from '@tanstack/query-core'
+import { atomsWithTanstackQuery } from 'jotai-tanstack-query'
```
Aslemammad, line 2: Can't the naming in jotai tanstack query be like the current naming in jotai? `atomWithQuery` instead of `atomWithTanstackQuery`?

docs/integrations/query.mdx
```diff
@@ -1,37 +1,61 @@
...
 
-`atomWithQuery` creates a new atom that implements a standard [`Query`](https://tanstack.com/query/v4/docs/guides/queries) from TanStack Query. This function combines both Jotai `atom` features and `useQuery` features in a single atom. This atom supports all features you'd expect to find when using the standard `@tanstack/react-query` package.
+- `atomsWithTanstackQuery` for [QueryObserver](https://tanstack.com/query/v4/docs/reference/QueryObserver)
+- `atomsWithTanstackInfiniteQuery` for [InfiniteQueryObserver](https://tanstack.com/query/v4/docs/reference/InfiniteQueryObserver)
+- `atomsWithTanstackMutation` for MutationObserver
+
+All three functions follow the same signature.
+
+```ts
+const [dataAtom, statusAtom] = atomsWithTanstackSomething(
```
Aslemammad, line 32: That's the ideal pattern I've ever wanted.

src/query/atomWithQuery.ts
dai-shi: Hm, I generally prefer shorter names. So, I'm not super happy with the current long name either.
At the same time, I think having a unique name is good for searchability.
`atoms` instead of `atom` is because we return two atoms.
Any ideas?

EduardoLopes:
I found another bug. The `fetchStatus`, `isInitialLoading` and others status never change. 

I did a reproduction, comparing to tanstack/query return: https://codesandbox.io/s/winter-water-kdysdj?file=/src/App.tsx

dai-shi:
Good catch, @EduardoLopes !
I will look into it. https://github.com/jotai-labs/jotai-tanstack-query/issues/9

src/query/atomWithQuery.ts
dai-shi: Yeah, we can do that.
You mean we do naming like this for all others `jotai-urql`, `jotai-apollo`, `jotai-relay`, right?

Let's have @EduardoLopes a comment.

nix6839:
Can we make this repository a monorepo later and add 'jotai-tanstack-query' to this repository? For more promotion, contribution and maintenance. It's just suggestion.

dai-shi:
@nix6839 Thanks for your comment.
We considered it as an option, but decided to go with separate repositories. That would make this repo too big if we accept more integrations. And, I also want to give different permissions for libs.
Feel free to drop some ideas to mitigate the promotion issue with multiple repos.

src/query/atomWithInfiniteQuery.ts
```diff
@@ -1,14 +1,13 @@
...
 import type {
   InfiniteData,
   InfiniteQueryObserverOptions,
   QueryKey,
-  QueryObserverResult,
   RefetchOptions,
   RefetchQueryFilters,
 } from '@tanstack/query-core'
+import { atomsWithInfiniteQuery } from 'jotai-tanstack-query'
 import { atom } from 'jotai'
```
wobsoriano, line 9: Hi @dai-shi! Are there plans of importing `atom` from `valtio/vanilla` instead? Thanks!

dai-shi: `jotai/query` will be deprecated soon.
So, you want to use `jotai-tanstack-query` and, in that case, we have a PR open.

https://github.com/jotai-labs/jotai-tanstack-query/pull/18

There, find codesandbox-ci comment and click `here`. You will then find "Local Install Instructions".

PR #1440, opened 2022-09-23
chore: fix the description in Overriding default atom values

masashi-sutou:
## Related Issues

- none

## Summary

- fix the description in [Overriding default atom values](docs/guides/composing-atoms.mdx)
- By the way, the code below is simpler than the example in this document. Why do we need to check the type of `typeof newValue === 'function'` ?

```ts
const rawNumberAtom = atom(10.1) // can be exported
const roundNumberAtom = atom((get) => Math.round(get(rawNumberAtom)))
const overwrittenAtom = atom<number | null>(null)
export const numberAtom = atom(
  (get) => get(overwrittenAtom) ?? get(roundNumberAtom),
  (_, set, newValue: number | null) => set(overwrittenAtom, newValue)
)
```

## Check List

### DIFF
--- a/docs/guides/composing-atoms.mdx
+++ b/docs/guides/composing-atoms.mdx
@@ -61,7 +61,7 @@ export const numberAtom = atom(
 ```
 
 The final `numberAtom` just works like a normal primitive atom like `atom(10)`.
-If you set a number value, it will override the `roundNumberAtom` value,
+If you set a number value, it will override the `overwrittenAtom` value,
 and if you set `null`, it will be the `roundNumberAtom` value.
 
 The reusable implementation is available as `atomWithDefault`


dai-shi (approved):
Thanks!

dai-shi:
> * By the way, the code below is simpler than the example in this document. Why do we need to check the type of `typeof newValue === 'function'` ?

Because we want function updates:
```js
const Component = () => {
  const [number, setNumber] = useAtom(numberAtom)
  const handleClick = () => {
    setNumber((prev) => prev + 1)
  }
  // ...
}
```

masashi-sutou:
> > * By the way, the code below is simpler than the example in this document. Why do we need to check the type of `typeof newValue === 'function'` ?
> 
> Because we want function updates:
> 
> ```js
> const Component = () => {
>   const [number, setNumber] = useAtom(numberAtom)
>   const handleClick = () => {
>     setNumber((prev) => prev + 1)
>   }
>   // ...
> }
> ```

I see. Thank you for teaching me.

PR #1446, opened 2022-09-25
chore: fix doc example in loadable api

masashi-sutou:
## Related Issues

- none

## Summary

- Fixed how variables are assigned when getting an asynchronous response using the loadable API in the sample docs.
- Since the response type is `[Awaited<Value>, never]`, we need to enclose the variable in square brackets. Or is it better to fix it with `useAtomValue`?

## Check List


dai-shi (approved):
👍

PR #1447, opened 2022-09-25
refactor(urql): infinite pending

douglaskariuki:


dai-shi (commented):
Thanks for working on this. I can merge this if tests are passing.
I'm also working on a full re-implementation, so this will be only short lived even if it's merged.

src/urql/atomWithMutation.ts
```diff
@@ -19,17 +19,33 @@ export function atomWithMutation<Data, Variables extends AnyVariables>(
...
+  const operationResultAtom = atom((get) => {
+    let resolve: ((result: Result) => void) | null = null
+    const makePending = () => 
+      new Promise<Result>((r) => {
+        resolve = r
+      })
+
+    const resultAtom = atom<Result | Promise<Result>>(makePending())
+    return { resultAtom, makePending }
+  })
```
dai-shi, line 32: On second look, this looks wrong. Do you plan to work on this refactor and fix?

douglaskariuki: I wanted to fix infinite pending in operationResultAtom by calling the pending function when there's no result

douglaskariuki: I'm also supposed to add typescript types to the values returned in operationResultAtom

dai-shi: I thought this was a refactoring, but do you have any issues? In such a case, you want to reproduce the issue with test. That would  be very valuable.

douglaskariuki: I will update test file and update this PR once all tests pass

dai-shi:
I'm still interested in reproducing the issue in tests.
For now, let me close this as the base implementation is now in `jotai-urql`.

douglaskariuki:
I'll be following up on the jotai-urql framework too and update this main one where needed or when all tests are successful

PR #1453, opened 2022-09-27
refactor(zustand): Removed unnecessary unknown extends

u-yas:
## Related Issues

Fixes #.

## Summary
The unknown part of <T extends unknown> in the atomWithStore function of the jotai/zustand package is unnecessary extends, so I deleted it. The reason is that the State type itself is marked as deprecated first. And since the StoreApi type generics does not extend anything in particular, this Generics T should be fine as is.

## Check List


dai-shi (approved):
Thanks for working on this. 
This is technically a type breaking for zustand v3 users, but it's unavoidable and we do this thing for jotai/* modules.

PR #1456, opened 2022-09-30
fix(devtools): double init call

### LINKED ISSUE #1445: useAtomsDevtools inside React.StrictMode shows twice in Redux Dev Tools
Hi there,

Firstly, it's entirely possible that I've done something wrong here, so apologies if so!

I've noticed that if I'm calling `useAtomsDevtools()` from *inside* of `React.StrictMode` then I get two different stores showing up in the Redix Dev Tools, where the one that is selected by default is not useful:

```
  const AtomsDevtools: React.FC<AtomsDevtoolsProps> = ({ children }) => {
    useAtomsDevtools('linaria');
    return children;
  };

  ReactDOM.createRoot(root).render(
    <JotaiProvider>
      <React.StrictMode>
        <AtomsDevtools>
          <AuthProvider>
            <Suspense fallback={'Loading'}>
              <UI />
            </Suspense>
          </AuthProvider>
        </AtomsDevtools>
      </React.StrictMode>
    </JotaiProvider>
  );
```

image

If I move it around so that `useAtomsDevtools` is instead *outside* of the `React.StrictMode` then everything works as expected:
```
  const AtomsDevtools: React.FC<AtomsDevtoolsProps> = ({ children }) => {
    useAtomsDevtools('linaria');
    return children;
  };

  ReactDOM.createRoot(root).render(
    <JotaiProvider>
      <AtomsDevtools>
        <React.StrictMode>
          <AuthProvider>
            <Suspense fallback={'Loading'}>
              <UI />
            </Suspense>
          </AuthProvider>
        </React.StrictMode>
      </AtomsDevtools>
    </JotaiProvider>
  );
```
image

I *assume* this is because `React.StrictMode` does the double-render trick to catch certain issues, but it seems to have a negative effect in this case.

Cheers

Aslemammad:
## Related Issues

Fixes #1445

## Summary
This tries to disconnect devtools on unmount which fixes the issue.

## Check List

### DIFF
--- a/src/devtools/useAtomsDevtools.ts
+++ b/src/devtools/useAtomsDevtools.ts
@@ -134,7 +134,11 @@ export function useAtomsDevtools(
 
     devtools.current = connection
     devtools.current.shouldInit = true
-    return devtoolsUnsubscribe
+    return () => {
+      (extension as any).disconnect()
+      devtoolsUnsubscribe?.()
+
+    }
   }, [extension, goToSnapshot, name])
 
   useEffect(() => {
--- a/src/devtools/useAtomsDevtools.ts
+++ b/src/devtools/useAtomsDevtools.ts
@@ -135,9 +135,8 @@ export function useAtomsDevtools(
     devtools.current = connection
     devtools.current.shouldInit = true
     return () => {
-      (extension as any).disconnect()
+      ;(extension as any).disconnect()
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Looks good! Thanks for the fix!

PR #1483, opened 2022-10-18
Fix docs mentioning incorrect key of data

paulxuca:
I think its meant to say `films` instead of `film` (Which doesn't exist as a key).

## Related Issues

Fixes #.

## Check List


dai-shi (approved):
Thanks!

PR #1489, opened 2022-10-21
fix(babel): Fix custom atom names for identifiers

Thisen:
## Summary
Small fix to make sure custom atom names works for identifiers.


dai-shi (approved):
LGTM

tests should help this kind of situation, but i'm not sure how easy it is to write such tests.

dai-shi:
Can anyone confirm if this works as expected?

jrsnyder:
I tested this (for plugin-debug-label) and it is working 👍

PR #1499, opened 2022-10-27
fix(docs): Fix typos and inconsistent headings in TanStack Query page

feute:
I think this is self-explanatory. 

The first hook was introduced as `atomsWithQuery`, but then the other two are prefixed with `atomsWithTansack` (which is a typo too because it should be TanStack).

### DIFF
--- a/docs/integrations/query.mdx
+++ b/docs/integrations/query.mdx
@@ -66,7 +66,7 @@ const UserData = () => {
 }
 ```
 
-## `atomsWithTansackInfiniteQuery` usage
+## `atomsWithInfiniteQuery` usage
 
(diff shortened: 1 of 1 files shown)


docs/integrations/query.mdx
```diff
@@ -179,7 +179,7 @@ const queryClient = new QueryClient()
 export const App = () => {
   return (
     <QueryClientProvider client={queryClient}>
-      {/* This Provider initalization step is needed so that we reference the same
+      {/* This Provider initialisation step is needed so that we reference the same
```
dai-shi, line 182: Can it be "initialization"?

dai-shi (approved):
Good catch. Thanks.

docs/integrations/query.mdx
feute: @dai-shi I thought so, but I decided to change it to be consistent with the rest of the page, where it's written as "initialisation".

dai-shi: Oh, I see. Okay, let's leave it for this PR. cc @sandren

PR #1506, opened 2022-10-28
fix(utils): modify atomWithHash to keep pathname when the replaceState option is set to true

### LINKED ISSUE #1110: [atomWithHash] replaceState should probably keep current path?
Setting `options.replaceState = true` causes the atom to overwrite current path with `#{jotai_hash}`. I would think what it should instead write `previous/path#{jotai_hash}`. Current implementation does not work well in SPAs that use `react-router` or similar libraries.

If this is the case I could write a PR to solve this.

jakkku:
## Related Issues

Fixes #1110 .

## Summary

If the replaceState option is set to true, the pathname is overwritten with a hash when returning to the page using `history.back`.

An error occurred in the codesandbox environment, but no error occurred in the test code environment.
But I added a test case.

This is my first time working in Jotai's codebase. Please let me know if there are any additional considerations

## Check List


tests/utils/atomWithStorage.test.tsx
```diff
@@ -470,4 +471,56 @@ describe('atomWithHash', () => {
     await findByText('visible')
     expect(window.location.hash).toEqual('')
   })
+
```
dai-shi, line 474:
```suggestion
})

describe('atomWithHash with react-router-dom', () => {
```

dai-shi (approved):
LGTM

dai-shi:
Works good: https://codesandbox.io/s/react-typescript-forked-kub7ji

Please check CI errors.

jakkku:
@dai-shi  When I check the CI errors, errors occur in `yarn install --frozen-lockfile`.
I added react-router-dom library to write test code, this seems to be the cause.

Would it be a good idea to modify the test code in a way that doesn't use the react-router-dom library?

dai-shi:
I think you just need to run `yarn` once.

dai-shi:
> Would it be a good idea to modify the test code in a way that doesn't use the react-router-dom library?

If it's reasonably possible, it'd be better anyways.

dai-shi:
How hard is it to write the test without react-router-dom?

jakkku:
> How hard is it to write the test without react-router-dom?

I think it is possible, so I will try

tests/utils/atomWithStorage.test.tsx
```diff
@@ -471,3 +471,47 @@ describe('atomWithHash', () => {
     expect(window.location.hash).toEqual('')
   })
 })
+
+describe('atomWithHash with page move', () => {
```
dai-shi, line 475:
```suggestion

```

dai-shi (approved):
Nice. Now, I think we can have it in the same "describe".

PR #1480, opened 2022-10-17
feat(babel): Support custom atom names

Thisen:
## Related Issues

Fixes #1414.

## Summary

This PR adds support for custom atom names for the Babel plugins.

PR #1507, opened 2022-10-29
test: custom atom names case

ahungrynoob:
## Related Issues

Fixes https://github.com/pmndrs/jotai/pull/1480

## Summary
The test case about custom atom names should test the node's initializer expression.

## Check List


Thisen (approved):
Thanks for fixing my quick hacking 👍🏼

PR #1513, opened 2022-10-30
[WIP] refactor(core): use thenable convention

dai-shi:
I was hoping to simplify core with mutable atom state. But, some tests are still failing...
(for the moment, I just gave up. so, this won't go.)


dai-shi:
I don't think I will continue on this road. #1515 should be the future.

PR #1516, opened 2022-10-31
docs: add info about yarn version

jakkku:
## Related Issues

not issues

## Summary

I added information about yarn version.
This may not be common, but I didn't know why ci was failing
So how about adding some info about yarn version?

## Check List


dai-shi (approved):
LGTM

PR #1515, opened 2022-10-31
feat(vanilla/react): New Jotai v2 API

dai-shi:
See the RFC: #1514 

- [x] initial implementation
- [x] convert tests/types.test.tsx
- [x] convert tests/basic.test.tsx
- [x] convert tests/async.test.tsx
- [x] convert tests/transition.test.tsx
- [x] convert tests/useAtomValue.test.tsx
- [x] convert tests/useSetAtom.test.tsx
- [x] convert tests/provider.test.tsx
- [x] convert tests/dependency.test.tsx
- [x] convert tests/items.test.tsx
- [x] convert tests/error.test.tsx
- [x] convert tests/onmount.test.tsx
- [x] convert tests/optimization.test.tsx
- [x] convert devtools
- [x] convert utils
- [x] migration guide `docs/guides/migrate-to-v2-api`
- [x] test with jotai-labs/jotai-location https://github.com/jotai-labs/jotai-location/pull/1
- [x] test with jotai-labs/jotai-form https://github.com/jotai-labs/jotai-form/pull/12
- [x] test with jotai-labs/jotai-tanstack-query https://github.com/jotai-labs/jotai-tanstack-query/pull/18
- [x] test with jotai-labs/jotai-urql https://github.com/jotai-labs/jotai-urql/pull/5
- [x] test with jotai-labs/jotai-relay https://github.com/jotai-labs/jotai-relay/pull/3
- [x] test wtih jotai-labs/jotai-optics https://github.com/jotai-labs/jotai-optics/pull/4
- [x] test with jotai-labs/jotai-immer https://github.com/jotai-labs/jotai-immer/pull/1
- [x] test with jotai-labs/jotai-trpc https://github.com/jotai-labs/jotai-trpc/pull/8
- [x] test with jotai-labs/jotai-xstate https://github.com/jotai-labs/jotai-xstate/pull/1
- [x] test with jotai-labs/jotai-redux https://github.com/jotai-labs/jotai-redux/pull/1
- [x] test with jotai-labs/jotai-zustand https://github.com/jotai-labs/jotai-zustand/pull/2
- [x] test with jotai-labs/jotai-valtio https://github.com/jotai-labs/jotai-valtio/pull/1
- [x] test with jotai-labs/jotai-jsx https://github.com/jotai-labs/jotai-jsx/pull/5
- [x] test with jotai-labs/jotai-game https://github.com/jotai-labs/jotai-game/pull/1
[...]

PR #1522, opened 2022-11-01
chore(tests): move some tests into core and improve

dai-shi:
...with removing setTimeout as much as possible like done in #1515.


tests/urql/atomWithQuery.test.tsx
```diff
@@ -26,7 +26,7 @@ const generateClient = (id: string | number, error?: () => boolean) =>
         fromValue(
           error?.() ? { error: new Error('fetch error') } : { data: { id } }
         ),
-        delay(100)
+        delay(100) // FIXME we want to use fake timer
```
dai-shi, line 29: @kitten Can you quickly guide how we can control this flow in tests?

kitten: Sorry that I didn't see this in time!

I'm not sure why it wouldn't use the fake timers as long as the global is modified in time 🤔 https://github.com/0no-co/wonka/blob/cc88647f65d0e24d84d2715b7af4eac4dc9df01f/src/operators.ts#L800-L817 That's at least true for the new major of Wonka.

dai-shi: Yeah, we could have used fake timers. (I did in other tests.)

From the URQL user perspective, I don't want to reach wonka. I just need a fake/mock client, in which case, fake timers are more appropriate. If you have a suggestion, it would be more than welcomed.

PR #1530, opened 2022-11-02
docs(query): add docs for react-query-devtools

### LINKED ISSUE #1105: [atomWithQuery] Queries are not displayed inside react-query devtools
Currently when using `atomWithQuery` the react-query devtools does not notice anything from the Query.

I believe this should be looked into, since the react-query devtools are really useful.

jakkku:
## Related Issues

Fixes #1105 

## Summary

Docs for `react-query-devtools`

## Check List

### DIFF
--- a/docs/integrations/query.mdx
+++ b/docs/integrations/query.mdx
@@ -223,9 +223,7 @@ With `dataAtom`,
 Fetch error will be thrown and can be caught with ErrorBoundary.
 Refetching may recover from a temporary error.
 
-FIXME: Update example
-
-See [a working example](https://codesandbox.io/s/04mepx) to learn more.
+See [a working example](https://codesandbox.io/s/joer59) to learn more.
 
 ## Devtools
 
--- a/docs/integrations/query.mdx
+++ b/docs/integrations/query.mdx
@@ -240,3 +240,66 @@ FIXME: Update demo
 FIXME: Update demo
 
 <CodeSandbox id="u4sli" />
+
+## Devtools
+
+In order to use the Devtools, you need to install it additionally.
+
+```bash
+$ npm i @tanstack/react-query-devtools
+# or
+$ pnpm add @tanstack/react-query-devtools
+# or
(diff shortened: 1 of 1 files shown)


docs/integrations/query.mdx
```diff
@@ -240,3 +240,66 @@ FIXME: Update demo
...
+
+export const App = () => {
+  const queryClient = useAtomValue(queryClientAtom)
+
+  return (
+    <QueryClientProvider client={queryClient}>
+      <ReactQueryDevtools />
+      <Provider>
+        <App />
+      </Provider>
```
dai-shi, line 271:
```suggestion
      <App />
```
I think omitting Provider here makes things more consistent.
But, I'm actually okay to drop the first example and only leave the second one.

dai-shi (commented):
Thanks for working on this!

Feel free to put a codesandbox demo link if you have one.

jakkku:
I leave only the second example and add devtools to hackernews example

dai-shi:
> Would it be better to leave the Hackernews without devtools as it is and make another example separately?

Hm, yeah, I would separate them, and use simpler example for devtools.

dai-shi:
I'd appreciate if you could fix those two existing demos too.

docs/integrations/query.mdx
```diff
@@ -227,16 +227,58 @@ FIXME: Update example
 
 See [a working example](https://codesandbox.io/s/04mepx) to learn more.
```
dai-shi, line 228: We have one more `FIXME` here. If you can help fixing it,  we'd be more than happy.
Otherwise, I can just merge it as is.

dai-shi (approved):

docs/integrations/query.mdx
jakkku: sure, i fixed

dai-shi: thanks so much!!!

PR #1535, opened 2022-11-03
fix(core): Override cancelled promises to avoid infinite loops

### LINKED ISSUE #1481: bug: promise handling can result in infinite loops
Hi! Thanks for all the great work on Jotai. 

I noticed some pretty severe breaking behavior in using async values in atoms. In particular, `store.ts` has two usages of the same pattern that result in an infinite loop:

In `writeGetter`: https://github.com/pmndrs/jotai/blob/0e610f920d65f18830e5769789ca9682add25c72/src/core/store.ts#L659-L666

In `asyncGet`: https://github.com/pmndrs/jotai/blob/0e610f920d65f18830e5769789ca9682add25c72/src/core/store.ts#L986-L987

In each of these, what happens is that we call the parent function, check if `atomState.p` exists, and then calls `p.then(() => writeGetter(atom))` again to invoke the parent function. The next time around, `atomState.p` still exists, so we end up in an infinite loop.

I'm not sure whether this is an oversight, or whether there is supposed to be code responsible for clearing out `atomState.p`. I think the repro was pretty simple, in that I just tried `get`ing an atom whose value was a pending promise. As far as I can tell from the code, this would result in an infinite loop any time this happens? Let me know how I can help!

hahnbiatfigma:
## Related Issues

Fixes #1481.

## Summary

Jotai's handling of promises results in some infinite loops when trying to evaluate an atom whose value is a pending promise. Let's say we have the following setup:

```ts
const baseAtom = atom(new Promise(() => {})); // never resolves

function App() {
  const val = useAtomValue(loadable(baseAtom));
  ...
}
```

I'm new to the Jotai codebase, but based on my analysis, I think this is what is happening:
- baseAtom and the loadable atom get mounted
- a suspensePromise is set to baseAtom's atomState.p
- when we mount a base atom whose value is a promise, we set atomState.p to be the corresponding suspensePromise which can be manually cancelled
- for some reason, the `baseAtom` gets unmounted and remounted, presumably due to re-rendering but I'm not completely sure why
- during unmounting, the suspensePromise (`atomState.p`) gets cancelled, BUT the underlying originalPromise (`suspensePromise.o`) is still pending
- the atom gets remounted, but atomState.p still refers to the same suspensePromise, which is now resolved
  - when we try to load the value of `baseAtom` in `loadable`, we go through the `writeGetter` path where `options.unstable_promise = true`
  - and because `atomState.p` has already been cancelled, `atomState.p.then` evaluates immediately, which gets us into that infinite loop `writeGetter` -> `atomState.p.then` -> `writeGetter`

This PR makes the change to write the _original_ promise to `aState.p` when unmounting. This way, when the atom gets remounted, the promise in `aState.p` is replaced by new suspensePromise (not yet resolved) instead of referring to the old already-resolved suspensePromise. Then, when loadable tries to get the value of `baseAtom`, we now have a non-resolved suspensePromise, so in `writeGetter`, that `atomState.p.then` will actually wait until the original promise is resolved instead of resolving immediately. This avoids the infinite loop.

## Check List


src/core/suspensePromise.ts
```diff
@@ -63,6 +63,10 @@ export const createSuspensePromise = (
   return suspensePromise
 }
 
+export const getOriginalPromise = (suspensePromise: SuspensePromise) => {
```
dai-shi, line 66: Let's call it `getBasePromise`.

src/core/store.ts
```diff
@@ -799,6 +800,12 @@ export const createStore = (
       // cancel suspense promise (and abort base promise)
       if ('p' in atomState) {
         cancelSuspensePromise(atomState.p)
+        setAtomPromiseOrValue(
+          version,
+          atom,
+          getOriginalPromise(atomState.p) as Value,
+          new Set(atomState.d.keys())
```
dai-shi, line 807: Can we remove this line?

dai-shi (commented):
Thanks for your investigation.
I don't yet fully understand how this fix solves the issue, but the code looks okay.
Let's see if tests pass.

src/core/store.ts
hahnbiatfigma: Removed!

src/core/suspensePromise.ts
hahnbiatfigma: Renamed!

hahnbiatfigma:
> I don't yet fully understand how this fix solves the issue, but the code looks okay.

I added some additional details to the description, so hopefully that helps!

dai-shi:
> > I don't yet fully understand how this fix solves the issue, but the code looks okay.
> 
> I added some additional details to the description, so hopefully that helps!

Hm, I think I get it, and wonder if there's another solution.

Something like this:
```diff
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -416,7 +416,10 @@ export const createStore = (
   ): AtomState<Value> => {
     const atomState = getAtomState(version, atom)
     if (atomState && 'p' in atomState) {
-      if (isEqualSuspensePromise(atomState.p, suspensePromise)) {
+      if (
+        isEqualSuspensePromise(atomState.p, suspensePromise) &&
+        !isSuspensePromiseAlreadyCancelled(atomState.p)
+      ) {
         // the same promise, not updating
         if (
           !atomState.y // invalidated
```

What do you think?


---

Please merge `main` branch, as there are some refactors in tests.

hahnbiatfigma:
> Hm, I think I get it, and wonder if there's another solution.
> 
> Something like this:
> 
> ```diff
> --- a/src/core/store.ts
> +++ b/src/core/store.ts
> @@ -416,7 +416,10 @@ export const createStore = (
>    ): AtomState<Value> => {
>      const atomState = getAtomState(version, atom)
>      if (atomState && 'p' in atomState) {
> -      if (isEqualSuspensePromise(atomState.p, suspensePromise)) {
> +      if (
> +        isEqualSuspensePromise(atomState.p, suspensePromise) &&
> +        !isSuspensePromiseAlreadyCancelled(atomState.p)
> +      ) {
>          // the same promise, not updating
>          if (
>            !atomState.y // invalidated
> ```
> 
> What do you think?

I ran it through the new test case locally, and unfortunately it fails (within a span of 500ms, we try to refetch the atom 800 times). With that patch, I think we're still running into the issue where atomState.p (suspensePromise) still refers to a fulfilled/cancelled promise with an underlying original promise that is still pending.

dai-shi:
I need to tweak one more thing. 02f3176
I like it better, because I don't think the bug is related with unmounting.
Please check if it fixes your case.

hahnbiatfigma:
@dai-shi works great! Thank you so much for taking a look into this.

tests/utils/loadable.test.tsx
```diff
@@ -256,6 +256,32 @@ it('loadable of a derived async atom with error does not trigger infinite loop (
...
+  render(
+    <StrictMode>
+      <Provider>
+        <LoadableComponent asyncAtom={derivedAtom} />
+      </Provider>
+    </StrictMode>
+  )
+
+  // wait some time to see how many calls are made to get the base atom
+  await new Promise((resolve) => setTimeout(resolve, 500))
```
dai-shi, line 278: We avoid using `setTimeout` in tests. Can this be improved?

hahnbiatfigma: Pushed up a new version of the test

tests/utils/loadable.test.tsx
```diff
@@ -256,6 +256,39 @@ it('loadable of a derived async atom with error does not trigger infinite loop (
...
+
+  // depending on provider-less mode or versioned-write mode, there will be
+  // either 2 or 3 calls.
+  await findByText('Loading...')
+  expect(callsToGetBaseAtom).toBeLessThanOrEqual(3)
+
+  callsToGetBaseAtom = 0
+  resolveAsync!(5)
+  await findByText('Data: 5')
+  expect(callsToGetBaseAtom).toBe(1)
```
dai-shi, line 289: Looks like we need "less than" here too?

hahnbiatfigma: Added!

dai-shi:
@hahnbiatfigma 
With the new test, it doesn't fail if we don't use copySuspensePromise. I think it failed with the previous test. Can you confirm?

```diff
diff --git a/src/core/store.ts b/src/core/store.ts
index d0ed1382..9953ed73 100644
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -590,7 +590,7 @@ export const createStore = (
         const suspensePromise =
           isSuspensePromise(errorOrPromise) &&
           isSuspensePromiseAlreadyCancelled(errorOrPromise)
-            ? copySuspensePromise(errorOrPromise)
+            ? createSuspensePromise(errorOrPromise, errorOrPromise)
             : createSuspensePromise(errorOrPromise, errorOrPromise)
         return setAtomSuspensePromise(
           version,
```

hahnbiatfigma:
Ah, you're right, the test doesn't actually catch the failing issue. I'm not sure how to restructure the test without using setTimeout, do you have any ideas? I'm not sure what is publicly exposed that lets us hook into an atom repeatedly querying its value in a synchronous way

dai-shi:
Maybe we can add an extra Counter component and click count up button a few times (or just once). Not sure if it works, but do you get the idea?

hahnbiatfigma:
Can you elaborate? Is the idea that the Counter component won't increment because we're busy in the infinite loop?

dai-shi:
I thought count component increment takes some ticks.
Hm, I just tried, but it doesn't work. (maybe it only consumes microtasks.)
My apologies. Please revert to `setTimeout`. It seems like no delay is necessary.
```js
await new Promise((r) => setTimeout(r))  
```

hahnbiatfigma:
I reverted the test to setTimeout. However, I found that with no delay, the test would still pass without using `copySuspensePromise`, so I added a smaller delay (10ms)

PR #1579, opened 2022-11-17
feat(utils): move atomWithHash to peer deps

dai-shi:
Like #1573, but only for `atomWithHash`.

Two motivations behind this:
- Jotai v2 API will change `atomWithStorage` slightly #1515 , and atomWithHash is dropped because it doesn't fit very well.
- `jotai-location` already provides `atomWithLocation` which is sometime better depending on use cases, so adding `atomWithHash` there is a natural fit.


src/utils.ts
```diff
@@ -15,10 +15,10 @@ export { waitForAll } from './utils/waitForAll'
...
   NO_STORAGE_VALUE as unstable_NO_STORAGE_VALUE,
   atomWithStorage,
-  atomWithHash,
   createJSONStorage,
 } from './utils/atomWithStorage'
 export { atomWithObservable } from './utils/atomWithObservable'
 export { useHydrateAtoms } from './utils/useHydrateAtoms'
 export { loadable } from './utils/loadable'
 export { abortableAtom } from './utils/abortableAtom'
+export { atomWithHash } from 'jotai-location'
```
dai-shi, line 24: Oops, this is breaking. Let's revert.

PR #1634, opened 2022-12-13
Example should be !item.online (?)

Ymirke:
Hey, just checking out this library. Unless I'm missing something this should be changed.


dai-shi (approved):
Nice catch. Thanks for your contribution!

PR #1675, opened 2023-01-10
breaking(build): drop module condition in exports

dai-shi:
"module" condition was added for non-native ESM to possibly support old bundlers.
Not sure, if/how it's used. Maybe, it was added for webpack 4, but not required for v2.


dai-shi:
what do you think? @barelyhuman

barelyhuman:
This was added for snowpack if I remember properly, since it didn't pick up the `import` key, let me see if it can pick it up now and we should be able to move forward with this, also , reminder, don't trust the codesandbox for this since it's flaky for snowpack package resolutions

barelyhuman:
Confirmed, 
Seems to be working 👍🏼

dai-shi:
Checking https://github.com/pmndrs/jotai/discussions/1756 if this is related.

PR #1677, opened 2023-01-10
feat(devtools): deprecate jotai/devtools to prepare for v2

dai-shi:
following #1672, similar to #1631

(pre-release v2 API devtools will be removed in a follow up patch release.)


dai-shi:
unfortunately, codesandbox ci doesn't seem stable.

PR #1701, opened 2023-01-19
fix(docs): Update typescript.mdx

meotimdihia:
Add an example about the case access to the return type

## Related Issues

https://github.com/pmndrs/jotai/discussions/1699

## Summary
Add an example about the case access to the return type

## Check List


dai-shi (commented):
Thanks for the improvement. Looks good other than one comment.

dai-shi (approved):
Thanks for your contribution!

PR #1720, opened 2023-01-30
feat(website): v2 update

sandren:
This PR upgrades the Jotai website for the v2 release.

jotai-v2-website


dai-shi (approved):
website only 😃

PR #1759, opened 2023-02-08
fix(react): avoid JSX transform for <= React 18 interop

CodyJasonBennett:
## Summary

Prefers to use React's classic transform `import { createElement } from 'react'` to the new JSX transform `import { jsx } from 'react/jsx-runtime'` which relies on use of `package#exports` not present before React 18 (https://github.com/facebook/react/issues/20235). In earlier versions, this would need to point to `react/jsx-runtime.js`, so it's not trivial to support React 17-18 when using automatic transform in an ESM context. I'd very much prefer for this to not require the use of the React namespace since that can be lost in refactor (e.g. https://github.com/kaqiinono/babel-jsx-pragma-module-auto-import), but I'll have to defer that to someone more familiar with the library and its build system.


dai-shi:
I thought it's fine as React 18 is still CJS and not a big difference, but I see "exports" is the difference.
If this causes too much trouble, we can go back to `createElement` as did in v1.

PR #1768, opened 2023-02-13
fix(vanilla): async derived atom not updated (race condition in an edge case)

gimelfarb:
## Related Issues

No issue was created.
See: https://github.com/pmndrs/jotai/discussions/1761

## Summary

When async derived atom is set immediately following an 'await' on get, it fails to get marked for re-calculation, so subsequent get will return the old value.

Included a unit test that has a minimal failure reproduction scenario, and a fix that makes the test pass.

## Check List


dai-shi (approved):
Nice! This looks perfect.
Thanks for your work!

PR #1774, opened 2023-02-15
fix(vanilla): use toStringTag symbol instaed of toString method

dai-shi:
This should improve toString behavior?


dai-shi:
This doesn't work as I hoped (which is to keep the current behavior). Closing for now.

PR #1781, opened 2023-02-18
Feat/doc atomWithDefault another way

t6adev:
## Summary

Hi, This is an alternative way to `atomWithDefault` for the documentation, not a request to add any functions.

Here is my motivation against atomWithDefault with its document,

```
As you can see above the example, the two atoms' relation is disconnected after updating created one, `count2Atom = atomWithDefault((get) => get(count1Atom) * 2)`.
Let's confirm what's occurred,
- 1. Click "increment count1", then count1 is 2 and count2 is 4
- 2. Click "increment count2", then count1 is 2 and count2 is 5 (Disconnected!!)
Those atoms have no relation after updating count2Atom. So,
- Click "increment count1", count1 is incremented only
- Even if you reset count2Atom, these dependency relation never come back

### Motivation
In some cases, 
- After disconnecting and resetting, they should come back to their relation
- Derived atoms should be reset based on updated the original atom
- We'd like to reset all derived atoms but just want to operate as simply as possible
How do we make those cases? 
Here is a declarative way to create a function to provide a refreshable atom instead of atomWithDefault.
```

## Check List

## My Check List

- [ ] What codesandbox id should I use (the current one is presented by my project)


dai-shi (commented):
Thanks for your contribution!

I wonder if this should be in atom creator recipe or resettable utility in docs.
I'm open for either case.
@sandren any thoughts?

sandren:
> Thanks for your contribution!
> 
> I wonder if this should be in atom creator recipe or resettable utility in docs. I'm open for either case. @sandren any thoughts?

@dai-shi The utils docs should focus on the content of the `jotai/utils` bundle and not alternate implementations. For example the persistence doc has alternatives to `atomWithStorage`, but they're not mentioned in Storage. So I think it would be best to add this as a new guide or reciple, unless you're open to adding a new `atomWithRefresh`?

dai-shi:
Yeah, let's move this to atom creator recipe. @t6adev

t6adev:
Thank you for your direction. Moved it :)
@dai-shi @sandren

How about the codesandbox project id I wrote? No problem?

dai-shi (approved):
LGTM

I think it's time to split atom creators recipes into multiple mdx files, isn't it?

PR #1794, opened 2023-02-25
fix(utils, types): added export of Loadable type

uaeio:
It would be nice for me  to have this type exported, since I use it in my typings

### DIFF
--- a/src/vanilla/utils.ts
+++ b/src/vanilla/utils.ts
@@ -12,5 +12,5 @@ export {
   createJSONStorage,
 } from './utils/atomWithStorage'
 export { atomWithObservable } from './utils/atomWithObservable'
-export { loadable } from './utils/loadable'
+export { loadable, Loadable } from './utils/loadable'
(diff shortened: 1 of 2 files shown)


src/vanilla/utils.ts
```diff
@@ -12,5 +12,5 @@ export {
   createJSONStorage,
 } from './utils/atomWithStorage'
 export { atomWithObservable } from './utils/atomWithObservable'
-export { loadable } from './utils/loadable'
+export { loadable, Loadable } from './utils/loadable'
```
dai-shi, line 15: Let's not export from the index. (Like  `AtomFamily`.)
```suggestion
export { loadable } from './utils/loadable'
```

dai-shi (approved):
Not exporting the Loadable type was intentional, but I see there's a little pain with it. Assuming this is fairly stable API now, I think it's okay to make it public.

dai-shi (changes requested):
I don't have a right to commit changes. Can you commit the suggestion?

PR #1805, opened 2023-03-02
fix(docs): clarify on atomWithStorage in v2

### LINKED ISSUE #1791: atomWithStorage behaviour change from v1 -> v2, and documentation
## Summary

After the v2 launch, atomWithStorage by default gets its initial value from `baseAtom.onMount ()` method, making it breaking existing logic. This would be better be mentioned in the documentation which currently **at least** indicates otherwise.

## Link to reproduction

https://codesandbox.io/s/vibrant-waterfall-p96z11?file=/src/App.tsx

Click on the SETVAL button in the link and get the value updated in the local storage, then refresh the page.

With Jotai v2.0.2, after each refresh of the page, `currText` will be logged twice in `useEffect ()`, the first one `{wow: 'init'}`, and the second one `{wow: 'updated'}`.
With Jotai v1.1.13, after each refresh of the page `currText` will only be logged once, namely `{wow: 'updated'}`.

The current [documentation](https://jotai.org/docs/guides/persistence#a-simple-pattern-with-localstorage) at least indicates that `getInitialValue ()` happens immediately at the point where the atom is created, which will probably lead to confusions.

Basically, it is used to be OK to put the `currText` atom in the dependency list of `useEffect ()`. Now it has to to be placed inside that of something like `useUpdateEffect ()`. I don't know whether this change is intentional to make v2 more "async-look-and-feel" or it is just a slip.

wenq1:

Fixes #1791 #1803

## Summary

See if the documentation makes it more clear

## Check List

### DIFF
--- a/docs/guides/migrating-to-v2-api.mdx
+++ b/docs/guides/migrating-to-v2-api.mdx
@@ -244,7 +244,7 @@ For more information, refer the following discussions:
 
 ### Utils
 
-- `atomWithStorage` util's `delayInit` is removed as being default. The new behavior can be breaking in some cases. See https://github.com/pmndrs/jotai/discussions/1737 for more information.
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Thanks for the clarification!

PR #1773, opened 2023-02-14
tests: improve test reliability

arjunvegda:
## Summary

- Improve tests for abortable atom

## Check List

- [x] `yarn run prettier` for formatting code and docs

PR #1814, opened 2023-03-04
fix(atomWithStorage, createJSONStorage): scope subscriptions to specified browser storage

### LINKED ISSUE #1812: createJSONStorage / atomWithStorage subscribes to all storage events, regardless of where value is stored
## Summary

When using `createJSONStorage` (or using it indirectly via `atomWithStorage`'s default settings) in a browser environment, a `storage` event handler is bound to the `window`.

https://github.com/pmndrs/jotai/blob/fed8d807a0a1a4c6361f6b30356e58895a26b9a3/src/vanilla/utils/atomWithStorage.ts#L77-L92

This `storage` event handler reacts to ALL storage events (both `localStorage` and `sessionStorage`) regardless of what storage is actually being used by the atom, as long as the `key` matches. This means an atom which stores its values in `localStorage` will get updated by changes to `sessionStorage` with the same key (and vice-versa). It also means that custom storage implementations, which use neither local or session storage, will get updated by both local and session storage events unexpectedly.

## Fix discussion

A naive fix would be to add a conditional check to the `storage` event handler, which checks if the `storageArea` property of the [StorageEvent](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event#event_properties):

```diff
const storageEventCallback = (e: StorageEvent) => {
-  if (e.key === key && e.newValue) {
+  if (e.storageArea === getStringStorage() && e.key === key && e.newValue)
    callback(JSON.parse(e.newValue))
  }
}
```

But this would break subscriptions for users who are providing a custom wrapper around `localStorage`, since the reference returned by `getStringStorage()` would not match the `storageArea` property of the event.

I'm not sure if there's a way to solve it for all cases while keeping the same API interface. It's likely that custom storage implementations will need to implement the internals of `createJSONStorage` on their own instead.

## Link to reproduction

I've created a minimal reproduction of the issue on codesandbox here:

https://codesandbox.io/s/jotai-storage-bug-9zlsho?file=/src/App.tsx

To reproduce the issue, you will need to open TWO separate tabs/windows of this demo.

After clicking the "Increment" button for any individual atom, ALL atoms get updated unexpectedly in the other window.

## Check List

nderscore:

Fixes #1812

## Summary

* Added check to `createJSONStorage` to ensure that it only listens to the specified window `storage` events.

* Added check to avoid adding an event handler if neither `localStorage` or `sessionStorage` is passed.

I didn't add any documentation around how to create a custom storage in this PR, because I found [another bug related to storage](https://codesandbox.io/s/jotai-storage-bug-reset-u9doo3?file=/src/App.tsx) (issue ticket coming soon) which will likely require more involved changes than this one. I think it makes sense to hold off on documentation updates until after that fix.

## Check List


src/vanilla/utils/atomWithStorage.ts
```diff
@@ -76,11 +76,16 @@ export function createJSONStorage<Value>(
   }
   if (
     typeof window !== 'undefined' &&
-    typeof window.addEventListener === 'function'
+    typeof window.addEventListener === 'function' &&
+    getStringStorage() instanceof window.Storage
```
dai-shi, line 80: I don't think we should call `getString()` on creation. (It will fail on server.)

dai-shi (changes requested):
Thanks for working on this. Can you also check failed CIs?

src/vanilla/utils/atomWithStorage.ts
nderscore: It wouldn't run on the server because the condition will short-circuit on the checks for `window` before it reaches the call to `getStringStorage()`

Happy to remove this optimization though if it's a concern.

nderscore:
> Can you also check failed CIs?

I pushed some updates to the unit tests:

* Tightened up the condition checking for `addEventListener` so that it will specifically look for `storage` event handlers and ignore `error` event handlers [being set by the fiber renderer](https://github.com/facebook/react/blob/v17.0.0/packages/shared/invokeGuardedCallbackImpl.js#L189)
  * This should (hopefully) fix the CI checks that were failing on React <18
  
* Added a test for the `StorageEvent.storageArea` condition of the event handler

src/vanilla/utils/atomWithStorage.ts
dai-shi: Yes, please remove it. Some configs on sever add dummy `window`.

nderscore: Got it. :+1:  I reverted that part of my change and squashed the git history.

dai-shi (approved):
The code looks good to me. The remaining is CI. Any clue?

nderscore:
Just pushed an update that should fix the CI failing on React 16.9.

The test was grabbing the wrong event handler for mocking storage events. 

React 16.9's window `error` handler was the last event handler bound to the window.

```diff
-    const storageEventHandler = mockAddEventListener.mock.calls[
-      mockAddEventListener.mock.calls.length - 1
-    ]?.[1] as (e: StorageEvent) => void
+    const storageEventHandler = mockAddEventListener.mock.calls
+      .filter(([eventName]) => eventName === 'storage')
+      .pop()?.[1] as (e: StorageEvent) => void
```

nderscore:
https://github.com/pmndrs/jotai/actions/runs/4333932972/jobs/7567663941#step:9:1826

This one is a real head-scratcher. It doesn't look like it should be related to my changes :thinking:

dai-shi:
> https://github.com/pmndrs/jotai/actions/runs/4333932972/jobs/7567663941#step:9:1826
> 
> This one is a real head-scratcher. It doesn't look like it should be related to my changes 🤔

Yeah, our tests are not very stable. We should fix them.
Like #1773 by @arjunvegda , maybe for everything.

PR #1818, opened 2023-03-05
fix(babel): improve debug label support for third-party jotai libraries

arjunvegda:
## Summary
- Add debug label support for third-party jotai libraries like `jotai-optics`, `jotai-location`, `jotai-form`, `jotai-cache`, `jotai-recoil`, etc.

## Check List


dai-shi (approved):
Wow, thanks so much for the fix!

PR #1856, opened 2023-03-29
docs(DevTools): add note about @emotion/react + misc. updates

arjunvegda:
- Relevant PR https://github.com/jotaijs/jotai-devtools/pull/42
- Ready to be merged at any time

## Check List

### DIFF
--- a/docs/tools/devtools.mdx
+++ b/docs/tools/devtools.mdx
@@ -11,16 +11,17 @@ Install `jotai-devtools` to your project to get started
 
 ```sh
 # yarn
-yarn add jotai-devtools
+yarn add jotai-devtools @emotion/react
 
 # npm
-npm install jotai-devtools --save
+npm install jotai-devtools @emotion/react --save
 ```
 
 ## Notes
 
-- All the components and the hooks are optimized to be tree-shakable for production builds and **only works in a non-production environment**
-- File issues or ask questions on the [Jotai DevTools GitHub Repo](https://github.com/jotaijs/jotai-devtools/discussions)
+- `<DevTools/>` is optimized to be tree-shakable for production builds, and **only works in a non-production environment**
+- Hooks are dev-only, and are designed to work in a non-production environment
+- Feedback welcome, please file issues or ask questions on the [Jotai DevTools GitHub Repo](https://github.com/jotaijs/jotai-devtools/discussions)
 
 ## Quick links
 


sandren:
@arjunvegda Thanks for update! ☺️

Could `@emotion/react` be moved to a direct dependency of `jotai-devtools` instead of a peer dependency? Also the docs should install `jotai-devtools` with the `-D` flag since it isn't used in production builds.

arjunvegda:
> @arjunvegda Thanks for update! ☺️
> 
> Could `@emotion/react` be moved to a direct dependency of `jotai-devtools` instead of a peer dependency? Also the docs should install `jotai-devtools` with the `-D` flag since it isn't used in production builds.

Thanks for taking a look! 🙏 

> Could `@emotion/react` be moved to a direct dependency of `jotai-devtools` instead of a peer dependency?

We moved away from that approach [recently](https://github.com/jotaijs/jotai-devtools/pull/42). We realized that it may cause issues like [these](https://github.com/Tintef/react-google-places-autocomplete/issues/176) where it loads multiple instances of the library if the user relies on `@emotion/react` on their end.

The maintainers [recommend](https://github.com/emotion-js/emotion/issues/2343#issuecomment-817265816) using it as a peer dependency too.

> Also the docs should install `jotai-devtools` with the `-D` flag since it isn't used in production builds

I thought about it, but it's not fully dev-only unlike `eslint`, `jest`, etc. They'll have to strip out the `jotai-devtools` code when they build a prod bundle to fully use `yarn install --production` otherwise, it'll throw reference errors on their production builds which may lead to confusion 😅

sandren:
@arjunvegda Thanks for those details. I'll go ahead and merge your changes, but in the future, we should see if `jotai-devtools` can be refactored to avoid such a heavy peer dependency. Thanks!

nicobatalla:
The problem with this is that if you're using any other styling solution out there + `jotai` + `jotai-devtools` you are forced to install `@emotion/react` in your product even if your app is not using it at all. 

Direct dependency was better IMO

arjunvegda:
> The problem with this is that if you're using any other styling solution out there + `jotai` + `jotai-devtools` you are forced to install `@emotion/react` in your product even if your app is not using it at all.
> 
> Direct dependency was better IMO

Is installation the only concern or did we run into any interop issues? We wanted to cater to users using `emotion` as their primary styling solution, too, and prevent version mismatch issues for them. 

FWIW, we're planning to move towards a native CSS solution and move away from`@emotion/react in the long term.

nicobatalla:
There should not be any interoperation issues but the build is failing on the jotai-devtools version update PR in our repo because we don't have `@emotion/react` installed.  In our case we are using `styled-components`, so we don't really need `emotion`. 

The previous approach was fine because it was tree-shaken anyway, but now we are forced to install `emotion` too if we want to update.

arjunvegda:
> There should not be any interoperation issues but the build is failing on the jotai-devtools version update PR in our repo because we don't have `@emotion/react` installed. In our case we are using `styled-components`, so we don't really need `emotion`.
> 
> The previous approach was fine because it was tree-shaken anyway, but now we are forced to install `emotion` too if we want to update.

Gotcha! We should still be okay from the tree-shaking standpoint, no? Or are you running into any issues there?

If it's a blocker for your workflow, I can try identifying a few workarounds that you could use on your end, until we completely move towards a native CSS solution.

nicobatalla:
Need to check if the bundle size would be increased, but we would definitely prefer not having to install emotion at all if possible.

Any ETA on the move to native CSS?

Issue #1387, opened 2022-08-30
Typescript issue with Atoms of different types


elz5star:
Hooks such as useAtom, useAtomValue and useSetAtom have a typescript error when using atoms of different types. As demonstrated in this typescript playground: https://tsplay.dev/N5Lk2w

dai-shi:
Would anyone like to tackle this? I'll take it otherwise. It may require decent TS skills. (Also, note that we don't generally export types.)

carlesba:
I think it's doable if types are exported. Is that an option?

dai-shi:
It doesn't sound like a good option, but if you have an idea, can you please try with typescript playground?

Here's something you can start with: https://tsplay.dev/WKRMgm

carlesba:
Great! Thanks for the playground.

Here's the **solution** I was mentioning https://tsplay.dev/mAQMRm

It works exporting a `AtomUnion` type to infer the values correctly. 
The **issue** I found is that typescript is not inferring  unions of generic values correctly (ie, `Foo<number> | Foo<string>`). 
Curiously,  it does it well for things like functions. Like, if you pass `value: (() => string) | (()=> number)` it will infer `() => string | number`.

A generic `AtomUnion` repackages the types converting things like `Atom<number> | Atom<string>` into `Atom<number | string>`.

Now, despite this is possible **I'm not sure this is something we should encourage**. There're some side effects to take into account. 
Users could find issues when setting values into the atoms as the new types are wider.  
Say,
```ts
const a = atom(0); 
const b = atom('');
function run(a: Atom<string|number>){
  const setA = useSetAtom(a)
  // ...
  setA("bad") // would break atom a
  setA(1) // would break atom b
```
Here, both cases are valid for the type-checker but can break in runtime. 

I'd love to know your opinion here @dai-shi

dai-shi:
Here's the solution I had in mind (but didn't confirm until now): https://tsplay.dev/WJ5MZm
It's a breaking change in types, but I think we'd do with minor version up, if we were to do it.
Do you see any drawbacks other than that?

carlesba:
Nice! way better.

dai-shi:
I did for the core types in #1427. Will continue.

dai-shi:
I'm not quite sure if/how this causes some troubles with existing app code.

dai-shi:
https://tsplay.dev/werx1N
Okay, deprecation works. Let's do that.

dai-shi:
[...]

PR #1866, opened 2023-04-04
fix(react): atom type inference in hooks

Thisen:

Fixes #1831

## Summary

Draft. Lets iterate on solving this inference issue with `useAtom`.

### DIFF
--- a/src/react/useAtom.ts
+++ b/src/react/useAtom.ts
@@ -32,11 +32,6 @@ export function useAtom<
   SetAtom<ExtractAtomArgs<AtomType>, ExtractAtomResult<AtomType>>
 ]
 
-export function useAtom<AtomType extends Atom<unknown>>(
-  atom: AtomType,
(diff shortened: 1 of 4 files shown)


src/react/useAtom.ts
```diff
@@ -32,11 +32,6 @@ export function useAtom<
   SetAtom<ExtractAtomArgs<AtomType>, ExtractAtomResult<AtomType>>
 ]
 
-export function useAtom<AtomType extends Atom<unknown>>(
```
dai-shi, line 35: I wonder if `unknown` is problematic. Can you try `any` instead? If that works, it was my bad then...

dai-shi:
Related PR is #1427 and the issue is #1387. Our fault is that we didn't add tests back then.

dai-shi:
Nice! Can you fix `useAtomValue` and `useSetAtom` too?

PR #1882, opened 2023-04-15
fix(atomWithStorage): handle RESET/removeItem from cross-tab storage updates

### LINKED ISSUE #1815: createJSONStorage does not properly subscribe to updates caused by RESET
## Summary

When an item is removed from storage, a `storage` event fires with `null` as the incoming value. The storage event handler in `createJSONStorage` ignores `null` values from incoming `storage` events. 

This is caused by a check for a non-falsy `e.newValue` here:
https://github.com/pmndrs/jotai/blob/main/src/vanilla/utils/atomWithStorage.ts#L83

When `setAtom(RESET)` is called on an `atomWithStorage`, subscribers in other windows do not get the reset update.

## Link to reproduction

I've created a minimal reproduction of the issue on codesandbox here:

https://codesandbox.io/s/jotai-storage-bug-reset-u9doo3?file=/src/App.tsx

Like with #1812 you might have to open the preview iframe itself (https://u9doo3.csb.app/) in separate windows to avoid permissions errors.

## Fix discussion

I think that in order to fix this, the interface of the `(Sync/Async)Storage.subscribe` argument `callback` needs to be updated to accept `NO_STORAGE_VALUE`, so that the Storage can inform `atomWithStorage` that a reset occurred in another window.

```diff
--- a/src/vanilla/utils/atomWithStorage.ts
+++ b/src/vanilla/utils/atomWithStorage.ts
@@ -15,14 +15,14 @@ export interface AsyncStorage<Value> {
   getItem: (key: string) => Promise<Value | typeof NO_STORAGE_VALUE>
   setItem: (key: string, newValue: Value) => Promise<void>
   removeItem: (key: string) => Promise<void>
-  subscribe?: (key: string, callback: (value: Value) => void) => Unsubscribe
+  subscribe?: (key: string, callback: (value: Value | typeof NO_STORAGE_VALUE) => void) => Unsubscribe
 }
 
 export interface SyncStorage<Value> {
   getItem: (key: string) => Value | typeof NO_STORAGE_VALUE
   setItem: (key: string, newValue: Value) => void
   removeItem: (key: string) => void
-  subscribe?: (key: string, callback: (value: Value) => void) => Unsubscribe
+  subscribe?: (key: string, callback: (value: Value | typeof NO_STORAGE_VALUE) => void) => Unsubscribe
 }
 
 export interface AsyncStringStorage {
@@ -83,10 +83,13 @@ export function createJSONStorage<Value>(
       const storageEventCallback = (e: StorageEvent) => {
         if (
           e.storageArea === getStringStorage() &&
-          e.key === key &&
-          e.newValue
+          e.key === key
         ) {
-          callback(JSON.parse(e.newValue))
+          try {
+            callback(JSON.parse(e.newValue ?? ''))
+          } catch (e) {
+            callback(NO_STORAGE_VALUE)
+          }
         }
       }
       window.addEventListener('storage', storageEventCallback)
@@ -126,7 +129,9 @@ export function atomWithStorage<Value>(
     }
     let unsub: Unsubscribe | undefined
     if (storage.subscribe) {
-      unsub = storage.subscribe(key, setAtom)
+      unsub = storage.subscribe(key, (nextValue) => {
+        setAtom(nextValue === NO_STORAGE_VALUE ? initialValue : nextValue)
+      })
     }
     return unsub
   }
```

Currently, the only place the `NO_STORAGE_VALUE` symbol is exposed to consumers is with an `unstable_` prefix. 

That change would technically leak `NO_STORAGE_VALUE` as part of the exposed types for `Sync/AsyncStorage`.

It might make sense to remove the unstable prefix from the exported `NO_STORAGE_VALUE` symbol because of that? 

**Semi-related:** Another reason it might make sense to expose `NO_STORAGE_VALUE` is because I don't think you can create a custom storage (one that doesn't use `createJSONStorage`, such as would be required to make a validated storage - #1198) without using that symbol either.

## Check List

nderscore:

Fixes #1815

## Summary

* **BREAKING CHANGE:** `SyncStorage` and `AsyncStorage` now receive `initialValue` as an argument to `getItem` and `subscribe`
* **BREAKING CHANGE:** Removed export `unstable_NO_STORAGE_VALUE`
* Updated `createJSONStorage` to correctly handle incoming `storage` events triggered by `removeItem` in another window
* Updated `createJSONStorage` to avoid attaching `storage` event handler when the storage is not browser storage (this happens on mount, so it shouldn't conflict with SSR)
* Added unit tests to cover both of the updates to `createJSONStorage` behavior.
* Updated mock "listeners" in unit tests to more correctly reflect browser storage behavior ([the `storage` event is not triggered by updates from the current window](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event))
* Updated documentation to reflect the new storage API
* Added documentation example for `atomWithStorage` with a custom storage that performs runtime validation

## Check List

### DIFF
--- a/docs/utilities/storage.mdx
+++ b/docs/utilities/storage.mdx
@@ -34,10 +34,14 @@ The `atomWithStorage` function creates an atom with a value persisted in `localS
 
 **initialValue** (required): the initial value of the atom
 
-**storage** (optional): an object with:
+**storage** (optional): an object with the following methods:
 
-- `getItem`, `setItem` and `removeItem` methods for storing/retrieving/deleting persisted state; defaults to using `localStorage` for storage/retrieval and `JSON.stringify()`/`JSON.parse()` for serialization/deserialization.
-- Optionally, the storage has a `subscribe` property, which can be used to synchronize storage. The default `localStorage` handles `storage` events for cross-tab synchronization.
+- **getItem(key, initialValue)** (required): Reads an item from storage, or falls back to the `intialValue`
+- **setItem(key, value)** (required): Saves an item to storage
+- **removeItem(key)** (required): Deletes the item from storage
+- **subscribe(key, callback, initialValue)** (optional): A method which subscribes to external storage updates.
+
+If not specified, the default storage implementation uses `localStorage` for storage/retrieval, `JSON.stringify()`/`JSON.parse()` for serialization/deserialization, and subscribes to `storage` events for cross-tab synchronization.
 
 <CodeSandbox id="vuwi7" />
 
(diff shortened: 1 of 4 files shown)


dai-shi (approved):
Looks great!
I'll merge it when we are ready to release a minor bump with other changes.

nderscore:
Just noticed a small mistake in the validated storage example I added to the docs - I had forgotten to actually call the callback :facepalm: 

Fixed in 6d62ef3 - sorry about that!

PR #1885, opened 2023-04-19
Improve typescript documentation

eikonomega:
Add additional explanation of how to correctly type derived atoms.

#1870 

## Check List

### DIFF
--- a/docs/guides/typescript.mdx
+++ b/docs/guides/typescript.mdx
@@ -33,11 +33,31 @@ const arrAtom = atom<string[]>([])
 
 ### Derived atoms can mostly have their types inferred
 
+In general, this is the recommended approach since typing derived
+atoms can get a little confusing, particularly for those who are
+doing it for the first time.
+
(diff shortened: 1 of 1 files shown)


docs/guides/typescript.mdx
```diff
@@ -31,13 +31,45 @@ const numAtom = atom<number | null>(0)
...
+const readWriteAtom = atom(
   (get) => get(strAtom),
+  (_get, set, num: number) => set(strAtom, String(num))
+)
+```
+
+### Derived atoms can also be explicitly typed
+
+If you encounter a situation where you need or want to explicitly
+type your derived atoms, you can do that as well.
```
dai-shi, line 48: I wonder if we can emphasize this more.
We should probably show more example with type inference.
(Currently, the section size seems bigger for manual typing.)

sandren (approved):
This is really helpful. Thanks! ☺️

PR #1887, opened 2023-04-19
feat(useLazyValues)

bitttttten:
I use this hook a lot, so I thought I'd contribute it back to the project. In short, it's a hook that lazyly lets you read the value of an atom, so a component could use this atom value in a callback, user event, mutation, etc. without needing to re-render the component every time.

In the below example, we have a button that posts some values to an API. We only need these values when the user clicks on the button, it would be a waste of renders to subscribe to all recent updates in the input and date which, in our example, might change often.

```tsx
function Submit() {
   let lazyInput = useLazyValue(inputAtom)
   let lazyDate = useLazyValue(dateAtom)

  function onClick() {
    let input = lazyInput()
    let date = lazyDate()

    sendForm({ input, date })
  }

  return (
    <button type="button" onClick={onClick}>Submit values</button>
  )
}
```


dai-shi:
While we appreciate for suggestions for new features, we strictly design primitive utils. For example, we basically don't have utils that depend on other utils. Such features should be provided as third-party libraries. The same applies for this case. Please consider developing it in a third-party  library. Thanks!

PR #1790, opened 2023-02-25
feat: add dev_subscribe_store

arjunvegda:
## Summary
- Add the `dev_subscribe_store` method to subscribe to the store lifecycle events (`state`,`sub`, `unsub`)
- Mark `dev_subscribe_state` as deprecated, to be removed in the upcoming jotai minor upgrade (2.**x**.x)

**Note:**
These methods are intended to be used by `jotai-devtools` and are subject to change without notice. Please avoid using them directly in your apps.

## Check List

- [x] `yarn run prettier` for formatting code and docs

PR #1893, opened 2023-04-23
breaking: remove deprecated store dev methods

dai-shi:

#1790

## Summary

Removed the deprecated one

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -108,7 +108,6 @@ type Mounted = {
 }
 
 // for debugging purpose only
-type StateListener = () => void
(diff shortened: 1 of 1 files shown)


arjunvegda (approved):
LGTM! 

I'll bump the `peerDependencies` of `jotai-devtools` to `jotai >=2.1.0`. 

Should we also throw an error [here](https://github.com/jotaijs/jotai-devtools/blob/8c63da0a3ff75aa3f4d0ab8aa21c6ffaf879db33/src/utils/useAtomsSnapshot.ts#L46-L48) to make it more explicit for the users 🤔?

dai-shi:
> Should we also throw an error

I think it's unnecessary.

PR #1898, opened 2023-04-26
fix(react): Infer useHydrateAtoms value types from atom.read ReturnType

### LINKED ISSUE #990: Infered types when using useHydrateAtoms
Hi,

Thanks for creating this elegant way of managing state, it's opened so many architectural doors for us to explore!

**Goal**
It would be really nice if the atoms passed to `useHydrateAtoms` could have their types inferred and checked against.

**Current situation**
This will compile perfectly fine:
```js
const volumeAtom = atom(100);
useHydrateAtoms([[volumeAtom, 'not-an-integer']]);
```

**Question**
Is it possible to infer the type of the atoms provided to `useHydrateAtoms`?

ericdowell:

Fixes #990

## Summary
Currently `useHydrateAtoms` allows any unknown value to be passed for all atoms being hydrated. Instead this should be type safe and infer the type for the value(s) passed based on the atom.read function return type.

Cleaned up test title and removed duplicate test.

## Screenshots
|     BEFORE      |      AFTER      |
|:---------------:|:---------------:|
| Screenshot 2023-04-27 at 12 09 19 PM | Screenshot 2023-04-27 at 12 08 29 PM |
| Screenshot 2023-04-27 at 12 09 36 PM | Screenshot 2023-04-27 at 12 10 24 PM |
| Screenshot 2023-05-01 at 9 11 40 PM | Screenshot 2023-05-01 at 9 12 23 PM |
| Screenshot 2023-05-01 at 9 14 45 PM | Screenshot 2023-05-01 at 9 15 43 PM |

## Check List

### DIFF
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -3,7 +3,7 @@ import type { WritableAtom } from '../../vanilla.ts'
 
 type Store = ReturnType<typeof useStore>
 type Options = Parameters<typeof useStore>[0]
-type AnyWritableAtom = WritableAtom<unknown, any[], any>
+export type AnyWritableAtom = WritableAtom<unknown, any[], any>
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -5,20 +5,18 @@ type Store = ReturnType<typeof useStore>
 type Options = Parameters<typeof useStore>[0]
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
 type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]
-type InferAtoms<
-  T extends Array<AtomTuple>,
-  S extends Array<AtomTuple> = []
-> = S['length'] extends T['length']
-  ? S
-  : T extends Array<AtomTuple<infer A>>
-  ? A extends AnyWritableAtom
-    ? InferAtoms<T, [AtomTuple<A, ReturnType<A['read']>>, ...S]>
-    : T
-  : T
+
+type InferAtoms<T extends Iterable<AtomTuple>> = {
+  [K in keyof T]: T[K] extends AtomTuple<infer A>
+    ? A extends AnyWritableAtom
+      ? AtomTuple<A, ReturnType<A['read']>>
+      : T[K]
+    : never
+}
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
-export function useHydrateAtoms<T extends Array<AtomTuple>>(
+export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
   values: InferAtoms<T>,
   options?: Options
 ) {
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -4,17 +4,26 @@ import type { WritableAtom } from '../../vanilla.ts'
 type Store = ReturnType<typeof useStore>
 type Options = Parameters<typeof useStore>[0]
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
+type AtomMap<A = AnyWritableAtom, V = unknown> = Map<A, V>
 type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]
 type InferAtoms<T extends Iterable<AtomTuple>> = {
   [K in keyof T]: T[K] extends AtomTuple<infer A>
     ? A extends AnyWritableAtom
       ? AtomTuple<A, ReturnType<A['read']>>
       : T[K]
-    : any
+    : never
 }
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
+export function useHydrateAtoms<T extends AtomMap>(
+  values: T,
+  options?: Options
+): void
+export function useHydrateAtoms<T extends Array<AtomTuple>>(
+  values: InferAtoms<T>,
+  options?: Options
+): void
 export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
   values: InferAtoms<T>,
   options?: Options
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -5,7 +5,6 @@ type Store = ReturnType<typeof useStore>
 type Options = Parameters<typeof useStore>[0]
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
 type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]
-
 type InferAtoms<T extends Iterable<AtomTuple>> = {
   [K in keyof T]: T[K] extends AtomTuple<infer A>
     ? A extends AnyWritableAtom
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -16,11 +16,15 @@ type InferAtoms<T extends Iterable<AtomTuple>> = {
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
+export function useHydrateAtoms<T extends Array<AtomTuple>>(
+  values: InferAtoms<T>,
+  options?: Options
+): void
 export function useHydrateAtoms<T extends AtomMap>(
   values: T,
   options?: Options
 ): void
-export function useHydrateAtoms<T extends Array<AtomTuple>>(
+export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
   values: InferAtoms<T>,
   options?: Options
 ): void
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -11,7 +11,7 @@ type InferAtoms<T extends Iterable<AtomTuple>> = {
     ? A extends AnyWritableAtom
       ? AtomTuple<A, ReturnType<A['read']>>
       : T[K]
-    : never
+    : any
 }
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -3,7 +3,7 @@ import type { WritableAtom } from '../../vanilla.ts'
 
 type Store = ReturnType<typeof useStore>
 type Options = Parameters<typeof useStore>[0]
-export type AnyWritableAtom = WritableAtom<unknown, any[], any>
+type AnyWritableAtom = WritableAtom<unknown, any[], any>
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -7,8 +7,8 @@ export type AnyWritableAtom = WritableAtom<unknown, any[], any>
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
-export function useHydrateAtoms(
-  values: Iterable<readonly [AnyWritableAtom, unknown]>,
+export function useHydrateAtoms<T extends AnyWritableAtom, V = T['read']>(
+  values: Iterable<readonly [T, V]>,
   options?: Options
 ) {
   const store = useStore(options)
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -4,11 +4,22 @@ import type { WritableAtom } from '../../vanilla.ts'
 type Store = ReturnType<typeof useStore>
 type Options = Parameters<typeof useStore>[0]
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
+type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]
+type InferAtoms<
+  T extends Array<AtomTuple>,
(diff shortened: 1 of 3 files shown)


tests/react/utils/useHydrateAtoms.test.tsx
```diff
@@ -145,52 +145,6 @@ it('useHydrateAtoms can only restore an atom once', async () => {
   await findByText('count: 44')
 })
 
-it('useHydrateAtoms can only restore an atom once', async () => {
```
ericdowell, line 148: Test was duplicated, still defined above:
https://github.com/pmndrs/jotai/blob/fb1158837023b4d75f7075f9f7d0a4ba3b45845d/tests/react/utils/useHydrateAtoms.test.tsx#L102-L103

src/react/utils/useHydrateAtoms.ts
```diff
@@ -3,12 +3,12 @@ import type { WritableAtom } from '../../vanilla.ts'
 
 type Store = ReturnType<typeof useStore>
 type Options = Parameters<typeof useStore>[0]
-type AnyWritableAtom = WritableAtom<unknown, any[], any>
+export type AnyWritableAtom = WritableAtom<unknown, any[], any>
```
ericdowell, line 6: I found myself duplicating this type, I feel it would be useful to export it for external usage.

dai-shi (commented):
Nice improvement. Thanks for your contribution!

src/react/utils/useHydrateAtoms.ts
ericdowell: Makes sense, it's not a big that I have to duplicate this type.

src/react/utils/useHydrateAtoms.ts
```diff
@@ -7,8 +7,8 @@ type AnyWritableAtom = WritableAtom<unknown, any[], any>
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
-export function useHydrateAtoms(
-  values: Iterable<readonly [AnyWritableAtom, unknown]>,
+export function useHydrateAtoms<T extends AnyWritableAtom, V = T['read']>(
```
dai-shi, line 10: It should be something like this. Not sure why it worked before.
```suggestion
export function useHydrateAtoms<T extends AnyWritableAtom, V = T extends WritableAtom<infer Value, any[], any> ? Value : never>(
```

ericdowell: Test the above and still not what we need. 

### Code
Screenshot 2023-04-27 at 9 25 13 AM
### Test
Screenshot 2023-04-27 at 9 25 39 AM

Need something recursive like this, but this is not working yet either.
```ts
type AnyWritableAtom = WritableAtom<unknown, any[], any>
type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]

type InferAtoms<T extends Iterable<AtomTuple>> = {
  [K in keyof T]: T[K] extends infer A
    ? A extends AnyWritableAtom
      ? AtomTuple<A, ReturnType<A['read']>>
      : [unknown, unknown]
    : [unknown, unknown]
```

### Code
Screenshot 2023-04-27 at 9 26 52 AM

### Test
Screenshot 2023-04-27 at 9 27 09 AM

ericdowell: I think I have it, but we'd have to use `Array<>` rather than `Iterable<>`
```ts
type InferAtoms<
  T extends Array<AtomTuple>,
  S extends Array<AtomTuple> = []
> = S['length'] extends T['length']
  ? S
  : T extends Array<AtomTuple<infer A>>
  ? A extends AnyWritableAtom
    ? InferAtoms<T, [AtomTuple<A, ReturnType<A['read']>>, ...S]>
    : T
  : T

export function useHydrateAtoms<T extends Array<AtomTuple>>(
  values: InferAtoms<T>,
  options?: Options
) {
```
Screenshot 2023-04-27 at 9 37 10 AM

src/react/utils/useHydrateAtoms.ts
```diff
@@ -4,11 +4,22 @@ import type { WritableAtom } from '../../vanilla.ts'
...
+  ? A extends AnyWritableAtom
+    ? InferAtoms<T, [AtomTuple<A, ReturnType<A['read']>>, ...S]>
+    : T
+  : T
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
-export function useHydrateAtoms(
-  values: Iterable<readonly [AnyWritableAtom, unknown]>,
+export function useHydrateAtoms<T extends Array<AtomTuple>>(
```
ericdowell, line 21: What do you think of using `Array` instead of `Iterable`?

ericdowell: This would be a breaking change for `useHydrateAtoms`, using `Map` doesn't work with `Array`. Need to work on the `InferAtoms` some more.

I'm seeing some other `Map` errors when testing, but passing `Map` gives me the following:
Screenshot 2023-04-27 at 11 22 39 AM

ericdowell: Getting close using `Iterable`, but not perfect:
```ts
type InferIterableAtoms<T extends Iterable<AtomTuple>> = {
  [K in keyof T]: T[K] extends AtomTuple<infer A>
    ? A extends AnyWritableAtom
      ? AtomTuple<A, ReturnType<A['read']>>
      : T[K]
    : AtomTuple
}
export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
  values: InferIterableAtoms<T>,
  options?: Options
) {
```
### Code
Screenshot 2023-04-27 at 11 52 48 AM

### Test
Screenshot 2023-04-27 at 11 52 48 AM

src/react/utils/useHydrateAtoms.ts
```diff
@@ -4,11 +4,20 @@ import type { WritableAtom } from '../../vanilla.ts'
...
+      : T[K]
+    : never
+}
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
-export function useHydrateAtoms(
-  values: Iterable<readonly [AnyWritableAtom, unknown]>,
+export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
+  values: InferAtoms<T>,
```
ericdowell, line 20: Are there any other tests you'd like me to add for this change?

ericdowell: Added this test after tweaking type to use `any` instead of `never` (needed to continue to support `Map`):
https://github.com/pmndrs/jotai/blob/b9ba05b704ab29cdfc9ba30133668e69cb9b97b4/tests/react/utils/useHydrateAtoms.test.tsx#L8

src/react/utils/useHydrateAtoms.ts
```diff
@@ -4,11 +4,20 @@ import type { WritableAtom } from '../../vanilla.ts'
...
 type Options = Parameters<typeof useStore>[0]
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
+type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]
+
+type InferAtoms<T extends Iterable<AtomTuple>> = {
+  [K in keyof T]: T[K] extends AtomTuple<infer A>
+    ? A extends AnyWritableAtom
+      ? AtomTuple<A, ReturnType<A['read']>>
+      : T[K]
+    : any
```
ericdowell, line 14: If this library drops support for `Map` this could be `never`. Feels less safe to allow `any`, not sure what else to do for continued support for passing `Map`.

dai-shi: For now, I'd like to keep supporting `Map`s.

tests/react/utils/useHydrateAtoms.test.tsx
```diff
@@ -7,9 +7,69 @@ import { atom } from 'jotai/vanilla'
...
+  await findByText('status: fulfilled')
+})
+
+it('useHydrateAtoms should only hydrate on first render using a Map', async () => {
+  const countAtom = atom(0)
 
   const Counter = ({ initialCount }: { initialCount: number }) => {
-    useHydrateAtoms([[countAtom, initialCount]])
+    useHydrateAtoms(
+      new Map<typeof countAtom, ReturnType<(typeof countAtom)['read']>>([
```
dai-shi, line 69: Maybe we can type it more explicitly.
```suggestion
      new Map<typeof countAtom, typeof initialCount>([
```

src/react/utils/useHydrateAtoms.ts
dai-shi: I think we should add tests for type errors, with expectType or ts-expect-error.

dai-shi (approved):
Overall, it looks very good.
Thanks for your contribution!

ericdowell:
I'll work on finishing up testing and function parameter type overloading today.

tests/react/utils/types.test.tsx
```diff
@@ -0,0 +1,16 @@
+import { it } from '@jest/globals'
+import { useHydrateAtoms } from 'jotai/react/utils'
+import { atom } from 'jotai/vanilla'
+
+it('useHydrateAtoms should not allow invalid atom types when array is passed', () => {
```
ericdowell, line 5: I thought about adding something for `Map` but I wasn't able to get this type safe because of the limitations Typescript compiler has for `Map`. I'd only be adding a test for `Map` and not `useHydrateAtom`. e.g.
```tsx
it('useHydrateAtoms should not allow Map without generics passed', () => {
  function Component() {
    const countAtom = atom(0)
    const activeAtom = atom(true)

    const map = new Map([
      [countAtom, 'foo'],
      [activeAtom, 0],
    ])
    useHydrateAtoms(map)
  }
  Component
})
```
Really only test that `Map` is not happy:
Screenshot 2023-05-01 at 9 55 21 PM

Unless you can think of something else I should be adding a test for.

src/react/utils/useHydrateAtoms.ts
```diff
@@ -4,11 +4,19 @@ import type { WritableAtom } from '../../vanilla.ts'
...
+}
 
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
-export function useHydrateAtoms(
-  values: Iterable<readonly [AnyWritableAtom, unknown]>,
+export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
```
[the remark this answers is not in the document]
ericdowell: Added in commit https://github.com/pmndrs/jotai/pull/1898/commits/8619dbfebc6540f106b87ea424d79ebc89e90a79

tests/react/utils/useHydrateAtoms.test.tsx
```diff
@@ -1,19 +1,92 @@
...
+    initialCount,
+  }: {
+    initialActive?: boolean
+    initialCount: number
+  }) => {
+    useHydrateAtoms(
+      new Map<
+        typeof activeAtom | typeof countAtom,
+        typeof initialActive | typeof initialCount
+      >([
```
ericdowell, line 79: Note that these generics passed to `Map` were required before this PR as well, I tested it and added screenshots in the PR description showing BEFORE/AFTER.

src/react/utils/useHydrateAtoms.ts
ericdowell: I added one test using `ts-expect-error` in commit https://github.com/pmndrs/jotai/pull/1898/commits/6b03f946edb41d4dcb70bc9686886820c4390e6c

tests/react/utils/types.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+  function Component() {
+    const countAtom = atom(0)
+    const activeAtom = atom(true)
+    // @ts-expect-error TS2769
```
ericdowell, line 9: Actually this is the wrong `ts-expect-error` 😞, ideally the test would be:
```tsx
it('useHydrateAtoms should not allow invalid atom types when array is passed', () => {
  function Component() {
    const countAtom = atom(0)
    const activeAtom = atom(true)
    useHydrateAtoms([
      // @ts-expect-error TS2322
      [countAtom, 'foo'],
      // @ts-expect-error TS2322
      [activeAtom, 0],
    ])
    useHydrateAtoms([
      [countAtom, 1],
      // @ts-expect-error TS2322
      [activeAtom, 0],
    ])
    useHydrateAtoms([
      // @ts-expect-error TS2322
      [countAtom, true],
      [activeAtom, false],
    ])
  }
  Component
})
```
but the `Map` overloading is preventing type script from correctly picking `Array`/`Iterable` over `T extends Map`:
https://github.com/pmndrs/jotai/blob/6b03f946edb41d4dcb70bc9686886820c4390e6c/src/react/utils/useHydrateAtoms.ts#L19-L22

I'm at a loss because the above test would be the most preferred and correct.

src/react/utils/useHydrateAtoms.ts
```diff
@@ -4,11 +4,32 @@ import type { WritableAtom } from '../../vanilla.ts'
...
+  options?: Options
+): void
+export function useHydrateAtoms<T extends AtomMap>(
+  values: T,
+  options?: Options
+): void
+export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
+  values: InferAtoms<T>,
+  options?: Options
+): void
```
ericdowell, line 30: Added this because when I was trying to get Typescript to pick the correct overload by moving `export function useHydrateAtoms<T extends AtomMap>` as the function implementation signature I noticed the following error in the tests:
``
useHydrateAtoms.ts(27, 17): The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.
``
Screenshot 2023-05-01 at 10 38 25 PM

dai-shi (approved):
You covered more than I expected. Great work!

dai-shi:
Oops, we need a trick for ts 3.8.3.

ericdowell:
> Oops, we need a trick for ts 3.8.3.

I've added `@ts-ignore` as a fallback since typescript 3.8.3 doesn't support `ts-expect-error`:
https://github.com/pmndrs/jotai/pull/1898/commits/d932ccbc6c512a863f632929e48b1a83183ec3d4#diff-173f142eee55311e79b1a049ac2c8c1299d42950568d2cc1ebc6ac24bcaa5071R17-R19

dai-shi:
That's a nice workaround!

Issue #1890, opened 2023-04-21
Showcase are broken


JulioJu:
## Summary

Some shocases are broken .

## Link to reproduction

Follow README.md of

- https://github.com/pmndrs/jotai/tree/1313a9d2fd07689e30a771ad5516d3de0e91f0a8/examples/hacker_news
- https://github.com/pmndrs/jotai/tree/1313a9d2fd07689e30a771ad5516d3de0e91f0a8/examples/todos

We have error in browser
```
Failed to compile

./src/App.tsx
Attempted import error: 'useSetAtom' is not exported from 'jotai'.

This error occurred during the build time and cannot be dismissed.
```
When we download this sample from https://jotai.org/docs/basics/showcase we have the same problem. Same problem with Yarn and Npm.

I notice that is Jotai 1.x that is used in this examples.

## Check List

Please do not ask questions in issues.

- [X] I've already opened a [discussion](https://github.com/pmndrs/jotai/discussions) before opening this issue, or already discussed in other media. See https://github.com/pmndrs/jotai/discussions/1889

Please include a minimal reproduction.

- [X] I've added a link to a [typescript playground](https://www.typescriptlang.org/play) or [codesandbox](https://codesandbox.io) with a minimal reproduction.

milhamm:
Hey @dai-shi, I want to fix this issue. Should the example be updated to Jotai v2, or should I keep it and just minor upgrade the v1?

dai-shi:
I think just updating v2 latest is fine. Thanks!

PR #1903, opened 2023-04-29
fix(examples): import issue and upgrade to v2

milhamm:

Fixes https://github.com/pmndrs/jotai/issues/1890

## Summary

After trying and fixing the issue, I encountered another error when trying to start the app. To fix that issue, I added the rule to each examples.

```json
 "eslintConfig": {
    "extends": [
      "react-app",
      "react-app/jest"
    ]
  },
```

I also updated the showcases to v2 except for `todos_with_atomFamily` since I can't figure out how to do the `serializeAtom` in v2.

## Check List


dai-shi:
> I can't figure out how to do the `serializeAtom` in v2.

What's blocking?
If it's about typing, we should type for `action`.
```ts
const serializeAtom = atom(null, (get, set, action: ...) => { 
```

milhamm:
Hey @dai-shi. Thanks for the help. Yea, it was a typing issue and it's fixed.

About the linting error. Is it okay for me to add `"ignorePatterns": ["examples/*"]` to the `.eslintrc.json`?
I'm unsure how to make the examples work without the newly added rules on each `package.json` and I feel like it's safe to ignore it since its "not quite related" to the core library

dai-shi:
> I think the code sandbox example from https://jotai.org/docs/guides/persistence needs to be updated to v2 too

Yes, please fork it and update it.

> About the linting error. Is it okay for me to ignore the ./example/* path on for the linting ci scripts? `"eslint '*.{js,json}' '{src,tests,benchmarks}/**/*.{ts,tsx}' --ignore-pattern 'examples/*'"`
> 
> The issue here is that the extend "react-app" config is only available when we `yarn install` the example app

Can you do the same as https://github.com/pmndrs/zustand/pull/1754 by @arjunvegda ?

milhamm:
Thanks for the reference, I've updated it and it works fine.

milhamm:
I will create a separate PR to update the docs if that's okay for you @dai-shi.

dai-shi (approved):
Thanks for your contribution!

PR #1906, opened 2023-05-01
docs(react): iterator type on es5 target (#1900)

Himself65:

Fixes #1900

## Check List

### DIFF
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -8,7 +8,7 @@ type AnyWritableAtom = WritableAtom<unknown, any[], any>
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
 export function useHydrateAtoms(
-  values: Iterable<readonly [AnyWritableAtom, unknown]>,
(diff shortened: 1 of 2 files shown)


src/react/utils/useHydrateAtoms.ts
```diff
@@ -8,7 +8,7 @@ type AnyWritableAtom = WritableAtom<unknown, any[], any>
 const hydratedMap: WeakMap<Store, WeakSet<AnyWritableAtom>> = new WeakMap()
 
 export function useHydrateAtoms(
-  values: Iterable<readonly [AnyWritableAtom, unknown]>,
+  values: (readonly [AnyWritableAtom, unknown])[],
```
dai-shi, line 11: Does this support `Map`?

Himself65: No. I didn't realize here accept a map

dai-shi: #1898 will be adding a test with `Map`.

Himself65: Ok i will update the document instead of code

dai-shi (approved):
Thanks!!

PR #1907, opened 2023-05-01
fix(docs): Update urql.mdx

RIP21:

https://github.com/jotaijs/jotai-urql/pull/12

## Summary

Update urql documentation to align with current version

## Check List

- [] `yarn run prettier` for formatting code and docs

### DIFF
--- a/docs/integrations/urql.mdx
+++ b/docs/integrations/urql.mdx
@@ -15,127 +15,194 @@ yarn add jotai-urql @urql/core wonka
 
 ## Exported functions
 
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Nice! Thanks for updating!

RIP21:
@dai-shi only issue I see is that codesandbox example is pointing to some old sandbox. Can we point to the one that is in sync with the actual repository? I have no idea how to do that, but that would be nice.

dai-shi:
It would be nice to have a codesandbox id. Can you update it? I mean we need to manually sync with `jotai-urql`.

PR #1920, opened 2023-05-07
fix: trigger dev store listener cb when promise resolves

arjunvegda:
## Summary
We're only triggering `'state'` change callbacks for non-async atoms. This change allows the DevTools to capture the snapshot history of the state before and after the promise is resolved/rejected

**Changes**
- Trigger dev store listener callback when promise is resolved or rejected. 

## Check List

- [ ] Check with @dai-shi if there is a better place to add these callbacks
- [ ] Test it with the new time travel feature of devtools

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -245,6 +245,9 @@ export const createStore = () => {
                 if (prevAtomState?.d !== nextAtomState.d) {
                   mountDependencies(atom, nextAtomState, prevAtomState?.d)
                 }
+                if (import.meta.env?.MODE !== 'production') {
+                  storeListeners.forEach((l) => l('state'))
+                }
               }
             },
             (e) => {
@@ -261,6 +264,9 @@ export const createStore = () => {
                 reject(e)
                 if (prevAtomState?.d !== nextAtomState.d) {
                   mountDependencies(atom, nextAtomState, prevAtomState?.d)
+                }
+                if (import.meta.env?.MODE !== 'production') {
+                  storeListeners.forEach((l) => l('state'))
                 }
               }
             }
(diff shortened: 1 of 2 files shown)


dai-shi:
This change isn't very ideal, because it means the way app works and the way devtool works are different. The store has no such capability that it notifies on promise resolution/rejection. This is an intentional behavior to match how React currently works. The worst case is that devtools show something inconsistent with the app.

I'm hesitant of this change, and hope devtools work like React. It means just observe those promises.

arjunvegda:
That's true, it is a little hacky! This does not rely on suspense unfortunately 😟. The problem I'm trying to solve is how to best show the diffs between two actions that touch the atom. 

> The worst case is that devtools show something inconsistent with the app.

This is exactly what I'm trying to solve 😓 

When a user triggers a fetch call, time travel would capture that action and mark the atom as "promise pending".  Once the call is completed, the app would show the correct results, however, the time travel tab would not receive any update until the user performs another action that changes an atom. 

BTW, This issue is also present in the `useAtomsDevTool` hook. Any thoughts?

dai-shi:
> Once the call is completed, the app would show the correct results, however, the time travel tab would not receive any update

So, can't we do `promise.then(...)` to know when it's completed?

> BTW, This issue is also present in the useAtomsDevTool hook.

Ah, good point, they aren't considered when we move to v2 API...

arjunvegda:
> So, can't we do promise.then(...) to know when it's completed?

Also feels a bit hacky but it should work with a delay, I suppose 🤔. We need to wait for the store to finish recomputing the atom's dependencies and set the value to the atom, hence the delay.  

Also, I wonder if we need to worry about canceling stale promises if the user triggers multiple requests while the other ones are pending.

```ts
  atomState.v.then(() => {
    setTimeout(() => {
      cb('state');
    }, 250); // <-- this is arbitrary time, maybe we could lower it
  });
```

dai-shi:
All dependent atoms are async, so you just can get the resolved value from the promise.
I might be missing something, but all I'm saying is we should make it work like React does.

arjunvegda:
Closing this as `promise.then(...)` works!

PR #1922, opened 2023-05-08
fix(react): Add HydrateAtoms Component

ericdowell:

Fixes https://github.com/pmndrs/jotai/discussions/1921

## Summary

After updating types for `useHydrateAtoms` in PR https://github.com/pmndrs/jotai/pull/1898 it's come to light it is much harder for developers to create a `HydrateAtoms` component within their codebases without duplicating a lot of the complex types used in the `useHydrateAtoms` function overloading.

- Moved types to `src/react/utils/typeUtils.ts`
- Created and exported `src/react/utils/HydrateAtoms.tsx` components
- Added tests for `HydrateAtoms` component.

## Check List


src/react/utils.ts
```diff
@@ -1,3 +1,4 @@
+export { HydrateAtoms } from './utils/HydrateAtoms.tsx'
```
ericdowell, line 1: I didn't export the types from `src/react/utils/typeUtils.ts` to be used externally, but if we think some some be I would suggest they be:
```ts
export {
  AnyWritableAtom,
  AtomMap,
  AtomTuple,
  InferAtoms,
} from './utils/typeUtils.ts'
```

dai-shi:
> After updating types for `useHydrateAtoms` in PR #1898 it's come to light it is much harder for developers to create a `HydrateAtoms` component within their codebases without duplicating a lot of the complex types used in the `useHydrateAtoms` function overloading.

Hmmm, that's very unfortunate. Is there any workaround?

> * Moved types to `src/react/utils/typeUtils.ts`

We are strongly against doing that. Because, right now, there's no way to prohibit using those internal types by lib users. If you look at other files, most types are defined locally in a module.

> * Created and exported `src/react/utils/HydrateAtoms.tsx` components

I've discussed about it with @sandren before, and concluded providing two similar functions is something we should avoid.

I consider the issue is #1898, and exporting `HydrateAtoms` only for that is a bad workaround.

ericdowell:
> > After updating types for `useHydrateAtoms` in PR #1898 it's come to light it is much harder for developers to create a `HydrateAtoms` component within their codebases without duplicating a lot of the complex types used in the `useHydrateAtoms` function overloading.
> 
> Hmmm, that's very unfortunate. Is there any workaround?

I'll look more closely at what https://github.com/pmndrs/jotai/discussions/1921 is trying to solve and think on possible typescript solution they can use for external usages.

> 
> > * Moved types to `src/react/utils/typeUtils.ts`
> 
> We are strongly against doing that. Because, right now, there's no way to prohibit using those internal types by lib users. If you look at other files, most types are defined locally in a module.

Good to know. This makes sense.

> > * Created and exported `src/react/utils/HydrateAtoms.tsx` components
> 
> I've discussed about it with @sandren before, and concluded providing two similar functions is something we should avoid.
> 
> I consider the issue is #1898, and exporting `HydrateAtoms` only for that is a bad workaround.

`concluded providing two similar functions is something we should avoid.`

I can agree with this, better to just have one function rather than two.

ericdowell:
Based on some testing locally I believe my answer https://github.com/pmndrs/jotai/discussions/1921#discussioncomment-5833973 should solve the persons issue

PR #1933, opened 2023-05-13
fix(vanilla): Stable promise

backbone87:

Fixes #1925

## Summary

Do not recreate the cancelable promise when the promise did not change

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -69,7 +69,7 @@ type NextDependencies = Map<AnyAtom, AtomState | undefined>
  */
 type AtomState<Value = AnyValue> = {
   d: Dependencies
-} & ({ e: AnyError } | { v: Value })
+} & ({ e: AnyError; o?: never } | { v: Value; o: Promise<Value> | undefined })
 
 const isEqualAtomValue = <Value>(a: AtomState<Value>, b: AtomState<Value>) =>
   'v' in a && 'v' in b && Object.is(a.v, b.v)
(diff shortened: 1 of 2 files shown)


src/vanilla/store.ts
```diff
@@ -225,6 +227,16 @@ export const createStore = () => {
     abortPromise?: () => void
   ): AtomState<Value> => {
     if (valueOrPromise instanceof Promise) {
+      const prevAtomState = getAtomState(atom)
+      if (prevAtomState?.o === valueOrPromise) {
+        return setAtomValue(
+          atom,
+          prevAtomState.v as Value,
+          nextDependencies,
```
dai-shi, line 235: This looks bad (it clears up dependencies). I'll think about the solution.

dai-shi (commented):
Thanks for your work. I will continue working on it later.

tests/vanilla/store.test.tsx
```diff
@@ -225,3 +225,45 @@ it('should update async atom with deps after await (#1905)', async () => {
   expect(await lastValue).toBe(3)
   unsub()
 })
+
+it('should not fire subscription when async atom promise is the same', async () => {
```
dai-shi, line 229: I wonder if it should simply bail out updating. If it wouldn't be sufficient, would you please add more tests to fail?

dai-shi: Maybe something like this:
```js
const promise = Promise.resolve()
const countAtom = atom(0)
const derivedAtom = atom((get) => {
  get(countAtom)
  return promise
})
```

src/vanilla/store.ts
backbone87: The dependencies could have changed even if the promise is the same

dai-shi: If dependencies are changed, we need to set it after the promise resolution, no?

tests/vanilla/store.test.tsx
dai-shi: After reading the comment https://github.com/pmndrs/jotai/pull/1933#discussion_r1193074433,
I don't think we can support the above case, or we shouldn't as dependencies could have been changed.

src/vanilla/store.ts
backbone87: The async deps most likely can not change (they can with some creativity, but let's consider this out of scope for now). But the sync deps can change

dai-shi: async deps include sync deps, so they can change. I'm afraid this isn't solvable easily.

dai-shi: Added some commits. Hope it works.

dai-shi (approved):
This should be good. I'll merge it later.

PR #1939, opened 2023-05-17
breaking(utils): predictable atomWithDefault

### LINKED ISSUE #1937: Weird behaviours when using atomWithDefault with async atom
Following the discussion https://github.com/pmndrs/jotai/discussions/1935
In this piece of code, every time we increase `field1` it renders the whole component (it starts by displaying `LOADING` then the components).

Updating the `atomWithDefault` first display the fallback component thus reload the entire component...

```ts
const anAtom = atom((get) =>
  Promise.resolve({
    field1: 1,
    field2: 1
  })
);

export const testAtom = atomWithDefault(async (get) => {
  const res = await get(anAtom);
  return res;
});

const field1Atom = selectAtom(testAtom, (p) => p.field1);
const field2Atom = selectAtom(testAtom, (p) => p.field2);

function Field1Display() {
  const field1 = useAtomValue(field1Atom);
  console.log('RENDER DISPLAY 1')
  return <div>VALUE FIELD 1 : {field1}</div>;
}

function Field2Display() {
  const field2 = useAtomValue(field2Atom);
  console.log('RENDER DISPLAY 2')
  return <div>VALUE FIELD 2 : {field2}</div>;
}

function IncreaseButtons() {
  const setTestAtom = useSetAtom(testAtom);
  return (
    <>
      <button
        onClick={() =>
          setTestAtom((prev) => ({ ...prev, field1: prev.field1 + 1 }))
        }
      >
        Increase field 1
      </button>
      <button
        onClick={() =>
          setTestAtom((prev) => ({ ...prev, field2: prev.field2 + 1 }))
        }
      >
        Increase field 2
      </button>
    </>
  );
}
const App = () => (
  <React.Suspense fallback={<div>LOADING</div>}>
    <Field1Display />
    <Field2Display />
    <IncreaseButtons />
  </React.Suspense>
);
```

Here is the sandbox to reproduce : 
https://codesandbox.io/s/with-jotai-2-forked-leejy2?file=/src/App.js

dai-shi:

Fixes #1937 

## Summary

As it turns out, the conversion of atomWithDefault from jotai v1 to jotai v2 wasn't very ideal. It's unnecessary complex and the behavior isn't very predictable.

This fixes it. However, it comes with breaking change.

## Migration guide

```js
// suppose we have this
const asyncAtom = atom(() => Promise.resolve(1))
const countAtom = atomWithDefault((get) => get(asyncAtom))
// and in component
const setCount = useSetAtom(countAtom)

// previously,
setCount((c) => c + 1) // it worked, but it will no longer work

// instead, you need to do this
setCount((countPromise) => countPromise.then((c) => c + 1))
```

## Check List


rikisamurai:
I encountered some confusion regarding a breaking change while upgrading from version 2.1.x. Does this imply that if atomWithDefault is initially asynchronous, it will remain asynchronous consistently?

I find the ability of atomWithDefault to transition from asynchronous to synchronous quite impressive. However, the current behavior seems weird:
```js
const asyncAtom = atomWithDefault(async () => {
  await sleep(1000);
  return "async";
});

const setAsyncAtom = atom(null, async (get, set) => {
  await sleep(1000);
  set(asyncAtom, Promise.resolve("update"));
});
```
here is codesandbox: https://codesandbox.io/p/sandbox/async-atom-and-update-hdp67v?file=%2Fsrc%2FApp.tsx%3A9%2C4

before this change, i could write in this way:
```js
const asyncAtom = atomWithDefault(async () => {
  await sleep(1000);
  return "async";
});

const setAsyncAtom = atom(null, async (get, set) => {
  await sleep(1000);
  set(asyncAtom, "update");
});
```
Did i do something wrong?

dai-shi:
> Does this imply that if atomWithDefault is initially asynchronous, it will remain asynchronous consistently?

Yes, and it's how Jotai v2 is designed for all async atoms, not specific only for atomWithDefault. And, `unwrap` is a solution to convert an async atom to sync atom.
You can even use `unwrap` conditionally. https://fosstodon.org/@daishi/111379830300795523

Though, I think you can set a non-promise value.

```ts
const anAtom = atomWithDefault<Promise<number> | number>(() => Promise.resolve(1));
  const set = useSetAtom(anAtom);
  set(2);
```

I haven't looked your codesandbox, but is it a type only issue?

rikisamurai:
> > Does this imply that if atomWithDefault is initially asynchronous, it will remain asynchronous consistently?
> 
> Yes, and it's how Jotai v2 is designed for all async atoms, not specific only for atomWithDefault. And, `unwrap` is a solution to convert an async atom to sync atom. You can even use `unwrap` conditionally. https://fosstodon.org/@daishi/111379830300795523
> 
> Though, I think you can set a non-promise value.
> 
> ```ts
> const anAtom = atomWithDefault<Promise<number> | number>(() => Promise.resolve(1));
>   const set = useSetAtom(anAtom);
>   set(2);
> ```
> 
> I haven't looked your codesandbox, but is it a type only issue?

Yes, i can write like this ```set(asyncAtom, "update");``` , but it will emit a typescript error.  Is this a type issue that needs to be fixed?

dai-shi:
https://tsplay.dev/wjlZlm

image

I don't see any type errors. Please make a reproduction with modifying the typescript playground and open a new discussion.

rikisamurai:
> https://tsplay.dev/wjlZlm
> 
> image
> I don't see any type errors. Please make a reproduction with modifying the typescript playground and open a new discussion.

Thank you for your response! Now I understand that I can solve this problem using TypeScript generics.
Before this pr:
``` js
const asyncAtom = atomWithDefault(async () => {
  await sleep(1000);
  return 'async';
});

const setSyncAtom = atom(null, async (_, set) => {
  await sleep(1000);
  set(asyncAtom, 'update');
});
```
Now i should add generics for compatiblity
``` js
const asyncAtom = atomWithDefault<<Promise<string> | string>>(async () => {
  await sleep(1000);
  return 'async';
});

const setSyncAtom = atom(null, async (_, set) => {
  await sleep(1000);
  set(asyncAtom, 'update');
});
```
🙏

PR #1958, opened 2023-05-26
breaking(utils): improve atomWithStorage

dai-shi:

#1879 #1784 #1737 #1803 #1791 #1968

## Summary

It's been requested quite some times since we dropped v1 `delayInit` option and make it default in v2. There was certain difficulties in v1, and that led us to the breaking change in v2. Recent discussion with @arjunvegda revealed that we can implement it differently in v2 and it's fairly maintainable. (I originally thought we would need two utils for async storage and sync storage separately.)

Tasks:
- [x] implementation
- [x] tests
- [x] docs
- [x] migration guide

## Migration guide

```js
// suppose we have async storage
const storage = createJSONStorage(() => AsyncStorage)
const countAtom = atomWithStorage('count-key', 0, storage)
  // in component
  const [count, setCount] = useAtom(countAom)

  // previously, countAtom is a sync atom, so you could update like this:
  setCount((c) => c + 1)

  // with the new version, it becomes async occasionally, so you need to resolve it:
  setCount(async (c) => (await c) + 1)
```

## unstable_options

This PR also introduces new unstable option `unstable_getOnInit`. It's the opposite of `delayInit` in v1.
We will make it stable after getting more feedback and remove `unstable_` prefix in the future version.


src/vanilla/utils/atomWithStorage.ts
```diff
@@ -13,6 +13,7 @@ export interface AsyncStorage<Value> {
   getItem: (key: string, initialValue: Value) => Promise<Value>
   setItem: (key: string, newValue: Value) => Promise<void>
   removeItem: (key: string) => Promise<void>
+  getItemOnInit?: boolean
```
dai-shi, line 16: I'm open for changing this property name.
It would be helpful if `createJSONStorage` takes it as an option.

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -114,25 +116,40 @@ const defaultStorage = createJSONStorage(() =>
...
+
 export function atomWithStorage<Value>(
   key: string,
   initialValue: Value,
   storage:
     | SyncStorage<Value>
     | AsyncStorage<Value> = defaultStorage as SyncStorage<Value>
-): WritableAtom<Value, [SetStateActionWithReset<Value>], void> {
-  const baseAtom = atom(initialValue)
+): any {
```
dai-shi, line 141: `any` is unfortunate here, but can't be helped as we overload functions and TS can't understand it.

HarunKilic:
Just tested it out. And can confirm that it is working.

docs/utilities/storage.mdx
```diff
@@ -116,6 +116,21 @@ const content = {} // anything JSON serializable
 const storedAtom = atomWithStorage('stored-key', content, storage)
 ```
 
+#### Notes with AsyncStorage (since v2.2.0)
```
dai-shi, line 119: It's going to be v2.2.0. (If it's not, let's update.)

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -143,15 +162,25 @@ export function atomWithStorage<Value>(
...
     (get) => get(baseAtom),
-    (get, set, update: SetStateActionWithReset<Value>) => {
+    (get, set, update: SetStateActionWithReset<Promise<Value> | Value>) => {
       const nextValue =
         typeof update === 'function'
-          ? (update as (prev: Value) => Value | typeof RESET)(get(baseAtom))
+          ? (
+              update as (
+                prev: Promise<Value> | Value
+              ) => Promise<Value> | Value | typeof RESET
```
dai-shi, line 171: I hope TS can infer this soon...

dai-shi (commented):
Just some minor comments.

arjunvegda (approved):
Nice stuff! LGTM!

PR #1965, opened 2023-05-29
fix: avoid triggering state change call when restoring state

arjunvegda:
## Summary
- Skip triggering `state` change callback when restoring the state

## Check List

- [ ] Get Daishi's feedback
- [ ] Add tests

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -109,7 +109,7 @@ type Mounted = {
 }
 
 // for debugging purpose only
-type StoreListener = (type: 'state' | 'sub' | 'unsub') => void
+type StoreListener = (type: 'state' | 'sub' | 'unsub' | 'restore') => void
 type MountedAtoms = Set<AnyAtom>
 
 /**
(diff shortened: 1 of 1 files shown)


dai-shi:
> Skip triggering state change callback when restoring the state

Hi, I'm curious the reason behind it. Is it because your devtools initiates `restore`, or someone does it just in case? If you initiate it, you know it so you can mute your listeners.

arjunvegda:
> Is it because your devtools initiates `restore`

That is precisely it! 

> If you initiate it, you know it so you can mute your listeners

Yeah, I tried taking that route. However, I'm not sure how to unmute them back without relying on if the value changed 😅 

So I figured that it would be nice to have `dev_subscribe_store` differentiate between different types of changes

dai-shi:
(Actually, restore is v2 isn't a real "restore" which was the case in v1. it's just bulk write. "write" can be async... I'll see what I can do in #1966)

dai-shi:
Closing in favor of #1966.

PR #1990, opened 2023-06-14
feat(utils/useHydrateAtoms) - Optionally rehydrate with force hydrate

SariSabouh:

The useHydrateAtom discussion has been live since August of 2021 in https://github.com/pmndrs/jotai/discussions/669.

## Summary

We found a use case within our project that required a rehydration after a specific event. We had to manually rely on conditional useEffect actions to manually update the store rather than hydrating back from a fully reset SSR page. After review, a forceHydration option meets that requirement and a lot of the cases mentioned within that option discussion.

## Check List

- [X] `yarn run prettier` for formatting code and docs

### DIFF
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -3,7 +3,7 @@ import type { WritableAtom } from '../../vanilla.ts'
 
 type Store = ReturnType<typeof useStore>
 type Options = Parameters<typeof useStore>[0] & {
-  forceHydrate?: boolean
+  dangerouslyForceHydrate?: boolean
 }
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
 type AtomMap<A = AnyWritableAtom, V = unknown> = Map<A, V>
@@ -38,12 +38,7 @@ export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
 
   const hydratedSet = getHydratedSet(store)
   for (const [atom, value] of values) {
-    const isHydratedAtom = hydratedSet.has(atom)
-    if (!isHydratedAtom || options?.forceHydrate) {
-      if (options?.forceHydrate && isHydratedAtom) {
-        hydratedSet.delete(atom)
-      }
-
+    if (!hydratedSet.has(atom) || options?.dangerouslyForceHydrate) {
       hydratedSet.add(atom)
       store.set(atom, value)
     }
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -2,7 +2,9 @@ import { useStore } from '../../react.ts'
 import type { WritableAtom } from '../../vanilla.ts'
 
 type Store = ReturnType<typeof useStore>
-type Options = Parameters<typeof useStore>[0]
+type Options = Parameters<typeof useStore>[0] & {
+  forceHydrate?: boolean
+}
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
 type AtomMap<A = AnyWritableAtom, V = unknown> = Map<A, V>
 type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]
@@ -36,7 +38,12 @@ export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
 
   const hydratedSet = getHydratedSet(store)
   for (const [atom, value] of values) {
-    if (!hydratedSet.has(atom)) {
+    const isHydratedAtom = hydratedSet.has(atom)
+    if (!isHydratedAtom || options?.forceHydrate) {
+      if (options?.forceHydrate && isHydratedAtom) {
+        hydratedSet.delete(atom)
+      }
+
       hydratedSet.add(atom)
       store.set(atom, value)
     }
--- a/tests/react/utils/useHydrateAtoms.test.tsx
+++ b/tests/react/utils/useHydrateAtoms.test.tsx
@@ -241,7 +241,7 @@ it('useHydrateAtoms should respect onMount', async () => {
   expect(onMountFn).toHaveBeenCalledTimes(1)
 })
 
-it.only('passing dangerouslyForceHydrate to useHydrateAtoms will re-hydrated atoms', async () => {
+it('passing dangerouslyForceHydrate to useHydrateAtoms will re-hydrated atoms', async () => {
   const countAtom = atom(0)
   const statusAtom = atom('fulfilled')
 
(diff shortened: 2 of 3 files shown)


src/react/utils/useHydrateAtoms.ts
```diff
@@ -36,7 +38,12 @@ export function useHydrateAtoms<T extends Iterable<AtomTuple>>(
 
   const hydratedSet = getHydratedSet(store)
   for (const [atom, value] of values) {
-    if (!hydratedSet.has(atom)) {
+    const isHydratedAtom = hydratedSet.has(atom)
+    if (!isHydratedAtom || options?.forceHydrate) {
+      if (options?.forceHydrate && isHydratedAtom) {
+        hydratedSet.delete(atom)
+      }
+
```
dai-shi, line 46: Can it be as simple as this?
```suggestion
    if (!hydratedSet.has(atom) || options?.dangerouslyForceHydrate) {
```

dai-shi (changes requested):
Thanks for your suggestion. Added some comments.

src/react/utils/useHydrateAtoms.ts
```diff
@@ -2,7 +2,9 @@ import { useStore } from '../../react.ts'
 import type { WritableAtom } from '../../vanilla.ts'
 
 type Store = ReturnType<typeof useStore>
-type Options = Parameters<typeof useStore>[0]
+type Options = Parameters<typeof useStore>[0] & {
+  forceHydrate?: boolean
```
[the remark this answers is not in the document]
SariSabouh: That's a great catch! Will definitely update 👍

SariSabouh:
Added all requested changes @dai-shi

tests/react/utils/useHydrateAtoms.test.tsx
```diff
@@ -240,3 +240,78 @@ it('useHydrateAtoms should respect onMount', async () => {
   await findByText('count: 42')
   expect(onMountFn).toHaveBeenCalledTimes(1)
 })
+
+it.only('passing dangerouslyForceHydrate to useHydrateAtoms will re-hydrated atoms', async () => {
```
dai-shi, line 244:
```suggestion
it('passing dangerouslyForceHydrate to useHydrateAtoms will re-hydrated atoms', async () => {
```

tests/react/utils/useHydrateAtoms.test.tsx
```diff
@@ -241,7 +241,7 @@ it('useHydrateAtoms should respect onMount', async () => {
   expect(onMountFn).toHaveBeenCalledTimes(1)
 })
 
-it.only('passing dangerouslyForceHydrate to useHydrateAtoms will re-hydrated atoms', async () => {
+it('passing dangerouslyForceHydrate to useHydrateAtoms will re-hydrated atoms', async () => {
```
SariSabouh, line 244: Great catch, forgot to remove while testing :)

dai-shi:
On second thought, we could do and suggest to use store api for such use cases.
```js
const Component = () => {
  const store = useStore()
  store.set(countAtom, countValue)
  // ...
}
```

What do you think?

SariSabouh:
> On second thought, we could do and suggest to use store api for such use cases.
> 
> ```js
> const Component = () => {
>   const store = useStore()
>   store.set(countAtom, countValue)
>   // ...
> }
> ```
> 
> What do you think?

Basically disregard the forceHydrate and suggest using the store API? If yes, that could technically fully replace the `useHydrateAtom` though and may cause many re-renders in some cases, especially if its an object compare.

dai-shi:
Basically disregard force hydrate, but store api wouldn't fully replace useHydreteAtoms which ensures only setting values if unset.

> may cause many re-renders in some cases

I'm not sure if I get those cases.

SariSabouh:
> Basically disregard force hydrate, but store api wouldn't fully replace useHydreteAtoms which ensures only setting values if unset.
> 
> > may cause many re-renders in some cases
> 
> I'm not sure if I get those cases.

I can try to re-produce it again but I remember going for the `forceHydrate` because useStore was was causing a re-render. Is there a reason you're opposed to going with some sort of re-hydrate method? I was thinking we can integrate atomsWithReset somehow instead.

dai-shi:
The reason to be against `dangerouslyForceHydrate` (the name is actually good to warn the usage) is that it bypasses the hydration check, and the result is just equal to `store.set(...)`. It means we provide two ways for the same capability which we carefully avoid in Jotai API design.

SariSabouh:
So you are suggesting that we do something like this:

```typescript
export const Component = ({
  componentPreferences,
}: ComponentPageProps) => {
  useHydrateAtoms([[componentPreferencesAtom, componentPreferences]])
  const store = useStore()
  componentPreferencesValue.reset &&
    store.set(componentPreferencesAtom, componentPreferences)
```

dai-shi:
I'm not sure if I understand the purpose of "reset", but my suggestion was if you use `store.set()` you don't use useHydrateAtoms.

Ah, I see you want to force hydrate if reset is true, and otherwise, only hydrate initially.
Hm, that use case supports this PR's change.
My suggestion to go against this PR would require `store.has()` method which doesn't exist.

SariSabouh:
I'm open to ideas. The issue is that we hydrate a page with data that is needed until a certain action completion. Once this action is completed, we want to fully reset that atom and get new data from SSR on next time this page loads.

dai-shi:
I feel like merging this PR as the code is clean and it has `dangerously` prefix to warn the usage.
After all, it is in `jotai/utils` which is a little less strict about API design.

> we want to fully reset that atom and get new data from SSR on next time this page loads.

I'm curious about your case. When you get new data from SSR on next time, how would the store be kept in memory (assumed so because you want to fully reset) ?

dai-shi:
> and completes a certain action

Hmm? If this is a user action, we should update atoms at the same time. Updating atoms in render is something we should really avoid. It's not just bad practice, but theoretically delayed.

SariSabouh:
So once the user completes the action it takes him out of the page. And they won't get the newest set of data until they reach that page.

Like once you complete checkout, you are taken out of checkout and your checkout data is reset. However, once you come back from checkout, you need the newest set of data from the server to be hydrated again. Client side won't know the data until they reach checkout, which is SSR.

dai-shi:
I'm not exactly sure how the second SSR&hydration work, but it sounds like you can only get the new data from props. In that case, the render time updating wouldn't be avoidable. Can't be helped because you don't have control, right? Sounds fair enough, because we usually suggest useEffect for sync with props, but that's far from ideal.

dai-shi:
I think there should be a better way, probably atoms-in-atom or jotai-molecules, and we want to educate such ones for common use cases.

This PR change helps as an escape hatch for edge cases, so I want to merge, but we shouldn't teach people to use it for common cases.

dai-shi (approved):
Thanks for the discussion. It was convincing. I will merge it later.

SariSabouh:
> I think there should be a better way, probably atoms-in-atom or jotai-molecules, and we want to educate such ones for common use cases.
> 
> This PR change helps as an escape hatch for edge cases, so I want to merge, but we shouldn't teach people to use it for common cases.

Haven't messed with jotai-molecules yet, but definitely will give them a shot!

PR #2017, opened 2023-07-04
docs(core): add a section about `signal`

dai-shi:
Restored the abortableAtom docs in v1 which is removed in #1685.


wootra:
I love this change! thanks for the simple and the gtreat change!

PR #2020, opened 2023-07-04
Update family.mdx

lveillard:
Adding pattern for atom-in-atom alternative

https://github.com/pmndrs/jotai/discussions/1942#discussioncomment-6336944

### DIFF
--- a/docs/utilities/family.mdx
+++ b/docs/utilities/family.mdx
@@ -75,6 +75,20 @@ There are two ways to remove params.
   - shouldRemove is a function that takes two arguments `createdAt` in milliseconds and `param`, and returns a boolean value.
   - setting `null` will remove the previously registered function.
 
+### Workaround: Atom-in-atom
+
+You may not need an atomFamily if you want to use params to retrieve a specific subsection of a larger atom. You can create an atom in your component using the following pattern:
+
+```js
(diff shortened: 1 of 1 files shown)


docs/utilities/family.mdx
```diff
@@ -75,6 +75,20 @@ There are two ways to remove params.
   - shouldRemove is a function that takes two arguments `createdAt` in milliseconds and `param`, and returns a boolean value.
   - setting `null` will remove the previously registered function.
 
+### Workaround: Atom-in-atom
```
dai-shi, line 78: Can we call something like an alternative approach instead of a workaround?

```suggestion
### Alternative approach: [Atom-in-atom](../guides/atoms-in-atom.mdx)
```

dai-shi: Wait, it's not an atoms-in-atom pattern. It's simply a big atom and deriving atoms. Close to https://github.com/pmndrs/jotai/blob/main/docs/recipes/large-objects.mdx

```suggestion
### Alternative approach: A big atom and slicing atoms
```
(Not sure if slicing is the right word here.)

dai-shi (commented):
Thanks for your work. Added some comments.

Issue #1956, opened 2023-05-25
Unexpected behavior when subscribing to derived atoms in a specific order


awaw00:
I've encountered a bug in Jotai when subscribing to derived atoms in a specific order. The issue arises when a derived atom's response function doesn't execute as expected due to the order of atom subscription.

Here's a code snippet to reproduce the issue:

```jsx
import { createStore, atom, useAtomValue, useSetAtom } from 'jotai';

const valueAtom = atom(0);
valueAtom.debugLabel = 'value'

const dep1Atom = atom((get) => get(valueAtom) * 2);
dep1Atom.debugLabel = 'dep1'

const dep2Atom = atom((get) => {
  return `${get(valueAtom)}/${get(dep1Atom)}`;
});
dep2Atom.debugLabel = 'dep2'

const dep3Atom = atom((get) => {
  return `${get(dep1Atom)}`;
})
dep3Atom.debugLabel = 'dep3'

const store = createStore();

const logAtom = (a) => {
  const val = store.get(a);

  const line = document.createElement('div');
  line.innerText = `${a.debugLabel}: ${val}`;
  document.body.appendChild(line);
}

store.sub(dep2Atom, () => logAtom(dep2Atom));
store.sub(dep1Atom, () => logAtom(dep1Atom));
store.sub(valueAtom, () => logAtom(valueAtom));
store.sub(dep3Atom, () => logAtom(dep3Atom));

const btn = document.createElement('button');
btn.innerText = 'add';
btn.onclick = () => store.set(valueAtom, store.get(valueAtom) + 1);
document.body.appendChild(btn);
```

In the above code, atoms are subscribed in the order of dep2, dep1, value, and dep3. When a new value is written to valueAtom, the response function of dep3 is not executed. This is because during the process of handling the write to valueAtom, dep2 is processed first due to the order of subscription. While processing dep2, the value of dep1 (a dependency of dep2) is also updated via readAtomState, causing the value of dep2 to be updated prematurely.

Because the recomputeDependents method for dep1 is not executed until after this update, the comparison between prevAtomState and nextAtomState for dep1 fails, leading to dep3 (which only depends on dep1) not being updated.

This behavior seems to be inconsistent with the expected behavior. I would appreciate it if you could look into this issue. Thank you.

----
Reproduction with useAtomValue can see https://codesandbox.io/s/throbbing-fire-5fq563?file=/src/App.tsx

dai-shi:
Hi, thanks for reporting.
[...]

PR #2029, opened 2023-07-07
refactor(vanilla): simplify store impl

dai-shi:
This is to fix the hack in #2016.
~As we update state wisely in #1956, we should be able to skip things.~
~Hopefully, this will improve the issue in #1901.~


src/vanilla/store.ts
```diff
@@ -355,35 +355,24 @@ export const createStore = () => {
...
-            (aState &&
-              !hasPromiseAtomValue(aState) &&
-              isEqualAtomValue(aState, s))
-          )
-        })
+        Array.from(atomState.d).every(
+          ([a, s]) => a === atom || readAtomState(a) === s
+        )
       ) {
         return atomState
```
dai-shi, line 377: I hope to remove this for #1901, but this breaks various behavior including async atoms on mount. Hope to revisit it in the far future.

src/vanilla/store.ts
```diff
@@ -601,6 +581,15 @@ export const createStore = () => {
...
     readAtomState(atom)
+    // mount self
+    const mounted: Mounted = {
+      t: new Set(initialDependent && [initialDependent]),
+      l: new Set(),
+    }
+    mountedMap.set(atom, mounted)
+    if (import.meta.env?.MODE !== 'production') {
+      mountedAtoms.add(atom)
+    }
```
dai-shi, line 592: It feels like our `mountAtom` has room for improvement (like propagating `changed` from dependencies).
It might be related with the previous comment, but not sure.

dai-shi (commented):
I think this PR is okay to merge as is.

PR #2078, opened 2023-08-16
fix(vanilla): should only mount dependencies if the atom is mounted

### LINKED ISSUE #2076: Read function of atom keeps running even after the component unmounts
## Summary
I'm running into a really weird issue where each time a component mounts and creates a new atom via useMemo(), I can see that the read function keeps running for atoms that should have been unmounted already

If you look at the code from the codesandbox, everytime the component unmounts, i can see that it will log "Unmounting atom 1" etc.

But when the component remounts, I can see it still logging running atom read function 1 and also running atom read function 2, and it keeps going from like 1-N where N is the amount of times the component has mounted and unmounted. This is causing some issues in our app where it slows down overtime as the amount of work jotai is doing seems to be increasing 

I know in this code is a bit trivial in that we didn't need to create the atom in a useMemo (can just define it outside of the scope) but sometimes we do need to pass parameters to the atoms

Link to conversation on discord: https://discord.com/channels/740090768164651008/752920140483526677/1140738335640408074

## Link to reproduction
https://codesandbox.io/s/upbeat-noether-xf8j7c?file=/src/App.tsx

## Check List

### LINKED ISSUE #2077: Accessing derived atoms from a write only atom can lead to atoms that never get cleaned up and degrade performance over time
## Summary
If a write only atom reads from a derived atom and that derived atom doesn't have a subscription via useAtomValue then that derived atom stays mounted and doesn't get cleaned up. This leads to an increasing number of atoms and starts to slow down `recomputeDependents` after a while

## Link to reproduction
https://codesandbox.io/s/xenodochial-worker-zwnspm?file=/src/App.tsx

Clicking the initialize button here will increase the mounted.t.size in vanilla.js (line 343) by 1 each time. Looks like onMount and unmount is never triggered though (which makes sense since there's no subscription to that new atom)

I debugged this by putting a breakpoint on vanilla.js on line 343 with this condition `mounted != null && mounted.t.size > 0`

## Link to convo
https://discord.com/channels/740090768164651008/752920140483526677/1141131076115378257

## Check List

dai-shi:

Fix #2076
Fix #2077 

## Summary

- should not call read function for unmounted atoms in StrictMode #2076
- should not mount atoms unnecessarily #2077

## Check List


tests/react/async2.test.tsx
```diff
@@ -141,42 +141,3 @@ describe('atom read function setSelf option test', () => {
     await findByText('text: hello1')
   })
 })
-
-describe('new async atom dependency behavior in v2', () => {
-  it('should override promise with a dependency', async () => {
```
dai-shi, line 146: This no longer is valid. I thought it's a good addition, but turns out it has worked because of the buggy logic.

PR #2083, opened 2023-08-20
docs: split atom creators in multiple files

### LINKED ISSUE #2082: Split atom creators recipe in docs into multiple markdown files
It's a single file for now and it's fairly big.

gerzonc:

Closes #2082

## Summary

Atom creators file in docs is too big so I just split it in a file per atom creator.

## Check List


gerzonc:
@dai-shi the other way I thought of doing it is that we can keep it as it was (just atom-creators.mdx) and import all the creators as components, e.g.

```
---
title: Atom creators
nav: 8.02
keywords: creators
---

<AtomWithToggle />
```

I did it this way first, but let me know which one you prefer since I already had both solutions in mind 🙈

dai-shi:
I think multiple files are better for search.

Maybe we should consider about the sidebar labels.
image

@sandren Do you have any idea?

docs/recipes/atom-with-storage.mdx
```diff
@@ -0,0 +1,44 @@
...
+  return derivedAtom
+}
+```
+
+And how it's used:
+
+```js
+import { atomWithToggleAndStorage } from 'XXX'
+
+// will have an initial value set to false & get stored in localStorage under the key "isActive"
```
himself65, line 40: set to false? I don't see the initial value on your code

docs/recipes/atom-with-listeners.mdx
```diff
@@ -0,0 +1,81 @@
...
+import { atom, useSetAtom, Getter, Setter, SetStateAction } from 'jotai'
+
+type Callback<Value> = (
+  get: Getter,
+  set: Setter,
+  newVal: Value,
+  prevVal: Value
+) => void
+
+export function atomWithListeners<Value>(initialValue: Value) {
```
himself65, line 29: I think `useCurrentStore` (non-exist API) and `store.sub(aAtom)` would be better

dai-shi: This recipe is written for v1 API, I guess. `useStore` exists, btw.

docs/recipes/atom-with-storage.mdx
dai-shi: I think L19 should be `initialValue = false,`.

dai-shi (commented):
This PR is to split markdown files. We should probably open a new PR for improving the contents.

gerzonc:
> This PR is to split markdown files. We should probably open a new PR for improving the contents.

I can do both if you don't mind. I'm still waiting for approval of current changes so I can open a new PR in order to update the docs

sandren:
> @sandren Do you have any idea?

Yes, let's change the frontmatter title of each of the new docs to reflect the recipe function name instead. For example, `atomWithToggle` instead of  `Atom creator with toggle`. Once that is done, we can merge. Thanks, @gerzonc! 🙂

sandren:
> Is ready now!

Merged. Thank you so much! 🙂

docs/recipes/atom-creators.mdx
```diff
@@ -1,521 +0,0 @@
...
-      {/* Clicking this button twice will only trigger one render */}
-      <button onClick={() => setStyles({ ...styles, fontSize: 32 })}>
-        Enlarge font
-      </button>
-    </div>
-  )
-}
-```
-
-## atomWithRefresh
```
dai-shi, line 160: @gerzonc It seems like this section is lost. Can you double check please?

gerzonc: True OMG 🥺 I'll add it right away!

PR #2088, opened 2023-08-22
fix(utils): Do not set a splitted atom when not actually modified

arnodb:

Fixes https://github.com/pmndrs/jotai/discussions/2079

## Summary

As agreed during the discussion, `set` should not be called when an item remains unmodified.

## Check List

### DIFF
--- a/src/vanilla/utils/splitAtom.ts
+++ b/src/vanilla/utils/splitAtom.ts
@@ -111,11 +111,13 @@ export function splitAtom<Item, Key>(
             const nextItem = isFunction(update)
               ? update(arr[index] as Item)
               : update
-            set(arrAtom as WritableAtom<Item[], [Item[]], void>, [
-              ...arr.slice(0, index),
-              nextItem,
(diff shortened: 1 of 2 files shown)


dai-shi (commented):
Please check CI errors.

src/vanilla/utils/splitAtom.ts
```diff
@@ -111,11 +111,13 @@ export function splitAtom<Item, Key>(
             const nextItem = isFunction(update)
               ? update(arr[index] as Item)
               : update
-            set(arrAtom as WritableAtom<Item[], [Item[]], void>, [
-              ...arr.slice(0, index),
-              nextItem,
-              ...arr.slice(index + 1),
-            ])
+            if (!Object.is(nextItem, arr[index])) {
```
dai-shi, line 114: Just a minor preference:
```suggestion
            if (!Object.is(arr[index], nextItem)) {
```

dai-shi (approved):
Thanks for your contribution!

PR #2099, opened 2023-08-29
fix(vanilla): should recompute state after onMount

### LINKED ISSUE #2098: atomWithStorage issue since 2.3.0+
This is a tracking issue for #2095. See discussions there.

Reproduction: https://codesandbox.io/s/sharp-lamport-k9s8mp?file=/src/App.tsx

I think it's a bug in https://github.com/pmndrs/jotai/pull/2029 and it's not specific to atomWithStorage.

Tasks:
1. Create a failing test (without atomWithStorage, hopefully only with store api)
2. Fix it

tmkx:

Fixes #2098 

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -364,7 +364,7 @@ export const createStore = () => {
         return atomState
       }
       // Otherwise, check if the dependencies have changed.
-      // If all dependencies havne't changed, we can use the cache.
+      // If all dependencies haven't changed, we can use the cache.
       if (
         Array.from(atomState.d).every(
           ([a, s]) => a === atom || readAtomState(a) === s
@@ -572,8 +572,9 @@ export const createStore = () => {
     atom: Atom<Value>,
     initialDependent?: AnyAtom
   ): Mounted => {
+    const prevState = getAtomState(atom)
     // mount dependencies before mounting self
-    getAtomState(atom)?.d.forEach((_, a) => {
+    prevState?.d.forEach((_, a) => {
       const aMounted = mountedMap.get(a)
       if (aMounted) {
         aMounted.t.add(atom) // add dependent
@@ -584,7 +585,7 @@ export const createStore = () => {
       }
     })
     // recompute atom state
-    readAtomState(atom)
+    const currentState = readAtomState(atom)
     // mount self
     const mounted: Mounted = {
       t: new Set(initialDependent && [initialDependent]),
@@ -593,6 +594,15 @@ export const createStore = () => {
     mountedMap.set(atom, mounted)
     if (import.meta.env?.MODE !== 'production') {
       mountedAtoms.add(atom)
+    }
+    // update `initialDependent` if the value has changed during mounting
+    if (
+      initialDependent &&
+      prevState &&
+      currentState !== prevState &&
+      !isEqualAtomValue(prevState, currentState)
+    ) {
+      readAtomState(initialDependent, true)
     }
     // onMount
     if (isActuallyWritableAtom(atom) && atom.onMount) {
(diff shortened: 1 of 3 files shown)


dai-shi (changes requested):
Thanks for working on it!

Can you add a test with store api, without atomWithStorage, and without React?

tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -445,6 +445,62 @@ describe('atomWithStorage (with browser storage)', () => {
...
+
+  it("should recompute dependents' state after onMount (#2098)", async () => {
+    const store = createStore()
+    const mockNativeStorage = Object.create(window.Storage.prototype)
+    let currentValue: string | null = 'true'
+    mockNativeStorage.setItem = vi.fn(
+      (_key: string, value: string) => (currentValue = value)
+    )
+    mockNativeStorage.getItem = vi.fn(() => currentValue)
+    mockNativeStorage.removeItem = vi.fn(() => (currentValue = null))
```
dai-shi, line 457: Can it be something like this?
```suggestion
    let currentValue: string | null = 'true'
    const mockNativeStorage = {
      setItem: vi.fn(
      (_key: string, value: string) => (currentValue = value)
    ),
      getItem: vi.fn(() => currentValue),
      removeItem: vi.fn(() => (currentValue = null)),
    }
```

tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -445,6 +445,62 @@ describe('atomWithStorage (with browser storage)', () => {
...
+
+    const isLoggedAtom = atom(false)
+    const _isDevModeState = atomWithStorage(
+      '_isDevModeState',
```
dai-shi, line 461: We reserve `_foo` for unused names:
```suggestion
    const isDevModeState = atomWithStorage(
      'isDevModeState',
```

tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -445,6 +445,62 @@ describe('atomWithStorage (with browser storage)', () => {
...
+
+    const { getByRole } = render(
+      <StrictMode>
+        <DummyComponent />
+      </StrictMode>
+    )
+
+    act(() => store.set(isLoggedAtom, true))
+
+    const $checkbox = getByRole('checkbox') as HTMLInputElement
```
dai-shi, line 495: Just a preference:
```suggestion
    const checkbox = getByRole('checkbox') as HTMLInputElement
```

src/vanilla/store.ts
```diff
@@ -594,6 +583,17 @@ export const createStore = () => {
     if (import.meta.env?.MODE !== 'production') {
       mountedAtoms.add(atom)
     }
+    // mount dependencies before mounting self
```
dai-shi, line 586: I wonder if we should mount dependencies before mounting self or after mounting self. There was a failing test in #2029 AFAIR, but hmm maybe it's fixed afterwards.

dai-shi (commented):
The tricky thing is CI tests.

tests/react/vanilla-utils/atomWithStorage.test.tsx
tmkx: updated, it's copied from https://github.com/pmndrs/jotai/blob/2ec0bb96533f827b2b6f38456bbeab9cc542eddc/tests/react/vanilla-utils/atomWithStorage.test.tsx#L380-L383, should I update it too?

dai-shi (approved):
It looks good. Nice work.

Let me try and see if I can refactor.

dai-shi:
I guess the fundamental issue is that dependent graph isn't ready when onMount is invoked. So, the queue is introduced. Not really sure if it's the ideal solution, but it works so far.

fturmel:
Thank you both!

PR #2100, opened 2023-08-29
fix(utils/atomWithStorage): Prevent createJSONStorage from adding subscribe method when window.Storage is not defined

byudaniel:
Related to https://github.com/pmndrs/jotai/pull/1375. In other environments (React Native in my case) `window.addEventListener` may be polyfilled, but `window.Storage` is not, so `instanceof window.Storage` fails.

## Summary
`storage.subscribe` should only be defined if `window.Storage` is available in the environment.

## Check List


dai-shi (approved):
LGTM (Though, it's getting tricker and tricker.)

PR #2105, opened 2023-08-31

dai-shi:
for v2.4.1


tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "jsx": "react-jsx",
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "nodenext",
+    "moduleResolution": "bundler",
```
dai-shi, line 8: I was hoping to do this:
```json
    "module": "nodenext",
    "moduleResolution": "nodenext",
```

However, because the package.json's `type` is undefined (=commonjs), it causes an error:
```
TS1470: The 'import.meta' meta-property is not allowed in files which will build into CommonJS output.
```

PR #2112, opened 2023-09-02
chore(tests/utils): suppress error boundary error log

tmkx:
## Summary

it's a bit verbose when debugging :)

image

## Check List

### DIFF
--- a/tests/react/vanilla-utils/atomWithObservable.test.tsx
+++ b/tests/react/vanilla-utils/atomWithObservable.test.tsx
@@ -8,6 +8,7 @@ import { useAtom, useAtomValue, useSetAtom } from 'jotai/react'
 import { atom, createStore } from 'jotai/vanilla'
 import { atomWithObservable } from 'jotai/vanilla/utils'
 
+const consoleError = console.error
 beforeEach(() => {
   vi.useFakeTimers({ shouldAdvanceTime: true })
   // A workaround for missing performance.mark after using fake timers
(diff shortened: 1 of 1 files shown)


tests/react/vanilla-utils/atomWithObservable.test.tsx
```diff
@@ -17,11 +26,14 @@ beforeEach(() => {
     performance.clearMarks = (() => {}) as any
     performance.clearMeasures = (() => {}) as any
   }
+  // suppress ErrorBoundary error log
+  consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
```
dai-shi, line 30: Nice idea. Can we only suppress the error boundary error, using some simple regex?

We have this, so maybe we should be consistent about using `spyOn`?
https://github.com/pmndrs/jotai/blob/ccba918281069f3ce003988f591389df4fefdde5/tests/react/error.test.tsx#L12

tmkx: atomWithObservable was excluded from older ts test, so it passed, but `error.test.tsx` it not:

https://github.com/pmndrs/jotai/blob/ccba918281069f3ce003988f591389df4fefdde5/.github/workflows/test-old-typescript.yml#L69

dai-shi: Hmmm, maybe we give up `spyOn` and just assign `console.error = ...`?

dai-shi (approved):
LGTM!

PR #2113, opened 2023-09-03
fix: text overflow under Recipes section in introduction page

Aslemammad:
Fixes #2109

As suggested in the discussion, the fix is simple using `break-all`.


sandren:
Thanks for the PR, but `break-all` introduces other issues:

image

Will have a more robust fix in shortly! 🙂

PR #2122, opened 2023-09-07
chore: fix ide complaints

tmkx:

1. Missing file extension "cjs" for "vitest/config"
2. Option 'allowImportingTsExtensions' can only be used when either 'noEmit' or 'emitDeclarationOnly' is set

## Check List


.eslintrc.json
```diff
@@ -36,7 +36,7 @@
     "prefer-const": "error",
     "curly": ["warn", "multi-line", "consistent"],
     "no-console": "off",
-    "import/extensions": ["error", "always"],
+    "import/extensions": ["error", "always", { "ignorePackages": true }],
```
dai-shi, line 39: Which line complains without this change?

tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "exactOptionalPropertyTypes": true,
     "baseUrl": ".",
     "skipLibCheck": true,
+    "noEmit": true,
```
dai-shi, line 14: Do we need this even if we have `--noEmit` option?

dai-shi (commented):
Thanks for the suggestion.

We should do the same in zustand and valtio too.

Meanwhile, please see these minor comments.

.eslintrc.json
tmkx: this line:
https://github.com/pmndrs/jotai/blob/cc35d1d222120e401d545e4991619eb49aea97b7/vitest.config.ts#L1

tsconfig.json
tmkx: not anymore

.eslintrc.json
dai-shi: Ohhhh, because it has a slash, right? I think it's the incompleteness of the plugin.
In that case, let's add an inline comment `// eslint-disable-next-line import/extensions` and revert the change in eslintrc.

dai-shi (approved):
The change looks good.
Can you also open the same PR in the two repos? https://github.com/pmndrs/zustand and https://github.com/pmndrs/valtio

tsconfig.json
tmkx: sorry i misunderstood, yes we need it because this flag is only allowed when --noEmit or --emitDeclarationOnly is enabled, since these import paths would not be resolvable at runtime in JavaScript output files.

https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions

tmkx:
there is a flaky test case :(

tsconfig.json
dai-shi: We used to have `--noEmit` in package.json and no "noEmit" in tsconfig.json. Does it cause any errors with this PR's change?
I mean I'm aware of https://www.typescriptlang.org/tsconfig#allowImportingTsExtensions and it's working before this PR, so I'm confused.

dai-shi:
> there is a flaky test case :(

Yeah... Maybe using user-event would help.

dai-shi:
> Yeah... Maybe using user-event would help.

Hmm, `it('does not show async stale result', async () => {` is already using `userEvent`.

Feel free to dig into it!
https://github.com/pmndrs/jotai/actions/runs/6111275676/job/16586153332

tsconfig.json
tmkx: it'ok, we are bundling with rollup instead of tsc. 
if you remove the `--noEmit` option in the pretest script, you will be able to see the complaint in the console.

tmkx: VSCode(tsserver) will show an error in the tsconfig.json

tmkx: image

dai-shi: Alright! thanks for the clarification!

PR #2147, opened 2023-09-21
Update index.js react-refresh document

Crayon-ShinChan:

Fixes #

## Summary

Improve the document, make it less confusing

## Check List


sandren:
The Next.js config can be either ESM or CommonJS so it is probably more clear to show only the property we are modifying. Thanks for the suggestion though. 🙂

PR #2159, opened 2023-09-30
docs: use correct pnpm command

ilyasmez:

None

## Summary

[pnpm install](https://pnpm.io/cli/install) is meant to install all the dependencies in the project (I believe it also works to add new dependencies just for compatibility with `npm`, but that's not the recommended way to add a dependency using pnpm).

[pnpm add](https://pnpm.io/cli/add) is what we need to add a dependency to the project.

## Check List

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -120,7 +120,7 @@ npm i jotai
 yarn add jotai
 
 # pnpm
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Thanks!

PR #2186, opened 2023-10-16
feat(vanilla): prefer using `this` for atom config

dai-shi:
It was intentional to avoid `this` so that it's never confusing when copying atom config objects.
(btw, this was the reason we avoided using class syntax suggested by @Thisen. so, it might be reconsiderable. Though, I'm not sure the final decision because of bundle size / little performance trade-off.)

As it turns out, not using `this` blocks some extended patterns. https://github.com/jotaijs/jotai-scope/issues/4
(And, I ended up with a suboptimal solution https://github.com/jotaijs/jotai-scope/pull/2.)

Let's see how it can change the situation.

### DIFF
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -92,14 +92,22 @@ export function atom<Value, Args extends unknown[], Result>(
     config.read = read as Read<Value, SetAtom<Args, Result>>
   } else {
     config.init = read
-    config.read = (get) => get(config)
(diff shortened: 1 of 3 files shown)


yf-yang:
Can you please give a short example to explain the difference when patching atom before/after this change?

dai-shi:
https://github.com/jotaijs/jotai-scope/pull/5 is open.

dai-shi:
https://github.com/jotaijs/jotai-scope/pull/5 seems to work well. Let's do this.

Issue #2191, opened 2023-10-17
`loadable` seems to tear


dmaskasky:
## Example:
https://codesandbox.io/s/epic-nash-pqkfvj

### Steps to reproduce: 
  (1) click the increment button multiple times in short succession.

### Expected
The final count is displayed (debounced by the async delay)

### Actual
If the async count value can lag behind the real count.

PR #2199, opened 2023-10-19
fix(vanilla): deal with promise resolving race condition

dmaskasky:
https://github.com/pmndrs/jotai/discussions/2192

## Related Issue
https://github.com/pmndrs/jotai/issues/2191

## Summary
Adds tests for the issue linked above.

## Description of Issue
A primitive atom (`atomP`) exists and has a value (`valueP`)
An async atom (`atomA`) gets the value (`valueAP`) from atomP and then has a delay (`delayA`) before returning the value from atomP.
A derived async atom (`atomB`) gets the value (`valueBP`) from atomP and after a microtask or greater delay (`delayB`) it gets the value (`valueBA`) from atomA.
The bug requires delayB < delayA.
So,
 - atomP is a synchronous dependency of atomA and atomB
 - atomA is an asynchronous dependency of atomB

After all atoms mount,
atomP is changed twice with a delay (`delayC`) less than delayA.
This first multi-change proceeds nominally. ✅

atomP is then changed twice again with a delay (`delayD`) less than delayA.
This second multi-change does NOT proceed nominally.
The first value of this multi-change is used for ALL subsequent values of this multi-change.
Therefore, subsequent values use a stale value. ❌

## Timing
See below for this timing in depth:
```
init valueP to 0
inside atomB.read: before await valueBA, valueBP is 0
change valueP to 1
inside atomB.read: before await valueBA, valueBP is 1
change valueP to 2
inside atomB.read: before await valueBA, valueBP is 2

inside atomA.read: after delayA, valueAP is 0
inside atomA.read: after delayA, valueAP is 1
inside atomA.read: after delayA, valueAP is 2

inside atomB.read: before await valueBA, valueBP is 2, valueBA is 2
inside atomB.read: before await valueBA, valueBP is 1, valueBA is 2
inside atomB.read: before await valueBA, valueBP is 0, valueBA is 2
===== above is nominal ✅ =====

change valueP to 3
inside atomB.read: before await valueBA, valueBP is 3
change valueP to 4
inside atomB.read: before await valueBA, valueBP is 4
inside atomA.read: after delayA, valueAP is 3
inside atomA.read: after delayA, valueAP is 4
inside atomB.read: before await valueBA, valueBP is 3, valueBA is 3 // Error: should be 4
inside atomB.read: before await valueBA, valueBP is 4, valueBA is 3 // Error: should be 4
===== above is anomalous ❌ =====
```

## Theory
Likely, as a performance optimization there is logic in [store.ts](https://github.com/pmndrs/jotai/blob/main/src/vanilla/store.ts) to reuse the existing value. The bug is probably due to caching the incorrect value.

The microsecond delay (delayB) in atomB is significant for some reason. Not sure why yet.

The `store.sub` is also significant for some reason.

## Unit Test
I'm having trouble getting a unit test to work with just the store. Subsequent fires of `setCount(increment)` do not cause asyncAtom.read to refire unless I use a React hook. Not sure why.

It would be preferable to write a unit test and not depend on the React library for testing core Jotai functionality.

## Check List

https://codesandbox.io/s/jovial-kepler-d9v27h

https://github.com/pmndrs/jotai/assets/5034873/0b873a61-5aa3-4f6a-a99e-cc51cc138258

### DIFF
--- a/tests/react/async.test.tsx
+++ b/tests/react/async.test.tsx
@@ -1,10 +1,10 @@
 import { StrictMode, Suspense, useEffect, useRef } from 'react'
 import { act, fireEvent, render, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
-import { expect, it } from 'vitest'
+import { assert, expect, it } from 'vitest'
 import { Provider, useAtom } from 'jotai/react'
 import { atom, createStore } from 'jotai/vanilla'
-import type { Atom, SetStateAction, Setter } from 'jotai/vanilla'
+import type { Atom, SetStateAction } from 'jotai/vanilla'
 
 const useCommitCount = () => {
   const commitCountRef = useRef(1)
@@ -1145,10 +1145,10 @@ it('multiple derived atoms with dependency chaining and async write (#813)', asy
 })
 
 it('[render] resolves dependencies reliably after a delay', async () => {
-  expect.assertions(2)
+  expect.assertions(1)
   const countAtom = atom(0)
-  const resultAtom = atom<number | null>(null)
 
+  let result: number | null = null
   const resolve: (() => void)[] = []
   const asyncAtom = atom(async (get) => {
     const count = get(countAtom)
@@ -1167,9 +1167,10 @@ it('[render] resolves dependencies reliably after a delay', async () => {
       console.log(`derived (${count})`)
       const resultCount = await get(asyncAtom)
       console.log(`derived (${count})`, resultCount)
-      setSelf(resultAtom, resultCount)
+      result = resultCount
+      if (resultCount === 2) setSelf() // <-- necessary
     },
-    (_get, set, ...args: Parameters<Setter>) => set(...args)
+    () => {}
   )
 
   const derivedSyncAtom = atom((get) => {
@@ -1192,7 +1193,7 @@ it('[render] resolves dependencies reliably after a delay', async () => {
   render(<TestComponent />, { wrapper: Wrapper })
 
   const setCount = (arg: SetStateAction<number>) => store.set(countAtom, arg)
-
+  const increment = (c: number) => c + 1
   await waitFor(() => assert(resolve.length === 1))
 
   resolve[0]!()
@@ -1203,8 +1204,7 @@ it('[render] resolves dependencies reliably after a delay', async () => {
   resolve[1]!()
   resolve[2]!()
 
-  await waitFor(() => assert(store.get(countAtom) === 2))
-  expect(store.get(resultAtom)).toBe(2)
+  await waitFor(() => assert(result === 2))
 
   await act(() => setCount(increment))
   await act(() => setCount(increment))
@@ -1213,72 +1213,5 @@ it('[render] resolves dependencies reliably after a delay', async () => {
   resolve[4]!()
 
   await waitFor(() => assert(store.get(countAtom) === 4))
-  expect(store.get(resultAtom)).toBe(4) // 3
+  expect(result).toBe(4) // 3
 })
-
-it('[unit] resolves dependencies reliably after a delay', async () => {
-  const countAtom = atom(0)
-  const resultAtom = atom<number | null>(null)
-
-  const resolve: (() => void)[] = []
-  const asyncAtom = atom(async (get) => {
-    const count = get(countAtom)
-    await new Promise<void>((r: { (): void; count?: number }) => {
-      r.count = count
-      resolve.push(r)
-    })
-    console.log(`resolved (${count})`)
-    return count
-  })
-
-  const derivedAtom = atom(
-    async (get, { setSelf }) => {
-      const count = get(countAtom)
-      await Promise.resolve()
-      console.log(`derived (${count})`)
-      const asyncCount = await get(asyncAtom)
-      console.log(`derived (${count})`, asyncCount)
-      setSelf(resultAtom, asyncCount)
-    },
-    (_get, set, ...args: Parameters<Setter>) => set(...args)
-  )
-
-  const store = createStore()
-  store.sub(derivedAtom, () => {})
-
-  await waitFor(() => assert(resolve.length === 1))
-
-  resolve[0]!()
-
-  store.set(countAtom, increment)
-  store.set(countAtom, increment)
-
-  await waitFor(() => assert(resolve.length === 3))
-
-  resolve[1]!()
-  resolve[2]!()
-
-  await waitFor(() => assert(store.get(countAtom) === 2))
-  expect(store.get(resultAtom)).toBe(2)
-
-  store.set(countAtom, increment)
-  store.set(countAtom, increment)
-
-  await waitFor(() => assert(resolve.length === 5))
-
-  resolve[3]!()
-  resolve[4]!()
-
-  await waitFor(() => assert(store.get(countAtom) === 4))
-  expect(store.get(resultAtom)).toBe(4) // 3
-})
-
-function assert(condition: boolean): asserts condition {
-  if (!condition) {
-    throw new Error('assertion failed')
-  }
-}
-
-function increment(c: number) {
-  return c + 1
-}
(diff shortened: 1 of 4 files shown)


dai-shi:
Thanks. It would be really nice to have store api tests.

Also, could you avoid `renderHook` somehow? All of our tests never depend on it.

tests/react/async.test.tsx
```diff
@@ -1143,3 +1144,229 @@ it('multiple derived atoms with dependency chaining and async write (#813)', asy
...
+  await microtask()
+  await microtask()
+  await microtask()
+  await macrotask()
+
+  expect(result.current.count).toBe(4)
+  expect(result.current.asyncCount).toBe(4) // 3
+})
+
+it('[render] resolves dependencies reliably after a delay', async () => {
```
dai-shi, line 1221: Sorry, you already have it? Can we remove renderHook tests?

tests/react/async.test.tsx
```diff
@@ -1143,3 +1144,229 @@ it('multiple derived atoms with dependency chaining and async write (#813)', asy
...
+  await macrotask()
+
+  resolve[0]!()
+  // is it possible to do this test without react hooks?
+  resolve[1]!() // <--- Error: resolve[1] is not defined, it should be.
+  resolve[2]!()
+
+  await microtask()
+  await microtask()
+  await microtask()
```
dai-shi, line 1343: Do we need three of them because we have three atoms?

tests/react/async.test.tsx
```diff
@@ -1143,3 +1144,229 @@ it('multiple derived atoms with dependency chaining and async write (#813)', asy
...
+
+  expect(store.get(countAtom)).toBe(4)
+  expect(store.get(asyncCountAtom)).toBe(4) // 3
+})
+
+function macrotask() {
+  return new Promise((r) => setTimeout(r, 0))
+}
+
+async function microtask() {}
```
dai-shi, line 1368: Does this change behaves the same?
```suggestion
function microtask() {
  return Promise.resolve()
}
```
Seems a bit consistent with other tests. Well, I would inline them, but that's just a tiny pref.

tests/react/async.test.tsx
```diff
@@ -1143,3 +1144,229 @@ it('multiple derived atoms with dependency chaining and async write (#813)', asy
...
+  await microtask()
+  await microtask()
+  await macrotask()
+
+  expect(store.get(countAtom)).toBe(4)
+  expect(store.get(asyncCountAtom)).toBe(4) // 3
+})
+
+function macrotask() {
+  return new Promise((r) => setTimeout(r, 0))
```
dai-shi, line 1365:
```suggestion
  return new Promise((r) => setTimeout(r))
```
Just to be consistent with other tests.

tests/react/async.test.tsx
dmaskasky: sure. I just left it here for posterity, in case we felt it was a more reliable way to test. I'll remove it.

tests/react/async.test.tsx
dmaskasky: I don't know.

In the other test, I arrived at this combination of micro / macro tasks experimentally. If I remove one, the expect doesn't see the update in time.

dmaskasky:
> Thanks. It would be really nice to have store api tests.

I agree. Could you please take a look? I don't understand why asyncAtom isn't re-firing when I increment countAtom in the store api test I provided.

https://github.com/pmndrs/jotai/pull/2199/files#diff-4b43a4541a1bdfe262a00485c9d88e670551bcbecade30b84fdb0571e492c23eR1338

tests/react/async.test.tsx
dmaskasky: I could be more general and use [waitFor](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor), but I figured it would be better to be more explicit with this timing delay.

But let me know if you think this approach is too brittle.

dai-shi: >  I figured it would be better to be more explicit with this timing delay.

Yeah, if the goal of test is about precise timing, it would be fine.
But, tests with `waitFor` can cover it, it sounds more reliable.

> But let me know if you think this approach is too brittle.

Yeah, I would like to see the `waitFor` version. If that works, it's probably better.

dmaskasky: ok, I switched to using `waitFor`.
And I got the unit test working with `store`.

So, right now both tests are failing, but if you add macrotask delays in places, they will pass.

tests/react/async.test.tsx
```diff
@@ -1143,3 +1143,142 @@ it('multiple derived atoms with dependency chaining and async write (#813)', asy
...
+  })
+
+  const derivedAtom = atom(
+    async (get, { setSelf }) => {
+      const count = get(countAtom)
+      await Promise.resolve()
+      console.log(`derived (${count})`)
+      const resultCount = await get(asyncAtom)
+      console.log(`derived (${count})`, resultCount)
+      setSelf(resultAtom, resultCount)
```
dai-shi, line 1170: Please do not use `setSelf` if we can reproduce the issue without it.

(Otherwise, we have a different file `async2.test.tsx` for v2 specific apis.)

tests/react/async.test.tsx
```diff
@@ -1143,3 +1143,142 @@ it('multiple derived atoms with dependency chaining and async write (#813)', asy
...
+  await act(() => setCount(increment))
+
+  resolve[3]!()
+  resolve[4]!()
+
+  await waitFor(() => assert(store.get(countAtom) === 4))
+  expect(store.get(resultAtom)).toBe(4) // 3
+})
+
+it('[unit] resolves dependencies reliably after a delay', async () => {
```
dai-shi, line 1219: Can you please move this test to `tests/vanilla/store.test.tsx`?

dai-shi (commented):
Sorry for minor requests. I will do the rest afterwards.

tests/react/async.test.tsx
dmaskasky: This is weird. When I comment out the setSelf line, the test passes.
```ts
const derivedAtom = atom(
    async (get, { setSelf }) => {
      const count = get(countAtom)
      await Promise.resolve()
      console.log(`derived (${count})`)
      const resultCount = await get(asyncAtom)
      console.log(`derived (${count})`, resultCount)
      result = resultCount
      setSelf(resultAtom, resultCount) // <--- comment out this line and the test fails ¯\_(ツ)_/¯
    },
    (_get, set, ...args: Parameters<Setter>) => set(...args)
  )
...
expect(result).toBe(4) // 3
```

dmaskasky: fails
```ts
if (resultCount === 2) setSelf(resultAtom, resultCount)
```

```
derived (0)
resolved (0)
derived (1)
derived (2)
resolved (1)
resolved (2)
derived (2) 2
derived (1) 2
derived (0) 2
derived (3)
derived (4)
resolved (3)
resolved (4)
derived (3) 3 <-- should be 4
derived (4) 3
```

dmaskasky: passes
```ts
if (resultCount === 4) setSelf(resultAtom, resultCount)
```

```
derived (0)
resolved (0)
derived (1)
derived (2)
resolved (1)
resolved (2)
derived (2) 2
derived (1) 2
derived (0) 2
derived (3)
derived (4)
resolved (3)
resolved (4)
derived (4) 4 <-- passes
derived (3) 4
```

dmaskasky: Since `setSelf`, itself does not introduce any micro or macro task delay. This difference between fails / passes is the result of a side-effect of setSelf and not actually caused by a race condition.

Although there could still be an async side-effect of setSelf that is racing.

dai-shi (approved):
Thanks so much for your help. This must be a huge bug fix.

dmaskasky:
> Thanks so much for your help. This must be a huge bug fix.

Thanks for your help. Merge is blocked for me, kindly merge at the next opportunity.

PR #2216, opened 2023-10-26
fix(utils): unwrap may not return previous value

satoren:

Fixes #2214

## Summary

Fixed to carry over previous values

## Check List

### DIFF
--- a/src/vanilla/utils/unwrap.ts
+++ b/src/vanilla/utils/unwrap.ts
@@ -95,7 +95,7 @@ export function unwrap<Value, Args extends unknown[], Result, PendingValue>(
       return atom(
         (get) => {
           const state = get(promiseAndValueAtom)
-          if ('f' in state) {
+          if ('f' in state) { // is pending
             return state.f
           }
           return state.v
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
I added 3744b21 and modified the description slightly.

src/vanilla/utils/unwrap.ts
```diff
@@ -95,10 +95,10 @@ export function unwrap<Value, Args extends unknown[], Result, PendingValue>(
       return atom(
         (get) => {
           const state = get(promiseAndValueAtom)
-          if ('v' in state) {
-            return state.v
+          if ('f' in state) {
+            return state.f
```
satoren, line 99: It looks a little odd that f(allback) seems to have priority over v(alue).

dai-shi: Yeah, but with 'f' and 'v', it means 'v' is obsolete.

```js
if ('v' in state && !('f' in state)) {
  return state.v
}
return state.f
```
that requires more bytes.

dai-shi, line 99: We can add a comment.
```suggestion
          if ('f' in state) { // is pending
            return state.f
```

PR #2222, opened 2023-10-31
fix: primitive atom should always have `init` value

himself65:

N/A

## Summary

I'm creating a library that will modify the primitive atom in the runtime, which means each runtime could have a different initial value,  onMount...

I read the code and found that primitiveAtom should always property `init` in any case.

In this PR, I update the PrimitiveAtom type, so that in my code, infer `init` property should not case a missing type error:

```ts
function f(atom: PrimitiveAtom<unknown>) {
  atom.init // <-- this does exist
}
```

image

## Check List

### DIFF
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -59,7 +59,9 @@ export type PrimitiveAtom<Value> = WritableAtom<
   Value,
   [SetStateAction<Value>],
   void
->
+> & {
+  init: Value
+}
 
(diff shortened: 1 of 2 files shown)


dai-shi:
It's intentional not to include `init` in `PrimitiveAtom`. We can create a derived atom that works exactly like a primitive atom.
So, `PrimitiveAtom` type represents both a real primitive atom and a primitive-like atom.

PR #2229, opened 2023-11-02
fix(utils): useHydrateAtoms type

dai-shi:
Just found a tiny issue for precise  typing.

### DIFF
--- a/src/react/utils/useHydrateAtoms.ts
+++ b/src/react/utils/useHydrateAtoms.ts
@@ -6,34 +6,37 @@ type Options = Parameters<typeof useStore>[0] & {
   dangerouslyForceHydrate?: boolean
 }
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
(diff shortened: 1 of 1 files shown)


src/react/utils/useHydrateAtoms.ts
```diff
@@ -6,34 +6,37 @@ type Options = Parameters<typeof useStore>[0] & {
   dangerouslyForceHydrate?: boolean
 }
 type AnyWritableAtom = WritableAtom<unknown, any[], any>
-type AtomMap<A = AnyWritableAtom, V = unknown> = Map<A, V>
-type AtomTuple<A = AnyWritableAtom, V = unknown> = readonly [A, V]
-type InferAtoms<T extends Iterable<AtomTuple>> = {
-  [K in keyof T]: T[K] extends AtomTuple<infer A>
-    ? A extends AnyWritableAtom
-      ? AtomTuple<A, ReturnType<A['read']>>
```
dai-shi, line 14: > `ReturnType<A['read']>`

is not correct.

PR #2231, opened 2023-11-03
Simplify `splitAtom()` by using the `Array.prototype.toSpliced()` method

MiroslavPetrik:
## Summary

A small refactor to use the new [immutable change-array-by-copy methods](https://github.com/tc39/proposal-change-array-by-copy). I believe this makes code more readable and performant, as less calls & copying is made.
The methods are available since TypeScript 5.2 and [Node 20](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md#v8-113) which is stable/current since October.

I've noticed untested behavior for the `move` action, that the first item can be moved like in a carousel to the end of the list. I've updated the test to make the behavior official.

For tests to pass, I had to upgrade the node from v18, to v20. I've signaled that in the `engine` field in the `package.json`.
Lastly, I've set the `packageManager` so it can be easily installed with `corepack install`. That should simplify further contributions.

## Check List

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "patch-readme": "shx sed -i 's/.*Jotai \\(dark mode\\).*//' dist/readme.md"
   },
   "engines": {
-    "node": ">=12.20.0"
+    "node": ">=20.0.0"
   },
   "prettier": {
     "semi": false,
@@ -176,5 +176,6 @@
     "@types/react": {
       "optional": true
     }
-  }
+  },
+  "packageManager": "yarn@1.22.19"
 }
--- a/src/vanilla/utils/splitAtom.ts
+++ b/src/vanilla/utils/splitAtom.ts
@@ -158,10 +158,10 @@ export function splitAtom<Item, Key>(
                   const index = get(splittedAtom).indexOf(action.atom)
                   if (index >= 0) {
                     const arr = get(arrAtom)
-                    set(arrAtom as WritableAtom<Item[], [Item[]], void>, [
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -83,7 +83,7 @@
     "patch-readme": "shx sed -i 's/.*Jotai \\(dark mode\\).*//' dist/readme.md"
   },
   "engines": {
-    "node": ">=12.20.0"
+    "node": ">=20.0.0"
```
dmaskasky, line 86: Would this create node incompatibility issues for versions less than 20?

MiroslavPetrik: Hmm, I realize that jotai runs even on SSR, so strictly speaking the package would be indicated as incompatible, given even the latest Next.js 14 runs on Node v18. Originally I missed this case, and thought only of the dev/build env.

In practice its unlikely that the code will run in SSR contexts, given the utility has UI actions. But we cannot make such assumption, so the polyfill would have to be recommended for users running below v20?

I would guess this is not desirable to have such installation 'pre-caution' step, as likely just a subset of users uses the `splitAtom`.

Maybe it could be a separate package `jotai-split` hehe.

dai-shi: Yeah, I think this is too early. We build CJS targeting ie11, even though it doesn't work well on ie11 anyway.

I think this could be done when we reach Jotai v3. It's not planned yet. Let me close this for now.

dai-shi:
@MiroslavPetrik Thanks for your suggestion. Wouk you pease open a new PR to add tests for splitAtom?

PR #2271, opened 2023-11-25
docs: update `atom.mdx`

himself65:

Fixes https://github.com/pmndrs/jotai/discussions/2268

## Check List

### DIFF
--- a/docs/core/atom.mdx
+++ b/docs/core/atom.mdx
@@ -39,7 +39,8 @@ const writeOnlyAtom = atom(
     // `update` is any single value we receive for updating this atom
     set(priceAtom, get(priceAtom) - update.discount)
     // or we can pass a function as the second parameter
-    // the function will be invoked, receiving the atom's current value as its first parameter
+    // the function will be invoked,
+    //  receiving the atom's current value as its first parameter
     set(priceAtom, (price) => price - update.discount)
   },
 )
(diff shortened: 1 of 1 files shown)


docs/core/atom.mdx
```diff
@@ -79,24 +82,24 @@ const Component = ({ value }) => {
...
+function atom<Value, Args extends unknown[], Result>(
   read: Value,
-  write: (get: Getter, set: Setter, update: Update) => void | Promise<void>,
-): WritableAtom<Value, Update>
+  write: (get: Getter, set: Setter, ...args: Args) => Result,
+): WritableAtom<Value, Args, Result>
 ```
 
 - `initialValue`: the initial value that the atom will return until its value is changed.
-- `read`: a function that's called on every re-render. The signature of `read` is `(get) => Value | Promise<Value>`, and `get` is a function that takes an atom config and returns its value stored in Provider as described below. Dependency is tracked, so if `get` is used for an atom at least once, the `read` will be reevaluated whenever the atom value is changed.
```
himself65, line 98: `called on every re-render` is incorrect. atom's read will be only call once if it's a pure function like `atom(() => 1)`

docs/core/atom.mdx
```diff
@@ -126,6 +133,20 @@ anAtom.onMount = (setAtom) => {
...
+
+const Component = () => {
+  // `onMount` will be called when the component is mounted in the following cases:
+  useAtom(anAtom)
+  useAtomValue(anAtom)
+
+  // however, in the following cases,
+  //  `onMount` will not be called because the atom is not subscribed:
+  useSetAtom(anAtom)
+  useAtomCallback(
```
himself65, line 145: https://github.com/pmndrs/jotai/discussions/2243

dai-shi (approved):
This is very good. Thanks for improving docs!

Issue #2269, opened 2023-11-25
Vite 5 / Rollup 4 build warning in v2.6.0


Kamahl19:
The combination of upgrading to Vite 5 (Rollup 4) and Jotai 2.6.0 results in an error during building
```
node_modules/jotai/esm/react.mjs (1:0) Error when using sourcemap for reporting an error: Can't resolve original location of error.
```

This error does not happen in 2.5.1 and also does not happen in Vite 4 + Jotai 2.6.0.

dai-shi:
@himself65 Was `rollup-plugin-banner2` too easy solution??

himself65:
> @himself65 Was `rollup-plugin-banner2` too easy solution??

Might be. I was worried about this before 
https://github.com/pmndrs/jotai/discussions/2241#discussioncomment-7514559

himself65:
Here is the minimum reproduction

https://stackblitz.com/edit/vitejs-vite-v2qh9v?file=src%2Fmain.ts

himself65:
@Kamahl19 Hi, would you try to use https://github.com/huozhi/rollup-plugin-swc-preserve-directives to see if the error disappear

himself65:
> I get even more errors. Maybe I misunderstood. Should I just add it to the vite plugin?

Could you please share you vite config?

himself65:
I just tried, and there's only one warning that won't cause building interrupt.

```
main.js → output...
(!) Module level directives cause errors when bundled, "use client" in "node_modules/.pnpm/jotai@2.6.0_react@18.2.0/node_modules/jotai/esm/react.mjs" was ignored.
node_modules/.pnpm/jotai@2.6.0_react@18.2.0/node_modules/jotai/esm/react.mjs (1:0)
1: 'use client';
   ^
2: import ReactExports, { createContext, useContext, useRef, createElement, useReducer, useEffect, useDebugValue, useCal...
3: import { getDefaultStore, createStore } from 'jotai/vanilla';
```

himself65:
Anyway, I'm working on a new implementation that won't cause this error

himself65:
I've published npm `jotai-test-react-server` with a patch version; could you please try that again?

dai-shi:
Can this be an upstream issue?

https://github.com/vitejs/vite/issues/15012

dai-shi:
Any updates on this? Anyone?

himself65:
https://stackblitz.com/edit/vitejs-vite-v2qh9v

[...]

PR #2274, opened 2023-11-26
build: fix bundler might cause error

himself65:

Fixes https://github.com/pmndrs/jotai/issues/2269

## Summary

image

## Check List

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,8 @@
         "default": "./esm/react.mjs"
       },
       "default": {
-        "react-server": "./error.js",
         "types": "./react.d.ts",
+        "react-server": "./error.js",
         "default": "./react.js"
       }
     },
--- a/package.json
+++ b/package.json
@@ -36,11 +36,37 @@
     ".": {
       "import": {
         "types": "./esm/index.d.mts",
+        "react-server": "./esm/vanilla.mjs",
         "default": "./esm/index.mjs"
       },
       "default": {
         "types": "./index.d.ts",
+        "react-server": "./vanilla.js",
         "default": "./index.js"
+      }
+    },
+    "./react": {
+      "import": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -36,13 +36,39 @@
...
+        "types": "./esm/*.d.mts",
+        "react-server": "./esm/error.mjs",
+        "default": "./esm/*.mjs"
+      },
+      "default": {
+        "types": "./*.d.ts",
+        "react-server": "./error.js",
+        "default": "./*.js"
+      }
+    },
```
himself65, line 71: I go through the code, and all react* files uses `useContext` or `useAtom` which is not accepted in react server environment

dai-shi:
While this might improve the RSC developer experience, I don't think it solves #2241.

At this point, I don't think adding `react-server` entry point to show an error is necessary. React should be showing errors anyway. Or, we can add `client-only` if we want an explicit error.

If we don't have a solution with `'use client'`, let's revert #2245.

himself65:
> While this might improve the RSC developer experience, I don't think it solves #2241.
> 
> At this point, I don't think adding `react-server` entry point to show an error is necessary. React should be showing errors anyway. Or, we can add `client-only` if we want an explicit error.
> 
> If we don't have a solution with `'use client'`, let's revert #2245.

revert would be better

himself65:
There might be another approach. We have `index.react-server.ts` that only exports `vanilla.ts.`

This might help if people did import `jotai` but only used `atom.` Otherwise, they have to import `jotai/vanilla.`

himself65:
I did some changes

dai-shi:
Unfortunately, it doesn't change my previous comment. I feel it's worse because it hides the error.

> Otherwise, they have to import jotai/vanilla,

That's recommended. So, it's a doc issue.

PR #2304, opened 2023-12-17
fix(utils): add init property to the return type of atomWithReset

jaesoekjjang:

Fixes #2303 

## Summary

Changed return type of `atomWithReset()`,  
``` typescript
return anAtom as WritableAtom<Value, [Update], void>
```
to
``` typescript
return anAtom as WritableAtom<Value, [Update], void> & WithInitialValue<Value>
```

## Check List

### DIFF
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -23,7 +23,7 @@ type Write<Args extends unknown[], Result> = (
   ...args: Args
 ) => Result
 
-type WithInitialValue<Value> = {
+export type WithInitialValue<Value> = {
   init: Value
 }
 
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -23,6 +23,8 @@ type Write<Args extends unknown[], Result> = (
   ...args: Args
 ) => Result
 
+// This is an internal type and not part of public API.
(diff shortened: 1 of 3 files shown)


src/vanilla/atom.ts
```diff
@@ -23,7 +23,7 @@ type Write<Args extends unknown[], Result> = (
   ...args: Args
 ) => Result
 
-type WithInitialValue<Value> = {
+export type WithInitialValue<Value> = {
```
dai-shi, line 26: This type is intentionally private. Please define it in atomWithReset.ts.
`init` property is an internal property too.

dai-shi (changes requested):
Thanks for working on it.

dai-shi (approved):
As I said in #2303, using `.init` is not recommended in the userland code. (Technically, there can be breaking changes without notice. If this is confusing, we should probably remove `WithInitialValue` type.)

This change is technically valid, and helps hacks. I will merge it later when preparing the next release.

dai-shi:
Let me add a commit to add a note in the source code and see how you feel.

jaesoekjjang:
> Let me add a commit to add a note in the source code and see how you feel.

Love it :)

PR #1687, opened 2023-01-15
chore(benchmark): fix tool for v2

dai-shi:
Continuing from #486.

PR #2313, opened 2023-12-30
fix: benchmarks

dai-shi:
fix https://github.com/pmndrs/jotai/pull/1687#issuecomment-1872456507


dai-shi:
@L-Qun This should fix it.

PR #2318, opened 2024-01-01
chore: benchmarks

L-Qun:

Fixes #

## Summary

I think it might be better to put all the results together.

## Check List


dai-shi (approved):
Thanks for improving it.

PR #2320, opened 2024-01-02
add configuration for vite react-wsc

SpringHgui:

Fixes #

## Check List


himself65 (approved):

website/src/pages/index.js
```diff
@@ -131,6 +131,30 @@ pnpm add jotai
         Then we recommend adding the optional SWC or Babel plugin to enable React Fast Refresh
         support for the best developer experience (framework specific).
       </p>
+      <h3>Vite (SWC)</h3>
+      <Code language="bash">{`# npm
+npm install --save-dev @swc-jotai/react-refresh
```
hyoban, line 136:
```suggestion
npm install --save-dev @swc-jotai/react-refresh @swc-jotai/debug-label
```

sandren (changes requested):
Thanks for looking into this.

Can we reduce the plugins to only `@swc-jotai/react-refresh` to match the instructions for the other environments? Thanks! 🙂 

Also we'll want to confirm that inline atom declarations are working since there has been some trouble with Next.js (SWC).

SpringHgui:
I don't quite understand what you mean, what else needs to be done @sandren

sandren (approved):

PR #2325, opened 2024-01-04
chore(docs): show only npm install

dai-shi:
It's 2024, and we can't list all npm alternatives. Let's simplify.

### DIFF
--- a/docs/extensions/immer.mdx
+++ b/docs/extensions/immer.mdx
@@ -7,12 +7,10 @@ keywords: immer
 
 ### Install
 
-You have to install `immer` and `jotai-immer` to use this feature.
+You have to install `immer` and `jotai-immer` to use this feature, with your preferred package manager.
 
 ```
-npm install immer jotai-immer
-# or
-yarn add immer jotai-immer
+npm i immer jotai-immer
 ```
 
 ## atomWithImmer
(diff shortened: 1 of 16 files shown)


docs/extensions/immer.mdx
```diff
@@ -109,6 +107,8 @@ const Controls = () => {
 
 It would be better if you don't use `withImmer` and `atomWithImmer` with `useImmerAtom` because they provide the immer-like `writeFunction` and we don't need to create a new one.
 
+You can use `useSetImmerAtom` if you need only the setter part of `useImmerAtom`.
```
dai-shi, line 110: New feature in jotai-immer v0.3.0.

sandren (commented):
Everyone should understand that dependencies can be installed "with your preferred package manager" so I think those lines can be omitted. 🙂

himself65 (commented):
Some document support select from multiple package managers as preferences.

This might be another idea but spends extra work

https://nextra.site/docs/guide/built-ins/tabs

dai-shi:
Well, that's different from my motivation. (But I'll consider such alternatives if this confuses people too much.)

sandren (approved):
LGTM! 🙂

PR #2332, opened 2024-01-08
feat(docs): add devtools setup example for react+vite projects

ahme-dev:
## Summary

Essentially adds another example for the setup of Jotai Devtools in a React project (with Vite). More convenient than searching for it, like in my case.

### DIFF
--- a/docs/tools/devtools.mdx
+++ b/docs/tools/devtools.mdx
@@ -59,9 +59,7 @@ export default defineConfig({
   plugins: [
     react({
       babel: {
-        presets: [
-          'jotai/babel/preset',
-        ],
+        presets: ['jotai/babel/preset'],
       },
     }),
   ],
--- a/docs/tools/devtools.mdx
+++ b/docs/tools/devtools.mdx
@@ -41,7 +41,7 @@ Enhance your development experience with the UI based Jotai DevTool.
 
 Use Jotai babel plugins for optimal debugging experience. Find the complete guide on the [babel](../tools/babel) page and/or [swc](../tools/swc) page.
 
-Eg.
+Example:
 
 ```ts
 {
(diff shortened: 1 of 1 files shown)


ahme-dev:
Prettier seems to run fine locally, not sure about the issue here.

dai-shi:
It's `docs/tools/devtools.mdx`. Try running `yarn run prettier` please.

dai-shi (commented):
@Thisen What do you think? Should we recommend the preset?

Thisen (approved):
I think it's good 👍🏼

himself65 (approved):

docs/tools/devtools.mdx
```diff
@@ -50,6 +50,25 @@ Eg.
...
+// vite.config.ts
+
+export default defineConfig({
+  plugins: [
+    react({
+      babel: {
+        plugins: [
+          'jotai/babel/plugin-debug-label',
+          'jotai/babel/plugin-react-refresh',
+        ],
```
dai-shi, line 65: Hi, thanks for the suggestion.
Can you try `jotai/babel/preset` instead? I think it's the way we recommend using the plugins.

dai-shi (changes requested):

docs/tools/devtools.mdx
ahme-dev: Hello, @dai-shi. I took it from the other example but sure.

docs/tools/devtools.mdx
```diff
@@ -50,6 +50,24 @@ Eg.
...
+Example for Vite + React project:
+
+```ts
+// vite.config.ts
+
+export default defineConfig({
+  plugins: [
+    react({
+      babel: {
+        plugins: [
```
dai-shi, line 62: Isn't it presets instead of plugins? Does it work with this?

docs/tools/devtools.mdx
dai-shi: Maybe we should fix the other example too??

docs/tools/devtools.mdx
ahme-dev: No, my bad. Not familiar at all with babel. Fixed it now

docs/tools/devtools.mdx
ahme-dev: Done

dai-shi (approved):
Thanks for your contribution!

PR #2333, opened 2024-01-09
readAtomState is called many times redundantly before atoms are mounted

edkimmel:

https://github.com/pmndrs/jotai/discussions/2334

## Summary

This PR is here to showcase the issue with a potential fix, not to be merged.

```
Test case without unmounted cache
Jotai store call counts
-----------------------
| readAtomState | 45
Test case with unmounted cache
Jotai store call counts
-----------------------
| readAtomState | 24
```

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -592,8 +592,8 @@ export const createStore = (opts?: { unmountedCacheEnabled: boolean }) => {
         l({ type: 'write', flushed: flushed as Set<AnyAtom> }),
       )
     }
-    if (unmountedCacheEnabled && !mountedAtoms.has(atom)) {
-      // A value was manually changed in the store, invalidate all unmounted atoms
+    if (unmountedCacheEnabled) {
+      // A value was changed in the store, invalidate all unmounted atoms
       unmountedAtomReadCache = new WeakSet<AnyAtom>()
     }
     return result
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -150,6 +150,7 @@ type MountedAtoms = Set<AnyAtom>
  * @returns A store.
  */
 export const createStore = () => {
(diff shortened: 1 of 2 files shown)


src/vanilla/store.ts
```diff
@@ -772,6 +802,8 @@ export const createStore = () => {
           l({ type: 'restore', flushed: flushed as Set<AnyAtom> }),
         )
       },
+      dev_print_call_counts: printCounts,
+      dev_clear_call_counts: resetCounts,
     }
```
himself65, line 807: is this necessary for our core library? I think making a wrapper in user side should make more sense

tests/react/optimization.test.tsx
```diff
@@ -271,3 +271,78 @@ it('no extra rerenders after commit with derived atoms (#1213)', async () => {
...
+    <Provider store={store}>
+      <MyPage />
+    </Provider>,
+  )
+
+  store.dev_print_call_counts?.()
+  store.dev_clear_call_counts?.()
+}
+it('Avoid redundant readAtomState while unmount', () => {
+  console.debug('Test case without unmounted cache')
```
himself65, line 333: remove console?

tests/react/optimization.test.tsx
```diff
@@ -271,3 +271,78 @@ it('no extra rerenders after commit with derived atoms (#1213)', async () => {
...
+  render(
+    <Provider store={store}>
+      <MyPage />
+    </Provider>,
+  )
+
+  store.dev_print_call_counts?.()
+  store.dev_clear_call_counts?.()
+}
+it('Avoid redundant readAtomState while unmount', () => {
```
himself65, line 332: in this test case we didn't expect anything. just run the code? I think we should add some counter to make sure behavior is stable

dai-shi:
As I understand, this is not something to be merged. Let's convert it to draft.

PR #2340, opened 2024-01-13
feat(react): add new hook `useAtomValueMemo`

himself65:
## Summary

Since some field of an object is never used but will trigger an update. This new hook will improve the performance of rendering the huge object atom.

Only jotai core library can archive this magic because it needs internal `useStore` hook

## Check List


dai-shi:
useStore is a public API. Please do it in [Jotai Labs](https://github.com/jotaijs). 😄

himself65:
> useStore is a public API. Please do it in [Jotai Labs](https://github.com/jotaijs). 😄

Oh I didn't know that

PR #2347, opened 2024-01-18
fix: atoms should not remount on recalculations (with async dependencies)

dmaskasky:
https://github.com/pmndrs/jotai/issues/2348

## Summary
Atoms having an onMount function will remount on another atom recalculation.

Please see the following example below:
countAtom will remount when anotherAtom is incremented.
```ts
const countAtom = atom(0)
countAtom.onMount = () => {}
const anotherAtom = atom(0)
  const derivedAtom = atom((get) => {
    void get(countAtom)
  })
const effectAtom = atom((get) => {
    get(derivedAtom)
    get(anotherAtom)
  })
})
```

Expected behavior is that countAtom should not remount when anotherAtom increments.

## Check List

- [o] `yarn run prettier` for formatting code and docs

### DIFF
--- a/tests/react/async.test.tsx
+++ b/tests/react/async.test.tsx
@@ -70,6 +70,8 @@ it('does not show async stale result', async () => {
   await findByText('loading')
   resolve1()
   resolve2()
+  await Promise.resolve()
+  resolve2()
   await waitFor(() => {
     getByText('count: 2')
     getByText('delayedCount: 2')
--- a/tests/vanilla/dependency.test.tsx
+++ b/tests/vanilla/dependency.test.tsx
@@ -93,3 +93,63 @@ it('correctly handles the same promise being returned twice from an atom getter 
   store.set(counterAtom, 1)
   await expect(store.get(derivedAtom)).resolves.toBe('Asynchronous Data')
 })
+
+it('keeps atoms mounted between recalculations', async () => {
+  const metrics1 = {
+    mounted: 0,
+    unmounted: 0,
+  }
+  const atom1 = atom(0)
+  atom1.onMount = () => {
+    ++metrics1.mounted
+    return () => {
+      ++metrics1.unmounted
+    }
+  }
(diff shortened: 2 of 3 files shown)


tests/vanilla/dependency.test.tsx
```diff
@@ -93,3 +93,45 @@ it('correctly handles the same promise being returned twice from an atom getter
...
+  store.set(countAtom, (c) => c + 1)
+  expect(metrics).toEqual({
+    mountCount: 1,
+    unmountCount: 0,
+  })
+  store.set(anotherAtom, (c) => c + 1)
+  await Promise.resolve().then()
+  expect(metrics).toEqual({
+    mountCount: 1,
+    unmountCount: 0,
```
dmaskasky, line 135: This test fails with `{ mountCount: 2, unmountCount: 1 }`

tests/vanilla/dependency.test.tsx
```diff
@@ -93,3 +93,61 @@ it('correctly handles the same promise being returned twice from an atom getter
...
+  })
+  store.set(atom1, (c) => c + 1)
+  await Promise.resolve().then()
+  expect(metrics1).toEqual({
+    mounted: 1,
+    unmounted: 0,
+  })
+  expect(metrics2).toEqual({
+    mounted: 1,
+    unmounted: 0,
```
dmaskasky, line 151: This test fails with `{ mounted: 2, unmounted: 1 }` for metrics2

tests/react/async.test.tsx
```diff
@@ -70,6 +70,8 @@ it('does not show async stale result', async () => {
   await findByText('loading')
   resolve1()
   resolve2()
+  await Promise.resolve()
+  resolve2()
```
dai-shi, line 74: This is required because the behavior is changed.

dmaskasky:
LGTM

PR #2353, opened 2024-01-22
feat(vanilla): Add a hack for jotai-scope

yf-yang:
Reverts #2186
Add a property `scopedOriginal` and add a intercepted function to ensure copy of scoped atoms can directly set itself without calling its write function again. Now all the write function in jotai-scope get executed **exactly once**.
Related to https://github.com/jotaijs/jotai-scope/pull/20

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -102,6 +102,14 @@ const returnAtomValue = <Value>(atomState: AtomState<Value>): Value => {
   return atomState.v
 }
 
+/**
+ * @returns Original atom if anAtom is a scoped atom in `jotai-scope`. Otherwise it
+ * returns anAtom itself.
+ */
+const getScopedOriginal = (anAtom: AnyAtom): AnyAtom => {
+  return anAtom.scopedOriginal ?? anAtom;
+}
+
 type Listeners = Set<() => void>
 type Dependents = Set<AnyAtom>
 
@@ -524,7 +532,7 @@ export const createStore = () => {
       ...args: As
     ) => {
       let r: R | undefined
-      if ((a as AnyWritableAtom) === atom) {
+      if (getScopedOriginal(a) === getScopedOriginal(atom)) {
         if (!hasInitialValue(a)) {
           // NOTE technically possible but restricted as it may cause bugs
           throw new Error('atom not writable')
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -107,7 +107,7 @@ const returnAtomValue = <Value>(atomState: AtomState<Value>): Value => {
  * returns anAtom itself.
  */
 const getScopedOriginal = (anAtom: AnyAtom): AnyAtom => {
-  return anAtom.scopedOriginal ?? anAtom;
+  return anAtom.scopedOriginal ?? anAtom
 }
 
 type Listeners = Set<() => void>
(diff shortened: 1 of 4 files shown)


dai-shi:
First of all, thanks for working on this. Here's some notes.
- My motivation behind jotai-scope is to completely avoid modifying Jotai core. However, what we found is it's extremely difficult and inefficient, right?
- We don't need to revert #2186 as it has its own capability (meaning it might be useful not only jotai-scope).
- I'm not 100% sure if `scopedOriginal` is a good abstraction.
- So, please allow me to draft a new PR.

yf-yang:
> completely avoid modifying Jotai core

As I state in https://github.com/jotaijs/jotai-scope/pull/20, the shallow equality comparison is insufficient. Yesterday I thought about hours and figured out the key is to **call atom.write** exactly once, which means this modification is unavoidable.

> We don't need to revert #2186 

I am not so sure, at least part of the change should be reverted. Let me illustrate the key difference:
Internally, if read function/write function is not specified, in #2186, it would be converted to something like: `read = get => get(this)`. However, in a scoped copy atom copy of atom `anAtom` (we call if `anAtomCopy`), when it internally calls `get(anAtom)`, with #2186, it would be transformed to `get(anAtomCopy)` instead of `get(anAtom)`. That's the difference with `this`.

dai-shi:
#2356 is something I would like to go with.
Would you please try it and see if it works for jotai-scope?

dai-shi:
> That's the difference with `this`.

And, it seems intuitive, no?
```js
const copyAtom = (a) => ({ ...a });

const countAtom = atom(0);
const copiedAtom = copyAtom(countAtom);

  set(copiedAtom, 1);
  get(countAtom) // ---> should be `0`
```

yf-yang:
Nope, the difference is here
https://github.com/pmndrs/jotai/blob/b5c38081e922b56621c6da9b9c32ef3d0f46e4b3/src/vanilla/atom.ts#L100-L102
and 
https://github.com/pmndrs/jotai/blob/b5c38081e922b56621c6da9b9c32ef3d0f46e4b3/src/vanilla/atom.ts#L103-L115

Originally I want to make sure the `target` argument of `getAtom` should not be the scoped copy, with them, it cannot be assured.

However, I think it can be solved by adding more conditions in jotai-scope.

yf-yang:
The difference is not in store access, but in the parameter passed to setter and getter. Very hard to explain this, maybe after I write a documentation to explain how the scope works now.

yf-yang:
``` javascript
anAtomCopy = {...anAtom};
store.set(anAtomCopy, 0);
// =>
anAtomCopy.write(getter, setter, 0);
// => 
anAtomCopy.write(getter, (a, 0) => set(a, 0), 0);
// with `this`, a here is `anAtomCopy`
// without `this`, a here is `anAtom`
```

dai-shi:
I still want to keep #2186 as it feels right and gives more capability. So, I wish you can find a solution on the jotai-scope end.

dai-shi:
by the way, you can override `.write` in jotai-scope, if you want. Maybe we should use the same function reference.

src/vanilla/store.ts
```diff
@@ -524,7 +532,7 @@ export const createStore = () => {
       ...args: As
     ) => {
       let r: R | undefined
-      if ((a as AnyWritableAtom) === atom) {
+      if (getScopedOriginal(a) === getScopedOriginal(atom)) {
```
dai-shi, line 535: The same thing can be done with #2356, no?
```js
anAtom.is = function(target) {
  return getScopedOriginal(this) === getScopedOriginal(target)
}
```

PR #2354, opened 2024-01-22
fix(utils/atomWithStorage): defaultStorage with disabled local storage crashes on mount

benediktschlager:

Fixes downstream issue for https://ui.shadcn.com/
It crashes when cookies are disabled with this exception:
image

## Summary
On Firefox accessing the property localStorage of window results in a `DOMException: The operation is insecure.`
This PR will handle this case gracefully and fall back to providing no storage implementation. This should be fine as we cannot store anything in a cookie-free browser.

## Check List

### DIFF
--- a/src/vanilla/utils/atomWithStorage.ts
+++ b/src/vanilla/utils/atomWithStorage.ts
@@ -105,7 +105,7 @@ export function createJSONStorage<Value>(
     } catch (e) {
       if (import.meta.env?.MODE !== 'production') {
         if (typeof window !== 'undefined') {
-          console.warn(e);
+          console.warn(e)
         }
       }
       return undefined
--- a/src/vanilla/utils/atomWithStorage.ts
+++ b/src/vanilla/utils/atomWithStorage.ts
@@ -99,8 +99,18 @@ export function createJSONStorage<Value>(
   getStringStorage: () =>
     | AsyncStringStorage
     | SyncStringStorage
-    | undefined = () =>
-    typeof window !== 'undefined' ? window.localStorage : undefined,
+    | undefined = () => {
+    try {
+      return window.localStorage
+    } catch (e) {
+      // Window could be undefined or local storage could be disabled
+      if (e instanceof TypeError || e instanceof DOMException) {
+        return undefined
+      }
+
+      throw e
+    }
+  },
   options?: JsonStorageOptions,
 ): AsyncStorage<Value> | SyncStorage<Value> {
   let lastStr: string | undefined
--- a/src/vanilla/utils/atomWithStorage.ts
+++ b/src/vanilla/utils/atomWithStorage.ts
@@ -103,12 +103,12 @@ export function createJSONStorage<Value>(
     try {
       return window.localStorage
     } catch (e) {
-      // Window could be undefined or local storage could be disabled
-      if (e instanceof TypeError || e instanceof DOMException) {
-        return undefined
+      if (import.meta.env?.MODE !== 'production') {
+        if (typeof window !== 'undefined') {
+          console.warn(e);
+        }
       }
-
-      throw e
+      return undefined
     }
   },
   options?: JsonStorageOptions,
--- a/tests/react/vanilla-utils/atomWithStorage.test.tsx
+++ b/tests/react/vanilla-utils/atomWithStorage.test.tsx
@@ -515,7 +515,7 @@ describe('atomWithStorage (with disabled browser storage)', () => {
   })
 
   afterAll(() => {
-    // TS < 4.5 causes type error
+    // TS < 4.5 causes type error without `as any`
     ;(window as any).localStorage = savedLocalStorage
   })
 
--- a/tests/react/vanilla-utils/atomWithStorage.test.tsx
+++ b/tests/react/vanilla-utils/atomWithStorage.test.tsx
@@ -515,7 +515,8 @@ describe('atomWithStorage (with disabled browser storage)', () => {
   })
 
   afterAll(() => {
-    window.localStorage = savedLocalStorage
+    // TS < 4.5 causes type error
+    (window as any).localStorage = savedLocalStorage
   })
 
   it('initial value of atomWithStorage can be used when cookies are disabled', async () => {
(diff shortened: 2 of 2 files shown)


src/vanilla/utils/atomWithStorage.ts
```diff
@@ -99,8 +99,18 @@ export function createJSONStorage<Value>(
...
     | AsyncStringStorage
     | SyncStringStorage
-    | undefined = () =>
-    typeof window !== 'undefined' ? window.localStorage : undefined,
+    | undefined = () => {
+    try {
+      const storage =
+        typeof window !== 'undefined' ? window.localStorage : undefined
+      return storage
+    } catch (e: DOMException | any) {
```
dai-shi, line 107:
```suggestion
    } catch (e) {
```
should work?

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -99,8 +99,18 @@ export function createJSONStorage<Value>(
   getStringStorage: () =>
     | AsyncStringStorage
     | SyncStringStorage
-    | undefined = () =>
-    typeof window !== 'undefined' ? window.localStorage : undefined,
+    | undefined = () => {
+    try {
+      const storage =
+        typeof window !== 'undefined' ? window.localStorage : undefined
+      return storage
```
dai-shi, line 106: Can we do this and catch "undefined" error below?
```suggestion
      return window.localStorage
```

tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -502,6 +502,44 @@ describe('atomWithStorage (with browser storage)', () => {
   })
 })
 
+describe('atomWithStorage (with disabled browser storage)', () => {
+  const localStorage = window.localStorage
```
dai-shi, line 506: Just a preference.
```suggestion
  const savedLocalStorage = window.localStorage
```

dai-shi (changes requested):
Hi! This looks like a nice improvement!
Please check comments.

dai-shi:
> On Firefox accessing the property localStorage of window results in a `DOMException: The operation is insecure.`

On Chrome, it's the same with slightly different message:
`DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

dai-shi:
@benediktschlager hi, are you still around?

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -99,8 +99,18 @@ export function createJSONStorage<Value>(
...
     | AsyncStringStorage
     | SyncStringStorage
-    | undefined = () =>
-    typeof window !== 'undefined' ? window.localStorage : undefined,
+    | undefined = () => {
+    try {
+      return window.localStorage
+    } catch (e) {
+      // Window could be undefined or local storage could be disabled
+      if (e instanceof TypeError || e instanceof DOMException) {
```
dai-shi, line 107: Hmm, this is what I tried in Node.js:
```
> try { window.localStorage } catch (e) { console.log(e instanceof TypeError) }
false
undefined
> try { window.localStorage } catch (e) { console.log(e instanceof ReferenceError) }
true
undefined
```
Maybe, it's not stable. On the other hand, when is the case we want to `throw e`?

benediktschlager:
Hi @dai-shi, I included your feedback.
Can you check again?
Some non-required steps in the pipeline are failing.
Is this fine, or should I add a _as any_ cast to the test to mute this error?

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -99,8 +99,18 @@ export function createJSONStorage<Value>(
...
+        return undefined
+      }
+
+      throw e
+    }
```
dai-shi, line 112: completely different idea:
```suggestion
    try {
      return window.localStorage
    } catch (e) {
      if (import.meta.env?.MODE !== 'production') {
        if (typeof window !== 'undefined') {
          console.warn(e);
        }
      }
      return undefined
    }
```

src/vanilla/utils/atomWithStorage.ts
benediktschlager: Oh, it seems on node the window instance is not even defined.
I'm wondering under which circumstances the window instance is not defined, but this library is used?
As far as I understand, Jotai only works alongside React?

tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -502,6 +502,44 @@ describe('atomWithStorage (with browser storage)', () => {
...
+    // Firefox and chromium based browser throw DOMException when cookies are disabled
+    Object.defineProperty(window, 'localStorage', {
+      get() {
+        throw new DOMException('The operation is insecure.')
+      },
+    })
+  })
+
+  afterAll(() => {
+    window.localStorage = savedLocalStorage
```
dai-shi, line 518:
```suggestion
    // TS < 4.5 causes type error
    (window as any).localStorage = savedLocalStorage
```

src/vanilla/utils/atomWithStorage.ts
dai-shi: SSR=Server Side Rendering.

dai-shi:
oops, can you run prettier please?

tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -502,6 +502,45 @@ describe('atomWithStorage (with browser storage)', () => {
...
+    // Firefox and chromium based browser throw DOMException when cookies are disabled
+    Object.defineProperty(window, 'localStorage', {
+      get() {
+        throw new DOMException('The operation is insecure.')
+      },
+    })
+  })
+
+  afterAll(() => {
+    // TS < 4.5 causes type error
```
dai-shi, line 518:
```suggestion
    // TS < 4.5 causes type error without `as any`
```

dai-shi (approved):
Thanks for your contribution!

PR #2356, opened 2024-01-23
feat(vanilla): customizable atom.unstable_is

dai-shi:
supersedes #2353

- introduce `atom.unstable_is`
- make default `read` & `write` functions stable

### DIFF
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -40,7 +40,7 @@ type OnMount<Args extends unknown[], Result> = <
 export interface Atom<Value> {
   toString: () => string
   read: Read<Value>
-  is?: (a: Atom<unknown>) => boolean
+  is?: (self: Atom<unknown>, target: Atom<unknown>) => boolean
   debugLabel?: string
   /**
    * To ONLY be used by Jotai libraries to mark atoms as private. Subject to change.
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -40,7 +40,7 @@ type OnMount<Args extends unknown[], Result> = <
 export interface Atom<Value> {
   toString: () => string
   read: Read<Value>
-  is?(a: Atom<unknown>): boolean
+  unstable_is?(a: Atom<unknown>): boolean
   debugLabel?: string
   /**
    * To ONLY be used by Jotai libraries to mark atoms as private. Subject to change.
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -40,6 +40,7 @@ type OnMount<Args extends unknown[], Result> = <
 export interface Atom<Value> {
   toString: () => string
   read: Read<Value>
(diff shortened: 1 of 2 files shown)


src/vanilla/atom.ts
```diff
@@ -40,6 +40,7 @@ type OnMount<Args extends unknown[], Result> = <
 export interface Atom<Value> {
   toString: () => string
   read: Read<Value>
+  is?: (a: Atom<unknown>) => boolean
```
dai-shi, line 43: I thought about requiring this property rather than making it optional, but as we want to release this as a patch, let's do this.
Eventually, if it makes sense (like moving to `class` syntax), we can consider changing it a required property.

yf-yang: A scenario needs to be considered: A global atom `anAtom` has two copies `copy1` and `copy2`, it is possible that `copy1` internally calls `set(copy2, value)`, so even with this change, we still needs to add another property in jotai-scope's intercepted copy to mark two copies as having the same original atom.

dai-shi: So, does it mean it requires to be `is?: (a: Atom<unknown>, b: Atom<unknown>) => boolean` and doesn't use `this`?

src/vanilla/atom.ts
```diff
@@ -40,6 +40,7 @@ type OnMount<Args extends unknown[], Result> = <
 export interface Atom<Value> {
   toString: () => string
   read: Read<Value>
+  is?: (self: Atom<unknown>, target: Atom<unknown>) => boolean
```
dai-shi, line 43: Hm, it doesn't seem to add any capability.
If it works, it should work with this too:
```suggestion
  is?: (this: Atom<unknown>, target: Atom<unknown>) => boolean
```

dai-shi:
I feel like adding `unstable_` prefix for now. 🤔

yf-yang:
I've made jotai-scope work with commit 3185a1. Once this PR is done, I'll give it another try.

src/vanilla/atom.ts
yf-yang: Well, for the jotai-scope PR, there is no need to do so.
Personally, I think the term should not be called `is`, as `is` suggests equality and equality is **symmetric**, while in this case it is not.
https://doc.rust-lang.org/std/cmp/trait.PartialEq.html

dai-shi:
Can you actually try the csb build in https://github.com/jotaijs/jotai-scope/pull/20 ?

https://ci.codesandbox.io/status/pmndrs/jotai/pr/2356/builds/463536
☝️ Find "Local Install Instructions"

I plan to merge this PR tomorrow, but it depends on other PRs in this repo.

yf-yang:
Done, ping me when this one get merged.

dai-shi:
@yf-yang https://www.npmjs.com/package/jotai/v/2.6.3

PR #2363, opened 2024-01-24
Improve performance of recomputeDependents

samkline:

N/A

## Summary

Hi. I've been working with @edkimmel on performance improvements to our app. It's a React Native ecommerce app that heavily relies on Jotai for managing most state and data. A hotspot that stood out recently was `recomputeDependents`—that's how Eddie identified the Hermes memory leak (https://github.com/pmndrs/jotai/discussions/2355) as a source of our problems.

Even with Eddie's workaround in place for that memory leak, we still found that `recomputeDependents` was very slow for us, so we took a closer look at what `recomputeDependents` is doing. It's something like this:

* loop1 is `O(n*m)` where `n` is the number of atoms and `m` is the number of dependency edges.
* loop2 is `O(n*n*m)` because it does the same thing as loop1 but also iterates over the `dependencyMap` for each iteration where the atom is unchanged (that is, roughly `n*m` times).

This can be improved by doing a topological sort of the atom dependency graph, which is `O(n)`. This PR changes the `recomputeDependents` implementation to do that topsort. All existing tests pass.

I ran through a test case for our app that represents a common customer flow. The total accumulated time in `recomputeDependents` is as follows:

| Platform | Time before | Time after | Percent improvement |
| --- | --- | --- | --- |
| Web/Chrome | 1,483ms | 295ms | 80% |
| iOS React Native (Hermes) | 3,581ms | 367ms | 90% | 
| Android React Native (Hermes) | 4,137ms | 402ms | 90% | 

Thanks again for all of your work on Jotai, we're big fans!

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -489,58 +489,64 @@ export const createStore = () => {
       return dependents
     }
 
-    // This is a topological sort via depth-first search, as described here:
+    // This is a topological sort via depth-first search, slightly modified from
+    // what's described here for performance reasons:
     // https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search
 
     // Step 1: find all atoms in the graph
     const allAtomsInDependencyGraph = new Array<AnyAtom>()
     const seen = new Set<AnyAtom>()
     const queue = [atom]
-    while (queue.length) {
-      const atom = queue.pop()!
-      if (!seen.has(atom)) {
-        seen.add(atom)
-        allAtomsInDependencyGraph.push(atom)
-        queue.push(...getDependents(atom))
+    while (true) {
+      const a = queue.pop()
+      if (!a) {
+        break
+      }
+      if (!seen.has(a)) {
+        seen.add(a)
+        allAtomsInDependencyGraph.push(a)
+        queue.push(...getDependents(a))
       }
     }
 
-    // Step 2: traverse the dependency graph to visit all atoms
+    // Step 2: traverse the dependency graph to build the topsorted atom list
+    // We don't bother to check for cycles, which simplifies the algorithm.
     const topsortedAtoms = new Array<AnyAtom>()
-    const permanentlyMarked = new Set<AnyAtom>()
-    const temporarilyMarked = new Set<AnyAtom>()
+    const markedAtoms = new Set<AnyAtom>()
     const visit = (n: AnyAtom) => {
-      if (permanentlyMarked.has(n)) {
+      if (markedAtoms.has(n)) {
         return
       }
-      if (temporarilyMarked.has(n)) {
-        throw new Error('atom graph has cycle')
-      }
-      temporarilyMarked.add(n)
+      markedAtoms.add(n)
       for (const m of getDependents(n)) {
         if (n !== m) {
           visit(m)
         }
       }
-      permanentlyMarked.add(n)
-      temporarilyMarked.delete(n)
+      // The algorithm calls for pushing onto the front of the list. Since
+      // Array.unshift is likely slower than pushing onto the end, we simply
+      // push and will have to iterate in reverse order later.
       topsortedAtoms.push(n)
     }
     while (allAtomsInDependencyGraph.length) {
       visit(allAtomsInDependencyGraph.pop()!)
     }
-    topsortedAtoms.reverse()
 
     // Step 3: use the topsorted atom list to recompute all affected atoms
     // Track what's changed, so that we can short circuit when possible
     const changedAtoms = new Set<AnyAtom>([atom])
-    for (const a of topsortedAtoms) {
+    for (let i = topsortedAtoms.length; i > 0; i--) {
+      const a = topsortedAtoms[i - 1]!
       const prevAtomState = getAtomState(a)
-      const hasChangedDeps =
-        prevAtomState?.d.size &&
-        Array.from(prevAtomState?.d.keys()).some(
-          (dep) => dep !== a && changedAtoms.has(dep),
-        )
+      let hasChangedDeps = false
+      if (prevAtomState?.d.size) {
+        for (const dep of prevAtomState?.d.keys()!) {
+          if (dep !== a && changedAtoms.has(dep)) {
+            hasChangedDeps = true
+            break
+          }
+        }
+      }
       if (hasChangedDeps) {
         const nextAtomState = readAtomState(a, true)
         if (!isEqualAtomValue(prevAtomState, nextAtomState)) {
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -500,7 +500,7 @@ export const createStore = () => {
       }
       markedAtoms.add(n)
       for (const m of getDependents(n)) {
-        if (!isSelfAtom(n, m)) {
+        if (n !== m) {
           visit(m)
         }
       }
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -517,8 +517,8 @@ export const createStore = () => {
     // Step 2: use the topsorted atom list to recompute all affected atoms
     // Track what's changed, so that we can short circuit when possible
     const changedAtoms = new Set<AnyAtom>([atom])
-    for (let i = topsortedAtoms.length; i > 0; i--) {
-      const a = topsortedAtoms[i - 1]!
+    for (let i = topsortedAtoms.length - 1; i >= 0; --i) {
+      const a = topsortedAtoms[i]!
       const prevAtomState = getAtomState(a)
       if (!prevAtomState) {
         continue
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -506,7 +506,7 @@ export const createStore = () => {
     // Step 2: traverse the dependency graph to build the topsorted atom list
     // We don't bother to check for cycles, which simplifies the algorithm.
     const topsortedAtoms = new Array<AnyAtom>(allAtomsInDependencyGraph.length)
-    let topsortedAtomsIndex = 0
+    let topsortedAtomsIndex = allAtomsInDependencyGraph.length
     const markedAtoms = new Set<AnyAtom>()
     const visit = (n: AnyAtom) => {
       if (markedAtoms.has(n)) {
@@ -518,10 +518,9 @@ export const createStore = () => {
           visit(m)
         }
       }
-      // The algorithm calls for pushing onto the front of the list. For
-      // simplicity, we simply append items in order, and will iterate
-      // in reverse order later.
-      topsortedAtoms[topsortedAtomsIndex++] = n
+      // The algorithm calls for pushing onto the front of the list, so we fill
+      // the array in reverse order.
+      topsortedAtoms[--topsortedAtomsIndex] = n
     }
     while (allAtomsInDependencyGraph.length) {
       visit(allAtomsInDependencyGraph.pop()!)
@@ -530,8 +529,7 @@ export const createStore = () => {
     // Step 3: use the topsorted atom list to recompute all affected atoms
     // Track what's changed, so that we can short circuit when possible
     const changedAtoms = new Set<AnyAtom>([atom])
-    for (let i = topsortedAtoms.length; i > 0; i--) {
-      const a = topsortedAtoms[i - 1]!
+    for (const a of topsortedAtoms) {
       const prevAtomState = getAtomState(a)
       let hasChangedDeps = false
       for (const dep of prevAtomState?.d.keys() || []) {
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -500,6 +500,7 @@ export const createStore = () => {
       }
       markedAtoms.add(n)
       for (const m of getDependents(n)) {
+        // we shouldn't use isSelfAtom here.
         if (n !== m) {
           visit(m)
         }
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -531,8 +531,11 @@ export const createStore = () => {
     const changedAtoms = new Set<AnyAtom>([atom])
     for (const a of topsortedAtoms) {
       const prevAtomState = getAtomState(a)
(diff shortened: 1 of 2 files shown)


src/vanilla/typeUtils.ts
samkline, line 1: The changes in this file are only from Prettier.

dai-shi: Does that mean you are using a different version/config of Prettier?

samkline: Whoops, you're right—I was somehow using an older version of Prettier. I just prettified using the correct version.

src/vanilla/store.ts
```diff
@@ -487,42 +485,72 @@ export const createStore = () => {
...
+
+    // This is a topological sort via depth-first search, slightly modified from
+    // what's described here for performance reasons:
+    // https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search
+
+    // Step 1: find all atoms in the graph
+    const allAtomsInDependencyGraph = new Array<AnyAtom>()
+    const seen = new Set<AnyAtom>()
+    const queue = [atom]
+    while (true) {
```
dai-shi, line 497: Can it be like `while (a = queue.pop())` ?

src/vanilla/store.ts
```diff
@@ -487,42 +485,72 @@ export const createStore = () => {
...
+      if (markedAtoms.has(n)) {
+        return
+      }
+      markedAtoms.add(n)
+      for (const m of getDependents(n)) {
+        if (n !== m) {
+          visit(m)
+        }
+      }
+      // The algorithm calls for pushing onto the front of the list. Since
```
dai-shi, line 523: I think we can create an array with size as we know the max size, and fill it from the front.

src/vanilla/store.ts
```diff
@@ -487,42 +485,72 @@ export const createStore = () => {
...
+
+    // Step 3: use the topsorted atom list to recompute all affected atoms
+    // Track what's changed, so that we can short circuit when possible
+    const changedAtoms = new Set<AnyAtom>([atom])
+    for (let i = topsortedAtoms.length; i > 0; i--) {
+      const a = topsortedAtoms[i - 1]!
+      const prevAtomState = getAtomState(a)
+      let hasChangedDeps = false
+      if (prevAtomState?.d.size) {
+        for (const dep of prevAtomState?.d.keys()!) {
```
dai-shi, line 540: Should be this?
```suggestion
        for (const dep of prevAtomState.d.keys()) {
```
Or maybe
```js
for (const dep of prevAtomState?.d.keys() || []) {
```
and remove `if (prevAtomState?.d.size) {`.

dai-shi (commented):
Thanks for working on this!
Yes, I was pretty sure that there should be better algorithm then loop1&loop2.
So great to see this.
Some comments below:

src/vanilla/store.ts
cltnschlosser:
```ts
const topsortedAtoms = new Array<AnyAtom>(allAtomsInDependencyGraph.length)
```
Is probably a good idea either way, but even then, I still think we want to avoid `unshift` here. I suppose this could keep track of an index an insert it that way:
```
const index = allAtomsInDependencyGraph.length

topsortedAtoms[--index] = n
```
But I'm not sure that is more clear than iterating `topsortedAtoms` in reverse in step 3.

dai-shi: I think I wasn't clear about my suggestion.
```js
const topsortedAtoms = new Array<AnyAtom>(allAtomsInDependencyGraph.length)
let topsortedAtomsIndex = 0
topsortedAtoms[topsortedAtomsIndex++] = n
```

src/vanilla/store.ts
samkline: Yes, updated

src/vanilla/store.ts
samkline: Good point, done

src/vanilla/store.ts
samkline: Yes, done

samkline:
Thanks for the quick review @dai-shi ! I updated the branch with your comments, and this should fix the lint violations as well.

src/vanilla/store.ts
```diff
@@ -520,10 +518,10 @@ export const createStore = () => {
           visit(m)
         }
       }
-      // The algorithm calls for pushing onto the front of the list. Since
-      // Array.unshift is likely slower than pushing onto the end, we simply
-      // push and will have to iterate in reverse order later.
-      topsortedAtoms.push(n)
+      // The algorithm calls for pushing onto the front of the list. For
+      // simplicity, we simply append items in order, and will iterate
+      // in reverse order later.
```
dai-shi, line 523: It's no longer in reverse order, right?
Oh, wait. That was my misunderstanding. So, your suggestion was correct. My bad.
I meant to fill it from the end...

samkline: No worries, I just updated it

src/vanilla/store.ts
```diff
@@ -487,42 +485,66 @@ export const createStore = () => {
...
+      visit(allAtomsInDependencyGraph.pop()!)
+    }
+
+    // Step 3: use the topsorted atom list to recompute all affected atoms
+    // Track what's changed, so that we can short circuit when possible
+    const changedAtoms = new Set<AnyAtom>([atom])
+    for (const a of topsortedAtoms) {
+      const prevAtomState = getAtomState(a)
+      let hasChangedDeps = false
+      for (const dep of prevAtomState?.d.keys() || []) {
```
dai-shi, line 535: On second thought, an early return may look better and performant.
```js
if (!prevAtomState) {
  continue
}
```

src/vanilla/store.ts
```diff
@@ -487,42 +485,66 @@ export const createStore = () => {
...
+      markedAtoms.add(n)
+      for (const m of getDependents(n)) {
+        if (n !== m) {
+          visit(m)
         }
-      })
+      }
+      // The algorithm calls for pushing onto the front of the list, so we fill
+      // the array in reverse order.
+      topsortedAtoms[--topsortedAtomsIndex] = n
```
dai-shi, line 523: Does `topsortedAtomsIndex` become `0` at last?

dai-shi (commented):
Looks great. A few minor comments:

src/vanilla/store.ts
samkline: Yes - I checked locally by putting an assertion in place and running the test suite, but let me know if you want me to commit it.

src/vanilla/store.ts
samkline: Yep, done

samkline:
For context on the latest commit (https://github.com/pmndrs/jotai/pull/2363/commits/4cb7b8361e137c1ea58f0f76794641bc5d48f889), see my comment in https://github.com/pmndrs/jotai/discussions/2368#discussioncomment-8274991:
> Nice find @cltnschlosser ! I just updated the PR https://github.com/pmndrs/jotai/pull/2363 with a fix. For those out of the loop, the issue applies when the code is transpiled to support older JS versions. The transpiled code has an issue with using Array.prototype.push with the Set iterator, such that the push does nothing. It worked fine in modern JS environments. This was the fix: https://github.com/pmndrs/jotai/commit/4cb7b8361e137c1ea58f0f76794641bc5d48f889

src/vanilla/store.ts
dai-shi: Thanks. No, if you checked it locally, you don't need to commit it.

dai-shi:
> For context on the latest commit ([4cb7b83](https://github.com/pmndrs/jotai/commit/4cb7b8361e137c1ea58f0f76794641bc5d48f889)), see my comment in [#2368 (reply in thread)](https://github.com/pmndrs/jotai/discussions/2368#discussioncomment-8274991):

Nice. Can you please add comment in source code so that we know it's the workaround for transpilation?

samkline:
All set @dai-shi , and I've rebased on top of the latest `main`

dai-shi (approved):
LGTM
Thanks so much for your contribution!

I will merge this when I prepare the next release.

samkline:
@dai-shi Could you please take one more look? I found another performance optimization, specifically, step 1 where we build `allAtomsInDependencyGraph` was entirely unnecessary. This is because `recomputeDependents` only ever has exactly one "root" atom in the graph (that is, one node without incoming edges), and we already know which atom it is. This is a nice simplification and avoids the code that had the transpilation issue earlier as well. The change is https://github.com/pmndrs/jotai/pull/2363/commits/94fe6137dae41846d77d8b762c150c5d81145674 .

src/vanilla/store.ts
```diff
@@ -487,42 +485,58 @@ export const createStore = () => {
...
+
+    // Visit the root atom. This is the only atom in the dependency graph
+    // without incoming edges, which is one reason we can simplify the algorithm
+    visit(atom)
+
+    // Step 2: use the topsorted atom list to recompute all affected atoms
+    // Track what's changed, so that we can short circuit when possible
+    const changedAtoms = new Set<AnyAtom>([atom])
+    for (let i = topsortedAtoms.length; i > 0; i--) {
+      const a = topsortedAtoms[i - 1]!
```
dai-shi, line 521: A tiny preference.
```suggestion
    for (let i = topsortedAtoms.length - 1; i >= 0; --i) {
      const a = topsortedAtoms[i]!
```

dai-shi (commented):
It looks simplified!

src/vanilla/store.ts
samkline: Sure, all set

src/vanilla/store.ts
```diff
@@ -487,42 +485,58 @@ export const createStore = () => {
...
+    // We don't bother to check for cycles, which simplifies the algorithm.
+    const topsortedAtoms = new Array<AnyAtom>()
+    const markedAtoms = new Set<AnyAtom>()
+    const visit = (n: AnyAtom) => {
+      if (markedAtoms.has(n)) {
+        return
+      }
+      markedAtoms.add(n)
+      for (const m of getDependents(n)) {
+        if (n !== m) {
```
dmaskasky, line 503: should this be `!isSelfAtom(n, m)`?

dai-shi: Yeah, probably. (Or `!isSelfAtom(m, n)`?)
But, jotai-scope seems to be working for now. cc @yf-yang
We need tests for `unstable_is`...

cltnschlosser: For what it’s worth, this line was brought from the previous implementation and it seems to be the a pretty common check in jotai (though I’m not sure why it’s necessary at all, seems like the set of dependents should never contain self, so we shouldn’t need to check for it. But it’s harmless enough.)

dai-shi: > this line was brought from the previous implementation

Yes, it's a leftover in #2356.

> seems like the set of dependents should never contain self

good point. can we simply remove the if statement?
(So, dependencies can contain self, but dependents can't, right? I mean that's true AFAIR.)

cltnschlosser: If dependencies can contain self, shouldn’t the opposite be true?

It didn’t seem important to remove it if unnecessary. So seems safer to just leave it there for now. I wouldn’t want proving whether we need or don’t need it to hold up this improvement.

dai-shi: I would like to remove it if possible. Let me check the code.

dai-shi: That was my misunderstanding. Dependents can contain self.

Let's do this for now.
```suggestion
        if (!isSelfAtom(n, m)) {
```

yf-yang: Intuitively it should be `!==`, not `isSelfAtom`, because the scoped atom copy could depend on the original atom internally. Empirically, I suggest we simply give it a try with all the jotai-scope examples. Just give me a commit id and I can try it.

yf-yang: 😂 latest commit 69b5557a does break jotai-scope, yes something needs to change.

dai-shi: yeah, I was skeptical about it. okay, let's revert.

src/vanilla/store.ts
```diff
@@ -487,42 +485,58 @@ export const createStore = () => {
...
+    // We don't bother to check for cycles, which simplifies the algorithm.
+    const topsortedAtoms = new Array<AnyAtom>()
+    const markedAtoms = new Set<AnyAtom>()
+    const visit = (n: AnyAtom) => {
+      if (markedAtoms.has(n)) {
+        return
+      }
+      markedAtoms.add(n)
+      for (const m of getDependents(n)) {
+        if (!isSelfAtom(n, m)) {
```
dai-shi, line 503:
```suggestion
        if (n !== m) {
```

src/vanilla/store.ts
yf-yang: @dai-shi [34a635a](https://github.com/pmndrs/jotai/pull/2363/commits/34a635a159045a4398163ab43e7203fffeb158a1) should be reverted, because [c991752](https://github.com/pmndrs/jotai/pull/2363/commits/c991752654bf382ec3d14e1fd85dd6cb5b048f9b) is the correct version.

yf-yang:
jotai-scope is good with 94e356b7

src/vanilla/store.ts
dai-shi, line 503:
```suggestion
        // we shouldn't use isSelfAtom here.
        if (n !== m) {
```

dmaskasky: > Intuitively it should be `!==`, not `isSelfAtom`, because the scoped atom copy could depend on the original atom internally. Empirically, I suggest we simply give it a try with all the jotai-scope examples. Just give me a commit id and I can try it.

I don't think this is true anymore. Scope uses the original atom as the key, so any scoped reference to that atom will always return the scoped version, even if its a self-reference.

dai-shi: sounds good.

PR #2367, opened 2024-01-25
refactor: modified Args type

ssi02014:
## Summary
I modified the `Args` type, which was defined in several places, to be uniform. 

## Check List

### DIFF
--- a/src/react/useAtom.ts
+++ b/src/react/useAtom.ts
@@ -8,7 +8,7 @@ import type {
 import { useAtomValue } from './useAtomValue.ts'
 import { useSetAtom } from './useSetAtom.ts'
 
-type SetAtom<Args extends any[], Result> = (...args: Args) => Result
+type SetAtom<Args extends unknown[], Result> = (...args: Args) => Result
 
 type Options = Parameters<typeof useAtomValue>[1]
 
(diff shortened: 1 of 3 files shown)


src/vanilla/atom.ts
```diff
@@ -5,9 +5,7 @@ type Setter = <Value, Args extends unknown[], Result>(
   ...args: Args
 ) => Result
 
-type SetAtom<Args extends unknown[], Result> = <A extends Args>(
-  ...args: A
-) => Result
```
ssi02014, line 10: After checking the usage of `SetAtom`, it seems that the `<A extends Args>` part is unnecessary.

src/react/useAtom.ts
```diff
@@ -8,7 +8,7 @@ import type {
 import { useAtomValue } from './useAtomValue.ts'
 import { useSetAtom } from './useSetAtom.ts'
 
-type SetAtom<Args extends any[], Result> = (...args: Args) => Result
+type SetAtom<Args extends unknown[], Result> = (...args: Args) => Result
```
ssi02014, line 11: I thought the `unknwon` was more stable than the `any` 
https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type

ssi02014: Also, `Args extends unknown[]` is already used in many places. 🤗

dai-shi (approved):
AFAIR, there were some cases that we need to use `any`, but if tests are passing, it should be okay. (Or, we will see issues reported later..)

PR #2371, opened 2024-01-31
fix(vanilla): fix unexpected cache in jotai-scope

yf-yang:

Fixes https://github.com/jotaijs/jotai-scope/issues/21

## Summary

Here `isSelfAtom` should not be used as it could lead to unexpected cache.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -377,7 +377,8 @@ export const createStore = () => {
       // If all dependencies haven't changed, we can use the cache.
       if (
         Array.from(atomState.d).every(([a, s]) => {
-          if (atom === a) {
+          // we shouldn't use isSelfAtom. https://github.com/pmndrs/jotai/pull/2371
+          if (a === atom) {
             return true
           }
           const aState = readAtomState(a)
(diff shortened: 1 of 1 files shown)


yf-yang:
The failed case can work with this patch.

dai-shi:
Hm, okay, this is tricky, but alright.

Can you check this https://github.com/pmndrs/jotai/pull/2363#discussion_r1472011121 too?

yf-yang:
Actually, I think `isSelfAtom` only matters when writing an atom (because we should write to the correct atom). For read(sub) purposes, since they are always connected via atom dependency, `isSelfAtom` does not matter from the correctness perspective (but maybe it could have some performance benefits)?

dai-shi:
> but maybe it could have some performance benefits

`isSelfAtom` is not only for `jotai-scope`, so some kind of "correctness" is important.

src/vanilla/store.ts
```diff
@@ -377,7 +377,7 @@ export const createStore = () => {
       // If all dependencies haven't changed, we can use the cache.
       if (
         Array.from(atomState.d).every(([a, s]) => {
-          if (isSelfAtom(atom, a)) {
+          if (atom === a) {
```
dai-shi, line 380:
```suggestion
          // we shouldn't use isSelfAtom. https://github.com/pmndrs/jotai/pull/2371
          if (a === atom) {
```

dai-shi (approved):
Thanks for your contribution! Looking forward to more suggestions.

PR #2378, opened 2024-02-04
chore: onUpdate

L-Qun:
## Summary
We have a requirement where certain actions are triggered when the state of an atom changes. Therefore, we added onUpdate, corresponding to onMount.

## Check List

### DIFF
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -37,6 +37,8 @@ type OnMount<Args extends unknown[], Result> = <
   setAtom: S,
 ) => OnUnmount | void
 
+type OnUpdate<Args extends unknown[]> = (get: Getter, ...args: Args) => void
+
 export interface Atom<Value> {
   toString: () => string
   read: Read<Value>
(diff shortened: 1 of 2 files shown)


AjaxSolutions:
> We have a requirement where certain actions are triggered when the state of an atom changes. Therefore, we added onUpdate, corresponding to onMount.

Why is it called onUpdate as opposed to onChange?

L-Qun:
> > We have a requirement where certain actions are triggered when the state of an atom changes. Therefore, we added onUpdate, corresponding to onMount.
> 
> Why is it called onUpdate as opposed to onChange?

I think calling it onChange would also be good.

L-Qun:
> I noticed that [Reatom](https://www.npmjs.com/package/@reatom/npm-react) atoms have the onChange method which returns a cleanup function.
> 
> `const cleanup = itemAtom.onChange((ctx, state) => setValue(state))`
> 
> Calling this cleanup function removes the callback from the atom's listeners(?)

These two are not the same. In Reatom, onChange is implemented by the framework, but onUpdate is implemented by the user themselves and does not require a return of onUpdate.

AjaxSolutions:
Thanks.  Also, the [doc](https://jotai.org/docs/core/atom#), should be updated.

dai-shi:
We specifically avoid adding `onUpdate` as a design principle #211. While we might add it in the future (Jotai v3?), for now, we suggest [jotai-effect](https://github.com/jotaijs/jotai-effect) which is developed on the current API.

PR #2380, opened 2024-02-05
add livecodes preview workflow

dai-shi:
based on:
- https://github.com/live-codes/preview-in-livecodes
- https://github.com/hatemhosny/preview-in-livecodes-demo

### DIFF
--- a/.github/workflows/livecodes-preview.yml
+++ b/.github/workflows/livecodes-preview.yml
@@ -1,0 +1,14 @@
+name: livecodes-preview
+
+on: [pull_request]
+
+jobs:
+  build_and_prepare:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: live-codes/preview-in-livecodes@v1
+        with:
+          install-command: "npm install"
+          build-command: "npm run build"
+          base-url: "https://{{LC::REF}}.preview-in-livecodes-demo.pages.dev"
--- a/.github/workflows/livecodes-preview.yml
+++ b/.github/workflows/livecodes-preview.yml
@@ -1,4 +1,4 @@
-name: livecodes preview
+name: livecodes
 
 on: [pull_request]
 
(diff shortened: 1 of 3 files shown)


dai-shi:
@hatemhosny Do you happen to know why comment workflow doesn't fire?

hatemhosny:
I think the comment workflow needs to be merged before it can run. It runs from the context of the PR base rather than head to avoid anybody creating a pull request to have access to post comments to your PR or gain access to secrets.

see https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

dai-shi:
Oh, yeah. Thanks!

hatemhosny:
if this is the case, I'll add a note about that in docs.

hatemhosny:
> Oh, yeah. Thanks!
> 
> Does it look good otherwise?

yes, otherwise looks good

.github/workflows/livecodes-preview.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  build_and_prepare:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: live-codes/preview-in-livecodes@v1
+        with:
+          install-command: "yarn install --frozen-lockfile --check-files"
+          build-command: "yarn build"
+          base-url: "https://{{LC::REF}}.preview-in-livecodes-demo.pages.dev"
```
hatemhosny, line 14: you may want to change this to:

```yml
base-url: "https://jotai-git-{{LC::REF}}-pmndrs.vercel.app"
```

in case you want to use `{{LC:TO_URL}}`, instead of data urls.

please make sure that CORS is enabled on your deployment to allow loading assets in the playground.

dai-shi: But we don't deploy anything for livecodes to vercel.app. I think DataURL is safer.

Issue #2240, opened 2023-11-07
Cann't access ﻿`atomWithStorage` value in spa initial


Innei:
## Summary

When I use `atomWithStorage`, the storage already has this value present, but when I use `jotaiStore.get(atom)` I only get the initialValue.

This is a spa scenario, not an ssr scenario. When atomWithStorage is executed, globally `window` exists.

```
import { createStore } from "jotai/vanilla";
import { atomWithStorage } from 'jotai/vanilla/utils'

if(!localStorage.getItem('theme')) {
  localStorage.setItem('theme', 'light') // set localstorage key value if not exist
}

const jotaiStore  = createStore()

const themeAtom = atomWithStorage('theme', undefined, window.localStorage, {
  // unstable_getOnInit: true         // this default value is `false`
})

const v = jotaiStore.get(themeAtom)

console.log(v)      // v undefined

```

But if I set `unstable_getOnInit` to `true`, it can get the correct value.

## Link to reproduction

https://codesandbox.io/s/serverless-thunder-s3gddk?file=/src/index.mjs

## Check List

Please do not ask questions in issues.

- [ ] I've already opened a [discussion](https://github.com/pmndrs/jotai/discussions) before opening this issue, or already discussed in other media.

Please include a minimal reproduction.

- [x] I've added a link to a [typescript playground](https://www.typescriptlang.org/play) or [codesandbox](https://codesandbox.io) with a minimal reproduction.

jiangmaniu:
The same problem is that the value of storage cannot be obtained at the time of initialization.

Innei:
> The same problem is that the value of storage cannot be obtained at the time of initialization.

The current solution is to use `unstable_getOnInit: true`. I think getOnInit should be a reasonable behavior in non-ssr scenarios.

dai-shi:
It's our design choice to have `getOnInit` false by default.

dai-shi:
That said, we should change `unstable_getOnInit` to `getOnInit` some time soon.

So, thanks for the feedback. Wish to get more feedback on it. How do you like the name `getOnInit`?

Innei:
> That said, we should change `unstable_getOnInit` to `getOnInit` some time soon.
> 
> So, thanks for the feedback. Wish to get more feedback on it. How do you like the name `getOnInit`?

Sounds great

ViktorQvarfordt:
[...]

PR #2391, opened 2024-02-10
docs: clarify getOnInit being false

henrikvilhelmberglund:

Fixes https://github.com/pmndrs/jotai/issues/2240 confusion

## Summary

Clarify that getOnInit is false by default and can be set to true in an SPA to get the stored value instead of the initial value.

## Check List

`yarn run prettier` caused 100 invisible changes so I didn't commit the result.


dai-shi (approved):
Thanks for your contribution!

PR #2386, opened 2024-02-09
fix(types): useAtom with one type argument for primitive atoms

dai-shi:
This has been asked several times. Hope this doesn't break some usages.

PR #2392, opened 2024-02-11
fix livecodes template

dai-shi:
https://github.com/pmndrs/jotai/pull/2386#issuecomment-1937380333

### DIFF
--- a/.livecodes/react.json
+++ b/.livecodes/react.json
@@ -14,7 +14,7 @@
     "content": "import { StrictMode, Suspense } from 'react';\nimport { createRoot } from 'react-dom/client';\nimport { atom, useAtom } from 'jotai';\n\nconst countAtom = atom(0);\n\nconst Counter = () => {\n  const [count, setCount] = useAtom(countAtom);\n  const inc = () => setCount((c) => c + 1);\n  return (\n    <>\n      {count} <button onClick={inc}>+1</button>\n    </>\n  );\n};\n\nconst App = () => (\n  <Suspense fallback=\"Loading...\">\n    <div className=\"App\">\n      <h1>Hello Jotai</h1>\n      <h2>Enjoy coding!</h2>\n      <Counter />\n    </div>\n  </Suspense>\n);\n\nconst rootElement = document.getElementById('root');\nconst root = createRoot(rootElement);\n\nroot.render(\n  <StrictMode>\n    <App />\n  </StrictMode>\n);\n"
   },
   "customSettings": {
-    "jotai commit sha": "{{LC::SHA}}",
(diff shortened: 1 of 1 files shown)


.livecodes/react.json
```diff
@@ -13,12 +13,15 @@
...
   },
-  "imports": {
-    "jotai": "{{LC::TO_DATA_URL(./dist/esm/index.mjs)}}",
-    "jotai/vanilla": "{{LC::TO_DATA_URL(./dist/esm/vanilla.mjs)}}",
-    "jotai/utils": "{{LC::TO_DATA_URL(./dist/esm/utils.mjs)}}",
-    "jotai/react": "{{LC::TO_DATA_URL(./dist/esm/react.mjs)}}",
-    "jotai/vanilla/utils": "{{LC::TO_DATA_URL(./dist/esm/vanilla/utils.mjs)}}",
-    "jotai/react/utils": "{{LC::TO_DATA_URL(./dist/esm/react/utils.mjs)}}"
+  "customSettings": {
+    "jotai commit number used": "can we show something here?",
```
dai-shi, line 17: @hatemhosny It doesn't have to be in `customSettings`, but is there a way to show the commit used in the imports?

hatemhosny, line 17: You can use `{{LC::SHA}}` for commit SHA

```
"jotai commit sha": "{{LC::SHA}}"
```

It can be used anywhere in the config JSON.
So, yes, it can be used in custom settings. Alternatively, in the HTML footer or may be console.log

dai-shi: Nice.
```suggestion
    "jotai commit sha": "{{LC::SHA}}",
```

hatemhosny:
> Alternatively, in the HTML footer or may be console.log

Please note that the editors (markup, style, script) can have a `hiddenContent` property that has code that runs without being shown in the code editor.

You may use this to log commit number without showing that code in the editor. 

https://livecodes.io/docs/configuration/configuration-object#markup

dai-shi:
Yeah, but I'm thinking about removing "imports" in custom settings, say, to reproduce a bug. So, it makes sense to have the sha close to "imports".

hatemhosny:
for convenience, you can now also use `{{LC::SHORT_SHA}}` to match the 7-character short SHA displayed on GitHub.

I have also [updated the docs](https://github.com/live-codes/preview-in-livecodes?tab=readme-ov-file#usage) to highlight the fact that [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) event is only triggered from the default branch.

Thank you

hatemhosny:
sorry, fixed short sha! 
It should work now.
you may want to re-run the action.

PR #2393, opened 2024-02-12
feat: atomWithRefresh

### LINKED ISSUE #2379: RFC: Add `atomWithRefresh` in `jotai/utils`
`atomWithRefresh` has been a recipe, not a util.
We wanted to understand various use cases and thought the naive implementation was somewhat limited. My hesitation is also it can be overused, but that's probably a doc issue.

It's been a while and it seems the basic `atomWithRefresh` seems valuable. https://github.com/pmndrs/jotai/discussions/849#discussioncomment-8359658

There might be several implementation ideas, like using a symbol like `atomWithReset` if that helps.

So, here's an RFC. Should we have `atomWithRefresh` in `jotai/utils`? 👍  / 👎

dai-shi:
close #2379

- [x] implementation
- [x] tests
- [x] docs


docs/utilities/resettable.mdx
```diff
@@ -146,3 +146,62 @@ is no longer used and any change in dependencies atoms will not trigger an updat
...
+
+Passing zero arguments to `set` will refresh.
+Passing one or more arguments to `set` will call "write" function.
+
+### Example
+
+Here's how you'd use it to implement an refresh-able source of data:
+
+```js
+import { atomWithRefresh } from 'XXX'
```
arronKler, line 179:
```
import { atomWithRefresh } from 'jotai/utils'
```

dai-shi: Would you send a PR?

PR #2397, opened 2024-02-13
docs(effect): add recipe for useAtomEffect

dmaskasky:
## Summary
Adds a doc for useAtomEffect under the doc [Custom useAtom hooks](https://jotai.org/docs/recipes/custom-useatom-hooks).

## Check List

### DIFF
--- a/docs/recipes/use-atom-effect.mdx
+++ b/docs/recipes/use-atom-effect.mdx
@@ -1,0 +1,64 @@
+---
+title: useAtomEffect
+nav: 8.03
+keywords: effect, atom effect, side effect, side-effect, sideeffect, hook, useAtomEffect
+---
+
+> `useAtomEffect` runs side effects in response to changes in atoms or props using [atomEffect](../extensions/effect.mdx).
+
+The effectFn reruns whenever the atoms it depends on change or the effectFn itself changes. Be sure to memoize the effectFn if it's a function defined in the component.
+
+⚠️ Note: always prefer to use a [stable version of useMemo and useCallback](https://github.com/alexreardon/use-memo-one) to avoid extra atomEffect recomputations. You may rely on useMemo as a performance optimization, but not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components.
+
+```ts
+import { useMemoOne as useStableMemo } from 'use-memo-one'
+import { useAtomValue } from 'jotai/react'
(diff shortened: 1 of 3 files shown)


docs/recipes/useAtomEffect.mdx
```diff
@@ -0,0 +1,66 @@
+---
+title: useAtomEffect
```
dai-shi, line 2: Do you want to add a link to `extensions/effect.mdx`?
Or, I was thinking of adding this recipe there.

dmaskasky: A link would be good.

dai-shi (commented):
You commit two files.

dmaskasky:
> You commit two files.

Fixed.

docs/recipes/use-atom-effect.mdx
```diff
@@ -0,0 +1,66 @@
...
+import { useMemo } from 'react'
+import { useAtomValue } from 'jotai/react'
+import { atomEffect } from 'jotai-effect'
+
+type CleanupFn = () => void
+
+type EffectFn = (
+  get: Getter & { peek: Getter },
+  set: Setter & { recurse: Setter },
+) => CleanupFn | void
```
dai-shi, line 21: Can this be:
```ts
type EffectFn = Parameters<typeof atomEffect>[0]
```

docs/recipes/use-atom-effect.mdx
```diff
@@ -0,0 +1,66 @@
...
+  (channelId: string) => {
+    return atom(new Channel(channelId))
+  },
+)
+const messagesAtom = atom<Message[]>([])
+
+function Messages({ channelId }: { channelId: string }) {
+  const [messages] = useAtom(messagesAtom)
+  useAtomEffect(
+    useCallback(
```
dai-shi, line 45: Maybe we should add a small note about useCallback/useMemo that they aren't for semantic guarantee by React. So, it **may** run the effect twice **in the future**. (But, if it really happens, we could just rewrite the recipe using useRef based memo/callback. For example: https://github.com/alexreardon/use-memo-one )

dmaskasky: Good point. I think the recipe should use the stable useCallback from use-memo-one and include this general note on useCallback.

dai-shi (approved):
LGTM

Now, it makes more sense to be a recipe instead of a library util.

PR #2417, opened 2024-02-23
fix(utils): make 'unwrap' and 'loadable' able to access async atom's resolved value immediately

iwoplaza:
## Summary

In our project, we had the need to await on a secure store to load proper auth information before initializing the rest of the app, and as soon as that atom resolved, we had to check whether we were previously logged in, or logged out.

```ts
const secureAuthStateAtom = atom(/* syncing with the async secure store */);

const userIdentityAtom = atom(/* fetching user identity from our backend, makes sure the auth information is correct */);

function usePrepare(onReady: () => void) {
  const store = useStore();

  useEffect(() => {
    // making sure the effect is ran only once
    // ...
    
    (async () => {
      await store.get(secureAuthStateAtom);

      // UNEXPECTED BEHAVIOR HERE, returns 'undefined', the pending value, not the resolved value.
      const userIdentity = store.get(unwrap(userIdentityAtom));

      if (userIdentity) {
        // we are logged in
      } else {
        // we are not logged in
      }
    })();
  }, []);
}
```

The same happens with the `loadable` utility. I managed to write a test for it and implement a fix, hope it does not stray to heavy into the 'experimental' promise usage, or does not break anything else unintentionally.

## Check List


dai-shi:
Thanks for the suggestion. Yeah, it depends on the internal promise behavior, but I think it's a good capability to add.
Unfortunately, the coding style is slightly off from my preference (and, especially, we can't let react depend on utils, because it's a different bundle), so let me add some commit for refactor.

dai-shi (approved):

iwoplaza:
No worries! I failed to notice the problem with separate bundles. Thanks for a fast response 🐸

PR #2419, opened 2024-02-25
build: change rollup target to ES2018

amirhhashemi:

https://github.com/pmndrs/jotai/issues/48

## Summary

Jotai causes a very hard to debug syntax error in browsers that don't support ES2019. I tracked down the issue to the two `catch` blocks that use [Optional catch Binding](https://www.javascripttutorial.net/es-next/optional-catch-binding). Optional catch Binding was introduced in ES2019 and is not supported in Chrome 65 and older. This PR fixes that.

Browser support: https://caniuse.com/mdn-javascript_statements_try_catch_optional_catch_binding

## Check List

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -33,10 +33,11 @@ function getBabelOptions(targets) {
   }
 }
 
-function getEsbuild(target, env = 'development') {
+function getEsbuild(env = 'development') {
   return esbuild({
     minify: env === 'production',
-    target,
+    target: 'es2018',
+    supported: { 'import-meta': true },
     tsconfig: path.resolve('./tsconfig.json'),
   })
 }
@@ -78,7 +79,7 @@ function createESMConfig(input, output, clientOnly) {
         delimiters: ['\\b', '\\b(?!(\\.|/))'],
         preventAssignment: true,
       }),
-      getEsbuild('node12'),
+      getEsbuild(),
       banner2(() => clientOnly && cscComment),
     ],
   }
@@ -157,7 +158,7 @@ function createSystemConfig(input, output, env, clientOnly) {
         delimiters: ['\\b', '\\b(?!(\\.|/))'],
         preventAssignment: true,
       }),
-      getEsbuild('node12', env),
+      getEsbuild(env),
       banner2(() => clientOnly && cscComment),
     ],
   }


vincentfretin:
Probably not related to this, I coincidentally just got a sentry error of a user with an iPad with iOS 12.5.7 so probably one of the 10 years old devices like iPad Air, mini 2 or mini 3 that can't update to iOS 13.
I asked gpt-4:
The optional catch binding feature in JavaScript, which allows omitting the parentheses and the error parameter in a catch block when the error is not used, was introduced in ECMAScript 2019 (ES10). Safari started supporting this feature starting from Safari 13.

dai-shi (approved):
Thanks!
Yeah, I generally avoid omitting `()` to catch, but forgot in these places.
I wonder if there's an eslint rule to prohibit it.

dai-shi:
@amirhhashemi 
That's a good insight actually.
We should probably change the target.
This seems to work. Can you try it, and change this PR?

```diff
diff --git a/rollup.config.js b/rollup.config.js
index 66c8f0e..afd185f 100644
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -33,10 +33,11 @@ function getBabelOptions(targets) {
   }
 }
 
-function getEsbuild(target, env = 'development') {
+function getEsbuild(env = 'development') {
   return esbuild({
     minify: env === 'production',
-    target,
+    target: 'es2018',
+    supported: { 'import-meta': true },
     tsconfig: path.resolve('./tsconfig.json'),
   })
 }
@@ -78,7 +79,7 @@ function createESMConfig(input, output, clientOnly) {
         delimiters: ['\\b', '\\b(?!(\\.|/))'],
         preventAssignment: true,
       }),
-      getEsbuild('node12'),
+      getEsbuild(),
       banner2(() => clientOnly && cscComment),
     ],
   }
@@ -157,7 +158,7 @@ function createSystemConfig(input, output, env, clientOnly) {
         delimiters: ['\\b', '\\b(?!(\\.|/))'],
         preventAssignment: true,
       }),
-      getEsbuild('node12', env),
+      getEsbuild(env),
       banner2(() => clientOnly && cscComment),
     ],
   }
```

dai-shi (changes requested):
https://github.com/pmndrs/jotai/pull/2419#issuecomment-1963327016

amirhhashemi:
Yeah I think it works.

Before:
```sh
$ yarn run build
$ grep -o 'catch{' ./dist/**/*.js
./dist/system/vanilla/utils.production.js:catch{
./dist/system/vanilla/utils.production.js:catch{
./dist/system/vanilla/utils.production.js:catch{
```
After:
```sh
$ yarn run build
$ grep -o 'catch{' ./dist/**/*.js
# There is no 'catch{` in the build output
```

dai-shi:
btw you want to grep `catch {`.

dai-shi (approved):
Thanks for the change.
(We should probably do the same for Zustand and Valtio.)

amirhhashemi:
> btw you want to grep `catch {`.

Isn't the build output minified?

dai-shi:
No, for esm and cjs builds, we don't minify.

amirhhashemi:
I tried `catch {` and the result is the same:
Before:
```sh
$ grep -o 'catch {' ./dist/**/*.js
./dist/system/vanilla/utils.development.js:catch {
./dist/system/vanilla/utils.development.js:catch {
```
After:
```sh
$ yarn run build
$ grep -o 'catch {' ./dist/**/*.js
# There is no 'catch {` in the build output
```

dai-shi:
Interesting. I get this:
```
$ ag 'catch {' dist|cat        
dist/esm/vanilla/utils.mjs:355:          } catch {
dist/esm/vanilla/utils.mjs:391:          } catch {
dist/system/vanilla/utils.development.js:379:                } catch {
dist/system/vanilla/utils.development.js:415:                } catch {
```

amirhhashemi:
Wired. I tried it again and it seems to be working. Did you run `yarn run build` before grep?

dai-shi:
This isn't super important, but just curious.

I run `yarn build` again and here's the result:
```sh
$ grep 'catch {' dist/esm/vanilla/utils.mjs 
          } catch {
          } catch {
```

dai-shi:
> (We should probably do the same for Zustand and Valtio.)

https://github.com/pmndrs/zustand/pull/2361
https://github.com/pmndrs/valtio/pull/863

PR #2426, opened 2024-02-28
fix(website): hide menu overlay properly

rothsandro:

Fixes #2425

## Summary

The mobile menu was hidden using `opacity: 0` which doesn't hide it for keyboard users and screen readers and also breaks scrolling of the navigation in Safari.

The menu is no properly hidden when closed and on desktop. The `opacity` is still used for the transition.

## Check List


sandren (approved):
Thanks! This is a better practice. 🙂

PR #2435, opened 2024-03-03
[BREAKING] selectAtom does not resolve promises internally

dmaskasky:
https://discord.com/channels/740090768164651008/1213966054125084672

## Summary
When selectAtom is based on an atom whose value was originally a promise but was later changed to a value, selectAtom would still return a promise. This is because the original calculation of slice is a promise, so all subsequent calculations of slice must also be promises.

[BREAKING] In this PR selectAtom does not internally resolve promises. Instead this must be handled outside of the utility.

### `selectAtom`
`selectAtom` will no longer internally unwrap promises. To migrate to the new api, use the `unwrap` utility from `jotai/utils` package.
```ts
// suppose we have this
const baseAtom = atom(Promise.resolve({ id: 0, name: 'test' }))

// previously selectAtom would internally unwrap promises.
const idAtom = selectAtom(
  baseAtom,
  ({ name }) => name,
  (prev, curr) => prev.id === curr.id
)

// instead, you need to import `unwrap` from 'jotai/utils' and pass the unwrapped atom 
import { unwrap } from 'jotai/utils'
  ...
const idAtom = selectAtom(
  unwrap(baseAtom),
  ({ name }) => name,
  (prev, curr) => prev.id === curr.id
)
```

## Check List


dmaskasky:
Recalculating downstream dependencies for new promises is an intentional and very important part of the jotai dependency handling behavior. Closing this PR.

src/vanilla/utils/selectAtom.ts
```diff
@@ -29,6 +29,11 @@ export function selectAtom<Value, Slice>(
   return memo3(
     () => {
       const EMPTY = Symbol()
+      const refAtom = atom(() => ({
```
dai-shi, line 32: We used to use `refAtom` to implement `selectAtom` and it's improved since then. We should be able to implement without `refAtom`. Otherwise, it comes with more hacks.
BTW, using `promise.status` might be a better hack.

tests/react/vanilla-utils/selectAtom.test.tsx
```diff
@@ -231,3 +231,35 @@ it('equality function works even if suspend', async () => {
...
 })
+
+it.only('should not return async value when the base atom values are synchronous', async () => {
+  expect.assertions(4)
+  type Base = { id: number; value: number }
+  const initialBase = Promise.resolve({ id: 0, value: 0 })
+  const baseAtom = atom<Base | Promise<Base>>(initialBase)
+  const idAtom = selectAtom(
+    baseAtom,
+    ({ id }) => id,
```
dai-shi, line 242: Or, I wonder if resolving promises inside selectAtom was a good idea...

dmaskasky: I was wondering that too. It would be a breaking change if we revert this behavior.

IMO, not resolving promises is not a very good DX. You still have the problem of promises being contagious

```ts
const baseAtom = atom(Promise.resolve({ id: 0, value: 0 }))

const derivedAsyncAtom = selectAtom(baseAtom,
  async (valueMaybePromise, sliceMaybePromise) => {
    const value = await valueMaybePromise
    const slice = await sliceMaybePromise
    ...do something with slice and value
  }
)
```

src/vanilla/utils/selectAtom.ts
dmaskasky: there is potentially a weakmap solution. I need to think more on this.

dai-shi: I would probably prefer the weakmap solution.

tests/react/vanilla-utils/selectAtom.test.tsx
dai-shi: > not a very good DX

That's what I thought too, but for `jotai/utils` we want straightforward impls (and open space for better DX for `jotaijs/*`.)

> It would be a breaking change if we revert this behavior.

Yeah, that's the problem. Not sure if we can provide a migration path (=only deprecate the behavior with warning, and remove it in the next major.)

dmaskasky: In this case, we should either:
A. create a weak map solution for replacing promises with resolved values
B. deprecate resolving promises inside selectAtom

What should we do?

dai-shi: Yes, and at this point, both options sound valid. I would like to take B, depending on the hackiness with the implementation of A, which is likely?

Yeah, so my current proposition is B, which is the philosophy of Jotai core (along with atomWithDefault change #1939). Well, I think we can and did add breaking changes with minor bump for `jotai/utils`. https://github.com/pmndrs/jotai/releases/tag/v2.2.0
We assume a theoretical bug of `selectAtom` resolving promises.

dmaskasky: The weakmap solution is ready for your review. Please let me know your final decision here. I'm ok either way.

dmaskasky: for my own knowledge, what is the issue with a refAtom?
atomEffect uses a ref, if its an issue, I'll update it.

dai-shi: It's not an issue, if it's done properly, but it's a hack and we want to avoid hacks as much as possible.

dai-shi (commented):
> The weakmap solution is ready for your review. Please let me know your final decision here. I'm ok either way.

Let's go with B. deprecate resolving promises inside selectAtom, with a minor bump.

This complexity shouldn't be in `jotai/utils`. We may consider supporting a new util for promise-capable selectAtom alternative in `jotaijs/*`. (But, I don't think it's very useful, tbh. I might be wrong.)

src/vanilla/utils/selectAtom.ts
```diff
@@ -29,6 +31,12 @@ export function selectAtom<Value, Slice>(
   return memo3(
     () => {
       const EMPTY = Symbol()
+      const initState = () => ({
+        version: 0,
+        prev: EMPTY as Slice | typeof EMPTY | Promise<Slice>,
+      })
+      const ref = getCached(initState, stateCache, anAtom)
```
dai-shi, line 38: This is shared across different Providers/stores.

dai-shi (commented):
I forgot to add this comment in the previous comment.

src/vanilla/utils/selectAtom.ts
dmaskasky: resolves in a follow-up commit, but then replaced by a commit that removes this approach entirely.

dmaskasky:
```ts
const baseAtom = atom(0)
const selectedAtom = selectAtom(
  baseAtom,
  (v) => v,
  (a, b) => a === b,
) 

selectAtom<number, number>(
  anAtom: Atom<number>,
  selector: (v: number, prevSlice?: number | undefined) => number,
  equalityFn?: ((a: number, b: number) => boolean) | undefined
): Atom<number>
```

```ts
const baseAtom = atom(Promise.resolve(0))
const selectedAtom = selectAtom(
  baseAtom,
  (v) => v,
  (a, b) => a === b,
) 

selectAtom<
  Promise<number>,
  Promise<number>
>(
  anAtom: Atom<Promise<number>>,
  selector: (v: Promise<number>, prevSlice?: Promise<number> | undefined) => Promise<number>,
  equalityFn?: ((a: Promise<number>, b: Promise<number>) => boolean) | undefined
): Atom<Promise<number>>```

File: src/vanilla/utils/selectAtom.ts [hunk 4]
```diff
@@ -15,39 +15,51 @@ const memo3 = <T>(
...
       ]) => {
         if (prevSlice === EMPTY) {
           return selector(value)
         }
         const slice = selector(value, prevSlice)
-        return equalityFn(prevSlice, slice) ? prevSlice : slice
+        const areEqual = equalityFn(prevSlice, slice)
+        if (areEqual instanceof Promise) {
+          return areEqual.then((areEqual) => (areEqual ? prevSlice : slice))
+        }
```
dai-shi, line 54: Are we still resolving promises for equalityFn? I think it's one of complexity we want to avoid. (And, I think we'd generally recommend to use `unwrap` for promise-usage with `selectoAtom`.)

dmaskasky: > Are we still resolving promises for equalityFn? I think it's one of complexity we want to avoid. (And, I think we'd generally recommend to use `unwrap` for promise-usage with `selectoAtom`.)

Well, the main difference is that for equalityFn to work, the user would need to await the prevSlice if it were a promise.

dai-shi: As we are making breaking changes anyway, I hope to drop that capability too and recommend `unwrap`ping promises in advance.

src/vanilla/utils/selectAtom.ts
```diff
@@ -17,37 +17,35 @@ const memo3 = <T>(
...
+        Value,
         Slice | typeof EMPTY,
       ]) => {
         if (prevSlice === EMPTY) {
           return selector(value)
         }
         const slice = selector(value, prevSlice)
-        return equalityFn(prevSlice, slice) ? prevSlice : slice
+        const areEqual = equalityFn(prevSlice, slice)
+        return areEqual ? prevSlice : slice
```
dai-shi, line 41: Let's revert.

dai-shi (approved):
Yay! Looks far cleaner and more maintainable.

dai-shi:
Would you please write a migration guide in the PR description?
It will be in the release note later.
https://github.com/pmndrs/jotai/releases/tag/v2.2.0 is an example, but it's a minimal guide. More descriptions would be nice.

PR #2447, opened 2024-03-11
fix(utils): make 'atomWithStorage' behave as expected when store changes.

iwoplaza:
## Summary

We recreate the store that is provided to the React component tree when logging out of the app to clear cache. When using `atomWithStorage` (with `getOnInit=true`), recreating the store causes the atom to revert to its initial value, bound at atom creation.

```ts
// the current value of `localStorage.getItem('app-theme')` gets bound, e.g. { primaryColor: 'blue' }.
const storedThemeAtom = atomWIthStorage('app-theme', null, undefined, { getOnInit: true });

// updating the value of the atom, therefore updating the storage
setAnAtom({ primaryColor: 'red' });

// recreating the store
// ...
///

// `storedThemeAtom` now holds the same value as when the app initialized, therefore being desynced from the Storage.
```

I am not sure whether this was the desired behavior, or if the `getOnInit=true` is just supposed to retrieve the data sooner, but still be synced with the inherent Storage when it is possible.

## Fixes
- The base atom within `atomWithStorage` is wrapped within an atom, meaning the initial value is part of the store, not the atom definition.

## Check List

### DIFF
--- a/src/vanilla/utils/atomWithStorage.ts
+++ b/src/vanilla/utils/atomWithStorage.ts
@@ -199,30 +199,38 @@ export function atomWithStorage<Value>(
   options?: { getOnInit?: boolean },
 ): any {
   const getOnInit = options?.getOnInit
-  const baseAtom = atom(
-    getOnInit
-      ? (storage.getItem(key, initialValue) as Value | Promise<Value>)
-      : initialValue,
-  )
 
-  if (import.meta.env?.MODE !== 'production') {
-    baseAtom.debugPrivate = true
-  }
+  // wrapping in an atom to reinitialize per store.
+  const baseWrapperAtom = atom(() => {
+    const baseAtom = atom(
+      getOnInit
+        ? (storage.getItem(key, initialValue) as Value | Promise<Value>)
+        : initialValue,
+    )
 
-  baseAtom.onMount = (setAtom) => {
-    if (!getOnInit) {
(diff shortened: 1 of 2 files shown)


dai-shi:
I think this is by design or a limitation with `getOnInit`. ("Init" here means when atom definition is created, not an atom is first used in a store. And, it's another reason why getOnInit is false by default.)
We prefer simpler implementation and this is probably overcomplicated for `jotai/utils` thing.

You can always create your own version of atomWithStorage, but maybe we should document this limitation and show some workarounds in docs.
A workaround would be to use `useEffect`, or maybe `useHydrateAtom` would work with `getOnInit: true`.
Another might be wrapping an atom created with atomWithStorage, with another atom.

Closing this, but feel free to continue discussion here.

iwoplaza:
I see, I was wondering whether that was the desired behavior or a use-case that was not explored. May I ask, what is the main reason not to populate the initial value of the atom with the storage by default? (like having getOnInit=true, but also updating the value in onMount. Currently it does not update its value in onMount if getOnInit is true).

dai-shi:
> like having getOnInit=true, but also updating the value in onMount

That's possible. I thought it would be exclusive, but yeah, that change would be acceptable. As long as it doesn't cause double render, it should be fine. It depends on a storage impl?
Would you like to open another PR?

PR #2448, opened 2024-03-12
fix(store2): "stuck" derived atoms where some derivations may never resolve

backbone87:
Just a failing test case for:
When an async derived atom picks up a dep asynchronously that never resolves, the derivation never settles even when a subsequent update of the dep does resolve.

I added another test case that demonstrates that picking up the dep synchronously does not lead to this issue


dai-shi:
Is it even fixable? Feel free to try.

backbone87:
the problem is that async deps are only recognized by store once the derivation settles. imho all deps should be recognized and tracked as soon as they occur. a consistent read would also be nice, e.g.
```
async (get) => {
  const a = get(anAtom);
  await delay(someTime);
  assert(a === get(anAtom));
}
```

dai-shi:
that's kind of by design: collect deps once in sync and once when settled. it's a computational task, and so it might require some tricks to cover more cases.
So, I'd say, it's currently a design limitation.

dai-shi (approved):
Thanks for the nice tests. The "sync" version actually reveals a bug in store2.ts.

PR #2450, opened 2024-03-12
fix(vanilla): store should notify subscriber if subscribing causes the value to change

iwoplaza:
## Summary

Not sure if this was the desired behavior, but currently, if we create an atom that changes their value synchronously in `onMount`, and we subscribe to that atom's changes for the first time, the listener does not get executed.

Given the atom:
```ts
const anAtom = atom(0)
anAtom.onMount = (setAtom) => {
  setAtom(1)
}
```

```ts
// hoping to keep `value` in sync with the atom
let value = store.get(anAtom)
store.sub(anAtom, () => {
  // ideally, getting all subsequent changes to the value of `anAtom`.
  value = store.get(anAtom)
})
```

The code above is not a foolproof solution to tracking the atom's value in the described situation. We would have to do:
```ts
store.sub(anAtom, () => {
  value = store.get(anAtom)
})
// place the initial value read here, after subscribing
let value = store.get(anAtom)
```

I found this behavior a bit confusing, but it might be desired. Let me know if this is something you would like to change.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -473,10 +473,10 @@ export const createStore = () => {
   const readAtom = <Value>(atom: Atom<Value>): Value =>
     returnAtomValue(readAtomState(atom))
 
-  const addAtom = (atom: AnyAtom): Mounted => {
+  const addAtom = (atom: AnyAtom, initialListener?: () => void): Mounted => {
     let mounted = mountedMap.get(atom)
     if (!mounted) {
-      mounted = mountAtom(atom)
+      mounted = mountAtom(atom, initialListener)
     }
     return mounted
   }
@@ -610,6 +610,7 @@ export const createStore = () => {
 
   const mountAtom = <Value>(
     atom: Atom<Value>,
+    initialListener?: () => void,
     initialDependent?: AnyAtom,
     onMountQueue?: (() => void)[],
   ): Mounted => {
@@ -621,12 +622,12 @@ export const createStore = () => {
         aMounted.t.add(atom) // add dependent
       } else {
         if (a !== atom) {
-          mountAtom(a, atom, queue)
+          mountAtom(a, undefined, atom, queue)
         }
       }
     })
     // recompute atom state
-    readAtomState(atom)
+    const preMountState = readAtomState(atom)
     // mount self
     const mounted: Mounted = {
       t: new Set(initialDependent && [initialDependent]),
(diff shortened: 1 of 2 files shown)


src/vanilla/store.ts
```diff
@@ -610,6 +610,7 @@ export const createStore = () => {
 
   const mountAtom = <Value>(
     atom: Atom<Value>,
+    initialListener?: () => void,
     initialDependent?: AnyAtom,
     onMountQueue?: (() => void)[],
```
dai-shi, line 615: Does this work?
```suggestion
    initialDependent?: AnyAtom,
    onMountQueue?: (() => void)[],
    initialListener?: () => void,
```

Hm, on second thought, this should be a task of `addAtom`. Have you tried it?

dai-shi (commented):
This is not necessary for `jotai/react`, because it always calls `callback` after subscribing.
So, the question is only about vanilla api. If both behaviors make sense, we'd prefer the simpler one.
Can we say current behavior is correct, if we say `onMount` runs before the subscription?

src/vanilla/store.ts
iwoplaza: The problem with pulling it up into `addAtom` is that we need to check whether the value of the atom changed after `onMount`, and if we read the atom's value before we mount its dependents, then the mount order will be changed, and will fail other tests.

dai-shi: can we peek the prev value? aka `getAtomState`?

iwoplaza: If the atom was not read before, then the state would be missing, if I understand it correctly. So we would run the callback even if `onMount` did not change the atom's initial value.

dai-shi: okay, I think then this PR isn't what I would do.

I would agree this might not be ergonomic, 
```js
// to keep `value` in sync with the atom
let value = store.get(anAtom)
const cb = () => {
  value = store.get(anAtom)
}
store.sub(anAtom, cb)
// fire it first after the subscription
cb()
```

but, it's in line with our binding:
https://github.com/pmndrs/jotai/blob/bae472500dea29ba876d80b8c066c54816970862/src/react/useAtomValue.ts#L97

dai-shi:
> This is not necessary for `jotai/react`

It also means, this introduce an extra task, which can worsen the performance with React usage.

iwoplaza:
Definitely understand, no worries then!

PR #2451, opened 2024-03-12
fix(vanilla): store should flush pending write triggered asynchronously and indirectly

iwoplaza:

Sorry for the onslaught of PRs, I believe this one to be an actual fix to a regression caused by the #2396 refactor

## Summary

For this bug to occur a few things have to happen:
1. An atom that we wish to update indirectly, e.g.
```ts
const emailAtom = atom('user@example.com')
```
2. An atom that simply writes to the atom above in its write function, e.g.:
```ts
const updateEmailAtom = atom(null, (_get, set, value: string) => set(emailAtom, value.trim()))
```
3. An atom that executes the above atom asynchronously, e.g.:
```ts
const initializeAndVerifyAtom = atom(null, async (_get, set, fields: FormFields) => {
  const email = fields.email
  await verifyEmail(email)
  
  // this write is no longer sync, so should be flushed immediately, but the
  // actual atom that will get set (emailAtom) thinks it is being set synchronously,
  // so it just adds itself to the non-existent top of `pendingStack`.
  set(updateEmailAtom, email)
})
```

Instead of tracking `isSync` in `writeAtomState`, we can use the length of `pendingStack` to determine if the atom we are currently updating will get flushed in a batch, or do we have to handle the flushing immediately.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -565,6 +565,10 @@ export const createStore = () => {
       a: WritableAtom<V, As, R>,
       ...args: As
     ) => {
+      const isSync = pendingStack.length > 0
+      if (!isSync) {
+        pendingStack.push(new Set([a]))
+      }
       let r: R | undefined
       if (isSelfAtom(atom, a)) {
         if (!hasInitialValue(a)) {
@@ -579,8 +583,8 @@ export const createStore = () => {
       } else {
         r = writeAtomState(a as AnyWritableAtom, ...args) as R
       }
-      if (pendingStack.length === 0) {
-        const flushed = flushPending([a])
+      if (!isSync) {
+        const flushed = flushPending(pendingStack.pop()!)
         if (import.meta.env?.MODE !== 'production') {
           storeListenersRev2.forEach((l) =>
             l({ type: 'async-write', flushed: flushed! }),
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -560,7 +560,6 @@ export const createStore = () => {
     atom: WritableAtom<Value, Args, Result>,
     ...args: Args
   ): Result => {
-    let isSync = true
     const getter: Getter = <V>(a: Atom<V>) => returnAtomValue(readAtomState(a))
     const setter: Setter = <V, As extends unknown[], R>(
       a: WritableAtom<V, As, R>,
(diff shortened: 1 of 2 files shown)


dai-shi:
This one looks good.
Did you confirm if the new test passes before #2396 refactor?

The fix seems a little bit unexpected to me, but probably can't be helped. I'll take another look when I'm relaxed.

iwoplaza:
I updated the solution a bit, so that it behaves more like how it did before #2396. Since before this fix, `setAtomState` was pushing to a pending stack item that did not exist (and therefore was never flushed). This new solution ensures there is a Set in the pending stack if the `setter` was called asynchronously (with no item in the pending stack)

dai-shi:
the new one feels a little bit better.

iwoplaza:
What do you think about adding a `console.warn('[Bug] tried to register flush onto an empty stack')` outside of production mode, so:

```ts
const pendingStackTop = pendingStack[pendingStack.length - 1];
if (pendingStackTop) {
  pendingStackTop.add(a)
} else if (import.meta.env?.MODE !== 'production') {
  console.warn('[Bug] tried to register flush onto an empty stack')
}
```

instead of

```ts
pendingStack[pendingStack.length - 1]?.add(a)
```

might make it easier to catch bugs like this in the future, or to note running `setAtomState` without providing a value on the stack first is erroneous.

iwoplaza:
Scratch that, there are places where setting the state of a mounted atom should not cause it to be flushed immediately, like in the `updatePromiseDependencies` function. The store implementation is still a little hard for me to maneuver, where is that update to the dependencies flushed?

src/vanilla/store.ts
```diff
@@ -581,7 +584,7 @@ export const createStore = () => {
         r = writeAtomState(a as AnyWritableAtom, ...args) as R
       }
       if (!isSync) {
-        const flushed = flushPending([a])
+        const flushed = flushPending(pendingStack.pop()!)
```
dai-shi, line 587: I was skeptical about `[a]`, so this looks good in code.

dai-shi (approved):
> The store implementation is still a little hard for me to maneuver

Well, it's same for me.

dai-shi:
I was planning to release it with v2.8.0.
But, as #2454 is reported, let's ship v2.7.1.

PR #2457, opened 2024-03-15
refactor(vanilla): moved (un)mount guards into their respective functions. (suggestion)

iwoplaza:
In my pursuit to understand the store implementation better, I found a few places where the code seems redundant (either because the check fails/succeeds before, likely unreachable, or the code is handling an undefined case even though the function is only called when the value is defined).

Changes:
- `canUnmountAtom` was being called at every call site of `unmountAtom`, so to reduce the risk of unmounting an atom prematurely at some call site by mistake, I moved the check into the function.
- Renamed `unmountAtom` to `tryUnmountAtom` to better reflect the new behavior
- Moved `canUnmountAtom` closer to the call site (`tryUnmountAtom`) to increase code legibility.
- Made `mountAtom` idempotent for already mounted atoms, which removed most "is this already mounted" checks.
- (needs verification) Removed the note about mounting dependencies only after the dependent is mounted, because it seems that we only call `mountDependencies` after checking if the dependent atom is already mounted.
- Removed `addAtom` and `delAtom`, since without the checks, they were just wrappers for `mountAtom` and `tryUnmountAtom`.

Let me know what you think about these changes! 💚 🐸 

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -473,26 +473,6 @@ export const createStore = () => {
   const readAtom = <Value>(atom: Atom<Value>): Value =>
     returnAtomValue(readAtomState(atom))
 
-  const addAtom = (atom: AnyAtom): Mounted => {
-    let mounted = mountedMap.get(atom)
-    if (!mounted) {
-      mounted = mountAtom(atom)
-    }
-    return mounted
-  }
-
-  // FIXME doesn't work with mutually dependent atoms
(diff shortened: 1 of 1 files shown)


dai-shi:
Thanks for suggestion.
At first glance, it looks good to me.
I'll take another deeper look at it later, but there's no hurry.

dai-shi (approved):
While I'm currently working on a new store implementation, this change to the current implementation looks good. Thanks.

PR #2462, opened 2024-03-21
fix(vanilla): improve dual instance warning in DEV

dai-shi:
https://github.com/pmndrs/jotai/discussions/2460#discussioncomment-8858974


scamden (commented):
Added benefit of being more elegant / less code! I'll try tomorrow!

PR #2463, opened 2024-03-25
New store implementation as `store2.ts`

dai-shi:
The current store implementation is based on the assumption I made in Jotai v1. Since then, it has changed and v2 design is slightly different. Technically speaking, in v1, we tried to support state branching (or state versioning) which we gave up in v2. So, the store implementation could be a little simpler. This PR introduces a new implementation with mutable atom state. Currently, we have a new `store2.ts` file and run tests to both `store.ts` and `store2.ts`. We hope to replace `store.ts` with `store2.ts` eventually. This is not the end, but the start. We might be able to simplify the new implementation and add some new capabilities. But, for now, the migration is important especially with/for jotai-devtools. It takes time.


dai-shi:
Tagging those who might be interested: @arjunvegda @dmaskasky @backbone87 @iwoplaza

dai-shi:
I will merge this for `v2.8.0` as it changes the built result slightly.

src/vanilla/store2.ts
```diff
@@ -0,0 +1,723 @@
...
+export const createStore = (): Store => {
+  const atomStateMap = new WeakMap<AnyAtom, AtomState>()
+
+  const getAtomState = <Value>(atom: Atom<Value>) => {
+    if (!atomStateMap.has(atom)) {
+      const atomState: AtomState<Value> = { d: new Map(), t: new Set() }
+      atomStateMap.set(atom, atomState)
+    }
+    return atomStateMap.get(atom) as AtomState<Value>
+  }
```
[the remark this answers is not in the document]
dai-shi: Yes, we can do that. We actually did it previously.

src/vanilla/store2.ts
```diff
@@ -0,0 +1,724 @@
...
+ * tracked for both mounted and unmounted atoms in a store.
+ */
+type AtomState<Value = AnyValue> = {
+  /**
+   * Map of atoms that the atom depends on.
+   * The map value is value/error of the dependency.
+   */
+  readonly d: Map<AnyAtom, { readonly v: AnyValue } | { readonly e: AnyError }>
+  /** Set of atoms that depends on the atom. */
+  readonly t: Set<AnyAtom>
```
dai-shi, line 195: As this holds a strong reference, it may cause memory leaks.
We should create a failing test case and fix it. (Probably by moving it to `Mounted`, or possibly with WeakRef but unlikely.)

dai-shi: #2477

Issue #2458, opened 2024-03-16
Idea: lazy initialization for primitive atoms


dai-shi:
For now, if we want to lazy initialize atom value, it has to be a derived atom.

I've been wondering how to avoid the limitation, and here's one idea.

```js
import { atom, lazy } from 'jotai/vanilla'; // or 'jotai'

const a = atom(lazy(() => Date.now())); // will be evaluated on the first use in the store.
```

dmaskasky:
Interesting idea. How about atomWithLazy?

dai-shi:
I was thinking about using a `LAZY` symbol, so `atomWithLazy` would be possible too. The point is that we need a new capability in core store impl.

iwoplaza:
Would it involve changes to the store implementation, or would it be a handy wrapper for the existing derived workaround?

Here is what I mean by a wrapper:
```ts
export function atomWithLazy<Value>(makeInitial: () => Value) {
  const wrappedAtom = atom(() => {
    // an atom is created per store
    return atom(makeInitial())
  })

  return atom((get) => get(get(wrappedAtom)))
}
```

Since `wrappedAtom` does not have any dependencies, it will compute its value only when first used, and the value happens to be a primitive atom that can then be written to and updated at will.

Then we could use it like initially proposed:
```ts
import { atomWithLazy } from 'jotai/vanilla'; // or 'jotai'

const a = atomWithLazy(() => Date.now()); // will be evaluated on the first use in the store.
```

dai-shi:
Very interesting. We don't need anything new in the store. How nice.

```js
export function atomWithLazy(makeInitial) {
  const wrappedAtom = atom(() => atom(makeInitial()))
  return atom(
    (get) => get(get(wrappedAtom)),
    (get, set, ...args) => set(get(wrappedAtom), ..args)
  )
}
```

dai-shi:
@iwoplaza Would you like to open a PR? It should be in utils.

iwoplaza:
Sure thing, I will do that either later today or tomorrow.

rothsandro:
The only drawback of the atom in atom approach is that it will break when used in combination with Jotai Scope. Lazy atoms used outside and inside of a Jotai Scope Provider without being a scoped atom will unexpectedly be created twice.

dai-shi:
Could it solved with `unstable_is`? 🤔

iwoplaza:
[...]

PR #2465, opened 2024-03-25
feat(utils): atomWithLazy for lazily initialized primitive atoms.

iwoplaza:

Realizes an idea proposed in this issue: https://github.com/pmndrs/jotai/issues/2458

## Check List


dai-shi (commented):
Nice. It looks great.

Would you like to add docs too? (If you are busy, a follow-up PR is also fine.)

iwoplaza:
Writing a doc for it right now

src/vanilla/utils/atomWithLazy.ts
```diff
@@ -0,0 +1,20 @@
...
+  const proxyAtom = atom(
+    (get) => get(get(wrappedAtom)),
+    (get, set, value: SetStateAction<Value>) => set(get(wrappedAtom), value),
+  )
+
+  wrappedAtom.init = atom(undefined)
+  // when writing to wrappedAtom through proxyAtom, the store
+  // thinks we are actually storing the value of `proxyAtom`.
+  wrappedAtom.unstable_is = (a) =>
+    a.unstable_is ? a.unstable_is(proxyAtom) : a === proxyAtom
```
dai-shi, line 17: It would be super nice if we had a test that fails without this. Is it too hard?

iwoplaza: I would be nice. It would require a barebones recreation of what `jotai-scope` is doing to intercept atoms. I wanted to understand how `jotai-scope` works better anyway, so I am happy to write it up. It will probably take a couple of days though.

iwoplaza: Although this hack works, I would like to better understand why it is necessary for `atomWithLazy` to function correctly with `jotai-scope`, both the `.init = atom(undefined)` and `unstable_is`.

dai-shi (approved):
It looks great. Thanks for your contribution!

src/vanilla/utils/atomWithLazy.ts
```diff
@@ -0,0 +1,12 @@
+import { PrimitiveAtom, atom } from '../../vanilla.ts'
+
+export function atomWithLazy<Value>(
+  makeInitial: () => Value,
+): PrimitiveAtom<Value> {
+  return {
+    ...atom(undefined as unknown as Value),
+    get init() {
+      return makeInitial()
```
himself65, line 9: should this function be called only once crossing the store?

iwoplaza: Yes, it should.

dai-shi: > should this function be called only once crossing the store?

No, it should call once for each store.
https://github.com/pmndrs/jotai/issues/2458#issuecomment-2024163904

iwoplaza: Oh, I thought that is what they meant. Once per each store is right.

PR #2476, opened 2024-04-02
fix(utils): improve freezeAtom and freezeAtomCreator

backbone87:

None

## Summary

Extends the `freezeAtomCreator` test to unveil a bug

## Check List

### DIFF
--- a/tests/react/vanilla-utils/freezeAtom.test.tsx
+++ b/tests/react/vanilla-utils/freezeAtom.test.tsx
@@ -1,43 +1,69 @@
 import { StrictMode } from 'react'
-import { render } from '@testing-library/react'
+import { fireEvent, render } from '@testing-library/react'
 import { it } from 'vitest'
 import { useAtom } from 'jotai/react'
 import { atom } from 'jotai/vanilla'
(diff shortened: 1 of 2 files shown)


tests/react/vanilla-utils/freezeAtom.test.tsx
```diff
@@ -40,4 +40,8 @@ it('freezeAtomCreator basic test', async () => {
   )
 
   await findByText('isFrozen: true')
+
+  act(() => getDefaultStore().set(objAtom, { count: 1 }))
```
dai-shi, line 44: Can you make it `<button>` and avoid using `getDefaultStore()` explicitly?

dai-shi:
Thanks for the catch. Can you also fix it?

backbone87:
added a fix and expanded the test case to check for deep freeze. also applied the same strategy that is used in freezeAtomCreator to `freezeAtom` since the latter didnt handle non-primitive atoms (non-derived atoms with custom setter).

dai-shi (approved):
LGTM
Thanks for working on this.

Now, I think `freezeAtomCreator` isn't very useful.
I don't think it's widely used, and it's easy to create one on users end.
I'll work on a separate PR.

PR #2481, opened 2024-04-05
fix: expose internal devstore types

arjunvegda:
## Summary
- Exposes `INTERNAL_DevStoreRev2`, `INTERNAL_DevStoreRev3` and `INTERNAL_PrdStore` for DevTools

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -14,12 +14,6 @@ export const getDefaultStore: GetDefaultStore = import.meta.env?.USE_STORE2
   : store.getDefaultStore
 
 export type {
-  INTERNAL_DevStoreRev2,
-  INTERNAL_DevStoreRev3,
-  INTERNAL_PrdStore,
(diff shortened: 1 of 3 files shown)


src/vanilla.ts
```diff
@@ -13,6 +13,12 @@ export const getDefaultStore: GetDefaultStore = import.meta.env?.USE_STORE2
   ? store2.getDefaultStore
   : store.getDefaultStore
 
+export type {
+  INTERNAL_DevStoreRev2,
+  INTERNAL_DevStoreRev3,
+  INTERNAL_PrdStore,
+} from './vanilla/store.ts'
+
```
dai-shi, line 21: We want to avoid exporting internal types from the entry point.
(As of now), we can directly import those types from `'jotai/vanilla/store'`

```suggestion
```

dai-shi (approved):
I consider this is a temporary thing during migration.

Issue #2477, opened 2024-04-04
Investigate and fix possible memory leaks in `store2`


dai-shi:
https://github.com/pmndrs/jotai/pull/2463#discussion_r1550772929

dai-shi:
@backbone87 Do you have any idea for the scenario where only `store2` fails with https://github.com/pmndrs/jotai/pull/2463#discussion_r1550772929 ?

backbone87:
the downstream deps (`t`) must only be tracked for mounted downstream deps

dai-shi:
because it leaks memory, or any other reason?

backbone87:
yes because it leaks memory. you could keep the non-mounted downstream in weakrefs but I think that is more expensive than just always "pulling" unmounted atoms on read

dai-shi:
Do you think you can write a test that passes with `store.ts` and fails with `store2.ts`?

dai-shi:
> Do you think you can write a test that passes with `store.ts` and fails with `store2.ts`?

I was able to do it and am working on it: #2533

PR #2484, opened 2024-04-05
fix(store): remove dev3 methods and add dev4 methods only in store2

dai-shi:
base pr #2474


src/vanilla/store2.ts
```diff
@@ -649,114 +584,31 @@ export const createStore = (): Store => {
...
-          setAtomStateValueOrPromise(getAtomState(atom), value)
-          recomputeDependents(pendingPair, atom)
-        }
-        const flushed = flushPending(pendingPair)
-        flushed?.forEach((a) => {
-          devListenersRev3.forEach((l) => l({ type: 'set', atom: a }))
-        })
+      dev4_get_internal_weak_map: () => atomStateMap,
+      dev4_override_method: (key, fn) => {
+        ;(store as any)[key] = fn
```
arjunvegda, line 608: Nice! A few questions

1. With this approach users would probably need to `composeWithDevTools(store || getDefaultStore())` and supply it via `<Provider/>` to their application. So the DevTools would no longer support the provider-less mode unless if we keep `composeWithDevTools` function within the main lib. Any thoughts around this? 🤔 

2. What would be the best way to wait for atoms resolving a promise (in DevTools)? i.e. values changing from sync -> async or async -> async (new promise). This did not exist in v1 btw, but would be nice to 
include it in v2

3. Could we still keep `mountedAtoms` around? Perhaps we could introduce `dev4_get_internal_mounted_atom_map`? If you feel strongly about not including it, then I could start with keeping references to the atoms changed with `subscribe/unsubscribe` 🤔  

4. Do we still need `dev4` prefix or should we call it `dev_get_internal_weak_map` etc.? From Jotai DevTools perspective, exposing `import { DevTools } from 'jotai-devtools/experimental` might make more sense that eliminates any conflict with the `dev` vs `dev4`

dai-shi: > 1. With this approach users would probably need to `composeWithDevTools(store || getDefaultStore())` and supply it via `<Provider/>` to their application. So the DevTools would no longer support the provider-less mode unless if we keep `composeWithDevTools` function within the main lib. Any thoughts around this?

I thought without `dev4_override_method`, that's the case, and hoped `dev4_override_method` could work it around.

dai-shi: > 2\. What would be the best way to wait for atoms resolving a promise (in DevTools)?

I assume you can simply `await` the promise value from `store.get()`.

dai-shi: > 3\. Could we still keep `mountedAtoms` around?

Hmm, that was one of the biggest things, I would like to avoid in dev4. Otherwise, we can re-consider dev3.

dai-shi: > 4\. Do we still need `dev4` prefix or

I don't understand what conflicts. I prefer the revision number for clearer communication. (we may have `dev5` in the future.)

arjunvegda: > I thought without dev4_override_method, that's the case, and hoped dev4_override_method could work it around.

Since it is updating the reference of the store it should work with the default store too 🤔

> Hmm, that was one of the biggest things, I would like to avoid in dev4. Otherwise, we can re-consider dev3.

I can work around it by patching `store.set` and manually keeping track of the atoms by sub and unsub. Not sure if I'm missing any edge cases here.

> I don't understand what conflicts. 

Mostly TypeScript related, I'm trying to avoid the number of type guards 😅 but I can circumvent it.

src/vanilla/store2.ts
```diff
@@ -282,9 +256,7 @@ type PrdStore = {
   ) => Result
   sub: (atom: AnyAtom, listener: () => void) => () => void
 }
-type Store =
-  | (PrdStore & Partial<DevStoreRev2>)
-  | (PrdStore & DevStoreRev2 & DevStoreRev3)
+type Store = PrdStore | (PrdStore & DevStoreRev4)
```
arjunvegda, line 259: Any chance we could export `PrdStore & DevStoreRev4`?

arjunvegda (approved):
Thanks a lot for adding this! I tested a bunch of scenarios out locally and seems promising so far.

src/vanilla/store2.ts
dai-shi: I can do that.

dai-shi:
> seems promising so far.

Thanks, and we can modify/improve it moving forward. I don't think store2 is ready yet anyway. for example #2477

PR #2493, opened 2024-04-08
fix: lazyMap on `atomWithLazy`

himself65:

Fixes https://github.com/pmndrs/jotai/pull/2465#discussion_r1556274937

## Check List


dai-shi:
It's the motivation and the requirement to initialize for each store.
https://github.com/pmndrs/jotai/blob/9ceaef6403a9d932590b37852734bd8bd4958b74/docs/utilities/lazy.mdx?plain=1#L38

dai-shi:
In case, one wants to share data, they could do something like this:
```js
let initializedData;
const dataAtom = atomWithLazy(() => (initializedData ??= Date.now()));
```

PR #2497, opened 2024-04-10
chore(tests): refactor memory leak tests and add some more tests

dai-shi:


dai-shi:
This doesn't yet cover #2477 but more like a refactor.

PR #2502, opened 2024-04-12
bulk update for splitAtom

zolboo-gtn:

Resolves [#2500](https://github.com/pmndrs/jotai/discussions/2500)

## Summary

Add bulk update feature for splitAtom.

## Check List

### DIFF
--- a/src/vanilla/utils/splitAtom.ts
+++ b/src/vanilla/utils/splitAtom.ts
@@ -37,6 +37,10 @@ type SplitAtomAction<Item> =
       atom: PrimitiveAtom<Item>
       before?: PrimitiveAtom<Item>
     }
+  | {
+      type: 'bulk'
+      atoms: PrimitiveAtom<Item>[]
+    }
 
 export function splitAtom<Item, Key>(
   arrAtom: WritableAtom<Item[], [Item[]], void>,
@@ -202,6 +206,13 @@ export function splitAtom<Item, Key>(
                       ])
                     }
                   }
+                  break
+                }
+                case 'bulk': {
+                  set(
+                    arrAtom,
+                    action.atoms.map((atom) => get(atom)),
+                  )
                   break
                 }
               }


src/vanilla/utils/splitAtom.ts
```diff
@@ -37,6 +37,10 @@ type SplitAtomAction<Item> =
       atom: PrimitiveAtom<Item>
       before?: PrimitiveAtom<Item>
     }
+  | {
+      type: 'bulk'
```
dai-shi, line 41: I'm not sure if this type is very common. Do you have any alternative ideas?

src/vanilla/utils/splitAtom.ts
```diff
@@ -204,6 +208,13 @@ export function splitAtom<Item, Key>(
                   }
                   break
                 }
+                case 'bulk': {
+                  set(
+                    arrAtom,
+                    action.atoms.map((atom) => get(atom)),
```
dai-shi, line 214: Hmm, this works if used properly. But, I think it's safer to find the index and create a new array based on it. Let's not assume the passed atoms are correct.

dai-shi (changes requested):
Thanks for working on this.

In addition to the two comments below, we need tests for this, with broader coverage. It will be quite a bit of work. Let's take time.

src/vanilla/utils/splitAtom.ts
zolboo-gtn: `reorder` might be better for my use case. But there is difference between `reorder` and `bulk`.
`bulk` might have removed or inserted items with or without reordering. We should think `bulk` as new atoms not related to old ones.
Should we consider `bulk` too as a separate type?

dai-shi: yeah, "reorder" sounds something different. For me "bulk" doesn't sound like a verb. "bulk-change", too long.
If you don't have any other idea, we can use `bulk`, or maybe `unstable_bulk` would be better for now.

src/vanilla/utils/splitAtom.ts
zolboo-gtn: In case of `reorder`, we can calculate `symmetrical difference` first and if it has zero length update the splitAtom.

```
case 'reorder': {
  const oldArr = get(arrAtom);
  const newArr = action.atoms;
  
  const difference = oldArr
    .filter(atom => !newArr.includes(atom))
    .concat(newArr.filter(atom => !oldArr.includes(atom)));

  if(difference.length === 0) {
    set(
      arrAtom,
      newArr.map((atom) => get(atom)),
    );
  }
  break
}
```

dai-shi: My request is slightly different. I want to avoid using the atom value for the final array, only to find the index.

So, I would like to avoid this usage:
```js
const arrAtom = atom([1, 2, 3])
const atoms = splitAtom(arrAtom)

  set(atoms, { type: 'bulk', atoms: [atom(-999)] })
```
(This has too much power and we should probably update the original arr.)

src/vanilla/utils/splitAtom.ts
zolboo-gtn: Are you saying, we do not need `reorder` and only `bulk?` will suffice? For the naming i will come out something proper later.

dai-shi: Yeah, it can cover the reorder use case, right?

src/vanilla/utils/splitAtom.ts
zolboo-gtn: > only to find the index

Correct me if i am wrong.
In case of `bulk?`, using only indexes would means we should calculate moved, inserted, removed atoms separately. I do not understand what is good about it.

```
const newArrMapping = [
  // moved
  {
    oldIndex: n,
    newIndex: m
  },
  // inserted
  {
    oldIndex: null,
    newIndex: m
  },
  // removed
  {
    oldIndex: n,
    newIndex: null
  },
]
```

> const arrAtom = atom([1, 2, 3])
> const atoms = splitAtom(arrAtom)
> 
>   set(atoms, { type: 'bulk', atoms: [atom(-999)] })

I do not understand this either. What if 1, 2, 3 are removed and -999 is inserted? Because it is `bulk?` not `reorder` after all.

src/vanilla/utils/splitAtom.ts
zolboo-gtn: Yes

zolboo-gtn: [About the naming ](https://react.dev/learn/updating-arrays-in-state#updating-arrays-without-mutation)

Possible operations on an array i can think of are:
- insert (Insert single item. Do not considers multiples.)
- remove (Remove single item. Do not considers multiples.)
- replace (Whole new array)
- reorder (This is a special case of the `replace`. Without `insert` and `remove`)
- move (Move single item. This is a special case of the `reorder`. `reorder` is equivalent of multiple `move`)

zolboo-gtn: insert, remove, move are atomic operations (for single item only). And for other cases `replace` would be sufficient.
Is it correct?

src/vanilla/utils/splitAtom.ts
dai-shi: okay then, I think I misunderstood `bulk`. I think what I want to support is `reorder` only.

It doesn't make much sense for me to insert arbitrarily items.

For "bulk", I'd update the original atom. We could add it as a shortcut, but my hesitation is it doesn't add any capability. We don't usually do it in Jotai codebase. 

If you need "bulk", please share your use case. It can be a doc issue.

zolboo-gtn: Thank you for your time.
Last question.

> I want to avoid using the atom value for the final array

I do not see any difference between `symmetrical difference with length zero` and `creating index mapping and construct an array from it`. Either way, we are creating an new array with exactly same items by comparing them. Is there any?
Or did i misunderstand it?

For the reordering operation name, if `reorder` is not good, how about `shuffle`. As in shuffling cards.

dai-shi: `shuffle` to me sounds like random sorting.

What I have in mind with the plain array is this:
```js
const reorder = (array, indexList) =>
  indexList.map((index) => {
    if (index < 0 || index >= arr.length) {
      throw new Error('index out of bounds');
    }
    return array[index];
  });

// usage
const arr = ['a', 'b', 'c'];
const arr2 = reorder(arr, [2, 0])
```
The operation name `reorder` is tentative.

src/vanilla/utils/splitAtom.ts
dai-shi: thanks for the summary. it made me clear that, _for now_ I would like to avoid `replace` (bulk updates) and only provide single operations.
For `replace`, I would mutate the original atom. I am happy to learn the pitfall with this approach.

zolboo-gtn: Okay, thank you.
I will close the PR for now.

PR #2508, opened 2024-04-14
fix(tests): add memory leak tests

backbone87:

https://github.com/pmndrs/jotai/issues/2477 

## Summary

Fails with store and store2

### DIFF
--- a/tests/vanilla/memoryleaks.test.ts
+++ b/tests/vanilla/memoryleaks.test.ts
@@ -1,14 +1,26 @@
 import LeakDetector from 'jest-leak-detector'
 import { expect, it } from 'vitest'
-import { atom, createStore } from 'jotai/vanilla'
+import { Atom, atom, createStore } from 'jotai/vanilla'
 
-it('should not have memory leaks with an atom', async () => {
+it('should not hold onto atoms that are not mounted', async () => {
   const store = createStore()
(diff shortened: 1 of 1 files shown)


dai-shi:
Thanks, I will have a closer look later, but:

> Fails with store and store2

So, it means this doesn't reveal the regression with `.t` in atomState in `store2`.

backbone87:
yeah, it seems to be already a problem in main

tests/vanilla/memoryleaks.test.ts
```diff
@@ -1,14 +1,26 @@
...
+import { Atom, atom, createStore } from 'jotai/vanilla'
 
-it('should not have memory leaks with an atom', async () => {
+it('should not hold onto atoms that are not mounted', async () => {
   const store = createStore()
-  let detector: LeakDetector
-  ;(() => {
-    const objAtom = atom({})
-    detector = new LeakDetector(store.get(objAtom))
-  })()
```
dai-shi, line 11: Hm, I somehow thought function closure would be necessary, but I think I got something wrong. Let me update my PR.

dai-shi:
Hi, I updated the main branch. Would you please pull it and adjust this PR so that the diff becomes minimal?

dai-shi:
Fixed by #2533

PR #2509, opened 2024-04-14
feat: add back dev4_restore_atoms to store v2

arjunvegda:
## Summary
- Add back `dev4_restore_atoms` method to store v2
- Remove internal types for store1 (I no longer need them for v1, just v2 suffices😅)


src/vanilla/store2.ts
```diff
@@ -610,6 +611,14 @@ export const createStore = (): Store => {
       dev4_override_method: (key, fn) => {
         ;(store as any)[key] = fn
       },
+      dev4_restore_atoms: (values) => {
+        const pendingPair = createPendingPair()
+        for (const [atom, value] of values) {
+          setAtomStateValueOrPromise(getAtomState(atom), value)
+          recomputeDependents(pendingPair, atom)
```
dai-shi, line 618: Please add the primitive atom check which we already have it in `store.ts`.
```suggestion
          if (hasInitialValue(atom)) {
            setAtomStateValueOrPromise(getAtomState(atom), value)
            recomputeDependents(pendingPair, atom)
          }
```

dai-shi (changes requested):
A side note: I wonder what would be the difference between "restore" and multiple "set"s.

src/vanilla/store2.ts
```diff
@@ -679,6 +680,23 @@ export const createStore = (): Store => {
...
+            const prevValue = aState.v
+            setAtomStateValueOrPromise(atom, aState, value)
+            mountDependencies(pendingPair, atom, aState)
+            if (!hasPrevValue || !Object.is(prevValue, aState.v)) {
+              addPendingAtom(pendingPair, atom, aState)
+              recomputeDependents(pendingPair, atom)
+            }
+          }
+        }
+        flushPending(pendingPair)
```
dai-shi, line 698:
```suggestion
        const pending = createPending()
        for (const [atom, value] of values) {
          if (hasInitialValue(atom)) {
            const aState = getAtomState(atom)
            const hasPrevValue = 'v' in aState
            const prevValue = aState.v
            setAtomStateValueOrPromise(atom, aState, value)
            mountDependencies(pending, atom, aState)
            if (!hasPrevValue || !Object.is(prevValue, aState.v)) {
              addPendingAtom(pending, atom, aState)
              recomputeDependents(pending, atom)
            }
          }
        }
        flushPending(pending)
```

dai-shi:
> A side note: I wonder what would be the difference between "restore" and multiple "set"s.

Yes, there's a difference. store.set can't update atom with customized write (like write-only atom and atomWithReducer).

dai-shi (approved):
It looks great! Thanks for investigating it. Nice work.

PR #2511, opened 2024-04-19
remove unnacessary any type

do7be:
## Summary

`key: K extends keyof PrdStore` and there is no problem with the type.

So `any` seems unnecessary.

## Check List


do7be:
There was no problem with build on local, but it seems that the global variable named `store` is interpreted as not existing, so I closed this PR.

do7be:
Probably, the any in various places was necessary for rollup.

PR #2514, opened 2024-04-21
Fix document layout flickering.

ouweiya:
When the content in the main container is too short, the container will shrink, causing layout flickering.

Setting `flex-grow` on main can solve this problem, always keeping main to fill its parent container.

https://jotai.org/docs/utilities/reducer

https://github.com/pmndrs/jotai/assets/28357731/cc0462e8-f97e-4019-8c0c-ca1c82b41e7c

Adding the notranslate class to the code block component can prevent the code text from being translated by Google.

image


sandren (approved):
@ouweiya Thanks for this fix! 🙂

PR #2517, opened 2024-04-23
Improve the document reading experience.

ouweiya:
Adjusting the main text to be centered is more conducive to reading, and the sidebar should not be too far from the main text, as it would lose the association between the navigation and the article.

image

image


sandren (approved):
Thanks! I simplified the changes to achieve a similar layout without adjusting the sizing of elements. 🙂

PR #2520, opened 2024-04-25
Fix Gatsby Link URL partial match.

ouweiya:
When multiple path URLs are partially the same, it can cause simultaneous highlighting.

chrome_zqEPBcj8Kr

 
[Fix 'key' prop error](https://github.com/pmndrs/jotai/pull/2520/commits/91ba68f3f4e4476c8dbe7430c202c6526d0fed55) 

> Warning: Each child in a list should have a unique "key" prop.

The 'introduction' section has no parent directory, `section.title` is `undefined`.

[When the title is inline code, the ID is undefined](https://github.com/pmndrs/jotai/pull/2520/commits/86eb61670394b896268436fa1c935084f8fcad85)
When the title is inline code, `value` is a React element, returning an empty string will cause the ID to be unavailable, making the anchor point invalid.

### DIFF
--- a/.github/workflows/test-multiple-builds.yml
+++ b/.github/workflows/test-multiple-builds.yml
@@ -65,6 +65,6 @@ jobs:
           NODE_ENV: ${{ matrix.env }}
       - name: Test ${{ matrix.build }} ${{ matrix.env }}
         run: |
-          yarn test:ci
+          yarn test:ci:build # test:ci
         env:
           NODE_ENV: ${{ matrix.env }}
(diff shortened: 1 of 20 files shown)


sandren:
@ouweiya Thanks for these fixes!

Could you please run Prettier and also revert the changes to `performance.mdx` in [bfb555b](https://github.com/pmndrs/jotai/pull/2520/commits/bfb555b9e4b06024468662f07e78cd9c39f36c28)
? Let's move more substantive revisions to the docs in a separate PR. Thanks. 🙂

sandren (changes requested):
@ouweiya Thanks for these fixes!

Could you please run Prettier and also revert the changes to `performance.mdx` in bfb555b? Let's move the more substantive revisions to the docs into a separate PR. Thanks. 🙂

sandren (approved):

sandren:
Thanks! 🙂

PR #2529, opened 2024-04-29
chore: use yarn@4

himself65:
## Summary

Upgrade the toolchain since there are some problems with yarn@1.

1. Corepack doesn't fully support yarn 1 (at least it always tries to install yarn 4)

## Check List


dai-shi:
Hey, thanks for working on it.
Sorry for not sharing ideas, but we want to migrate to pnpm. We already do it for jotaijs, zustandjs, and valtiojs.
So, we want to follow for jotai, zustand, and valtio too.

dai-shi:
#2543

PR #2530, opened 2024-04-29
refactor: avoid extra rerender

hjiog:
## Summary
useAtomValue called rerender inside useEffect, which led to an additional render, and this PR optimized that performance issue.

## Check List

### DIFF
--- a/src/react/useAtomValue.ts
+++ b/src/react/useAtomValue.ts
@@ -94,7 +94,6 @@ export function useAtomValue<Value>(atom: Atom<Value>, options?: Options) {
       }
       rerender()
     })
-    rerender()
     return unsub
   }, [store, atom, delay])
 
--- a/src/vanilla/store2.ts
+++ b/src/vanilla/store2.ts
@@ -61,8 +61,12 @@ const flushPending = (pendingPair: PendingPair, isAsync?: true) => {
         pending()
       } else {
         const [atom, atomState] = pending
-        if (!flushed.has(atom) && atomState.m) {
-          atomState.m.l.forEach((listener) => listener())
+        if (!flushed.has(atom)) {
+          if (atomState.m && atomState.m.l.size) {
+            atomState.m.l.forEach((listener) => listener())
+          } else {
+            atomState.u = true
+          }
           flushed.add(atom)
         }
       }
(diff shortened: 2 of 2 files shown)


dai-shi (commented):
Thanks for the suggestion.
We can't remove `rerender` in useAtomValue, but you are right about the extra rerender (without commits), which isn't desirable.

I'm still working on store2, which will come with a big change, so please be patient.

That said, I think the optimization like what you suggest might be possible and meanwhile what you can help is to create a test case. Thanks for understanding.

src/react/useAtomValue.ts
```diff
@@ -94,7 +94,6 @@ export function useAtomValue<Value>(atom: Atom<Value>, options?: Options) {
       }
       rerender()
     })
-    rerender()
```
dai-shi, line 97: This is required. I wonder if we could make a contrived test to cover it.

(I thought there was a rejected PR or something to refer, but can't find it now.)

hjiog: yeah, maybe we should add more test. but the rerender should be optimized.

dai-shi: Did you have a real problem in your usage? If you have, memoization might mitigate the issue for the meantime.

dai-shi:
Please create a new test case in `tests/react/optimization.test.tsx`.
We can't guarantee if it will be solved.

gurkerl83:
> Thanks for the suggestion. We can't remove `rerender` in useAtomValue, but you are right about the extra rerender (without commits), which isn't desirable.
> 
> I'm still working on store2, which will come with a big change, so please be patient.
> 
> That said, I think the optimization like what you suggest might be possible and meanwhile what you can help is to create a test case. Thanks for understanding.

@dai-shi Could you clarify if store2 is intended to fix the underlying cause requiring a secondary render, or have I misunderstood? From your last comment about the test case, it seems this might still be an open question.

Thx!

dai-shi:
> if store2 is intended to fix the underlying cause requiring a secondary render

No, I only mentioned store2 because this PR made changes only to store2.
This behavior/problem isn't specific to store2 or store. It's working as expected.

PR #2539, opened 2024-05-03
feat(utils): allow providing a subscribe method to createJSONStorage util

mhsattarian:

Related to https://github.com/pmndrs/jotai/discussions/1833#discussioncomment-9288746

## Summary

previously for providing subscribe functionality to a custom persist storage, one should've used something like the code below which btw was not documented.

```ts
const stringStorage = {
  getItem: (key: string) => {
    return //
  },
  setItem: (key: string, newValue: string) => {
    //
  },
  removeItem: (key: string) => {
    ///
  },
}

const customStorage = createJSONStorage<number>(() => stringStorage)

customStorage.subscribe = (key, callback, initialValue) => {
  /// add listener to detected chanegs then call callback:
  callback()
  return () => {
    /// remove listener
  }
}
```

it's more convenient to just accept the subscribe method along with `getItem`, `setItem`, etc when using `createJSONStorage` to create a custom persist functionality like the cookies' in the linked discussed. here is an example:

```ts
const stringStorage = {
  getItem: (key: string) => {
    return; //
  },
  setItem: (key: string, newValue: string) => {
    //
  },
  removeItem: (key: string) => {
    //
  },
  subscribe(key, callback, initialValue) {
    /// add listener to detected chanegs then call callback:
    callback();
    return () => {
      /// remove listener
    };
  },
};

const customStorage = createJSONStorage<number>(() => stringStorage);

```

## Check List

- [x] provide tests for the added functionality


src/vanilla/utils/atomWithStorage.ts
```diff
@@ -114,6 +129,42 @@ export function createJSONStorage<Value>(
 ): AsyncStorage<Value> | SyncStorage<Value> {
   let lastStr: string | undefined
   let lastValue: Value
+
+  const webStorageSubscribe: Subscribe<Value> = (key, callback) => {
```
dai-shi, line 133: This is still string based, right?
```suggestion
  const webStorageSubscribe: Subscribe<string> = (key, callback) => {
```

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -114,6 +129,42 @@ export function createJSONStorage<Value>(
...
+    if (!(getStringStorage() instanceof window.Storage)) {
+      return () => {}
+    }
+    const storageEventCallback = (e: StorageEvent) => {
+      if (e.storageArea === getStringStorage() && e.key === key) {
+        callback((e.newValue || '') as Value)
```
dai-shi, line 139: This seems wrong:
```suggestion
        callback(e.newValue || '')
```

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -5,8 +5,21 @@ import { RESET } from './constants.ts'
...
+
 type Unsubscribe = () => void
 
+type SubscribeHandler<Value> = (
+  subscribe: Subscribe<Value>,
+  key: string,
+  callback: (value: Value) => void,
+  initialValue: Value,
+) => Unsubscribe
+
```
dai-shi, line 22: Maybe, we don't need this type alias. (But, not 100% sure.)

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -114,6 +129,42 @@ export function createJSONStorage<Value>(
...
+  const handleSubscribe: SubscribeHandler<Value> = (
+    subscriber,
+    key,
+    callback,
+    initialValue,
+  ) => {
+    function callbackWithParser(v: Value) {
+      let newValue: Value
+      try {
+        newValue = JSON.parse((v as string) || '')
```
dai-shi, line 157: `v` must be a string always.
```suggestion
        newValue = JSON.parse(v || '')
```

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -141,30 +192,18 @@ export function createJSONStorage<Value>(
...
-            newValue = initialValue
-          }
-          callback(newValue)
-        }
-      }
-      window.addEventListener('storage', storageEventCallback)
-      return () => {
-        window.removeEventListener('storage', storageEventCallback)
-      }
+    if (getStringStorage()?.subscribe) {
```
dai-shi, line 200: AFAIR, this can throw. So, we can't call it here.

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -141,30 +192,18 @@ export function createJSONStorage<Value>(
...
+      storage.subscribe = handleSubscribe.bind(
+        null,
+        getStringStorage()!.subscribe as unknown as Subscribe<Value>,
+      )
```
dai-shi, line 204: I don't think we use `bind` like this in our code base. Please either use an inline function or a higher-order function.
But, restructure will be needed necessary anyway because of the previous comment.

dai-shi (changes requested):
Thanks for working on this!
The idea seems good. We just need some fixes.
Please take your time.

src/vanilla/utils/atomWithStorage.ts
mhsattarian: would something like this be more acceptable?:

```ts
const handleSubscribe = (
    subscriber: Subscribe<Value>,
    ...params: Parameters<Subscribe<Value>>
  ) => {
    const [key, callback, initialValue] = params

    function callbackWithParser(v: Value) {
      let newValue: Value
      try {
        newValue = JSON.parse((v as string) || '')
      } catch {
        newValue = initialValue
      }

      callback(newValue as Value)
    }

    return subscriber(key, callbackWithParser, initialValue)
  }
```

src/vanilla/utils/atomWithStorage.ts
mhsattarian: before this PR, subscribe method was defined like this:

```ts
subscribe?: (
    key: string,
    callback: (value: Value) => void,
    initialValue: Value,
  ) => Unsubscribe
```
and the callback would accept value of type `Value`. should we change this to accept only strings?

src/vanilla/utils/atomWithStorage.ts
dai-shi: Yes, I meant to inline it as it's used just once.

src/vanilla/utils/atomWithStorage.ts
dai-shi: No, I mean `v` must be `string` type without any hacks/casts. It feels like there are some lies in types.
I can help for typing later. Please work on other comments first.

dai-shi:
Hope you find time to address this.

mhsattarian:
@dai-shi I would love to.
sorry this is taking some time, I'm a bit busy with family these days and I hope I can find time to work on this in a couple of days.

src/vanilla/utils/atomWithStorage.ts
mhsattarian: hey @dai-shi. I finally got some time to work on this. I would appreciate your opinion (and advice) on typing issues.

Regarding this code suggestion, `webStorageSubscribe` was previously typed as `Subscribe<Value>` and I didn't change the type here. how should we address these typings? Isn't the `Value` type always a subset of `string`?

src/vanilla/utils/atomWithStorage.ts
mhsattarian: I believe the `removeEventListener` method won't throw unless the given parameters are wrong which is not happening in our case. am I wrong?

and this code is the same as before:

src/vanilla/utils/atomWithStorage.ts
mhsattarian: applied a few changes to fix this. I would appreciate a review of them.

src/vanilla/utils/atomWithStorage.ts
dai-shi: Oh, my bad. I was thinking of old implementation. We catch it already.

dai-shi:
@mhsattarian Hey, I just added a commit to refactor. Please review it.

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -141,31 +147,42 @@ export function createJSONStorage<Value>(
...
+        }
+        callback(newValue)
+      })
+
+  const subscriber: StringSubscribe | undefined =
+    getStringStorage()?.subscribe ||
+    (typeof window !== 'undefined' &&
+      typeof window.addEventListener === 'function' &&
+      ((key, callback) => {
+        if (!(getStringStorage() instanceof window.Storage)) {
```
dai-shi, line 169: It feels like we can simply this, if we are allowed to call `getStringStorage` in sync.

dai-shi (approved):

mhsattarian:
Hey @dai-shi, sorry for the late response.
this looks clean! thank you.

Are we ready for the merge?

dai-shi:
Cool. I will merge it next week before preparing a new patch release.

mhsattarian:
Awesome!

since this PR started from syncing atoms with cookies, I will soon open a new PR for an example in the persistence guide.
Is there anything I should know before that? (I remember you said something about re-structuring the persistence guide)

dai-shi:
I don't remember exactly what I said, but the persistence guide (and other docs) isn't very polished. So, not only adding new things, but also improving the guide entirely would be appreciated.

src/vanilla/utils/atomWithStorage.ts
dai-shi: My first comment was actually right. We can't call `getStringStorage()` on creation. #2581

PR #2558, opened 2024-05-17
perf: avoid extra rerender

hjiog:
## Summary

useAtomValue called rerender inside useEffect, which led to an additional render, and this PR optimized that performance issue add a optimization test.

## Check List


tests/react/optimization.test.tsx
```diff
@@ -271,3 +271,24 @@ it('no extra rerenders after commit with derived atoms (#1213)', async () => {
   })
   expect(renderCount1).toBe(renderCount1AfterCommit)
 })
+
+// TODO
+it.skip('no extra rerenders after mounted (#2530)', async () => {
```
dai-shi, line 276:
```suggestion
it('no extra rerenders after mounted (#2558)', async () => {
```

dai-shi:
We should probably add another test to ensure that `rerender()` in useEffect is required, before tackling this.

dai-shi (changes requested):
> We should probably add another test to ensure that rerender() in useEffect is required, before tackling this.

Please create a test for `rerender()` first before tackling this. Otherwise, you don't know why we need it. It thinks it's reasonable to create a separate PR.

> Another point is that this optimization should be handled with React binding. There's nothing wrong in store implementation.

Please do not modify store implementation.

hjiog:
If I have uderstood correctly, `rerender()` is to ensure that all the components have executed the subscription logic, so that there will be no problem when the onMounted method is called. But I don't know how to create a test for `rerender()`. Yea,  this optimization can be handled with React binding, but what if we do it better? This PR can be used as a reference idea. If someone is interested, welcome to improve it.

dai-shi:
Yeah, let's leave it for now. Here's some comments for the future.

`rerender()` exists because `useEffect` is async.
I think `useLayoutEffect` instead might be a fix (but I haven't confirmed),
but it will probably be a major bump with React 19 requirement (or maybe 18).

If we were to optimize it in Jotai v2, it must be within the hook.
It should be technically possible with something like `use-sync-external-store` shim.

dai-shi:
Closing this for now.

If anyone is interested, feel free to develop a third-party library (that provides its own `useAtomValue`) and do some experiments.
I can invite someone to https://github.com/jotaijs if they work on it.

suguanYang:
> Yeah, let's leave it for now. Here's some comments for the future.
> 
> `rerender()` exists because `useEffect` is async. I think `useLayoutEffect` instead might be a fix (but I haven't confirmed), but it will probably be a major bump with React 19 requirement (or maybe 18).
> 
> If we were to optimize it in Jotai v2, it must be within the hook. It should be technically possible with something like `use-sync-external-store` shim.

Sorry for bring this up, maybe we could write something like this?
```javascript

const useStoreSubSync = (atom) => {
  const store = useStore();
  const listenerRef = useRef(null)
  {
    listenerRef.current = listenerRef.current || store.sub(atom, () => {
      // ...
      rerender()
    })
  }

  useEffect(() => {
    return listenerRef.current
  }, [])
}
```

dai-shi:
Unfortunately, useEffect doesn't fire in some cases.

suguanYang:
I did't get it. we already rely on useEffect, right?

dai-shi:
What's guaranteed is if useEffect runs once, the cleanup runs once.
You shouldn't make side effects in the render function. There should be lots of discussions and articles about it.

PR #2569, opened 2024-05-24
fix(core): expose unstable_resolve instead of unstable_is

dai-shi:
For jotai-scope, we don't need to expose unstable_is, because they use store api.
Exposing unstable_resolve might be better?


dai-shi:
Superseded by #2609.

PR #2572, opened 2024-05-24
Revert "fix(vanilla): fix unexpected cache in jotai-scope (#2371)"

dai-shi:
This reverts commit fdc2324f19d15827fcb2419279b6481f3fc5e9c0 #2371.

jotai-scope has got a new impl, so this might be correct, or not?


dai-shi:
this causes a regression.

PR #2585, opened 2024-05-29
Fix #2581: subscribe method of createJSONStorage should only run in the client

### LINKED ISSUE #2581: `createJSONStorage`: localStorage is not defined
## Summary
```console
ReferenceError: localStorage is not defined

  1 | import { atomWithStorage, createJSONStorage } from 'jotai/utils'
  2 |
> 3 | const storage = createJSONStorage(() => localStorage)
    |                                         ^
  4 |
  5 | const darkModeState = atomWithStorage('darkMode', true, storage)
```

## Details
Next.js: `14.2.3`
jotai: `2.8.2`

It correctly works with jotai <= 2.8.1.

Issue probably generated by #2539

Similar issue #949

mhsattarian:

Fixes #2581

## Summary

Fixes some issues created by previous PR (#2539) subscribe method of createJSONStorage should only run in the client

## Check List

### DIFF
--- a/src/vanilla/utils/atomWithStorage.ts
+++ b/src/vanilla/utils/atomWithStorage.ts
@@ -161,30 +161,34 @@ export function createJSONStorage<Value>(
         callback(newValue)
       })
 
-  let subscriber = getStringStorage()?.subscribe
   if (
-    !subscriber &&
     typeof window !== 'undefined' &&
-    typeof window.addEventListener === 'function' &&
-    window.Storage &&
-    getStringStorage() instanceof window.Storage
+    typeof window.addEventListener === 'function'
   ) {
-    subscriber = (key, callback) => {
-      const storageEventCallback = (e: StorageEvent) => {
-        if (e.storageArea === getStringStorage() && e.key === key) {
-          callback(e.newValue)
+    let subscriber = getStringStorage()?.subscribe
+    if (
+      !subscriber &&
+      window.Storage &&
+      getStringStorage() instanceof window.Storage
+    ) {
+      subscriber = (key, callback) => {
+        const storageEventCallback = (e: StorageEvent) => {
+          if (e.storageArea === getStringStorage() && e.key === key) {
+            callback(e.newValue)
+          }
+        }
+        window.addEventListener('storage', storageEventCallback)
+        return () => {
+          window.removeEventListener('storage', storageEventCallback)
         }
       }
-      window.addEventListener('storage', storageEventCallback)
-      return () => {
-        window.removeEventListener('storage', storageEventCallback)
-      }
+    }
+
+    if (subscriber) {
+      storage.subscribe = createHandleSubscribe(subscriber)
     }
   }
 
-  if (subscriber) {
-    storage.subscribe = createHandleSubscribe(subscriber)
-  }
   return storage
 }
 


mhsattarian:
@dai-shi Hi again. 

this is the simplest fix I could think of. is it enough in your opinion?

should we add a try/catch block to ensure subscribe method won't fail?

dai-shi:
Unfortunately, it's not what I expect. Let me try and see how it goes.

dai-shi:
oh, this is tricky. what I expect doesn't work.

dai-shi:
How does 6ca781f look?

I wanted to keep this: https://github.com/pmndrs/jotai/pull/2539/files#diff-4bb714dba559e7b7df1c7cb13113df5aefbafe023dac97e346bb39e8db159a1aL150-L152

mhsattarian:
Awesome. this looks better.

mhsattarian:
I should remove the test I added to check the subscriber not be called on server.

mhsattarian:
but now can we make sure such a case won't happen in the future?

dai-shi:
> but now can we make sure such a case won't happen in the future?

Ideally, there should be a test that throw on `window.localStorage` access. I'm not sure if we can write such a test.

dai-shi:
can we? Please give it a try!

mhsattarian:
honestly, I couldn't wrap my head around this problem as much as I wanted.

since we are already calling `getStringStorage` in the try block, checking `localStorage` access is not what we want. am I right? So I checked for access to `getItem` method instead.

the last commit seems eligible. what do you think @dai-shi?

mhsattarian:
I feel like I'm commit-polluting (if such a word exists) but in my defense, I should say these TS errors were not thrown in my local tests. anyways, sorry. :)

dai-shi:
How about this?

```js
    Object.defineProperty(window, 'localStorage', {
      get() {
         throw new Error('localStorage in not available.')
      },
    })
```

It's probably different from Node.js case, but if we removed try-catch, the test would fail.

mhsattarian:
same behavior with less code. I'm in.

src/vanilla/utils/atomWithStorage.ts
```diff
@@ -161,15 +161,23 @@ export function createJSONStorage<Value>(
         callback(newValue)
       })
 
-  let subscriber = getStringStorage()?.subscribe
+  let subscriber: StringSubscribe | undefined
+  try {
+    subscriber = getStringStorage()?.subscribe
+    // getStringStorage() //?.getItem('')
```
dai-shi, line 167:
```suggestion
```

tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -353,20 +353,49 @@ describe('atomWithStorage (in non-browser environment)', () => {
...
+    })
   })
 
   afterAll(() => {
     window.addEventListener = addEventListener
+    Object.defineProperty(window, 'localStorage', {
+      get() {
+        return localStorage
+      },
+    })
```
dai-shi, line 379: This doesn't really restore, but maybe it's fine.

tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -353,20 +353,49 @@ describe('atomWithStorage (in non-browser environment)', () => {
...
   })
 
   it('createJSONStorage with undefined window.addEventListener', async () => {
     const storage = createJSONStorage(() => asyncDummyStorage)
-
     expect(storage.subscribe).toBeUndefined()
   })
+
+  it('createJSONStorage with localStorage', async () => {
+    expect(() => createJSONStorage(() => window.localStorage)).not.toThrow()
```
dai-shi, line 393: What we should check is not only createJSONStorage, but also atomWithStorage.
You want to confirm if the test works as expected by removing the try-catch we added.

dai-shi (commented):
comments below.

src/vanilla/utils/atomWithStorage.ts
mhsattarian: Ah, this wasn't meant to be in the commit.

tests/react/vanilla-utils/atomWithStorage.test.tsx
mhsattarian: I tested with the try/catch removed and it did throw an error:

CleanShot 2024-05-30 at 15 17 44@2x

should we add a new test or modify this one in you opinion?

tests/react/vanilla-utils/atomWithStorage.test.tsx
mhsattarian: you're right. I removed this part of the code and nothing changes. doesn't it affect using the `localStorage` in other tests?

dai-shi: It may. the best is to leave it for now.

tests/react/vanilla-utils/atomWithStorage.test.tsx
dai-shi: thanks. that's my oversight. yeah, the current test seems good. no need to add new one.

dai-shi: oh, I think what we need to test, if possible, is `createJSONStorage()`. Do you have any idea?

mhsattarian: you mean instead of `createJSONStorage(() => window.localStorage)` we should test `createJSONStorage()`?

dai-shi: Yes, if possible, we want to test both. Otherwise, it's fine for now.

mhsattarian: I simply added another expect to check for it, but I have a feeling you mean something else.
there is also a `console.warn` happening here due to the checking we have in `createJSONStorage`. is it ok?

here is a screenshot:

CleanShot 2024-06-02 at 13 06 43@2x

dai-shi: Thanks.

`except(() => createJSONStorage()).not.toThrow()` seems good. (It fails if we remove the fix, right?)

As for the console.warn, let's mock it for this test and hide the warning.

mhsattarian: > It fails if we remove the fix, right?

Yes


Also, I added `console.warn` patching in the last commit, please check it.

dai-shi (approved):
LGTM
Thanks for your contribution!

mhsattarian:
I'm glad I could be of any help. 🙏

PR #2594, opened 2024-06-02
feat(store2): add dev4_get_mounted_atoms

arjunvegda:
## Summary
- Adds `dev4_get_mounted_atoms` to return Iterable list of mounted atoms

## Check List

### DIFF
--- a/src/vanilla/store2.ts
+++ b/src/vanilla/store2.ts
@@ -243,7 +243,7 @@ const flushPending = (pending: Pending) => {
 // for debugging purpose only
 type DevStoreRev4 = {
   dev4_get_internal_weak_map: () => WeakMap<AnyAtom, AtomState>
-  dev4_get_mounted_atoms: () => IterableIterator<AnyAtom>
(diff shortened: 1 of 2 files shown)


src/vanilla/store2.ts
```diff
@@ -243,6 +243,7 @@ const flushPending = (pending: Pending) => {
 // for debugging purpose only
 type DevStoreRev4 = {
   dev4_get_internal_weak_map: () => WeakMap<AnyAtom, AtomState>
+  dev4_get_mounted_atoms: () => IterableIterator<AnyAtom>
```
dai-shi, line 246: I think it's fine to return the Set, given that our contract is it's read-only.
```suggestion
  dev4_get_mounted_atoms: () => Set<AnyAtom>
```

arjunvegda: Makes sense! Updated it in [this commit](https://github.com/pmndrs/jotai/pull/2594/commits/b10f78126fa5a941aefa25284ea6d9950c46d7a2)

PR #2609, opened 2024-06-07
[WIP] fix(core): expose unstable_resolve instead of unstable_is

dmaskasky:
Replaces PR: https://github.com/pmndrs/jotai/pull/2569

## Context
[jotai-scope](https://github.com/jotaijs/jotai-scope) implementation can be made simpler if the store exposed the ability to customize which atom config to reference based on another atom config.

## Summary
`unstable_resolve` is a new store method. This method customizes how atoms are resolved by their atom config (e.g. using an atom as a key to look up another atom). By default the resolve behavior is `atom => atom`.

### ResolveAtom Invocation Sites
The following functions convert their atom arg to a resolved atom. Since these six functions cover the store interface, it is not necessary to call resolveAtom elsewhere.
1. `readAtom`
1. `writeAtom`
1. `subscribeAtom`
1. `readAtomState -> getter`
1. `writeAtomState -> getter`
1. `writeAtomState -> setter`

Note: store dev methods do not call resolveAtom. The caller of these dev methods is expected to handle atom resolution.

## Example
```js
const pseudo = atom('pseudo')
const a = atom('a')
const store = createStore()
store.unstable_resolve = (atom) => atom === pseudo ? a : atom
expect(store.get(pseudo)).toBe('a')
```

## TODO
1. Determine if this change is sufficient for jotai-scope

## Check List

- [X] `pnpm run prettier` for formatting code and docs

### DIFF
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -40,7 +40,6 @@ type OnMount<Args extends unknown[], Result> = <
 export interface Atom<Value> {
   toString: () => string
   read: Read<Value>
-  unstable_is?(a: Atom<unknown>): boolean
   debugLabel?: string
   /**
    * To ONLY be used by Jotai libraries to mark atoms as private. Subject to change.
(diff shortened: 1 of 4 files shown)


dai-shi (approved):
This looks super cool! Thanks for your contribution.
Now, this PR fixes both store and store2, I think it can be released as a patch.

dmaskasky:
[DO_NOT_MERGE] I am still trying to understand whether this change will satisfy the needs of jotai-scope. I might go in a slightly different direction.

PR #2627, opened 2024-06-20
fix: do not abort atoms on unmount

### LINKED ISSUE #2625: Atoms aborted due to unmount are unrecoverable
## Summary

The signal passed to read functions is aborted when an atom is unmounted. When the atom is read again (even after mounting again), the behaviour from the aborted signal remains (which is usually a thrown AbortError).

This is similar to #2610 (and may in fact be the same bug)

I can think of two solutions:

- Unmounted atoms shouldn't abort (but atoms that have their dependencies change should abort); or
- Unmounted atoms that haven't resolved should have their values discarded (this sounds like a breaking change)

## Reproduction

Here's a test case:

```js
import { createStore, atom } from 'jotai/vanilla';
const store = createStore();

let resolve = () => {};

const a = atom(1);

const b = atom(async (get, { signal }) => {
  await new Promise((r) => (resolve = r));

  signal.throwIfAborted();

  // This also breaks
  // if (signal.aborted) {
  //   return new Promise<never>(() => {})
  // }

  return get(a);
});

let unsub = store.sub(b, () => {});

// triggers an abort
unsub();
resolve();

// we've re-mounted, the aborted error should be irrelevant
unsub = store.sub(b, () => {});

// Should be 1, instead throws "Error: This operation was aborted"
console.log(await store.get(b));
```

Stackblitz reproduction: https://stackblitz.com/edit/stackblitz-starters-tzmn59

## Check List

- [x] I've already opened a [discussion](https://github.com/pmndrs/jotai/discussions) before opening this issue, or already discussed in other media -- given #2610 I didn't see the need to open a discussion

Pinpickle:

Fixes #2625

## Summary

@dai-shi agreed that not aborting promises on unsubscription was the best solution here.

When added to the continuable/cancellable promise idiom, it makes quite a lot of sense. Only abort the promise if you can replace it with another one.

I added the behaviour to store and store2

## Check List

### DIFF
--- a/tests/react/abortable.test.tsx
+++ b/tests/react/abortable.test.tsx
@@ -119,7 +119,7 @@ describe('abortable atom test', () => {
     expect(abortedCount).toBe(1)
   })
 
-  it('can abort on unmount', async () => {
+  it('does not abort on unmount', async () => {
     const countAtom = atom(0)
     let abortedCount = 0
     const resolve: (() => void)[] = []
@@ -169,7 +169,7 @@ describe('abortable atom test', () => {
     await findByText('hidden')
 
     resolve.splice(0).forEach((fn) => fn())
-    await waitFor(() => expect(abortedCount).toBe(1))
+    await waitFor(() => expect(abortedCount).toBe(0))
   })
 
   it('throws aborted error (like fetch)', async () => {
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -362,8 +362,12 @@ export const createStore = (): Store => {
       registerCancelPromise(promise, (next) => {
         if (next) {
           continuePromise(next as Promise<Awaited<Value>>)
+
+          // Only abort promises that aren't user-facing. When next is set,
+          // we can replace the current promise with the next one, so we don't
+          // see any abort-related errors.
+          abortPromise?.()
         }
-        abortPromise?.()
       })
       return setAtomValue(atom, promise as Value, nextDependencies, true)
     }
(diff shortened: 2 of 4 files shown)


Pinpickle:
@dai-shi 

```
 ❯ |jotai| tests/react/abortable.test.tsx  (4 tests | 1 failed) 4208ms
   ❯ tests/react/abortable.test.tsx > abortable atom test > can abort on unmount
     → expected +0 to be 1 // Object.is equality
```

What do we want to do here? As far as I can tell, it's impossible to satisfy this test case.

dai-shi:
> can abort on unmount

Yeah, it's the test case we will break. Just remove it?

dai-shi:
☝️ @Pinpickle

Pinpickle:
@dai-shi sorry! I've pushed the fix now. I've kept the test to assert that there are _no aborts_ on unmount so this behaviour doesn't get reverted :)

Pinpickle:
Oops, had a TS error. Fixed now!

tests/react/abortable.test.tsx
```diff
@@ -119,7 +119,7 @@ describe('abortable atom test', () => {
     expect(abortedCount).toBe(1)
   })
 
-  it('can abort on unmount', async () => {
+  it('does not abort on unmount', async () => {
```
dai-shi, line 122:
```suggestion
  it('does not abort on unmount (#2627)', async () => {
```

dai-shi (approved):
LGTM
Thanks for your contribution!

tests/react/abortable.test.tsx
Pinpickle: This makes sense! Though I see you've already merged!

Pinpickle:
Thanks @dai-shi!

tests/react/abortable.test.tsx
dai-shi: Ah, I forgot to apply changes. 😓

PR #2631, opened 2024-06-23
refactor: atom with observable

tien:
- Simplify subscription logic
- Fix #2225

### DIFF
--- a/src/vanilla/utils/atomWithObservable.ts
+++ b/src/vanilla/utils/atomWithObservable.ts
@@ -42,6 +42,8 @@ type OptionsWithInitialValue<Data> = {
   unstable_timeout?: number
 }
 
+type Result<Data> = { d: Data } | { e: AnyError }
+
 export function atomWithObservable<Data>(
   getObservable: (get: Getter) => SubjectLike<Data>,
   options: OptionsWithInitialValue<Data>,
@@ -66,13 +68,7 @@ export function atomWithObservable<Data>(
   getObservable: (get: Getter) => ObservableLike<Data> | SubjectLike<Data>,
   options?: Options<Data>,
 ) {
-  type Result = { d: Data } | { e: AnyError }
-  const returnResultData = (result: Result) => {
-    if ('e' in result) {
-      throw result.e
-    }
-    return result.d
-  }
+  const EMPTY = Symbol()
 
   const observableResultAtom = atom((get) => {
     let observable = getObservable(get)
@@ -81,43 +77,51 @@ export function atomWithObservable<Data>(
       observable = itself
     }
 
-    let resolve: ((result: Result) => void) | undefined
-    const makePending = () =>
-      new Promise<Result>((r) => {
-        resolve = r
-      })
-    const initialResult: Result | Promise<Result> =
-      options && 'initialValue' in options
-        ? {
-            d:
-              typeof options.initialValue === 'function'
-                ? (options.initialValue as () => Data)()
-                : (options.initialValue as Data),
-          }
-        : makePending()
+    // TODO: replace with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers
+    // once CI Node version is bump to 22+
+    let resolve: (value: Data | PromiseLike<Data>) => void
+    let reject: (reason?: any) => void
+    const promise = new Promise<Data>((_resolve, _reject) => {
+      resolve = _resolve
+      reject = _reject
+    })
 
-    let setResult: ((result: Result) => void) | undefined
-    let lastResult: Result | undefined
-    const listener = (result: Result) => {
-      lastResult = result
-      resolve?.(result)
-      setResult?.(result)
-    }
+    const initialResult =
+      options && 'initialValue' in options
+        ? typeof options.initialValue === 'function'
+          ? (options.initialValue as () => Data)()
+          : (options.initialValue as Data)
+        : promise
+
+    const resultAtom = atom<Result<Data | Promise<Data>>>({ d: initialResult })
+
+    let preMountData: Data | typeof EMPTY = EMPTY
+
+    let setResult: ((result: Result<Data | Promise<Data>>) => void) | undefined
+    const isMounted = () => setResult !== undefined
 
     let subscription: Subscription | undefined
     let timer: Timeout | undefined
-    const isNotMounted = () => !setResult
-    const start = () => {
-      if (subscription) {
+    const startSubscription = () => {
+      if (subscription !== undefined) {
         clearTimeout(timer)
         subscription.unsubscribe()
       }
+
       subscription = observable.subscribe({
-        next: (d) => listener({ d }),
-        error: (e) => listener({ e }),
+        next: (value) => {
+          resolve(value)
+          preMountData = value
+          setResult?.({ d: value })
+        },
+        error: (error) => {
+          reject(error)
+          setResult?.({ e: error })
+        },
         complete: () => {},
       })
-      if (isNotMounted() && options?.unstable_timeout) {
+
+      if (!isMounted() && options?.unstable_timeout) {
         timer = setTimeout(() => {
           if (subscription) {
             subscription.unsubscribe()
@@ -126,62 +130,60 @@ export function atomWithObservable<Data>(
         }, options.unstable_timeout)
       }
     }
-    start()
-
-    const resultAtom = atom(lastResult || initialResult)
 
     if (import.meta.env?.MODE !== 'production') {
       resultAtom.debugPrivate = true
     }
 
-    resultAtom.onMount = (update) => {
-      setResult = update
-      if (lastResult) {
-        update(lastResult)
+    startSubscription()
+
+    resultAtom.onMount = (setAtom) => {
+      setResult = setAtom
+
+      if (preMountData !== EMPTY) {
+        setAtom({ d: preMountData })
       }
-      if (subscription) {
-        clearTimeout(timer)
+
+      if (subscription === undefined) {
+        startSubscription()
       } else {
-        start()
+        clearTimeout(timer)
       }
+
       return () => {
-        setResult = undefined
-        if (subscription) {
-          subscription.unsubscribe()
-          subscription = undefined
-        }
+        subscription?.unsubscribe()
+        subscription = undefined
       }
     }
-    return [resultAtom, observable, makePending, start, isNotMounted] as const
+
+    return { atom: resultAtom, observable }
   })
 
-  if (import.meta.env?.MODE !== 'production') {
-    observableResultAtom.debugPrivate = true
-  }
-
-  const observableAtom = atom(
+  return atom(
     (get) => {
-      const [resultAtom] = get(observableResultAtom)
-      const result = get(resultAtom)
-      if (result instanceof Promise) {
-        return result.then(returnResultData)
-      }
-      return returnResultData(result)
-    },
-    (get, set, data: Data) => {
-      const [resultAtom, observable, makePending, start, isNotMounted] =
-        get(observableResultAtom)
-      if ('next' in observable) {
-        if (isNotMounted()) {
-          set(resultAtom, makePending())
-          start()
+      const unwrapResult = (result: Result<unknown>) => {
+        if ('e' in result) {
+          throw result.e
         }
+
+        return result.d
+      }
+
+      const result = get(get(observableResultAtom).atom)
+
+      if (result instanceof Promise) {
+        return result.then(unwrapResult)
+      }
+
+      return unwrapResult(result)
+    },
+    (get, _, data: Data) => {
+      const observable = get(observableResultAtom).observable
+      if ('next' in observable) {
         observable.next(data)
       } else {
         throw new Error('observable is not subject')
       }
     },
   )
-
-  return observableAtom
 }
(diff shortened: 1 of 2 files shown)


tests/react/vanilla-utils/atomWithObservable.test.tsx
```diff
@@ -408,7 +367,9 @@ it('synchronous subscription with BehaviorSubject', async () => {
 
   const { findByText } = render(
     <StrictMode>
-      <Counter />
+      <Suspense fallback="loading">
```
tien, line 370: Idk why, but this is needed when testing against React 17, which is used by the CI work flow.

tien, line 370: Idk how, but tests were running perfectly fine locally. Only on CI where testing against React 17 will it trigger a suspense, hence needing a fallback :p

tests/react/vanilla-utils/atomWithObservable.test.tsx
```diff
@@ -183,47 +183,6 @@ it('writable count state with delayed value', async () => {
   await findByText('count: 9')
 })
 
-it('only subscribe once per atom', async () => {
```
dai-shi, line 186: This is the requirement and the reason why `atomWithObservable` exists in the Jotai core repo. Can you preserve this behavior?

dai-shi:
> At the expense of setting up more than one subscription.

So, that's not acceptable.

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,107 +74,79 @@ export function atomWithObservable<Data>(
...
+        ? typeof options.initialValue === 'function'
+          ? (options.initialValue as () => Data)()
+          : (options.initialValue as Data)
+        : promise
+
+    const resultAtom = atom(initialResult)
+
+    let preMountData: Data | typeof EMPTY = EMPTY
+
+    const initialSubscription = observable.subscribe({
```
tien, line 97: @dai-shi the extra subscription is here.

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,107 +74,79 @@ export function atomWithObservable<Data>(
...
-        setResult = undefined
-        if (subscription) {
-          subscription.unsubscribe()
-          subscription = undefined
-        }
+      if (preMountData !== EMPTY) {
+        update(preMountData)
       }
+
+      initialSubscription.unsubscribe()
```
tien, line 117: Which will be unsubscribed right away on mount, effective becoming a promise and not really is a subscription.

src/vanilla/utils/atomWithObservable.ts
tien: Thread continue here 👉 https://github.com/pmndrs/jotai/pull/2631/files#r1650220462

tien:
@dai-shi I have re-added the subscription test, and modified it a little to show that there will only ever be one active subscription per atom.

tests/react/vanilla-utils/atomWithObservable.test.tsx
tien: I have updated the test to show that there will be in fact one active subscription per atom.

dai-shi:
The requirement is not one "active" subscription. It's just one subscription without unsubscribing for the normal use case.

tien:
> The requirement is not one "active" subscription. It's just one subscription without unsubscribing for the normal use case.

Hmm 🤔, alright, I'll have a go at keeping that same behaviour 👌

dai-shi:
btw, I'm aware that the current implementation is a hack and not ideal. I'm not very happy with it. So, suggestions are very welcome.

dai-shi (changes requested):
We need to keep the current behavior.

tien:
> btw, I'm aware that the current implementation is a hack and not ideal. I'm not very happy with it. So, suggestions are very welcome.

Honestly, I would like to convince you to accept the one additional subscription if possible :(

It simplifies the logic by a lot and help eliminate that bug from #2225, and you only have to create one additional (short live) subscription. I.e. before, mounting/unmounting an observable atom 20 times create 20 subscriptions (only 1 active), now it will create 21 subscriptions (only 1 active).

I'm still trying to figure out a better way in the meantime.

tien:
@dai-shi okay, I was able to figure out a way to remove that 1 additional subscription.

dai-shi:
> Honestly, I would like to convince you to accept the one additional subscription if possible :(

That's what tried before. I agree such implementation works for many cases and if you want we would create another library in https://github.com/jotaijs/ and publish a separate package.
But, the rationale behind one subscription is it can be a side effect and we want to fire it only once.

tien:
@dai-shi Okay that make sense, thanks for the explanation.

tien:
@dai-shi let me know if these changes are all good now btw 🙏

dai-shi:
> * Simplify subscription logic

Can you briefly explain why/how it's possible?


> * Fix [atomWithObservable unmounts unexepectedly #2225](https://github.com/pmndrs/jotai/discussions/2225)

Is it possible to add a test for #2225?

tien:
@dai-shi 
> Can you briefly explain why/how it's possible?

I just start over again with the goal of adding as little moving parts as possible, tested against #2225 and it works :p
Got to be frank to sure why tbh :))

> Is it possible to add a test for https://github.com/pmndrs/jotai/discussions/2225?

I'm trying right now, but I wasn't able to reproduce this bug at all in `react-testing-library` env :(

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,107 +74,86 @@ export function atomWithObservable<Data>(
...
+          resolve(value)
+          preMountData = value
+          setResult?.(value)
+        },
+        error: (error) => {
+          reject(error)
+          if (error instanceof Error) {
+            // @ts-expect-error TypeScript doesn't allow setting atom to error
+            // If we use Promise.reject here instead, Vitest will complain about an uncaught Error
+            setResult?.(error)
```
dai-shi, line 113: We can't distinguish error and value. Theoretically, value can be `Error`.
Please consider to revert to `{ d: Data } | { e: AnyError }`

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,107 +74,86 @@ export function atomWithObservable<Data>(
...
+          } else {
+            setResult?.(Promise.reject(error))
+          }
+        },
         complete: () => {},
       })
-      if (isNotMounted() && options?.unstable_timeout) {
```
dai-shi, line 120: `unstable_timeout` is a hack to avoid memory leaks when atom never mounts.
I'm not happy with this hack, but it's the only way instructed.
Maybe we should add memory leak tests.

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,107 +74,86 @@ export function atomWithObservable<Data>(
...
-    (get, set, data: Data) => {
-      const [resultAtom, observable, makePending, start, isNotMounted] =
-        get(observableResultAtom)
+  return atom(
+    (get) => get(get(observableResultAtom).atom),
+    (get, _, data: Data) => {
+      const observable = get(observableResultAtom).observable
       if ('next' in observable) {
-        if (isNotMounted()) {
-          set(resultAtom, makePending())
```
dai-shi, line 176: I wonder how it's resolved this case.

dai-shi (commented):
> I just start over again with the goal of adding as little moving parts as possible

It generally looks good and some refactors seems reasonable.
But, it feels you include both the refactor and the fix.
It would be nice if you could separate them into two PRs so that I can review from different perspectives.

src/vanilla/utils/atomWithObservable.ts
tien: For that, I did the below part: `setResult?.(Promise.reject(error))`. Maybe we should always do that?

src/vanilla/utils/atomWithObservable.ts
tien: Not sure about how to test for memory leak, but I'll add this `unstable_timeout` part back.

tien:
> But, it feels you include both the refactor and the fix.

I have no idea what was the actual fix tbh :))

Just testing out random stuffs until it works.

src/vanilla/utils/atomWithObservable.ts
dai-shi: I recently added `tests/vanilla/memoryleaks.test.ts`.
So, we should be able to follow the same technique.
Anyway, you don't need to include the memory leak test in this PR.

tien: @dai-shi also when timing out should we throw an error instead of just un-subscribing?

dai-shi:
> Just testing out random stuffs until it works.

I hope you can spot the exact change that fixes the behavior.
It would be really nice if we could reproduce #2225 behavior in a test.

src/vanilla/utils/atomWithObservable.ts
tien: K I've added back the timeout thing

dai-shi: when timing out, users won't notice anything and it just behaves like two subscriptions (like what you suggested originally.)

src/vanilla/utils/atomWithObservable.ts
tien: Okay, I've cleaned up this bit and also added a test for it 🙏

tien:
> > Just testing out random stuffs until it works.
> 
> I hope you can spot the exact change that fixes the behavior. It would be really nice if we could reproduce #2225 behavior in a test.

Yes I'm trying my best rn, reproducing this in a test turn out to be wayyy harder than fixing it :))

src/vanilla/utils/atomWithObservable.ts
tien: Cool, thanks for explaining it to me 🙏

src/vanilla/utils/atomWithObservable.ts
tien: I don't believe we need to re-trigger the promise here if it has already been resolved?

dai-shi (changes requested):

dai-shi:
> but I wasn't able to reproduce this bug at all in `react-testing-library` env :(

#2638 might or might not help.

tien:
Hey @dai-shi sorry I wasn't keeping up to date with this PR 😢

Had a look again today, I wasn't able to reproduce this error in a unit test, but have resolved your other comments 🙏.

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,43 +81,53 @@ export function atomWithObservable<Data>(
...
+        },
+        error: (error) => {
+          reject(error)
+          setResult?.(WrappedError.from(error))
+        },
         complete: () => {},
       })
-      if (isNotMounted() && options?.unstable_timeout) {
+
+      if (!isMounted() && options?.unstable_timeout !== undefined) {
```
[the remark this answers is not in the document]
tien: I have also reverted this to just a simple truthy check

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,43 +77,51 @@ export function atomWithObservable<Data>(
...
     }
 
-    let resolve: ((result: Result) => void) | undefined
-    const makePending = () =>
-      new Promise<Result>((r) => {
-        resolve = r
-      })
-    const initialResult: Result | Promise<Result> =
+    // TODO: replace with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers
+    // once CI Node version is bump to 22+
```
dai-shi, line 81: It's not just about CI Node version. We support old Node version specified in package.json.

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,43 +77,51 @@ export function atomWithObservable<Data>(
...
+          ? (options.initialValue as () => Data)()
+          : (options.initialValue as Data)
+        : promise
+
+    const resultAtom = atom<Result<Data | Promise<Data>>>({ d: initialResult })
+
+    let preMountData: Data | typeof EMPTY = EMPTY
+
+    let setResult: ((result: Result<Data | Promise<Data>>) => void) | undefined
+    const isMounted = () => setResult !== undefined
```
dai-shi, line 101:
```suggestion
    const isMounted = () => !!setResult
```

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -81,43 +77,51 @@ export function atomWithObservable<Data>(
...
 
     let subscription: Subscription | undefined
     let timer: Timeout | undefined
-    const isNotMounted = () => !setResult
-    const start = () => {
-      if (subscription) {
+    const startSubscription = () => {
+      if (subscription !== undefined) {
```
dai-shi, line 106:
```suggestion
      if (subscription) {
```

dai-shi (commented):
Hmm, this PR is a bit confusing to me. It includes both a) refactoring that keeps basically the same logic, b) fixing an issue, and c) adding a new test.

I would like to review them separately, otherwise can miss something easily. Can you separate PRs?

dai-shi:
> I wasn't able to reproduce this error in a unit test

That's unfortunate. As atomWithObservable is test driven, if we can't add tests, your change might be reverted in the future. Is the newly adding test for the fix?

tien:
Yeah, you are right. I'll close this PR for now until I can find a way to write a unit test for this :(

PR #2632, opened 2024-06-24
feat: atom.toString returns its debugLabel if defined

dmaskasky:
## Summary
If an atom's debugLabel property is defined, the atom toString method will return the debugLabel otherwise the atom's key.
```ts
const baseAtom = atom(0)
baseAtom.debugLabel = 'base'
console.log(String(baseAtom)) // 'base'
```

## Check List

### DIFF
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -92,7 +92,9 @@ export function atom<Value, Args extends unknown[], Result>(
 ) {
   const key = `atom${++keyCount}`
   const config = {
-    toString: () => key,
+    toString() {
+      return this.debugLabel ?? key
+    },
   } as WritableAtom<Value, Args, Result> & { init?: Value }
   if (typeof read === 'function') {
     config.read = read as Read<Value, SetAtom<Args, Result>>


dai-shi:
It's probably controversial, but we shouldn't add "debug" code to production code.
`toString()` isn't for "debug" actually, it's to use an atom for `key={}` prop. (which might not be a very good idea though.)

dmaskasky:
> It's probably controversial, but we shouldn't add "debug" code to production code.
> 
> `toString()` isn't for "debug" actually, it's to use an atom for `key={}` prop. (which might not be a very good idea though.) So, for this usage, `toString()` must return a unique identifier. (which again isn't ideal...)

Oh, ok. I'll abandon this then.

dmaskasky:
> It's probably controversial, but we shouldn't add "debug" code to production code.
> 
> `toString()` isn't for "debug" actually, it's to use an atom for `key={}` prop. (which might not be a very good idea though.) So, for this usage, `toString()` must return a unique identifier. (which again isn't ideal...)

Oh, ok. I'll abandon this then.

PR #2652, opened 2024-07-09
experimental(core): expose unstable_derive

dai-shi:
Another take related with #2609.

This is an experimental unstable feature. Please do not use it in production. Can change/remove it at anytime without notice.


dmaskasky (approved):
I compared with https://github.com/pmndrs/jotai/pull/2609. resolveAtom looks good.

Can you also add [tests/vanilla/store.test.tsx](https://github.com/pmndrs/jotai/pull/2609/files#diff-48e03c388e7ed5d11021367c48a94ccc79febd8d1105af9922e7429279a04ed6)?

dai-shi:
image

Ah, I don't like the code increase and the complexity... Hope there's a better idea.

dai-shi:
I'm not even sure if it's correctly implemented. test cases may help, but it feels like a bad design.

dai-shi:
image

It's a bit better now. The code isn't very explicit compared to `context`, but it's less complicated I suppose.

dmaskasky:
Any concern with `store.unstable_derive` being confused with `jotai-derive`? https://github.com/pmndrs/jotai/pull/2656

dai-shi:
I think it's fine.

dai-shi:
@dmaskasky I will merge this. It doesn't mean we finalize it yet. We hope it though.

dmaskasky:
Ok, makes sense. In a few weeks I will try again to make this work. I might need some more help though.

PR #2655, opened 2024-07-13
Benchmarks

L-Qun:
## Summary

I think benchmarks are really important; we also have a lot of them in the source code. However, we need a workflow to check every PR for any potential performance degradation.

image

## Check List

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -6,11 +6,10 @@ on:
   pull_request:
     types: [opened, synchronize]
 
-permissions:
-  issues: write
-
 jobs:
   benchmarks:
+    permissions:
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


L-Qun:
image

@dai-shi It seems we need to config the jotai repo, can you help to take a look at this issue?

dai-shi:
Hi, thanks for your suggestion.

I agree that benchmarks are important, but the one we have now is very specific and doesn't cover real use cases. I don't see much benefit of adding a workflow with it.
That said, it would be nice if you create some benchmark suites. I actually think we can drop what we have now.

I don't think a workflow has a permission to write a comment from a forked repo. This happens with `compressed-size-action.yml` too. Anyway, the workflow isn't very important. The benchmark itself would be important.

L-Qun:
> Hi, thanks for your suggestion.
> 
> I agree that benchmarks are important, but the one we have now is very specific and doesn't cover real use cases. I don't see much benefit of adding a workflow with it. That said, it would be nice if you create some benchmark suites. I actually think we can drop what we have now.
> 
> I don't think a workflow has a permission to write a comment from a forked repo. This happens with `compressed-size-action.yml` too. Anyway, the workflow isn't very important. The benchmark itself would be important.

Got it thanks for your reply

PR #2260, opened 2023-11-20
docs: add `ai.mdx`

himself65:
## Summary

https://github.com/jotaijs/jotai-ai

This PR adds a third party section that the packages from the community.

The document hasn't been finished yet; you can give me feedback. Thanks.

## Check List

- [ ] `yarn run prettier` for formatting code and docs

PR #2656, opened 2024-07-14
docs: add `jotai-derive` extension documentation

iwoplaza:

Adds documentation about an extension discussed in #2640 .

## Check List


dai-shi:
There is a distinction between "semi-official" libs and "third-party" libs.

https://github.com/pmndrs/jotai/pull/2260#issuecomment-1819992624

Now, this is a great discussion point.
I think eventually we need some guidance about moving from a third-party to a semi-official,
and also moving the other way around.
(We could even consider having "semi-official" and "third-party" in the docs for more visibility. cc @sandren I was hoping https://github.com/jotaijs/links to work, but it doesn't.)

On the other hand, I personally think `jotai-derive` is worth a semi-official library. But, it also feels too new. Someone has to review the code too.

Let's see what others think. @Aslemammad @dmaskasky @arjunvegda @DavidJFelix

docs/extensions/derive.mdx
```diff
@@ -0,0 +1,145 @@
...
+  )
+);
+```
+
+### Conditional dependency (multiple conditions)
+
+```ts
+// pipes allow for cleaner code when using `soon` directly.
+import { pipe } from 'remeda';
+import { soon, soonALl } from 'jotai-derive';
```
dmaskasky, line 122: nit: `soonAll`

docs/extensions/derive.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+// Atom<boolean | Promise<boolean>>
+const enabledAtom = ...;
+
+// Atom<null | RestrictedItem | Promise<null | RestrictedItem>>
+const restrictedItemAtom = atom((get) =>
+  pipe(
+    soonAll(get(isAdminAtom), get(enabledAtom)),
+    soon(([isAdmin, enabled]) => (isAdmin && enabled ? get(queryAtom) : null))
+  )
```
sebinsua, line 138: Out of interest, how would these examples be written as normal procedural-looking JavaScript without usage of `remeda`?

iwoplaza: Here is the example in question, with `remeda` removed.

```ts
import { soon, soonAll } from 'jotai-derive';

// Atom<RestrictedItem | Promise<RestrictedItem>>
const queryAtom = ...;

// Atom<boolean | Promise<boolean>>
const isAdminAtom = ...;

// Atom<boolean | Promise<boolean>>
const enabledAtom = ...;

// Atom<null | RestrictedItem | Promise<null | RestrictedItem>>
const restrictedItemAtom = atom((get) => {
  const dependencies = soonAll(get(isAdminAtom), get(enabledAtom));

  return soon(
    dependencies,
    ([isAdmin, enabled]) => (isAdmin && enabled ? get(queryAtom) : null)
  );
});

```

dai-shi:
Just noticed that https://github.com/jotaijs/jotai-derive is still a private repo. Is it intentional?

iwoplaza:
> Just noticed that https://github.com/jotaijs/jotai-derive is still a private repo. Is it intentional?

@dai-shi Definitely an oversight on my part, it's public now! 🙌

docs/extensions/derive.mdx
jaerod95: Would love to see this non-remeda example added to the documentation, as it was very foggy on how to use this without `remeda` during my initial investigation.

dai-shi: Does README help meanwhile? https://github.com/jotaijs/jotai-derive/blob/main/README.md

dai-shi (approved):
Yeah, it feels more comfortable under the third-party category. Thanks @sandren !

PR #2660, opened 2024-07-15
fix(core): Unmount an atom that is no longer dependent within a derived atom

nogaten:

Fix #2658 

## Summary
Add test for "Unmount an atom that is no longer dependent within a derived atom (#2658)" and fix.

## Check List


dai-shi:
Thanks for the failing test!

I can take a closer look later, but if anyone is interested, feel free to take the challenge.

nogaten:
Fix #2658.
When executing unmountAtom, it is confirmed that the dependent atom is not mounted, or even if it is mounted, it does not exist in dependent.

dai-shi (approved):
wow, it looks good to me! thanks for the fix.
there might be some other edge cases, this doesn't cover, but as now we have the test, we can refactor in the future.

PR #2678, opened 2024-07-29
idea1: add originator prop to atomFamily atoms

dmaskasky:

Fixes #

## Summary
```jsx
<ScopedProvider atoms={[fooFamily]}>{children}</ScopeProvider>
```
This PR demonstrates one approach in which the atomFamily annotates atoms created by this approach with an originator symbol set to the createAtom function.

The following pseudo code represents logic that would live in jotai-scope
```js
const isExplicitlyScoped = atomSet.has(atom) || atomSet.has(atom[SymbolOriginator])
```

## Check List

### DIFF
--- a/src/vanilla/utils/atomFamily.ts
+++ b/src/vanilla/utils/atomFamily.ts
@@ -2,6 +2,8 @@ import type { Atom } from '../../vanilla.ts'
 
 type ShouldRemove<Param> = (createdAt: number, param: Param) => boolean
 
+export const SymbolOriginator = Symbol('Originator')
+
 export interface AtomFamily<Param, AtomType> {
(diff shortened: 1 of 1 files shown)


src/vanilla/utils/atomFamily.ts
```diff
@@ -43,6 +45,7 @@ export function atomFamily<Param, AtomType extends Atom<unknown>>(
     }
 
     const newAtom = initializeAtom(param)
+    ;(newAtom as any)[SymbolOriginator] = createAtom
```
dai-shi, line 48: Is `createAtom` just used as a reference key?
It doesn't seem very straightforward and useful.

PR #2679, opened 2024-07-29
idea2: add getParams to createAtom

dmaskasky:
Discussion: https://github.com/pmndrs/jotai/discussions/2056

Fixes #

## Summary
```jsx
<ScopedProvider atoms={[fooFamily]}>{children}</ScopeProvider>
```
This PR demonstrates one approach in which the atomFamily supports a new `getParams` function.
The following pseudo code represents logic that would live in jotai-scope. For obvious complexity reasons, I find this solution less favorable to https://github.com/pmndrs/jotai/pull/2678.
```js
const isExplicitlyScoped = atomSet.has(atom) || 
  Array.from(atomSet)
    .filter(isAtomFamily)
    .some((af) => af.getParams().map(af).includes(atom))
```

## Check List

### DIFF
--- a/src/vanilla/utils/atomFamily.ts
+++ b/src/vanilla/utils/atomFamily.ts
@@ -47,6 +47,10 @@ export function atomFamily<Param, AtomType extends Atom<unknown>>(
     return newAtom
   }
 
+  createAtom.getParams = () => {
+    return Array.from(atoms.keys())
+  }
+
   createAtom.remove = (param: Param) => {
     if (areEqual === undefined) {
       atoms.delete(param)


src/vanilla/utils/atomFamily.ts
```diff
@@ -47,6 +47,10 @@ export function atomFamily<Param, AtomType extends Atom<unknown>>(
     return newAtom
   }
 
+  createAtom.getParams = () => {
```
dai-shi, line 50: This looks good as it's sometimes requested.

src/vanilla/utils/atomFamily.ts
```diff
@@ -47,6 +47,10 @@ export function atomFamily<Param, AtomType extends Atom<unknown>>(
...
+    return Array.from(atoms.keys())
```
dai-shi, line 51: Maybe we should just return the iterator:
```suggestion
    return atoms.keys()
```

dmaskasky:
I don't think this change will satisfy jotai-scope's use-case. Closing this PR for now, we can reopen when a community need arises.

dai-shi:
I think this should be reopened and we consider merging, even if it doesn't work for `jotai-scope`. But, yeah, we can do it when there's a need.

PR #2681, opened 2024-07-30
idea3: atomFamily supports instanceof

dmaskasky:

Fixes #

## Summary
```jsx
<ScopedProvider atomFamilies={[fooFamily]}>{children}</ScopeProvider>
```
This PR demonstrates one approach in which the atomFamily supports use of instanceof language feature.

The following pseudo code represents logic that would live in jotai-scope.
```js
const isExplicitlyScoped = atomSet.has(atom) || Array.from(atomFamilySet).some(af => atom instanceof af)
```

## Check List

### DIFF
--- a/src/vanilla/utils/atomFamily.ts
+++ b/src/vanilla/utils/atomFamily.ts
@@ -42,10 +42,15 @@ export function atomFamily<Param, AtomType extends Atom<unknown>>(
       }
     }
 
-    const newAtom = initializeAtom(param)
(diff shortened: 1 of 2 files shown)


dai-shi:
I don't think this complexity and the trick should be added in this util.
Please consider patching `atomFamily` in `jotai-scope`.

dai-shi:
If we were to go this road, I think I'd prefer adding `.has(atom)` method. But it feels too specific to `jotai-scope`.

PR #2683, opened 2024-07-30
idea4: atomFamily supports has

dmaskasky:
idea3: https://github.com/pmndrs/jotai/pull/2681

Fixes #

## Summary
```jsx
<ScopedProvider atomFamilies={[fooFamily]}>{children}</ScopeProvider>
```
This PR demonstrates one approach in which the atomFamily supports `has` for O(1) atom lookup.

The following pseudo code represents logic that would live in jotai-scope.
```js
const isExplicitlyScoped = atomSet.has(atom) || Array.from(atomFamilySet).some(af => af.has(atom))
```

## Check List

### DIFF
--- a/src/vanilla/utils/atomFamily.ts
+++ b/src/vanilla/utils/atomFamily.ts
@@ -4,6 +4,7 @@ type ShouldRemove<Param> = (createdAt: number, param: Param) => boolean
 
 export interface AtomFamily<Param, AtomType> {
   (param: Param): AtomType
+  has(atom: AtomType): boolean
   remove(param: Param): void
   setShouldRemove(shouldRemove: ShouldRemove<Param> | null): void
 }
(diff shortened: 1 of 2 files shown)


src/vanilla/utils/atomFamily.ts
```diff
@@ -4,6 +4,7 @@ type ShouldRemove<Param> = (createdAt: number, param: Param) => boolean
 
 export interface AtomFamily<Param, AtomType> {
   (param: Param): AtomType
+  has(atom: AtomType): boolean
```
dai-shi, line 7: Thanks. This is exactly what I imagined including the implementation.

I'm a bit hesitant with this new api, though.
- I don't see other use cases than jotai-scope
- I would like to see another solution "injecting callback" and compare with it

```suggestion
  onCreate(callback: (param: Param, atom: AtomType) => void): void
  getParams(): Iterator<Param>
```

dai-shi: > "injecting callback"

It requires `onRemove` too.

PR #2695, opened 2024-08-07
breaking(core): avoid continuable promise in store api

dai-shi:
fix #2682

This introduces a behavioral breaking change in the promise behavior with Store API.


dai-shi:
Hm, this requires `AbortController` always, whereas previously it's only used when `{ signal }` is used.

dai-shi:
https://github.com/pmndrs/jotai/discussions/2682#discussioncomment-10451718
Waiting for @martinhath response.

martinhath:
I’m unable to work on this for the next two weeks, but will happily continue after that @dai-shi 😄

dai-shi:
@martinhath I made a small change, which may not relate to your issue, but just in case.

dmaskasky:
> fix #2682
> 
> This introduces a behavioral breaking change in the promise behavior with Store API.

What is the change in behavior?

dai-shi:
> What is the change in behavior?

Previously, `store.get(atom)` returns a "continuable" promise, but with this PR it's not the case. So, the change is required for #2682. See some tests are changed, some of which might have been added by you.

martinhath:
I've been testing it again, and can't reproduce the error I got last time 🤔

dai-shi:
Thanks for confirming. I think ba0a731 improved it.

PR #2698, opened 2024-08-08
chore: update typescript-eslint

### LINKED ISSUE #2699: 🛠️ Repo: ESLint is running TypeScript rules on .json files
👋 Coming over from https://x.com/dai_shi/status/1821338388957388859 -> #2698, I see you're running ESLint on all `.js`, `.json`, and `.ts` files:

https://github.com/pmndrs/jotai/blob/20e0cde354c77dd7a9aee7bee4f54f83ea56ebb4/package.json#L77

...which means you're running all sorts of JavaScript- and TypeScript-specific rules on `.json` files like `tsconfig.json`. Was this intentional?

The applying JS/TS rules when linting `.json` files is why #2698 has a [`plugin:@typescript-eslint-no-unused-expressions`](https://typescript-eslint.io/rules/no-unused-expressions) failures in the `tsconfig.json` file:

```plaintext
/home/runner/work/jotai/jotai/tsconfig.json
Error:   1:1  error  Expected an assignment or function call and instead saw an expression  @typescript-eslint/no-unused-expressions
```

---

Posting some suggested next steps in case it's useful, sorry if you already know all this 😅 ...

If you _don't_ want to lint `.json` files, you can remove the `json` from the linting scripts.

If you _do_ want to lint `.json` files, I'd suggest:

1. Switching all JS- and TS-related plugins to be in [eslintrc `overrides`](https://eslint.org/docs/latest/use/configure/configuration-files-deprecated#configuration-based-on-glob-patterns) that apply to `.js` and `.ts` files
2. Onboarding `.json`-specific plugins such as [`eslint-plugin-jsonc`](https://github.com/ota-meshi/eslint-plugin-jsonc) in `overrides` for `.json` files

Hope this is helpful - cheers! 💙 

Edit: oh, and if this makes no sense, I can send a draft PR showing suggested changes. 🙂

dai-shi:
close #2699


tests/react/utils/types.test.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint @typescript-eslint/no-unused-expressions: off */
```
dai-shi, line 1: Hm, I now wonder if we need `function Component() {`...

tests/vanilla/utils/types.test.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint @typescript-eslint/no-unused-vars: off */
```
dai-shi, line 1: @JoshuaKGoldberg is there an option to mark `typeof foo` as used?

JoshuaKGoldberg: Not at the moment, no. But I think this is an interesting use case that _might_ be enough to warrant one?

We'd talked about it before in:
* https://github.com/typescript-eslint/typescript-eslint/issues/9679
* https://github.com/typescript-eslint/typescript-eslint/issues/9697

...but neither is _exactly_ what's happening here. 

For this specific file, I think the user need here is that you...

* Need to have the typings tests code wrapped in functions so that they doesn't actually run
* Are using `function Component`s for that, which are getting flagged as unused

...is that right?

If so: I'd suggest naming them something like `function _test` per the `varsIgnorePattern`. That way you can keep `no-unused-vars` enabled & have it just not care about the functions.

Example benefit: `types.test.tsx` has an unused import (`expect`) right now that would be caught by the rule. 🧹

dai-shi: Hm, that's unfortunate.

I was talking about `typeof foo`, not about `function test`, which I already resolved.

Thanks for all of your help. I will go with overrides for tests.

JoshuaKGoldberg: Oh! Sorry, I misread. But great!

PR #2700, opened 2024-08-08
fix(store): derived atom not updating when its value changes if another atom is read it conditionally

yuneco:

Fixes #2697

## Summary

Seems a regression in store2 introduced in v2.9.0.

## Check List


dai-shi:
https://github.com/pmndrs/jotai/discussions/2697#discussioncomment-10280687

src/vanilla/store.ts
```diff
@@ -509,10 +509,12 @@ export const createStore = (): Store => {
     // Track what's changed, so that we can short circuit when possible
     const changedAtoms = new Set<AnyAtom>([atom])
     const isMarked = (a: AnyAtom) => markedAtoms.has(a)
+    const prevEpochNumbers: number[] = topsortedAtoms.map(
+      (a) => getAtomState(a).n,
+    )
```
dai-shi, line 514: Looks good. Let me add a commit to refactor something.

dai-shi (approved):
This should be a good fix. Let's ship it.

PR #2712, opened 2024-08-22
docs: removes stop observing pattern section from performance page

mxthxngx:
[Linked Discussion](https://github.com/pmndrs/jotai/discussions/2395)

Fixes #

## Summary
Removed the "Stop observing pattern" section from the performance page as it uses hooks incorrectly and does not work.

## Check List

- [✅] `pnpm run prettier` for formatting code and docs


dai-shi (approved):
LGTM
Feel free to have suggestions for other parts of the guide, and adding new sections is also good.

PR #2713, opened 2024-08-23
fix(store2): reusing continuable promise causes dependencies to be lost

dmaskasky:
https://github.com/jotaijs/jotai-effect/issues/42

## Summary
When a promise is reused on subsequent recomputations, the dependencies are not preserved.
~This is affecting the functionality of atomEffect.~ A [workaround](https://github.com/jotaijs/jotai-effect/pull/43) has been found.
This test is broken since https://github.com/pmndrs/jotai/commit/1a40452738cc40176d44d27b9e7f52dbd156f0ba

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -385,12 +385,8 @@ const buildStore = (getAtomState: StoreArgs[0]): Store => {
       }
     }
 
-    if (isContinuablePromise(atomState.v) && atomState.v.status === PENDING) {
-      // If the atom has a pending promise, we should preserve the dependents
-    } else {
-      // Compute a new state for this atom.
-      atomState.d.clear()
-    }
+    // Compute a new state for this atom.
+    atomState.d.clear()
     let isSync = true
     const getter: Getter = <V>(a: Atom<V>) => {
       if (isSelfAtom(atom, a)) {
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -384,8 +384,13 @@ const buildStore = (getAtomState: StoreArgs[0]): Store => {
         return atomState
       }
     }
-    // Compute a new state for this atom.
-    atomState.d.clear()
+
+    if (isContinuablePromise(atomState.v) && atomState.v.status === PENDING) {
+      // If the atom has a pending promise, we should preserve the dependents
+    } else {
+      // Compute a new state for this atom.
+      atomState.d.clear()
+    }
     let isSync = true
     const getter: Getter = <V>(a: Atom<V>) => {
       if (isSelfAtom(atom, a)) {
--- a/tests/vanilla/store.test.tsx
+++ b/tests/vanilla/store.test.tsx
@@ -600,13 +600,9 @@ it('should preserve dependencies when returning a reused promise', async () => {
     promise: Promise.resolve(-1),
     inProgress: false,
   }))
-  const derivedAtom = atom<Promise<number>, [], void>(
+  const internalAtom = atom<Promise<number>, [], void>(
     (get, { setSelf }) => {
-      get(refreshAtom)
       const ref = get(refAtom)
-      if (ref.inProgress) {
-        return ref.promise
-      }
       ref.promise = Promise.resolve().then(() => {
         const value = get(baseAtom)
         ref.inProgress = true
@@ -616,6 +612,7 @@ it('should preserve dependencies when returning a reused promise', async () => {
       })
       Promise.resolve().then(() => {
         ref.inProgress = false
+        setSelf()
       })
       return ref.promise
     },
@@ -623,6 +620,14 @@ it('should preserve dependencies when returning a reused promise', async () => {
       set(refreshAtom, (c) => c + 1)
     },
   )
+  const derivedAtom = atom((get) => {
+    get(refreshAtom)
+    const ref = get(refAtom)
+    if (ref.inProgress) {
+      return ref.promise
+    }
+    return get(internalAtom)
+  })
   const store = createStore()
 
   store.sub(derivedAtom, () => {})
--- a/tests/vanilla/store.test.tsx
+++ b/tests/vanilla/store.test.tsx
@@ -595,19 +595,13 @@ it('should update derived atom even if dependances changed (#2697)', () => {
 
 it('should preserve dependencies when reusing continuable promise', async () => {
   const baseAtom = atom(0)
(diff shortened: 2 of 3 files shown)


dmaskasky:
I think I understand the issue from the outside better.

When derivedAtom is read, it adds the dependency `baseAtom`.
Then it returns a promise that resolves after the next microtask.
Before the promise resolves, it calls setSelf, which updates `baseAtom`. This causes `derivedAtom`  to recompute.
Because inProgress is now true, it early returns the original promise _before_ `baseAtom` is added as a dependency in this run.
Before 2.9.0, the dependencies from prior runs would be preserved if the original promise is returned.
However in 2.9.0, the dependencyMap is reset on each recursive run.
```ts
const baseAtom = atom(0)
let promise
const derivedAtom = atom<Promise<number>, [], void>(
  (get, { setSelf }) => {
    if (inProgress) {
      // bailing out when inProgress is true causes the original dependencies to be lost
      return promise
    }
    const value = get(baseAtom)
    promise = Promise.resolve().then(() => {
      inProgress = true
      // causes the derivedAtom to be re-evaluated, which triggers the inProgress bail
      setSelf()
      return value
    })
    return promise
  },
  (_get, _set) => {
    // setting baseAtom to Infinity causes the derivedAtom to be re-evaluated
    _set(baseAtom, c => c + 1)
  },
)
```

src/vanilla/store.ts
```diff
@@ -384,8 +384,13 @@ const buildStore = (getAtomState: StoreArgs[0]): Store => {
         return atomState
       }
     }
-    // Compute a new state for this atom.
-    atomState.d.clear()
+
+    if (isContinuablePromise(atomState.v) && atomState.v.status === PENDING) {
+      // If the atom has a pending promise, we should preserve the dependents
```
dai-shi, line 389: I'm not sure if I understand the rationale behind it.
We are recomputing anyway, right? Is it because the "continuable promise" doesn't change its reference after the recomputation?

tests/vanilla/store.test.tsx
```diff
@@ -592,3 +592,43 @@ it('should update derived atom even if dependances changed (#2697)', () => {
   store.set(primitiveAtom, 1)
   expect(onChangeDerived).toHaveBeenCalledTimes(1)
 })
+
+it('should preserve dependencies when reusing continuable promise', async () => {
+  const baseAtom = atom(0)
+  let promise = Promise.resolve(-1)
+  let inProgress = false
```
dai-shi, line 599: I think this test is overcomplicated.
Can you simply without using `setSelf` and maybe `inProgress`?
Not depending on micro/macro tasks, but do manual control like: `resolve()` or `resolve.splice(0).forEach((fn) => fn())`

tests/vanilla/store.test.tsx
```diff
@@ -592,3 +592,43 @@ it('should update derived atom even if dependances changed (#2697)', () => {
...
+  const derivedAtom = atom<Promise<number>, [], void>(
+    (get, { setSelf }) => {
+      if (inProgress) {
+        return promise
```
dai-shi, line 603: Does this change the behavior?
```suggestion
        return Promise.resolve(-1)
```

dai-shi (commented):
Thanks for your investigation. Comments below.

tests/vanilla/store.test.tsx
dmaskasky: I'm struggling to find a simpler test.
`setSelf` is used to cause a recomputation in the same microtask

tests/vanilla/store.test.tsx
dmaskasky: `return Promise.resolve(store.get(baseAtom))` does not change behavior.
Do you think it should?

dmaskasky: Ah, it must be the identically same promise in 2.8.4. So the logic needs to be updated to see if the promise is exactly the same.

src/vanilla/store.ts
dmaskasky: in `2.8.4`, if the returned promise doesn't change reference, the dependents are not cleared.

dai-shi: yeah, I mean, this is _kind of_ an intentional change in v2.9.0.

tests/vanilla/store.test.tsx
dai-shi: I misunderstood something. Forget about it.

tests/vanilla/store.test.tsx
dai-shi: > `setSelf` is used to cause a recomputation in the same microtask.

oh, i see.

tests/vanilla/store.test.tsx
```diff
@@ -592,3 +592,45 @@ it('should update derived atom even if dependances changed (#2697)', () => {
...
+      promise = Promise.resolve().then(() => {
+        const value = get(baseAtom)
+        inProgress = true
+        // causes the derivedAtom to be re-evaluated, which triggers the inProgress bail
+        setSelf()
+        return value
+      })
+      setTimeout(() => {
+        inProgress = false
+      })
```
dai-shi, line 616: If this is a microtask delay, doesn't it reproduce the issue?
```suggestion
      Promise.resovle().then(() => {
        inProgress = false
      })
```

dmaskasky: Yes, still reproducible with a microtask delay.

dai-shi:
> let inProgress = false

It feels like violating the contract. Can we somehow avoid it?

dmaskasky:
> > let inProgress = false
> 
> It feels like violating the contract. Can we somehow avoid it?
> 
> [194cdc2](https://github.com/pmndrs/jotai/commit/194cdc2aeb120662a7771286e3f174de57995e34) for example.

Ok, I moved it to a ref atom.

dai-shi:
> > It feels like violating the contract. Can we somehow avoid it?
> > [194cdc2](https://github.com/pmndrs/jotai/commit/194cdc2aeb120662a7771286e3f174de57995e34) for example.
> 
> Ok, I moved it to a ref atom.

Can you also divide derivedAtom into two atoms?

dai-shi:
>  reusing continuable promise causes dependencies to be lost

is the expected behavior. I'll cherry-pick my commit and open a new PR.

dmaskasky:
Thanks for looking into this further.

PR #2734, opened 2024-09-16
style: removed duplicate width property in CSS files

YuHyeonWook:
## Summary
- Removed the duplicate  'width' property in the CSS files.

## Check List


dai-shi (approved):
👍

PR #2741, opened 2024-09-19
fix(unstable_derive): trap atom methods

dai-shi:
This enables to trap `atom.read`, `atom.write` and `atom.onMount`.
This eliminates `originAtomState` for `getAtomState`.


src/vanilla/store.ts
```diff
@@ -718,7 +729,7 @@ const buildStore = (getAtomState: StoreArgs[0]): Store => {
   }
 
   const unstable_derive = (fn: (...args: StoreArgs) => StoreArgs) =>
-    buildStore(...fn(getAtomState))
+    buildStore(...fn(getAtomState, atomRead, atomWrite))
```
dmaskasky, line 732: might not be necessary, not sure though.

src/vanilla/store.ts
```diff
@@ -245,7 +251,12 @@ type Store = PrdStore | (PrdStore & DevStoreRev4)
 export type INTERNAL_DevStoreRev4 = DevStoreRev4
 export type INTERNAL_PrdStore = PrdStore
 
-const buildStore = (getAtomState: StoreArgs[0]): Store => {
+const buildStore = (
+  getAtomState: StoreArgs[0],
+  atomRead: StoreArgs[1],
+  atomWrite: StoreArgs[2],
+  atomOnMount: StoreArgs[3],
+): Store => {
```
dmaskasky, line 259:
```suggestion
const buildStore = (
  ...[getAtomState, atomRead, atomWrite, atomOnMount]: StoreArgs
): Store => {
```

dmaskasky:
I like the atomOnMount trap.

I'm not sure about removing originalAtomState just yet though. It could be important for jotai-scope if it could be used to reliably resolve atomState. Earlier implementation was consistent for readAtomState and writeAtomState, but inconsistent / missing in other cases such as subscribe.

src/vanilla/store.ts
dai-shi: nice!

dai-shi:
> I'm not sure about removing originalAtomState just yet though.

As it's technically impossible to solve missing cases, I think it was an insufficient idea. Let's drop it and merge this.

dmaskasky:
> As it's technically impossible to solve missing cases, I think it was an insufficient idea. Let's drop it and merge this.

Yes I agree.

PR #2742, opened 2024-09-19
fix(core): recompute dependents in edge cases

dai-shi:
Fix #2738


src/vanilla/store.ts
```diff
@@ -288,13 +288,14 @@ const buildStore = (getAtomState: StoreArgs[0]): Store => {
     pending: Pending | undefined,
     atom: Atom<Value>,
     atomState: AtomState<Value>,
-    force?: (a: AnyAtom) => boolean,
+    dirtyAtoms?: Set<AnyAtom>,
+    force?: boolean,
   ): AtomState<Value> => {
     // See if we can skip recomputing this atom.
-    if (!force?.(atom) && isAtomStateInitialized(atomState)) {
+    if (!force && isAtomStateInitialized(atomState)) {
```
Pinpickle, line 295: @dai-shi I'm not sure why but removing `!force` from this makes all tests pass... do you know why this would work?

dai-shi: cd67e40 --- yeah. I think we miss some optimization tests. I'll try to add one.

dai-shi: Hmm, it actually seems fine, because in normal cases (no reordering), dirtyAtoms doesn't include dependencies. So, no performance drawback.

dai-shi (commented):
There might exist some other edge cases we miss now, but the change is small and looks reasonable.

Thanks to @Pinpickle

xc1427:
This change breaks my app related to some atomEffect usage, anyway I have to figure out why and post an minimum viable repro.

dai-shi:
Oh... Yeah, a repro would be helpful and we need a test case to investigate further.

dmaskasky:
I installed both `2.10.1` and `https://pkg.csb.dev/pmndrs/jotai/commit/79adf231/jotai` as dev dependency for jotai-effect. All tests pass with both versions.

I might be missing some test cases. Please create a minimally viable repro and I can take a look.

na2hiro:
I also got exceptions with `2.10.1` that doesn't happen in `2.10.0`. Right now I don't have time to reduce this down to repro but here's the stack trace

```
Uncaught RangeError: Maximum call stack size exceeded
    at isAtomStateInitialized (chunk-CGZX2XVC.js?v=367fa982:63:30)
    at readAtomState (chunk-CGZX2XVC.js?v=367fa982:164:9)
    at chunk-CGZX2XVC.js?v=367fa982:172:11
    at Array.every (<anonymous>)
    at readAtomState (chunk-CGZX2XVC.js?v=367fa982:168:35)
    at chunk-CGZX2XVC.js?v=367fa982:172:11
    at Array.every (<anonymous>)
    at readAtomState (chunk-CGZX2XVC.js?v=367fa982:168:35)
    at chunk-CGZX2XVC.js?v=367fa982:172:11
    at Array.every (<anonymous>)
```

dai-shi:
It feels like a bug with this PR. Hope someone can create a failing test case!

xc1427:
@dai-shi @na2hiro @dmaskasky 

FYI, I fixed the issue (Uncaught RangeError: Maximum call stack size exceeded) in my app.

It seems I may have introduced "circular dependencies" between an atom and the effect of it, like below,

```typescript
const atomA = atom((get) => {
  get(atomAEffect);
})
const atomAEffect = atomEffect((get) => {
   get(atomA)
})
```
I think I had exploited a hidden feature by accident. This "feature" is apparently not allowed since version 2.10.1.

I am now surprised that my app worked 😂😂. Is that usage legal ? I remember the purpose of organizing code like this is to colocate the effect with the atom self.

dai-shi:
https://github.com/pmndrs/jotai/discussions/2823#discussioncomment-11281929

na2hiro:
It turned out that I was making a direct circular usage of `get()` between an atom and an atom affect, similarly to what @xc1427 described. Removing it fixed my exception. Thanks folks for the help.

> I remember the purpose of organizing code like this is to colocate logically atom and its effect, and make sure I do not forget to mount the effect. Because when atomA is mounted, atomAEffect will also take effect. The purpose of atomAEffect is to let atom's change be propagated to some other places.
> 
> Is that usage legal ? If not, what is the best practice ?

To answer @xc1427's question, @dmaskasky posted in #2823 how to always hook effect into an atom. With exporting `countWithEffect` without `countAtom`, we can force a client code to use always with an effect:

> You can use the withAtomEffect api.
> 
> ```
> const countAtom = atom(0)
> export const countWithEffect = withAtomEffect(countAtom, (get, set) => {
>   console.log('count is now', get(countAtom))
> })
> ```

(I'd name the first one `_countAtom` and the second one `countAtom` when effect is a must btw)

PR #2753, opened 2024-09-27
:white_check_mark: test: add unit test for atomWithReset utility

vangie:
## Summary

Add unit test for src/vanilla/utils/atomWithReset.ts

## Check List

### DIFF
--- a/tests/vanilla/utils/atomWithReset.test.ts
+++ b/tests/vanilla/utils/atomWithReset.test.ts
@@ -12,21 +12,11 @@ describe('atomWithReset', () => {
     testAtom = atomWithReset(initialValue)
   })
 
-  it('should create an atom with initial value', () => {
-    const { init } = testAtom
-    expect(init).toBe(initialValue)
-  })
-
   it('should reset to initial value using RESET', () => {
     const store = createStore()
     store.set(testAtom, 123)
     store.set(testAtom, RESET)
     expect(store.get(testAtom)).toBe(initialValue)
-
-    const set = vi.fn()
-    const get = vi.fn(() => 20)
-    testAtom.write(get, set, RESET)
-    expect(set).toHaveBeenCalledWith(testAtom, initialValue)
   })
 
   it('should update atom with a new value', () => {
@@ -34,11 +24,6 @@ describe('atomWithReset', () => {
     store.set(testAtom, 123)
     store.set(testAtom, 30)
     expect(store.get(testAtom)).toBe(30)
-
-    const set = vi.fn()
-    const get = vi.fn(() => 20)
-    testAtom.write(get, set, 30)
-    expect(set).toHaveBeenCalledWith(testAtom, 30)
   })
 
   it('should update atom using a function', () => {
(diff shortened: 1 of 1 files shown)


tests/vanilla/utils/atomWithReset.test.ts
```diff
@@ -0,0 +1,48 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { RESET, atomWithReset } from 'jotai/vanilla/utils'
+
+vi.mock('jotai/vanilla', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('jotai/vanilla')>()
+  return {
+    ...actual,
+    atom: vi.fn(actual.atom),
+  }
+})
```
dai-shi, line 10: we haven't don this before, but it looks nice.

tests/vanilla/utils/atomWithReset.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('should create an atom with initial value', () => {
+    const { init } = testAtom
+    expect(init).toBe(initialValue)
+  })
+
+  it('should reset to initial value using RESET', () => {
+    const set = vi.fn()
+    const get = vi.fn(() => 20)
+    testAtom.write(get, set, RESET)
```
dai-shi, line 30: `write` is an implementation detail, and we would like to test with store api.
can you try this?
```js
const store = createStore();
store.set(testAtom, RESET);
```

dai-shi (changes requested):
Thanks for working on adding tests. That would a great contribution.

Please see comments.

tests/vanilla/utils/atomWithReset.test.ts
vangie: I have remove it, it is useless in this case.

tests/vanilla/utils/atomWithReset.test.ts
vangie: @dai-shi I have already added the test code for using `createStore` and kept the previous test code, as it can isolate the dependency on `createStore`.

tests/vanilla/utils/atomWithReset.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  beforeEach(() => {
+    vi.clearAllMocks()
+    initialValue = 10
+    testAtom = atomWithReset(initialValue)
+  })
+
+  it('should create an atom with initial value', () => {
+    const { init } = testAtom
+    expect(init).toBe(initialValue)
+  })
```
dai-shi, line 18: This is testing the internal behavior. While it's correct, let's remove from the atomWithReset test.

tests/vanilla/utils/atomWithReset.test.ts
dai-shi: It's testing the internal behavior, which isn't guaranteed to work as it is now (it's likely though), so let's remove it.

dai-shi (commented):
we don't test the internal behavior.

tests/vanilla/utils/atomWithReset.test.ts
vangie: OK

tests/vanilla/utils/atomWithReset.test.ts
vangie: OK

dai-shi (approved):
LGTM
Thanks for your contribution! Feel free to send more PRs to improve test coverage.

PR #2755, opened 2024-09-29
feat: use React 19 `use` in `useStore` api

himself65:

Fixes #

## Check List

### DIFF
--- a/src/react/Provider.ts
+++ b/src/react/Provider.ts
@@ -1,5 +1,12 @@
-import { createContext, createElement, useContext, useRef } from 'react'
-import type { FunctionComponentElement, ReactNode } from 'react'
+import ReactExports, {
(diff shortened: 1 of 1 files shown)


src/react/Provider.ts
```diff
@@ -13,8 +20,13 @@ type Options = {
   store?: Store
 }
 
+// In React 19, `use` will allow calling within conditionals and loops.
+//  Refs: https://19.react.dev/reference/react/use#reading-context-with-use
```
dai-shi, line 24: But it doesn't allow calling `useStore` in a condition. This is pretty unfortunate, but can't be helped.

So, I don't think this PR helps. It rather misleads to misusage.
If you want to use `use`, the current recommendation is to use custom context on your end.

PR #2759, opened 2024-10-02
test(dependency): Improve test coverage

ts1994tw:
## Summary
This pull request enhances the test suite for Jotai's dependency management, particularly focusing on complex scenarios. The following improvement has been made:

## New Test Cases
1. **Complex Dependency Chains**
   - Introduced a test case that verifies Jotai's ability to handle and update complex chains of derived atoms, including async derivations.

## Coverage Improvements
- **Multi-level Derivations**: Verifies that updates properly propagate through multiple levels of derived atoms, including async operations.

## Check List
- [x] pnpm run prettier for formatting code and docs

Appreciated for having me your time on reviewing changes.

### DIFF
--- a/tests/vanilla/dependency.test.tsx
+++ b/tests/vanilla/dependency.test.tsx
@@ -285,3 +285,20 @@ it('handles errors in async atoms', async () => {
   await expect(store.get(errorAtom)).rejects.toThrow('Test error')
 })
 
+it('handles complex dependency chains', async () => {
+  const baseAtom = atom(1)
+  const derived1 = atom((get) => get(baseAtom) * 2)
+  const derived2 = atom((get) => get(derived1) + 1)
+  const asyncDerived = atom(async (get) => {
+    const value = get(derived2)
+    await new Promise((resolve) => setTimeout(resolve, 10))
+    return value * 2
+  })
+
+  const store = createStore()
+  expect(await store.get(asyncDerived)).toBe(6)
+
+  store.set(baseAtom, 2)
+  expect(await store.get(asyncDerived)).toBe(10)
+})
+
--- a/tests/vanilla/dependency.test.tsx
+++ b/tests/vanilla/dependency.test.tsx
@@ -274,3 +274,14 @@ it('refreshes deps for each async read', async () => {
   resolve.splice(0).forEach((fn) => fn())
   expect(values).toEqual([0, 1])
 })
+
+it('handles errors in async atoms', async () => {
+  const errorAtom = atom(async () => {
+    throw new Error('Test error')
+  })
+
+  const store = createStore()
+
+  await expect(store.get(errorAtom)).rejects.toThrow('Test error')
+})
+
--- a/tests/vanilla/dependency.test.tsx
+++ b/tests/vanilla/dependency.test.tsx
@@ -302,3 +302,15 @@ it('handles complex dependency chains', async () => {
   expect(await store.get(asyncDerived)).toBe(10)
 })
 
+it('handles large number of atoms efficiently', () => {
+  const store = createStore()
+  const atoms = new Array(1000).fill(null).map((_, i) => atom(i))
+  const derivedAtom = atom((get) => atoms.reduce((sum, a) => sum + get(a), 0))
+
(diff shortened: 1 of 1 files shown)


tests/vanilla/dependency.test.tsx
```diff
@@ -274,3 +274,43 @@ it('refreshes deps for each async read', async () => {
...
+
+it('handles errors in async atoms', async () => {
+  const errorAtom = atom(async () => {
+    throw new Error('Test error')
+  })
+
+  const store = createStore()
+
+  await expect(store.get(errorAtom)).rejects.toThrow('Test error')
+})
```
dai-shi, line 286: This doesn't seem a dependency test. Maybe move to some other file?

tests/vanilla/dependency.test.tsx
```diff
@@ -274,3 +274,43 @@ it('refreshes deps for each async read', async () => {
...
+
+it('handles complex dependency chains', async () => {
+  const baseAtom = atom(1)
+  const derived1 = atom((get) => get(baseAtom) * 2)
+  const derived2 = atom((get) => get(derived1) + 1)
+  const asyncDerived = atom(async (get) => {
+    const value = get(derived2)
+    await new Promise((resolve) => setTimeout(resolve, 10))
```
dai-shi, line 294: We don't use timers in the test.
Please do something like this:
https://github.com/pmndrs/jotai/blob/9bd8ed7b0e523af5366f6b762104241a0bb71203/tests/vanilla/dependency.test.tsx#L11

tests/vanilla/dependency.test.tsx
```diff
@@ -274,3 +274,43 @@ it('refreshes deps for each async read', async () => {
...
+  store.set(baseAtom, 2)
+  expect(await store.get(asyncDerived)).toBe(10)
+})
+
+it('handles large number of atoms efficiently', () => {
+  const store = createStore()
+  const atoms = new Array(1000).fill(null).map((_, i) => atom(i))
+  const derivedAtom = atom((get) => atoms.reduce((sum, a) => sum + get(a), 0))
+
+  const start = performance.now()
```
dai-shi, line 310: Hmm, I don't think we do performance tests in our unit tests.
Can you try something with `./benchmarks`? It's not a test that runs on CI, but you can try more complex cases and share the result in a discussion.

dai-shi (changes requested):
Thanks for working on this.
Here are some comments.

tests/vanilla/dependency.test.tsx
backbone87: I think the existing benchmark suite also needs some refinement. I did some resource tracking a while back and memory consumption was not lining up with what I thought the tests are testing for

dai-shi: > I think the existing benchmark suite also needs some refinement.

I actually agree. Not just refinement, you can create a brand new benchmark suite. We can even delete existing one, if yours covers it. That would be pretty helpful.

tests/vanilla/dependency.test.tsx
ts1994tw: @dai-shi Appreciated for indication and this below is adjustment which follows comment here.
[6186a57](https://github.com/pmndrs/jotai/pull/2759/commits/6186a5731564688e3cd2c2e317e67b5902d9ac0c)

tests/vanilla/dependency.test.tsx
ts1994tw: @backbone87 Appreciated for sharing detections you noticed before. @dai-shi I prefer to leave this topic aside and it should independently handle by another clear and goal-focused pull request. These changes have been removed at this moment. 🙇

tests/vanilla/dependency.test.tsx
ts1994tw: @dai-shi Appreciated for comments. You're absolutely right, it should be moved out of here. These changes have been removed from this pull request at this moment, after that, this pull request will be goal-focused enough and for dependency test only.

dai-shi (approved):
Thanks for your contribution.

PR #2771, opened 2024-10-15
fix(provider): make react compiler happy

dai-shi:
If we were to follow eslint-react-compiler, should this be a solution?


grigorischristainas:
@dai-shi could this be a solution too ? 

```tsx
export const Provider = ({
  children,
  store,
}: {
  children?: ReactNode
  store?: Store
}): FunctionComponentElement<{ value: Store | undefined }> => {
  const [value] = useState(() => store || createStore())
  return createElement(StoreContext.Provider, { value }, children)
}
```

The `value` is stable across re-renders and is only created once per component lifecycle.

dai-shi:
Unfortunately not, because `store` can be changed after the initial render.

sienki-jenki:
```
export const Provider = ({
  children,
  store,
}: {
  children?: ReactNode
  store?: Store
}): FunctionComponentElement<{ value: Store | undefined }> => {
  const value = useMemo(() => ({ value : store ?? createStore() }) ,[store]);
  return createElement(StoreContext.Provider, value, children)
}
```

@dai-shi what about this one? value is stable, store can change, if not supplied it would fallback to createStore()

dai-shi:
It works for 99.9% use cases, but React docs told us that we shouldn't use `useMemo` as semantic guarantee.

dai-shi:
https://github.com/pmndrs/jotai/pull/3061/files#r2075165611

PR #2773, opened 2024-10-16
docs: migrate Codesandbox embed for next.js swc example to Stackblitz

niklasbec:
## Summary

- Moves the code examples from Codesandbox to Stackblitz .
https://stackblitz.com/orgs/custom/Jotai/projects (Team access needed to view the whole collection).
- Added disclaimer to Persistence 1 Example because I wasn't sure add first what I had to do and other users might be confused as well.
- Same disclaimer for the second example.
- Jotai example atomWithBroadcast had a bug in the broadcast mount function which broke the example (fixed it).

## Issues in code examples

- For trpc the example is broken because the pokemon api is offline, do we know the dev who built it do you have a replacement api in mind I can use so we don't run into the same issue again?
- The swc example doesn't work because Stackblitz struggles with swc's, if you clone the Stackblitz it works @dai-shi suggested adding a disclaimer.
- Didn't manage to get Relay working => https://stackblitz.com/edit/vitejs-vite-divyhe
- I didn't move the Babel and Parcel example to StackBlitz, I have no clue about either and how they intergrate with StackBlitz.

## Check List

### DIFF
--- a/docs/core/atom.mdx
+++ b/docs/core/atom.mdx
@@ -207,9 +207,9 @@ It's a special function to invoke the write function of the self atom.
 
 ⚠️ It's provided primarily for internal usage and third-party library authors. Read the source code carefully to understand the behavior. Check release notes for any breaking/non-breaking changes.
 
-## codesandbox
+## Stackblitz
 
-<CodeSandbox id="hg665r" />
+<Stackblitz id="vitejs-vite-ccqxix" file="src%2FApp.tsx" />
 
 ```tsx
 import { Suspense } from 'react'
--- a/docs/extensions/cache.mdx
+++ b/docs/extensions/cache.mdx
@@ -115,6 +115,6 @@ const App = () => {
 }
 ```
 
-### Codesandbox
+### Stackblitz
 
-<CodeSandbox id="h09znd" />
+<Stackblitz id="vitejs-vite-p86ajq" file="src%2FApp.tsx" />
--- a/docs/extensions/effect.mdx
+++ b/docs/extensions/effect.mdx
@@ -77,7 +77,7 @@ function MyComponent() {
 }
 ```
 
-<CodeSandbox id="tg9xsf" />
+<Stackblitz id="vitejs-vite-fj5zjp" file="src%2FApp.tsx" />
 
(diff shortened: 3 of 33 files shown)


dai-shi:
Hi, thanks for working on this.
Are you able to run codesandbox one? I just tried and I can't.

https://stackblitz.com/edit/stackblitz-starters-ha5txx?file=app%2Fpage.tsx

Running it, it shows `undefined: 1` in the console log. Can you confirm?
image

I think it should show `countAtom: 1`. Maybe we need to update packages?

niklasbec:
> Hi, thanks for working on this. Are you able to run codesandbox one? I just tried and I can't.
> 
> https://stackblitz.com/edit/stackblitz-starters-ha5txx?file=app%2Fpage.tsx
> 
> Running it, it shows `undefined: 1` in the console log. Can you confirm? image
> 
> I think it should show `countAtom: 1`. Maybe we need to update packages?

I will see if I can fix this.

niklasbec:
@dai-shi there used to be a problems with swcs on earlier next versions, but pulling that up doesn't fix it. I cloned the stackblitz and ran it locally without issues so I am relatively confident this is an issue with stackblitz handling swcs.

I will see if I can do anything about it later, and will replace the code examples running on codesandbox that don't use swcs.

dai-shi:
> I cloned the stackblitz and ran it locally without issues

That's at least good to know.

If you don't find a solution, we can probably merge this if you add a note somewhere saying it has to be downloaded to use.

niklasbec:
> > I cloned the stackblitz and ran it locally without issues
> 
> That's at least good to know.
> 
> If you don't find a solution, we can probably merge this if you add a note somewhere saying it has to be downloaded to use.

Please see my updated comment :)

dai-shi (approved):
wow, this great. let's merge it and see how it goes.

dai-shi:
Thanks for your contribution! Feel free to improve docs more.

niklasbec:
Right now it is not showing the preview by default but only the code and you have to do a click to see the build preview. I will make a PR for that tomorrow.

PR #2778, opened 2024-10-19
Throw error on `useAtom(undefined)` or `useAtom(null)`

kevinschaich:

Fixes #2777 

## Summary

`useAtom(undefined)` throws `Invalid value used as weak map key`

## Check List


src/react/useAtomValue.ts
```diff
@@ -113,6 +113,10 @@ export function useAtomValue<AtomType extends Atom<unknown>>(
 export function useAtomValue<Value>(atom: Atom<Value>, options?: Options) {
   const store = useStore(options)
 
+  if (atom === undefined || atom === null) {
```
dai-shi, line 116: Let's make this DEV-only:

```suggestion
  if (import.meta.env?.MODE !== 'production' && (atom === undefined || atom === null)) {
```

dai-shi: Maybe this is simpler:

```suggestion
  if (import.meta.env?.MODE !== 'production' && !atom) {
```

dai-shi: In case we want stricter check, this is possible:

```suggestion
  if (import.meta.env?.MODE !== 'production' && typeof atom?.read !== 'function') {
```

dai-shi (changes requested):
Thanks for opening up a PR.

Here are some suggestions.

kevinschaich:
Nice, I like the second one @dai-shi. Simple and won't break apps in production.

dai-shi (commented):
Thanks for the change.

I reviewed it again, and now wonder if this check should be in `store.ts`, because the issue isn't specific to hooks.

kevinschaich:
@dai-shi would that error get thrown in the place we expect are hitting the bug now if we move it?

If that's the preferred path forward I'd need more guidance on where specifically you'd like the check implemented

dai-shi:
I added two commits illustrating two ideas. Do they work as expected?

dai-shi:
@kevinschaich whenever you have time.

kevinschaich:
Hey @dai-shi, changes working as expected for me. I think this is OK to move forward with.

dai-shi (approved):
Thanks for confirming. Will merge it later.

PR #2785, opened 2024-10-26
test: improve code coverage for atomWithDefault

kretajak:
## Summary

Added unit test with missing scenario of setting `atomWithDefault` with primitive value. Provides 100% of code coverage for this utility.

## Check List

### DIFF
--- a/tests/react/vanilla-utils/atomWithDefault.test.tsx
+++ b/tests/react/vanilla-utils/atomWithDefault.test.tsx
@@ -1,6 +1,6 @@
 import { StrictMode, Suspense } from 'react'
 import { render, screen, waitFor } from '@testing-library/react'
-import { userEvent } from '@testing-library/user-event'
+import userEvent from '@testing-library/user-event'
 import { expect, it } from 'vitest'
 import { useAtom } from 'jotai/react'
 import { atom } from 'jotai/vanilla'
@@ -209,8 +209,6 @@ it('refresh async atoms to default values', async () => {
 it('can be set synchronously by passing value', async () => {
   const countAtom = atomWithDefault(() => 1)
 
-  const user = userEvent.setup()
-
   const Counter = () => {
     const [count, setCount] = useAtom(countAtom)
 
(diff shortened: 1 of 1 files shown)


tests/react/vanilla-utils/atomWithDefault.test.tsx
```diff
@@ -205,3 +205,28 @@ it('refresh async atoms to default values', async () => {
...
+        <button onClick={() => setCount(10)}>Set to 10</button>
+      </>
+    )
+  }
+
+  const { getByText, getByRole } = render(<Counter />)
+
+  expect(getByText('count: 1')).toBeDefined()
+
+  await user.click(getByRole('button', { name: 'Set to 10' }))
```
dai-shi, line 229: How is it different from `await userEvent.click(...)`?

kretajak: Using `userEvent.setup` is recommended by the [userEvent docs](https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent).

dai-shi: I see. As it doesn't support old TypeScript versions that we still support, we need some workarounds. How about using `await userEvent.clieck()` in this PR, and then tackling for `userEvent.setup` in a separate PR because we also use it elsewhere?

kretajak: Using `import { userEvent }` made things even worse. I will, as you said, use `userEvent.click` to make things work with older Typescript versions.

tests/react/vanilla-utils/atomWithDefault.test.tsx
```diff
@@ -205,3 +205,26 @@ it('refresh async atoms to default values', async () => {
...
+    )
+  }
+
+  render(<Counter />)
+
+  expect(screen.getByText('count: 1')).toBeDefined()
+
+  await userEvent.click(screen.getByRole('button', { name: 'Set to 10' }))
+
+  expect(screen.getByText('count: 10')).toBeDefined()
```
dai-shi, line 229: Feel free to send PRs to improve test code using `screen.*`. Thx.

dai-shi (approved):
Thanks for your contribution!

PR #2790, opened 2024-10-31
fix(utils): make 'loadable' update immediate after resolve

e7h4n:
## Summary
This commit aims to update the Loadable state derived from `loadable` immediately after the Promise resolves.

The original code used the following approach to update the loadable cache, which resulted in a delay of one tick for the cache update:

```typescript
export function loadable<Value>(anAtom: Atom<Value>): Atom<Loadable<Value>> {
  return memo1(() => {
    // ...
    const derivedAtom = atom(
      (get, { setSelf }) => {
        get(refreshAtom);
        // ...
        promise
          .then(
              // ...
          )
          .finally(setSelf); // <---------- Note that here, .finally is applied to the Promise derived from .then(), not the original promise itself.
        // ...
      },
      (_get, set) => {
        set(refreshAtom, (c) => c + 1);
      },
    );
    // ...
    return atom((get) => get(derivedAtom));
  }, anAtom);
}
```

The use of `.finally` as shown above causes the following case to fail:

```typescript
const delayPromise = new Promise((r) => setTimeout(r, 10));
const store = createStore();
const asyncAtom = atom(delayPromise);
const loadableAtom = loadable(asyncAtom);

await store.get(asyncAtom);

expect(store.get(loadableAtom).state).toBe('hasData'); // <---- Error: it's still loading, although asyncAtom has been resolved.
```

To resolve this issue, `setSelf` needs to be executed within the `.then()` callback.

## Check List

- [x] Ran `pnpm run prettier` for formatting code and documentation.


src/vanilla/utils/loadable.ts
```diff
@@ -44,16 +44,17 @@ export function loadable<Value>(anAtom: Atom<Value>): Atom<Loadable<Value>> {
...
-            },
-            (error) => {
-              loadableCache.set(promise, { state: 'hasError', error })
-            },
-          )
-          .finally(setSelf)
+        promise.then(
+          (data) => {
+            loadableCache.set(promise, { state: 'hasData', data })
+            setSelf()
```
dai-shi, line 50: Yeah, this is good. We fixed store.ts as well in the past.

tests/vanilla/utils/loadable.test.ts
```diff
@@ -17,4 +17,22 @@ describe('loadable', () => {
       data: 'concrete',
     })
   })
+
+  it('should get the latest loadable state after the promise resolves', async () => {
+    const delayPromise = new Promise((r) => setTimeout(r, 10))
```
dai-shi, line 22: We don't rely on timers. Use `resolve` hack like other tests.

tests/vanilla/utils/loadable.test.ts
```diff
@@ -17,4 +17,22 @@ describe('loadable', () => {
...
+
+    expect(store.get(loadableAtom)).toEqual({
+      state: 'loading',
+    })
+
+    await store.get(asyncAtom)
+
+    expect(store.get(loadableAtom)).toEqual({
+      state: 'hasData',
+      data: undefined,
```
dai-shi, line 35: Maybe, it's more readable to have some value instead of `undefined`.

dai-shi (commented):
This is nice catch!

dai-shi (approved):
LGTM. I'll merge this when we prepare the next version.

PR #2794, opened 2024-11-01
fix(utils): make 'unwrap' update immediate after resolve

organize:

Identical to https://github.com/pmndrs/jotai/pull/2790, but for `unwrap`
This regression is from the change here: https://github.com/pmndrs/jotai/pull/2695

## Summary
 
When upgrading from 2.8.4 to 2.10.1, `unwrap` hangs in certain cases. I narrowed it down to atoms such as this one:

`atomWithDefault((get) => get(unwrap(anAsyncAtom)))`

Applying https://github.com/pmndrs/jotai/pull/2790 to `unwrap` fixes things.

Note that the recommended pattern for such atoms is to use two atoms:

```
const atomA = unwrap(anAsyncAtom);
const atomB = atomWithDefault((get) => get(atomA));
```

In this case, inlining `atomA` into `atomWithDefault` only works if `unwrap`'s default fallback is used, meaning the following code will hang:

`const atomB = atomWithDefault((get) => get(unwrap(anAsyncAtom, (prev) => prev ?? 'MAGIC_DEFAULT')));`

## Check List

### DIFF
--- a/src/vanilla/utils/unwrap.ts
+++ b/src/vanilla/utils/unwrap.ts
@@ -60,12 +60,16 @@ export function unwrap<Value, Args extends unknown[], Result, PendingValue>(
             return { v: promise as Awaited<Value> }
           }
           if (promise !== prev?.p) {
-            promise
-              .then(
-                (v) => promiseResultCache.set(promise, v as Awaited<Value>),
-                (e) => promiseErrorCache.set(promise, e),
-              )
-              .finally(setSelf)
+            promise.then(
+              (v) => {
+                promiseResultCache.set(promise, v as Awaited<Value>)
+                setSelf()
+              },
+              (e) => {
+                promiseErrorCache.set(promise, e)
+                setSelf()
+              },
+            )
           }
           if (promiseErrorCache.has(promise)) {
             throw promiseErrorCache.get(promise)
--- a/tests/vanilla/utils/unwrap.test.ts
+++ b/tests/vanilla/utils/unwrap.test.ts
@@ -137,4 +137,16 @@ describe('unwrap', () => {
     await new Promise((r) => setTimeout(r)) // wait for a tick
     expect(store.get(unwrap(asyncAtom))).toEqual('concrete')
   })
+
+  it('should get a fulfilled value after the promise resolves', async () => {
+    const store = createStore()
+    const asyncAtom = atom(Promise.resolve('concrete'))
+    const syncAtom = unwrap(asyncAtom)
+
+    expect(store.get(syncAtom)).toEqual(undefined)
+
+    await store.get(asyncAtom)
+
+    expect(store.get(syncAtom)).toEqual('concrete')
+  })
 })


src/vanilla/utils/unwrap.ts
```diff
@@ -35,6 +38,11 @@ export function unwrap<Value, Args extends unknown[], Result, PendingValue>(
   anAtom: WritableAtom<Value, Args, Result> | Atom<Value>,
   fallback: (prev?: Awaited<Value>) => PendingValue = defaultFallback as never,
 ) {
+  const stableFallback =
+    fallback === defaultFallback
+      ? fallback
+      : getStableFallback(fallback, anAtom)
```
dai-shi, line 44: This doesn't work for certain cases if we want dynamic callbacks.
Let's educate users to define callbacks outside component or use `useCallback`.

src/vanilla/utils/unwrap.ts
```diff
@@ -60,12 +68,16 @@ export function unwrap<Value, Args extends unknown[], Result, PendingValue>(
...
-            promise
-              .then(
-                (v) => promiseResultCache.set(promise, v as Awaited<Value>),
-                (e) => promiseErrorCache.set(promise, e),
-              )
-              .finally(setSelf)
+            promise.then(
+              (v) => {
+                promiseResultCache.set(promise, v as Awaited<Value>)
+                setSelf()
```
dai-shi, line 74: This is good.

dai-shi (changes requested):
Nice catch. Unfortunately the second one is not technically solvable as I understand.

src/vanilla/utils/unwrap.ts
organize: Alright, I can drop the commit. But could you elaborate just so I understand, what would be the use case? And is it something you would prefer to support over the one in my example above?

Thinking further, this definitely needs to be documented as a breaking change - a Jotai eslint plugin could be useful to disallow inline callbacks here if that's what you mean 🤔

dai-shi: Not a specific use case in mind, but the function can be anything and `prev => prev ?? DEFAULT` is only one use case.

It's not a change as I see it. Not only `unwrap`, but also `useAtomCallback` always expect stable selectors. It's kind of surprising inline callbacks worked in previous versions as it's not intended.

That said, because `atomWithDefault((get) => get(unwrap(anAsyncAtom, (prev) => prev ?? DEFAULT)))` worked previously by chance, it's better to document it for explicitly. Would you like to open another PR for doc improvement?

My recommendation is to separate it into two atoms:
```ts
const atomA = unwrap(anAsyncAtom, (prev) => prev ?? getSomeDefaultValueThatMayChange());
const atomB = atomWithDefault((get) => get(atomA));
```

organize: That's fair, thank you. I can make another PR to update the documentation. I'll drop the commit now.

dai-shi (commented):
The code diff looks good to me.
Please check CI errors and update the PR description.

dai-shi (approved):
Will merge it next week.

PR #2799, opened 2024-11-06
test: incorporate eslint-plugin-testing-library and fix reported issues

kretajak:
## Summary

Testing Library provides [Eslint plugin ](https://testing-library.com/docs/ecosystem-eslint-plugin-testing-library) which encourages to use best practices.

## Check List

- [ x] `pnpm run prettier` for formatting code and docs

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,8 @@
     "react-hooks",
     "import",
     "vitest",
-    "eslint-plugin-react-compiler"
+    "eslint-plugin-react-compiler",
+    "testing-library"
   ],
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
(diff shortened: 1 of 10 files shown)


dai-shi:
Thanks for working on it. Definitely seems like an improvement.
As always, we need to keep things sync between zustand, jotai and valtio.
Please file PRs to two other repos.

kretajak:
> Thanks for working on it. Definitely seems like an improvement. As always, we need to keep things sync between zustand, jotai and valtio. Please file PRs to two other repos.

I have added similar PRs to `zustand` and `valtio`.

dai-shi:
Thanks!!!

https://github.com/pmndrs/zustand/pull/2844
https://github.com/pmndrs/valtio/pull/992

PR #2800, opened 2024-11-08
proposal: synchronous effects

dmaskasky:
## Summary
- Adds `atomSyncEffect` function and `SyncEffectAtom` type to the api.
- extends atom definition to support `effect` property.
- updates store.ts to support synchronous effects
- adds initial test for synchronous effect.

## Check List

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -1,5 +1,10 @@
-export { atom } from './vanilla/atom.ts'
-export type { Atom, WritableAtom, PrimitiveAtom } from './vanilla/atom.ts'
+export { atom, atomSyncEffect } from './vanilla/atom.ts'
+export type {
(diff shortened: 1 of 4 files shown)


dai-shi:
Unfortunately, it's not my preference of the api. I will consider the capability. So, let's keep this open for a while.

dmaskasky:
I added this PR for reference only. atomEffects should be handled by jotai-effect as much as possible. Using this PR as a starting point, I hope we can find an acceptable implementation that satisfies all requirements.

dai-shi:
> I hope we can find an acceptable implementation that satisfies all requirements.

Adding a new API isn't acceptable. I haven't even started thinking about the api. Let's start with a discussion if necessary.

PR #2801, opened 2024-11-08
`atom.INTERNAL_onInit` hook to support sync effects

dmaskasky:
## Summary
1. Adds a new atom api, `atom.INTERNAL_onInit = (store: Store, atomState: AtomState) => void`.
Fires when the atomState is first created in the store for that atom.

2. Updates `store.unstable_derive` api to add a new callback param, `atomOnInit`. This is the interceptor for the `atom.unstable_onInit` api. Since getAtomState is where `atom.unstable_onInit` is called, we must add `atomOnInit` to the getAtomState params. `atomOnInit` accepts a store arg so that the current store is always passed.

## Supplemental
- adds initial test for synchronous effect

## Check List


dai-shi:
nothing is finalized yet. you are going too far.

src/vanilla/store.ts
```diff
@@ -513,11 +544,14 @@ const buildStore = (getAtomState: StoreArgs[0]): Store => {
       } else {
         r = writeAtomState(pending, a, aState, ...args) as R
       }
-      flushPending(pending)
+      flushPending(pending, !isSync)
```
dai-shi, line 547: This makes me feel like there's a bug in an edge case without this PR or we can improve something without this PR.

src/vanilla/store.ts
```diff
@@ -165,16 +165,23 @@ type Pending = readonly [
   dependents: Map<AnyAtom, Set<AnyAtom>>,
   atomStates: Map<AnyAtom, AtomState>,
   functions: Set<() => void>,
+  finalizers: Set<AnyAtom>,
```
dai-shi, line 168: I am hoping to use `functions` Set.

dmaskasky: Do we want finalization to happen after onMount is called?

dai-shi: > Do we want finalization to happen after onMount is called?

Good question. That's what I thought I should ask. onMount is basically like useEffect, which can be async. So, syncEffect should be done in advance to onMount.

(So, I even wonder if "after flushPending" is the right approach.)

dmaskasky: Good point. React doesn't differentiate between useEffect and componentDidMount. Even the order of effects will affect output.

dai-shi: Maybe we can just design store api apart from React lifecycle.
I'm fine to make `functions` an ordered list.

src/vanilla/store.ts
dmaskasky: I actually wonder if we can delete the synchronous call to flush pending.

dai-shi: can you elaborate?

dai-shi: Oh, do you mean like `if (!isSync) flushPending(pending)`? That's exactly my implication with the first comment.

dmaskasky: I mean if we remove synchronous flushPending from writeAtomState, then we should a flushPending after this line:
https://github.com/pmndrs/jotai/pull/2801/files#diff-849e7b1b832999901f65b02d61703f4e35f157d3f24387c13bf1e00e5a701dd3R617

src/vanilla/store.ts
```diff
@@ -514,10 +543,16 @@ const buildStore = (getAtomState: StoreArgs[0]): Store => {
         r = writeAtomState(pending, a, aState, ...args) as R
       }
       flushPending(pending)
+      if (!isSync) {
+        processOnChangeHandlers(pending)
+      }
```
dai-shi, line 548: This looks bad in the long term maintenance. It should be always paired.
Let's see how we can fix it before working on this PR.

src/vanilla/store.ts
dai-shi: #2804 Let's focus on it separately. If you can help it, I appreciate it. Otherwise, I'll tackle when I have time.

dai-shi: https://github.com/pmndrs/jotai/pull/2804 Let's focus on it separately. If you can help it, I appreciate it. Otherwise, I'll tackle when I have time.

dai-shi (commented):
preliminary task.

src/vanilla/store.ts
```diff
@@ -200,8 +194,11 @@ const registerBatchAtom = (
...
   if (!batch.D.has(atom)) {
     batch.D.set(atom, new Set())
-    addBatchFuncMedium(batch, () => {
-      atomState.m?.l.forEach((listener) => addBatchFuncMedium(batch, listener))
+    addBatchFunc(batch, 'H', () => {
+      atomState.m?.l.forEach(
+        (listener: (() => void) & { INTERNAL_priority?: BatchPriority }) =>
+          addBatchFunc(batch, listener.INTERNAL_priority || 'M', listener),
+      )
     })
```
dmaskasky, line 202: 1. Raise the priority of the listener insertion function so that it can insert to the high priority channel. Synchronous effects should run after recomputeDependents and before all listeners and mount hooks.
2. Determine the listener priority by reading an internal property tag on the listener.

src/vanilla/store.ts
```diff
@@ -198,8 +194,14 @@ const registerBatchAtom = (
 ) => {
   if (!batch.D.has(atom)) {
     batch.D.set(atom, new Set())
-    addBatchFunc(batch, 'M', () => {
-      atomState.m?.l.forEach((listener) => addBatchFunc(batch, 'M', listener))
+    addBatchFunc(batch, 'H', () => {
+      for (const listener of atomState.m?.l || []) {
+        let priority: BatchPriority = 'M'
+        if ('INTERNAL_priority' in listener) {
+          priority = listener.INTERNAL_priority as BatchPriority
```
dmaskasky, line 201: I couldn't think of a way to register a high priority listener without adding a new api, so I went with an `INTERNAL_priority` property on the listener function itself. I don't think we should include this in the typescript definition though, because I don't really think there's a use-case outside of sync effects.

dmaskasky: Update: I refactored this in a later commit. Now INTERNAL_onInit returns the store and the atomState. There is a new property on the atomState for registering additional event listeners.

PR #2804, opened 2024-11-10
fix: flushPending in async write

dai-shi:
We used to have `isSync` in writeAtomState to conditionally flush pending, but when I did #2463, I removed it. I thought it worked without it, but the behavior wasn't something expected. This PR introduces `isSync` again in writeAtomSync, and also in mountAtom and unmountAtom.


src/vanilla/store.ts
```diff
@@ -488,36 +488,44 @@ const buildStore = (getAtomState: StoreArgs[0]): Store => {
...
-        const prevValue = aState.v
-        const v = args[0] as V
-        setAtomStateValueOrPromise(a, aState, v)
-        mountDependencies(pending, a, aState)
-        if (!hasPrevValue || !Object.is(prevValue, aState.v)) {
-          addPendingAtom(pending, a, aState)
-          recomputeDependents(pending, a, aState)
+        if (!isSync) {
+          flushPending(pending)
         }
```
dmaskasky, line 520: Edge case: do we care if this changes the behavior a little bit for subscriber listeners?

```ts
const a = atom(0)
  console.log('before write')
  set(a, v=> v + 1)
  console.log('after write')
})
w.onMount = (setSelf) => {
  setSelf()
  console.log('after setSelf')
}
store.sub(a, () => {
  console.log('a listener fired')
})
store.sub(w, () => {})
/*
  CURRENT:
    1. 'before write'
    2. 'a listener fired'
    3. 'after write'
    4. 'after setSelf'

  NEW:
    1. 'before write'
    2. 'after write'
    3. 'after setSelf'
    4. 'a listener fired'
*/ 
```

dai-shi: Is it why the current test is failing?

Do you know how to make a failing test if we don't have `isSync`?

dmaskasky: > Is it why the current test is failing?

Yes

> Do you know how to make a failing test if we don't have `isSync`?

Thinking on this more...
I think we need to keep this synchronous call to `flushPending` in `set`. Some third-party libraries depend on this behavior, so we need some way to call all subscribers.

dmaskasky: Yes, I agree. This is better.

Hmm, it looks like atomWithObservable and atomWithStorage are not failing after all.

dai-shi (commented):
most work is done by @dmaskasky. thanks!

PR #2814, opened 2024-11-13
fix/flush pending async onmount finally

dmaskasky:
## Summary
call onMount flushPending in a finally block

## Check List


src/vanilla/store.ts
```diff
@@ -584,11 +584,13 @@ const buildStore = (
           let isSync = true
           try {
             const onUnmount = atomOnMount(atom, (...args) => {
-              const result = writeAtomState(pending, atom, ...args)
-              if (!isSync) {
-                flushPending(pending)
+              try {
+                return writeAtomState(pending, atom, ...args)
+              } finally {
```
dai-shi, line 589: What would be the technical difference? Does it mean, if `writeAtomState` throws, we still want to flush pending?

dmaskasky: Yes because some atoms could have been updated before the error was thrown.

dai-shi: Cool. That makes sense.

> some atoms could have been updated before the error was thrown

Is it too hard to add such a test?

PR #2815, opened 2024-11-14
fix: onMount setSelf does not notify listeners

dmaskasky:
- https://github.com/pmndrs/jotai/discussions/2810

## onMount setSelf does not notify listeners (https://github.com/pmndrs/jotai/discussions/2810)
- adds failing test: when changing the atom value with onMount setSelf, subscribers are not called
- fixes failing test

### Why aren't listeners called when calling setAtom in onMount?
Because the listener is added only after flushPending is called.

Fix: swap the order of flush pending and listeners.add(listener).

### Why aren't listeners called when calling setAtom in onUnmount?
Because the listener is deleted before flushPending is called.

There's a logic cycle here:
1. the listener is deleted _before_ calling unmountAtom, because unmountAtom only processes if atomState.m.l.size === 0
2. unmountAtom adds the onUnmount pending function
3. flushPending calls onUnmount pending function
4. setAtom fires in onUnmount pending function, which updates the atom
5. subscriber doesn't fire because it was removed in 1

👉 NOTE: This is expected behavior.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -666,9 +666,9 @@ const buildStore = (
     const pending = createPending()
     const atomState = getAtomState(atom)
     const mounted = mountAtom(pending, atom, atomState)
-    flushPending(pending)
     const listeners = mounted.l
     listeners.add(listener)
+    flushPending(pending)
     return () => {
       listeners.delete(listener)
       const pending = createPending()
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Very good. This is something I missed it.
The test is reasonable and the fix is surprisingly simple.

PR #2816, opened 2024-11-14
Fix/flush pending finally everywhere

dmaskasky:

Fixes #

## Check List


src/vanilla/store.ts
```diff
@@ -331,17 +331,21 @@ const buildStore = (
...
-        addDependency(pending, atom, atomState, a, aState)
-      } else {
-        const pending = createPending()
-        addDependency(pending, atom, atomState, a, aState)
-        mountDependencies(pending, atom, atomState)
-        flushPending(pending)
+      let aState: AtomState<V>
+      try {
+        // a !== atom
+        aState = readAtomState(pending, a, dirtyAtoms)
```
dai-shi, line 337: readAtomState shouldn't throw. so, this is okay?
```suggestion
      // a !== atom
      const aState = readAtomState(pending, a, dirtyAtoms)
      try {
```

dai-shi: if readAtomState throws, aState can be undefined, which is unexpected, so I think the change ☝️ should be better.

src/vanilla/store.ts
```diff
@@ -492,29 +496,32 @@ const buildStore = (
...
       ...args: As
     ) => {
-      const aState = getAtomState(a)
-      let r: R | undefined
-      if (isSelfAtom(atom, a)) {
-        if (!hasInitialValue(a)) {
-          // NOTE technically possible but restricted as it may cause bugs
-          throw new Error('atom not writable')
+      let aState: AtomState<V>
+      aState = getAtomState(a)
```
dai-shi, line 500:
```suggestion
      const aState = getAtomState(a)
```

src/vanilla/store.ts
```diff
@@ -641,7 +652,8 @@ const buildStore = (
   const subscribeAtom = (atom: AnyAtom, listener: () => void) => {
     const pending = createPending()
     const atomState = getAtomState(atom)
-    const mounted = mountAtom(pending, atom, atomState)
+    let mounted: Mounted
+    mounted = mountAtom(pending, atom, atomState)
```
dai-shi, line 656:
```suggestion
    const mounted = mountAtom(pending, atom, atomState)
```

dai-shi (approved):
Not a big fan of many try-finally syntax, but it feels safer.

PR #2818, opened 2024-11-14
Fix/flush pending finally everywhere 4

dmaskasky:

Fixes #

## Check List


src/vanilla/store.ts
```diff
@@ -199,14 +199,27 @@ const addPendingFunction = (pending: Pending, fn: () => void) => {
...
 
 const flushPending = (pending: Pending) => {
+  let error: unknown | undefined
+  const call = (fn: () => void) => {
+    try {
+      fn()
+    } catch (e) {
+      if (!error) {
+        error = e
+      }
```
dai-shi, line 209: Only throw the first error, right?

dai-shi (approved):
I understand the edge case to cover, but not super happy with the store code becoming complicated for edge cases. Hope to come up with a nicer solution in the future.

src/vanilla/store.ts
dmaskasky: That is correct.

I wish there was some way to throw the next error after the currently thrown error was caught. There isn't.

PR #2820, opened 2024-11-14
fix: rethrow falsy errors thrown in flushPending

dmaskasky:
## Summary
Errors thrown in flushPending are checked for truthyness. This change allows for falsy errors to be rethrown.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -199,13 +199,15 @@ const addPendingFunction = (pending: Pending, fn: () => void) => {
 }
 
 const flushPending = (pending: Pending) => {
-  let error: unknown | undefined
+  let error: AnyError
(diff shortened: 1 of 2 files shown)


src/vanilla/store.ts
```diff
@@ -199,13 +199,13 @@ const addPendingFunction = (pending: Pending, fn: () => void) => {
 }
 
 const flushPending = (pending: Pending) => {
-  let error: unknown | undefined
+  const ref: { e?: AnyError } = {}
```
dai-shi, line 202: Let's avoid using extra object.
```suggestion
  let error: unknown | undefined
  let hasError = false
```

dai-shi (approved):
LGTM

PR #2821, opened 2024-11-15
refactor(core): Use iterative approach in recompute dependents

dmaskasky:
## Summary
Replaces the recursive topo search in recomputeDependents with an iterative one. An iterative approach theoretically is faster and does not suffer from stack overflow issues. This makes it possible to process very large atom graphs.

## Benchmark
- Recursive: ~200ms
-  Iterative: ~ 350ms
TODO: Currently the new iterative approach is _slower_ than the recursive one.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -439,41 +439,62 @@ const buildStore = (
     return dependents
   }
 
+  // This is a topological sort via depth-first search, slightly modified from
+  // what's described here for simplicity and performance reasons:
+  // https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search
+  function getSortedDependents(
+    pending: Pending,
+    rootAtom: AnyAtom,
+    rootAtomState: AtomState,
+  ): [[AnyAtom, AtomState, number][], Set<AnyAtom>] {
+    const sorted: [atom: AnyAtom, atomState: AtomState, epochNumber: number][] =
+      []
+    const visiting = new Set<AnyAtom>()
+    const visited = new Set<AnyAtom>()
+    // Visit the root atom. This is the only atom in the dependency graph
+    // without incoming edges, which is one reason we can simplify the algorithm
+    const stack: [a: AnyAtom, aState: AtomState][] = [[rootAtom, rootAtomState]]
+    while (stack.length > 0) {
+      const [a, aState] = stack[stack.length - 1]!
+      if (visited.has(a)) {
+        // All dependents have been processed, now process this atom
+        stack.pop()
+        continue
+      }
+      if (visiting.has(a)) {
+        // The algorithm calls for pushing onto the front of the list. For
+        // performance, we will simply push onto the end, and then will iterate in
+        // reverse order later.
+        sorted.push([a, aState, aState.n])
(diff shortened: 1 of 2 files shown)


src/vanilla/store.ts
```diff
@@ -454,24 +454,40 @@ const buildStore = (
       epochNumber: number,
     ])[] = []
     const markedAtoms = new Set<AnyAtom>()
-    const visit = (a: AnyAtom, aState: AtomState) => {
-      if (markedAtoms.has(a)) {
-        return
+    // Visit the root atom. This is the only atom in the dependency graph
+    // without incoming edges, which is one reason we can simplify the algorithm
+    const stack: Array<{ a: AnyAtom; aState: AtomState; visited: boolean }> = [
```
dai-shi, line 459: Using object increases the bundle size. Let's use a tuple instead.

dai-shi (commented):
This looks like a great improvement.

src/vanilla/store.ts
```diff
@@ -439,46 +439,66 @@ const buildStore = (
...
+      }
+      visiting.add(a)
+      // Push unvisited dependents onto the stack
+      for (const [d, s] of getDependents(pending, a, aState)) {
+        if (a !== d && !visiting.has(d)) {
+          stack.push([d, s])
+        }
+      }
+    }
+    return [sorted.reverse(), visited]
```
dai-shi, line 482: reverse() is generally fast, but if we can avoid it, let's avoid it.

Can you revert the for loop?
```js
    for (let i = topsortedAtoms.length - 1; i >= 0; --i) {
      const [a, aState, prevEpochNumber] = topsortedAtoms[i]!
```

tests/vanilla/store.test.tsx
```diff
@@ -933,3 +933,33 @@ it('should call subscribers after setAtom updates atom value on mount but not on
...
+      ++depth
+      try {
+        return d()
+      } catch (error) {
+        return depth
+      }
+    }
+    return d()
+  }
+  const maxDepth = getMaxDepth()
```
dai-shi, line 950: Not sure this is ideal in the test, but it looks like a nice hack in the meantime.

dai-shi (approved):
LGTM

dmaskasky:
Oh, would you be able to provide these benchmarks? I only used vitest test duration using stack limit size nodes in a linked dependency chained configuration.

I'd love to hear your ideas on how to make this faster. The original recursive solution suffers from stack overflow issues, so I'm keen on keeping an iterative approach. I tried both DFS and BFS and found they both have the same number of iterations but the DFS was more memory efficient.

dmaskasky:
Oh, I found a typo in the PR description. The iterative DFS solution in around 250 ms in my testing, not 350 ms.

cltnschlosser:
We should be able to shuffle some things around to remove one of the visiting or visited sets.

dmaskasky:
I'm not sure how to improve on the post-order DFS topological sort solution we have currently. I tried implementing Kahn's, but there's too much administrative work required to beat the current solution.

https://github.com/pmndrs/jotai/pull/2833

PR #2824, opened 2024-11-16
add tests: can read sync derived atom in write without initializing

dmaskasky:
https://github.com/pmndrs/jotai/discussions/2789#discussioncomment-11275111

## Summary
This PR adds two tests to enforce the behavior:
 - _atoms accessed with write get should have their dependencies updated_

## Check List

### DIFF
--- a/tests/vanilla/dependency.test.tsx
+++ b/tests/vanilla/dependency.test.tsx
@@ -1,6 +1,5 @@
 import { expect, it, vi } from 'vitest'
 import { atom, createStore } from 'jotai/vanilla'
-import type { Getter } from 'jotai/vanilla'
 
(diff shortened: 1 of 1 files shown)


tests/vanilla/dependency.test.tsx
```diff
@@ -1,5 +1,6 @@
 import { expect, it, vi } from 'vitest'
 import { atom, createStore } from 'jotai/vanilla'
+import type { Getter } from 'jotai/vanilla'
```
dai-shi, line 3:
```suggestion
```

dai-shi (approved):
Yeah, this is the behavior we should preserve.

PR #2825, opened 2024-11-16
[Idea] tweak atom type naming

dmaskasky:
## Summary

I'm wondering if we could improve the clarity of the api by tweaking the type naming a bit.

The current naming conventions have inconsistencies:
- The `get` function in `atom.write` does not add atom's dependencies to the top-level.
- The `setSelf` function in `atom.read` options only supports asynchronous updates.
- The `setAtom` function in `atom.onMount` has the same behavior and type signature as `setSelf`, but it also supports synchronous updates. Additionally, the name `setAtom` does not clearly indicate that it can only update the self atom.

---

## Checklist

- [x] Run `pnpm run prettier` to format code and documentation.

### DIFF
--- a/src/vanilla/atom.ts
+++ b/src/vanilla/atom.ts
@@ -1,24 +1,28 @@
 type Getter = <Value>(atom: Atom<Value>) => Value
 
+type Peeker = Getter
+
 type Setter = <Value, Args extends unknown[], Result>(
   atom: WritableAtom<Value, Args, Result>,
   ...args: Args
 ) => Result
 
-type SetAtom<Args extends unknown[], Result> = <A extends Args>(
+type SetSelf<Args extends unknown[], Result> = <A extends Args>(
   ...args: A
 ) => Result
+
+type SetSelfAsync<Args extends unknown[], Result> = SetSelf<Args, Result>
 
 /**
  * setSelf is for internal use only and subject to change without notice.
  */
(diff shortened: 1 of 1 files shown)


src/vanilla/atom.ts
```diff
@@ -1,24 +1,28 @@
...
-type Read<Value, SetSelf = never> = (
+type Read<Value, SetSelfAsync = never> = (
   get: Getter,
-  options: { readonly signal: AbortSignal; readonly setSelf: SetSelf },
+  options: { readonly signal: AbortSignal; readonly setSelf: SetSelfAsync },
 ) => Value
 
 type Write<Args extends unknown[], Result> = (
-  get: Getter,
+  peek: Peeker,
```
dmaskasky, line 25: should we revert this line change?

dai-shi:
I would like to keep `SetSelf` type as is. It's for specifically for `setSelf` option. Let's assume that setSelf is async only without naming change.

If `SetAtom` type name is confusing, we could consider changing it, but not to `SetSelf`.

dmaskasky:
> I would like to keep `SetSelf` type as is. It's for specifically for `setSelf` option. Let's assume that setSelf is async only without naming change.
> 
> If `SetAtom` type name is confusing, we could consider changing it, but not to `SetSelf`.

In that case, let's just keep things as it. I will close this PR.

PR #2827, opened 2024-11-17
lazy recomputeDependents

dmaskasky:

https://github.com/pmndrs/jotai/discussions/2782

## Summary

Currently, recomputeDependents runs for each call to set. This causes overlapping dependents to recompute multiple times.

We still need to recompute some dependents if they are accessed with `get`. But the remaining dependents can be recomputed at the end of the transaction in flushPending.
```ts
const a = atom(0)
const b = atom((get) => get(a))
const fetch = vi.fn()
const c = atom((get) => fetch(get(a)))
  set(a, 1)
  expect(get(b)).toBe(1)
  expect(fetch).toHaveBeenCalledTimes(0)
})
const store = createStore()
store.sub(b, () => {})
store.sub(c, () => {})
fetch.mockClear()
store.set(w)
expect(fetch).toHaveBeenCalledOnce()
expect(fetch).toBeCalledWith(1)
expect(store.get(a)).toBe(1)
```

image

## Proposed Fix

1. on `set`
    - keep a list of all atoms that have been changed
    - mark the atom's state and its dependents' state as dirty
2. on `get`, if the atom's state is dirty
    - recompute the atom and all it's dependencies that are dirty
3. on `flushPending`, recompute all remaining dirty dependents

## More complex example
Suppose you write to node 17. Itself and it's dependents are `{17, 15, 16, 12, 13, 14, 9, 10, 11, 6, 3, 1}`.
We save these nodes as pending recompute.
Then suppose you read from node 1. Itself and it's dependencies are `{1, 2, 3, 4, 5, 6, 7, 8, 9, 12, 15, 17}`.
The intersection of these two sets is `{1, 3, 6, 9, 12, 15, 17}`.
On `get`, only these nodes will be eagerly recomputed.
On `flushPending`, the remaining nodes `{16, 13, 14, 10, 11}` will be recomputed.

image

### `getMountedDependents`
Was previously `getDependents`. Is called by `getAllDependents`. Returns a set of all mounted dependents.

### `getAllDependents`
Is called by `markRecomputePending` and `getSortedDependents`. Returns a map of atoms to a set of their dependents (`Map<AnyAtom, Set<AnyAtom>>`).

### `markRecomputePending`
Is called by `recomputeDependents` and `setter`. Sets `atomState.x` to true.

### `getSortedDependents`
Is called by `recomputeDependents`. Returns a topological sorted list of atoms and its dependent atoms (deep).

### `recomputeDependents`
Is called by `recomputeDependencies` and `flushPending`. Receives a set of atoms for which to recompute dependents on. Iteratively calls `readAtomState` and `mountDependencies` on all atoms and their dependents (deep) whose `atomState.x` is true.

### `recomputeDependencies`
Is called by the `getter` in `writeAtomState`. Recomputes the dependents (deep) of all dependencies (deep) whose `atomState.x` is true.

### `flushPending`
Is called in the following places
1. In the `readAtomState` `getter` finally block if `getter` is called async.
2. In `complete` which is called when a valueOrPromise is a promise and has settled.
3. In `setter` finally block if `setter` is called async.
4. In `writeAtom` finally block.
5. In `setAtom` finally block if `setAtom` is called async.
6. At the end of `subscribeAtom`
7. At the end of `unsubscribeAtom`

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -166,20 +166,24 @@ const addDependency = <Value>(
 type Batch = Readonly<{
   /** Atom dependents map */
   D: Map<AnyAtom, Set<AnyAtom>>
+  /** High priority functions */
+  H: Set<() => void>
   /** Medium priority functions */
   M: Set<() => void>
   /** Low priority functions */
   L: Set<() => void>
 }>
 
-const createPending = (): Pending => [
-  /** dependents */
-  new Map(),
-  /** atomStates */
-  new Map(),
-  /** functions */
-  new Set(),
-]
+const createBatch = (): Batch => ({
+  D: new Map(),
+  H: new Set(),
+  M: new Set(),
+  L: new Set(),
+})
+
+const addBatchFuncHigh = (batch: Batch, fn: () => void) => {
+  batch.H.add(fn)
+}
 
 const addBatchFuncMedium = (batch: Batch, fn: () => void) => {
   batch.M.add(fn)
@@ -220,6 +224,30 @@ const copySetAndClear = <T>(origSet: Set<T>): Set<T> => {
   const newSet = new Set(origSet)
   origSet.clear()
   return newSet
+}
+
+const flushBatch = (batch: Batch) => {
+  let error: AnyError
+  let hasError = false
+  const call = (fn: () => void) => {
+    try {
+      fn()
+    } catch (e) {
+      if (!hasError) {
+        error = e
+        hasError = true
+      }
+    }
+  }
+  while (batch.M.size || batch.L.size) {
+    batch.D.clear()
+    copySetAndClear(batch.H).forEach(call)
+    copySetAndClear(batch.M).forEach(call)
+    copySetAndClear(batch.L).forEach(call)
+  }
+  if (hasError) {
+    throw error
+  }
 }
 
 // internal & unstable type
@@ -273,33 +301,6 @@ const buildStore = (
     debugMountedAtoms = new Set()
   }
 
-  const flushPending = (pending: Pending) => {
-    let error: AnyError
-    let hasError = false
-    const call = (fn: () => void) => {
-      try {
-        fn()
-      } catch (e) {
-        if (!hasError) {
-          error = e
-          hasError = true
-        }
-      }
-    }
-    while (pending[0].size || pending[1].size || pending[2].size) {
-      recomputeDependents(pending, new Set(pending[0].keys()))
-      const atomStates = new Set(pending[1].values())
-      pending[1].clear()
-      const functions = new Set(pending[2])
-      pending[2].clear()
-      atomStates.forEach((atomState) => atomState.m?.l.forEach(call))
-      functions.forEach(call)
-    }
-    if (hasError) {
-      throw error
-    }
-  }
-
   const setAtomStateValueOrPromise = (
     atom: AnyAtom,
     atomState: AtomState,
@@ -314,11 +315,11 @@ const buildStore = (
         addPendingPromiseToDependency(atom, valueOrPromise, getAtomState(a))
       }
       atomState.v = valueOrPromise
-      delete atomState.e
     } else {
       atomState.v = valueOrPromise
-      delete atomState.e
     }
+    delete atomState.e
+    delete atomState.x
     if (!hasPrevValue || !Object.is(prevValue, atomState.v)) {
       ++atomState.n
       if (pendingPromise) {
@@ -347,7 +348,7 @@ const buildStore = (
           ([a, n]) =>
             // Recursively, read the atom state of the dependency, and
             // check if the atom epoch number is unchanged
-            readAtomState(pending, a).n === n,
+            readAtomState(batch, a).n === n,
         )
       ) {
         return atomState
@@ -370,7 +371,7 @@ const buildStore = (
         return returnAtomValue(aState)
       }
       // a !== atom
-      const aState = readAtomState(pending, a)
+      const aState = readAtomState(batch, a)
       try {
         return returnAtomValue(aState)
       } finally {
@@ -431,6 +432,7 @@ const buildStore = (
     } catch (error) {
       delete atomState.v
       atomState.e = error
+      delete atomState.x
       ++atomState.n
       return atomState
     } finally {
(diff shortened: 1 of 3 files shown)


dai-shi:
> * Unrelated: aState.m.l subscribers are not deduped in flushPending [Fixed](https://github.com/pmndrs/jotai/pull/2827/commits/8c9258fe2726a9e735d813763cdb41f4b1e0ca94)

The commit link is not found. (because of force pushing?)

dmaskasky:
> > * Unrelated: aState.m.l subscribers are not deduped in flushPending [Fixed](https://github.com/pmndrs/jotai/pull/2827/commits/8c9258fe2726a9e735d813763cdb41f4b1e0ca94)
> 
> The commit link is not found. (because of force pushing?)

Thinking on this more. aState.m.l subscribers do not need to be deduped. I reverted the code for this.

src/vanilla/store.ts
```diff
@@ -104,6 +104,8 @@ type AtomState<Value = AnyValue> = {
   v?: Value
   /** Atom error */
   e?: AnyError
+  /** Whether the atom value is has been changed */
+  isDirty?: boolean
```
dmaskasky, line 108: TODO: find a better property name

src/vanilla/store.ts
```diff
@@ -198,33 +207,6 @@ const addPendingFunction = (pending: Pending, fn: () => void) => {
   pending[2].add(fn)
 }
 
-const flushPending = (pending: Pending) => {
```
dmaskasky, line 201: needed to move flushPending in to buildStore in order to put invoke recomputeDependents from flushPending.

src/vanilla/store.ts
```diff
@@ -276,6 +258,33 @@ const buildStore = (
...
+        fn()
+      } catch (e) {
+        if (!hasError) {
+          error = e
+          hasError = true
+        }
+      }
+    }
+    while (pending[0].size || pending[1].size || pending[2].size) {
+      recomputeDependents(pending, new Set(pending[0].keys()))
```
dmaskasky, line 275: Actual change
```diff
- while (pending[1].size || pending[2].size) {
-   pending[0].clear()
+ while (pending[0].size || pending[1].size || pending[2].size) {
+ recomputeDependents(pending, new Set(pending[0].keys()))
```

src/vanilla/store.ts
```diff
@@ -104,6 +104,8 @@ type AtomState<Value = AnyValue> = {
   v?: Value
   /** Atom error */
   e?: AnyError
+  /** Indicates that the atom value has been changed */
+  x?: true
```
dai-shi, line 108: Looks good.

src/vanilla/store.ts
dai-shi: While I'm not sure which is better, I would try keeping it outside. I'm not sure about the future, and we may put "pending" in buildStore. In general, I prefer isolating it.

src/vanilla/store.ts
```diff
@@ -418,118 +426,156 @@ const buildStore = (
...
-    const [topsortedAtoms, markedAtoms] = getSortedDependents(
-      pending,
-      atom,
-      atomState,
-    )
-
-    // Step 2: use the topsorted atom list to recompute all affected atoms
-    // Track what's changed, so that we can short circuit when possible
-    const changedAtoms = new Set<AnyAtom>([atom])
-    for (let i = topsortedAtoms.length - 1; i >= 0; --i) {
```
dai-shi, line 504: Why not keep using the for-i loop and avoid .reverse()?

src/vanilla/store.ts
```diff
@@ -418,118 +426,156 @@ const buildStore = (
...
   const writeAtomState = <Value, Args extends unknown[], Result>(
     pending: Pending,
     atom: WritableAtom<Value, Args, Result>,
     ...args: Args
   ): Result => {
     let isSync = true
-    const getter: Getter = <V>(a: Atom<V>) =>
-      returnAtomValue(readAtomState(pending, a))
+    const getter: Getter = <V>(a: Atom<V>) => {
+      recomputeDependencies(pending, atom)
```
dai-shi, line 576: I'm not sure why this is necessary.

dai-shi (commented):
Thanks so much for working on it. The commit history tells how challenging this has been.

I only check the diff, and assuming that this keeps the basic logic and only defers the execution, it's pretty close to what I would do. What's blocking is the reasonable understanding of the `store.test.tsx` change and failed CI tests.

You can stop it here and I can take it over, if you like.

src/vanilla/store.ts
dmaskasky: We can keep it outside if we complicate pending by queuing a special function that runs before the others.

But I think I prefer flushPending to be part of the store. To me it makes sense that the commit phase of the transaction is internal to buildStore.

src/vanilla/store.ts
dmaskasky: I didn't like how getSortedDependents returns the opposite of what you would expect, which is a topologically sorted list. It has the same complexity, I think the performance will be comparable.

src/vanilla/store.ts
dmaskasky: Perhaps I should leave a comment here. Before we get any atom, we need to recompute the dependents of any of its dependencies that have been marked as dirty. Tonensure we are returning the latest value. This is the exceltikn to the lazy recomputeDependents rule.

Example:
```ts
const a = atom(0)
const b = atom(0)
const c = atom((get) => [get(a), get(b)])
const d = atom((get) => get(c))
const e = atom((get) => get(a))
  get(d) // [0, 0]

  // does not immediately recompute
  // marks a as dirty
  set(a, 1)

  // setA: dependencies of d are {c,a,b}
  // setB: dirty atoms of setA are {a}
  // setC: dependents setB are {a,c,e}
  // setD: intersection of setC and setB are {a,c}
  // recompute setD
  get(d) // [1, 0]

  // recompute remaining dependents of dirty atoms in flushPending
})
store.set(w)
```

dai-shi: I'm not convinced but maybe I need to play with it to get better understanding. It's my problem.

dmaskasky: I made this graphic to help me reason about this better.
image

src/vanilla/store.ts
dai-shi: I agree with the readability. I thought the performance matters and we are targeting deep dependencies like the case that recursion doesn't work. Are you sure that the performance is comparable?

dmaskasky:
I think I figured out why this passes the peek test. Here's a better test. This test also fails on main, but passes on fix/deep-peek.

```ts
it('can read in write function without changing dependencies', () => {
  const a = atom(0)
  let bReadCount = 0
  const b = atom(
    (get) => {
      ++bReadCount
      return get(a)
    },
    () => {},
  )
  let bMountCalled = false
  b.onMount = () => {
    bMountCalled = true
  }
  const c = atom((get) => get(a))
    expect(bReadCount).toBe(0)
    get(b)
    expect(bReadCount).toBe(1)
    set(a, 1)
    expect(bReadCount).toBe(1)
  })
  const store = createStore()
  store.sub(c, () => {}) // mounts c,a
  store.set(w)
  
  expect(bMountCalled).toBe(false) // ✅ PASSES (should fail if b atomState.m is defined
  expect(bReadCount).toBe(1) // ❌ FAILS - received 2
  expect(getAtomState(b)!.m).toBeUndefined() // ✅ PASSES (could fail)
  expect(getAtomState(b)!.d.size).toBe(0) // ❌ FAILS - 1 (a atomState)
  expect(getAtomState(a)!.m!.d.has(b)).toBe(false) // ❌ FAILS - true
})
```

dai-shi (approved):
Thanks so much for your effort! This is a big improvement.

PR #2828, opened 2024-11-18
refactor(store): prefer epoch number comparisons to indicate value change

dmaskasky:
## Summary
Replaces value comparisons with epoch number changes. This may help prevent memory leaks.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -538,12 +538,11 @@ const buildStore = (
             // NOTE technically possible but restricted as it may cause bugs
             throw new Error('atom not writable')
           }
-          const hasPrevValue = 'v' in aState
-          const prevValue = aState.v
+          const prevEpochNumber = aState.n
           const v = args[0] as V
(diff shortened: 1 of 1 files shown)


src/vanilla/store.ts
```diff
@@ -538,12 +538,11 @@ const buildStore = (
...
             throw new Error('atom not writable')
           }
-          const hasPrevValue = 'v' in aState
-          const prevValue = aState.v
+          const prevEpochNumber = aState.n
           const v = args[0] as V
           setAtomStateValueOrPromise(a, aState, v)
           mountDependencies(pending, a, aState)
-          if (!hasPrevValue || !Object.is(prevValue, aState.v)) {
+          if (prevEpochNumber !== aState.n) {
```
dai-shi, line 545: This looks really good and I probably missed it when introducing the epoch number.

My concern was for the error case, but the epoch number is incremented with `delete atomState.v`, so it should be fine.

dai-shi (approved):
LGTM! Nice finding.

PR #2830, opened 2024-11-18

dai-shi:
for v2.10.3

with patching RTL for react 19-rc.1.


package.json
```diff
@@ -158,17 +158,17 @@
     "jsdom": "^25.0.1",
     "json": "^11.0.0",
     "prettier": "^3.3.3",
-    "react": "19.0.0-rc.0",
-    "react-dom": "19.0.0-rc.0",
-    "rollup": "^4.25.0",
+    "react": "19.0.0-rc.1",
```
dai-shi, line 161: Updating react to rc.1 makes tests to fail.

@kretajak I wonder if you are familiar with this case. Do we misuse RTL for async tests?

kretajak: I will take a look at that. The way RTL is used here to test async stuff is fine from my perspective. Maybe some tweaks mentioned in [React 19 upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide) are required.

kretajak: Looks like broken test are using `Suspense` and `Suspense` behavior was changed in RC1 (https://github.com/facebook/react/issues/29898#issuecomment-2477449973).

dai-shi: Thanks for taking a look. I know the Suspense behavior was changed, but didn't consider if it really broke it.

https://stackblitz.com/edit/vitejs-vite-hbi7w9?file=src%2FApp.tsx&terminal=dev This looks like working. So, it's probably test only issue.

.github/workflows/test-multiple-versions.yml
```diff
@@ -33,9 +33,9 @@ jobs:
           - 18.1.0
           - 18.2.0
           - 18.3.1
-          - 19.0.0-rc.0
-          - 19.0.0-rc-5c56b873-20241107
-          - 0.0.0-experimental-5c56b873-20241107
+          - 19.0.0-rc.1
+          - 19.0.0-rc-e1ef8c95-20241115
+          - 0.0.0-experimental-e1ef8c95-20241115
```
dai-shi, line 38: Here's the list of commits for the change:

https://github.com/facebook/react/compare/5c56b873...e1ef8c95

dai-shi: More precisely, it's between these commit that cause the error:
https://github.com/facebook/react/compare/380f5d67...b01722d5

dai-shi: So, `enableSiblingPrerendering` should only be the meaningful change.

tests/react/basic.test.tsx
```diff
@@ -34,6 +36,28 @@ const useCommitCount = () => {
   return commitCountRef.current
 }
 
+/* eslint-disable testing-library/no-unnecessary-act */
+const renderRoot = async (element: ReactNode) => {
```
dai-shi, line 40: This is the only hack I came up with. If it looks good, I'll work on other files too.

dai-shi: It seems like we should await all `act()`. `render` from RTL throws away the promise from `act()` call.
https://github.com/testing-library/react-testing-library/blob/main/src/pure.js#L145

This workaround isn't ideal because we can't test the initial "Loading...".

patches/@testing-library__react@16.0.1.patch
```diff
@@ -0,0 +1,22 @@
...
++async function renderRoot(ui, {
+   baseElement,
+   container,
+   hydrate,
+@@ -177,7 +177,7 @@ function renderRoot(ui, {
+   root,
+   wrapper: WrapperComponent
+ }) {
+-  (0, _actCompat.default)(() => {
++  await (0, _actCompat.default)(() => {
```
dai-shi, line 19: As it turns out, this helps.

tests/react/abortable.test.tsx
```diff
@@ -1,10 +1,15 @@
...
+import { act, render, screen, waitFor } from '@testing-library/react'
+import userEventOrig from '@testing-library/user-event'
 import { describe, expect, it } from 'vitest'
 import { useAtomValue, useSetAtom } from 'jotai/react'
 import { atom } from 'jotai/vanilla'
 
+const userEvent = {
+  // eslint-disable-next-line testing-library/no-unnecessary-act
+  click: (element: Element) => act(() => userEventOrig.click(element)),
+}
```
dai-shi, line 11: I hoped to patch `user-event` too, but didn't find an easy way.

dai-shi (commented):
It's _very_ tricky, but for now, it seems to pass the CI.

PR #2831, opened 2024-11-20
fix(website): fix doc site sidebar is not displayed

zakke27:
## Summary

fix doc site sidebar is not displayed

## Check List


sandren:
Thanks for the PR, but linking back to the table of contents was an intentional change in anticipation of a new docs navigation menu. Stay tuned! 🙂

PR #2833, opened 2024-11-22
[DO NOT MERGE] try Kahn's toposort

dmaskasky:
*FOR DEMONSTRATION PURPOSES ONLY*

## Summary
Kahn's Topological Sort uses an iterative BFS approach to sort nodes in a DAG. It uses in-degrees a count of the number of remaining unsorted dependencies a node as. As nodes are sorted, in-degrees is decremented. When in-degrees reaches zero for a node, that node is appended to the sorted set.

The difficulty with using Kahn's in our application is that we first need to do a bunch of expensive admin work.
- construct a map of all dependent nodes to their dependents. Implemented here is an iterative BFS solution, but a DFS approach has the same complexity
- create a map of every atom to their in-degree
- iterate through all atoms and enqueu where `in-degree === 0`

The result averages (100x) show that Kahns runs ~25% slower than iterative DFS. 
```
--------------------------------------------------------------------------------
LINEAR
kahnsToposort: 25.4166
getSortedDependents: 20.7627
kahn times are 22.4148 percent slower than iterative DFS
--------------------------------------------------------------------------------
STAR
kahnsToposort: 22.9688
getSortedDependents: 19.8854
kahn times are 15.5060 percent slower than iterative DFS
--------------------------------------------------------------------------------
K-ARY
kahnsToposort: 26.1628
getSortedDependents: 21.1882
kahn times are 23.4780 percent slower than iterative DFS
```

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -482,6 +482,87 @@ const buildStore = (
     return [sorted, visited]
   }
 
+  function kahnsToposort(
+    pending: Pending,
+    rootAtom: AnyAtom,
+    rootAtomState: AtomState,
+  ): [[AnyAtom, AtomState, number][], Set<AnyAtom>] {
+    function createQueue<T>() {
+      let head = 0
+      let tail = 0
+      const items: Record<number, T> = {}
+      const isEmpty = () => head === tail
+      const enqueue = (item: T) => {
+        items[tail++] = item
+      }
+      const dequeue = () => {
+        if (head === tail) {
+          return undefined
+        }
+        const item = items[head]
+        items[head++] = undefined as any
+        return item
+      }
+      return { enqueue, dequeue, isEmpty }
+    }
+    // Kahn's algorithm for topological sorting
+    const graph = new Map<
+      AnyAtom,
+      [dependents: AnyAtom[], atomState: AtomState]
+    >()
+    const inDegree: Map<AnyAtom, number> = new Map([[rootAtom, 0]])
+    {
+      // 1: build the dependency graph
+      // and calculate in-degrees (number of incomming edges aka dependencies)
+      const queue = createQueue<[a: AnyAtom, aState: AtomState]>()
+      queue.enqueue([rootAtom, rootAtomState])
+      while (!queue.isEmpty()) {
+        const [a, aState] = queue.dequeue()!
+        if (graph.has(a)) {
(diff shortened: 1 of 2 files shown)


dmaskasky:
Update: if I use indexes instead of map, its a little faster but more ugly.

```
LINEAR --------------------------------------------------------------------------------
kahnsToposort: 18.4981
getSortedDependents: 18.0733
kahn times are 2.3504 percent slower than toposort
STAR --------------------------------------------------------------------------------
kahnsToposort: 19.8348
getSortedDependents: 20.3942
kahn times are 2.7431 percent faster than toposort
K-ARY --------------------------------------------------------------------------------
kahnsToposort: 20.2995
getSortedDependents: 20.3445
kahn times are 0.2209 percent faster than toposort
```

dmaskasky:
Update: using arrays seems to help
```
LINEAR --------------------------------------------------------------------------------
kahnsToposort: 20.1562
getSortedDependents: 20.0564
kahn times are 0.4973 percent slower than toposort
STAR --------------------------------------------------------------------------------
kahnsToposort: 17.1281
getSortedDependents: 20.2366
kahn times are 15.3605 percent faster than toposort
K-ARY --------------------------------------------------------------------------------
kahnsToposort: 19.4536
getSortedDependents: 19.6978
kahn times are 1.2398 percent faster than toposort
```

src/vanilla/store.ts
```diff
@@ -482,13 +482,123 @@ const buildStore = (
...
+          inDegree[adjacent]!--
+          // If indegree becomes 0, push it to the queue
+          if (inDegree[adjacent] === 0) {
+            queue.enqueue(adjacent)
+          }
+        }
+      }
+    }
+    return [sorted, visited]
+  }
```
dmaskasky, line 573: @dai-shi Can I get your thoughts on this approach?

Should we keep the current iterative DFS approach, or do you think the performance gains are worth the added complexity of Kahns?

Kahns does marginally better on Star topologies, which I believe would be the common case.

dmaskasky:
Closing this for now. While Kahn's algorithm is slightly faster for extremely large K-Ary and Star graphs, it comes with higher memory overhead and performs worse on smaller atom graphs. Given that the average case involves smaller atom graphs, I don't believe the added complexity of Kahn's algorithm is justified.

PR #2834, opened 2024-11-22
add ecosystem-ci workflow

dai-shi:
using https://github.com/jotaijs/jotai-ecosystem-ci


dai-shi:
Does it work without merging?

PR #2839, opened 2024-11-26
fix(workflow): ecosystem-ci

dai-shi:
There were some errors.


dai-shi:
hmm, i need merge to test anyway.

PR #2845, opened 2024-12-05
test: it can read in write fns w/o changing deps

dmaskasky:

Fixes #

## Check List

### DIFF
--- a/tests/vanilla/dependency.test.tsx
+++ b/tests/vanilla/dependency.test.tsx
@@ -346,3 +346,33 @@ it('can read sync derived atom in write without initializing', () => {
   // note: this is why write get needs to update deps
   expect(store.get(a)).toBe(2)
 })
+
+it('can read in write function without changing dependencies', () => {
+  // https://github.com/pmndrs/jotai/discussions/2789
+  const a = atom(0)
+  let bReadCount = 0
+  const b = atom(
+    (get) => {
+      ++bReadCount
+      return get(a)
+    },
+    () => {},
+  )
+  let bIsMounted = false
+  b.onMount = () => {
+    bIsMounted = true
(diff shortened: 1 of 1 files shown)


src/vanilla/store.ts
```diff
@@ -519,14 +519,83 @@ const buildStore = (
     }
   }
 
+  const peekAtomState = <Value>(
+    pending: Pending | undefined,
+    atom: Atom<Value>,
+    dirtyAtoms?: Set<AnyAtom>,
+  ): AtomState<Value> => {
+    const atomState = getAtomState(atom)
```
dmaskasky, line 527: note: we cannot skip recomputing the atom value each time because we do not update dependencies. This may make deep peek a deal-breaker.

src/vanilla/store.ts
```diff
@@ -519,14 +519,83 @@ const buildStore = (
     }
   }
 
+  const peekAtomState = <Value>(
```
dmaskasky, line 522: a lot of `peekAtomState` code looks similar to `readAtomState`. Is it cleaner to have this separate function or to add more conditionals to `readAtomState`?

dai-shi:
> Refactor Options

I don't prefer 1. 2 looks good.
Let me play with this problem. And, I may find 3 or 4 could be options.

hi-ogasawara:
@dai-shi @dmaskasky 

Since there was a test for unmounting, I wrote a test for getting a mounted read-only atom with a write atom. Is this the expected movement?

e7f02e7bd05d95945f73e8e6d4ab386f41224ca8

dai-shi:
Closing in favor of #2849.

Thanks for the prior art anyway!

PR #2852, opened 2024-12-11
feat: dev store with unstable_derive

dai-shi:
Thanks to `unstable_derive`, we can implement dev store with it.


dai-shi:
@arjunvegda Let's add jotai-devtools to https://github.com/jotaijs/jotai-ecosystem-ci ! Can you open a PR there?

dmaskasky:
Separately, I think createPending should be part of deriveStore.

src/vanilla/store.ts
```diff
@@ -706,43 +691,72 @@ const buildStore = (
...
         }
-        flushPending(pending)
+        return atomState
       },
-    }
-    Object.assign(store, devStore)
+    }),
+    dev4_get_mounted_atoms: () => debugMountedAtoms,
+    dev4_restore_atoms: (values) => {
+      throw new Error('TODO: not implemented yet' || values)
```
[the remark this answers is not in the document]
dmaskasky: https://github.com/pmndrs/jotai/pull/2827 introduces batch inside the write function only.
So after #2827 lands, it could look something like:
```ts
{
  ...
  dev4_restore_atoms: (values) => {
    store.set(atom(null, (get, set) => {
      values.forEach(([a, v]) => set(a, v))
    })
  },
}
```

dai-shi:
> Separately, I think createPending should be part of deriveStore.

If necessary, I generally want to make functions as pure as reasonably possible.

dai-shi:
> "jotai-devtools": "FAIL",

Yes, it's expected. Cool.

src/vanilla/store.ts
```diff
@@ -725,43 +710,88 @@ const buildStore = (
...
+    }),
+    dev4_get_mounted_atoms: () => debugMountedAtoms,
+    dev4_restore_atoms: (values) => {
+      const restoreAtom: WritableAtom<null, [], void> &
+        Record<typeof RESTORE_ATOM, typeof values> = {
+        [RESTORE_ATOM]: values,
+        read: () => null,
+        write: () => {},
+      }
+      derivedStore.set(restoreAtom)
```
dmaskasky, line 786: I'm curious why not simply?
```ts
store.set(atom(null, (get, set) => {
  values.forEach((args) => set(...args))
})
```

dai-shi:
> "jotai-devtools": "FAIL",

Hmm, it's unexpected.

src/vanilla/store.ts
dai-shi: `atom` in `values` can have a custom `write` function, which we need to bypass.

dai-shi:
@arjunvegda Can you help here? My expectation is this is compatible behavior-wise, but there might be some cases to break.

arjunvegda:
> @arjunvegda Can you help here? My expectation is this is compatible behavior-wise, but there might be some cases to break.

It appears that whenever a user attempts to restore a specific snapshot, the store ends up creating another snapshot in the "Time Travel" tab in DevTools UI. The intended behavior is to restore the chosen snapshot without adding any new entries to the list.

This is likely because we're calling `store.set` [here](https://github.com/pmndrs/jotai/blob/20e6b3728602f4573c8ff37c9e6610c05dfd05e3/src/vanilla/store.ts#L786), that triggers [a store listener](https://github.com/jotaijs/jotai-devtools/blob/main/src/utils/internals/compose-with-devtools.ts#L116) in dev tools to add a new entry. I'll see if there is a way to fix this in Devtools.

dai-shi:
Hm, I see. Thanks!!

dai-shi:
🎉  Thanks @arjunvegda! I'm planning to release this next week.

src/vanilla/store.ts
```diff
@@ -725,43 +710,89 @@ const buildStore = (
...
+        atomRead,
+        (atom, getter, setter, ...args) => {
+          if (RESTORE_ATOM in atom) {
+            const values = atom[RESTORE_ATOM] as Iterable<
+              readonly [AnyAtom, AnyValue]
+            >
+            for (const [atom, value] of values) {
+              if (hasInitialValue(atom)) {
+                setter(atom as never, value)
+              }
```
dmaskasky, line 756: I think this is still an issue. `atom(0, () => {})`, needs the `init` value, but its write function should not be called.

I think you will need to add createPending to unstable_derive. Then you can update the atomState and adjust batch.

dai-shi: > I think this is still an issue.

Thanks for the catch. I didn't notice there was a test case missing. Added in #2873.

src/vanilla/store.ts
```diff
@@ -725,43 +710,90 @@ const buildStore = (
...
+  return store
+}
+
+const deriveDevStore = (store: Store): Store & DevStoreRev4 => {
+  const proxyAtomStateMap = new WeakMap()
+  const debugMountedAtoms = new Set<AnyAtom>()
+  let savedGetAtomState: StoreArgs[0]
+  let inRestoreAtom = false
+  const derivedStore = store.unstable_derive(
+    (getAtomState, atomRead, atomWrite, atomOnMount) => {
```
dmaskasky, line 722: For future compat.
```suggestion
    (getAtomState, atomRead, atomWrite, ...rest) => {
```

dai-shi: Not so important as it's the same file.

src/vanilla/store.ts
```diff
@@ -723,43 +708,90 @@ const buildStore = (
...
+      }
+      savedStoreSet(restoreAtom)
+    },
   }
-  return store
+  return Object.assign(derivedStore, devStore)
 }
 
-export const createStore = (): Store => {
+type PrdOrDevStore = Store | (Store & DevStoreRev4)
```
dmaskasky, line 792: I think type `Store` should represent the complete store union. It makes the type cleaner in user land.
image

```suggestion
type Store = PrdStore | (PrdStore & DevStoreRev4)
```

dai-shi: It just looks ugly, right?
I'm planning to change the `createStore` to return just store without dev methods, in the near future (kind of breaking change to `jotai-devtools`). So, this is somewhat intentional.

dmaskasky: > It just looks ugly, right?

Yes

> I'm planning to change the `createStore` to return just store without dev methods,

I just think the type returned from `createStore` should be called `Store`.

dai-shi: > I just think the type returned from `createStore` should be called `Store`.

So, it will in the near future. Meanwhile, this ugly type is kind of warning. I don't know if anybody realize it though.

PR #2854, opened 2024-12-12
[idea] add pending to `store.unstable_derive`

dmaskasky:
## Summary
Improves capability of unstable_derive store api. Pending is a good candidate to represent a transaction primitive. These changes make it possible to have fine-grained control over the atom lifecycle.

### Details
- Adds createPending to StoreArgs
- Shares pending with all StoreArgs
- Adds flushPending to StoreArgs

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -226,20 +226,23 @@ const flushPending = (pending: Pending) => {
 
 // internal & unstable type
 type StoreArgs = readonly [
-  getAtomState: <Value>(atom: Atom<Value>) => AtomState<Value>,
+  getAtomState: <Value>(pending: Pending | undefined, atom: Atom<Value>) => AtomState<Value>,
   atomRead: <Value>(
+    pending: Pending | undefined,
     atom: Atom<Value>,
     ...params: Parameters<Atom<Value>['read']>
   ) => Value,
   atomWrite: <Value, Args extends unknown[], Result>(
+    pending: Pending,
     atom: WritableAtom<Value, Args, Result>,
     ...params: Parameters<WritableAtom<Value, Args, Result>['write']>
   ) => Result,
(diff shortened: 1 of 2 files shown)


src/vanilla/store.ts
```diff
@@ -227,19 +226,27 @@ const flushPending = (pending: Pending) => {
...
+    pending: Pending,
     atom: WritableAtom<Value, Args, Result>,
     ...params: Parameters<WritableAtom<Value, Args, Result>['write']>
   ) => Result,
   atomOnMount: <Value, Args extends unknown[], Result>(
+    pending: Pending,
     atom: WritableAtom<Value, Args, Result>,
     setAtom: (...args: Args) => Result,
   ) => OnUnmount | void,
+  createPending: (prevPending?: Pending | undefined) => Pending,
```
dmaskasky, line 248: prevPending is the in-context upstream pending. For instance, unsubscribe's invocation of createPending receives subscribe's pending.

dai-shi:
This is pretty close to our rejected idea (ref #2741), so I have a big hesitation.
Technically, it's good and makes things flexible, but my concern is if developers other than you can understand it. Unless we have urgent use cases, I believe we shouldn't rush. (Just curious if this helps jotai-scope.)

In any case, my plan for this is after we finalize "sync effect". So, converting to draft for now.

dmaskasky:
> This is pretty close to our rejected idea (ref https://github.com/pmndrs/jotai/pull/2741), so I have a big hesitation.

The big difference in my mind is that pending gives a place for extra details about the transaction to be processed. The difficulty we had with the earlier idea with passing `originalAtomState` was that the notion of _from_ was a bit opaque.

Say we had `A -> B -> C`. C's _from_ would be `B`, but originalAtomState would be A's state.

Now with flushPending, we are free to implement our own linked list or whatever.

This might be something to consider if the need ever arises.

PR #2867, opened 2024-12-16
[Not For Merge] alternate impl for #2827

dai-shi:
does this work?

ref: #2827


dai-shi:
All tests are passing. We need to create a test to break it.

dai-shi:
2 was my concern too. it requires a priority queue instead of a set. though, with the react use case, it's not a problem because useReducer dispatch is async. a new test with store api should fail.

1 should be fine unless we want to make different behaviors between mounted atoms and unmounted atoms.

dmaskasky:
Let me continue to iterate on #2827. I'll have some failing tests later today that we can use as a North star.

> unless we want to make different behaviors between mounted atoms and unmounted atoms

I was surprised to see no failing test, tbh. There is definitely a gap in our testing here. The `get` in write function is always expected to return the latest value. If the atom is a dependent on something that changed, we should to recompute that atom in place.

dai-shi:
> Let me continue to iterate on #2827.

That'd be good, and I will cherry-pick commits into this branch.

> I'll have some failing tests later today that we can use as a North star.

We are rowing our boats in the ocean from different directions.

dmaskasky:
Update: I am unable to break this implementation. Nice work.

dai-shi:
I will try to break 2.

dai-shi:
moved to https://github.com/dmaskasky/jotai/pull/1

PR #2871, opened 2024-12-17
fix(store): robust flush batch

dmaskasky:
## Summary
The recent refactor #2868 to batching causes a subtle regression in which errors thrown in listeners could cause the remaining listeners for that atom not to run.

This PR flattens listeners to the pending function so errors thown are detected but only rethrown after the remaining listeners have been processed.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -201,7 +201,7 @@ const registerBatchAtom = (
   if (!batch.D.has(atom)) {
     batch.D.set(atom, new Set())
     addBatchFuncMedium(batch, () => {
-      atomState.m?.l.forEach((listener) => listener())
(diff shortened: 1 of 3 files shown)


dai-shi:
Nice catch. I forgot about that calling a listener can throw.

src/vanilla/store.ts
```diff
@@ -241,9 +235,12 @@ const flushBatch = (batch: Batch) => {
   }
   while (batch.M.size || batch.L.size) {
```
dai-shi, line 236: Sorry, I forgot this in previous PR.

```suggestion
  while (batch.H.size || batch.M.size || batch.L.size) {
```

dai-shi (approved):
Thanks for the fix!

PR #2880, opened 2024-12-19
fix: unwrap synchronously returns awaited values of resolved promise atoms

dmaskasky:
https://github.com/pmndrs/jotai/discussions/2877

## Summary
Synchronously return atom values whose promises have already resolved.
```ts
const asyncAtom = atom(Promise.resolve(0)
await store.get(asyncAtom)
store.get(unwrap(asyncAtom)) // 0
```

## Check List

### DIFF
--- a/tests/vanilla/utils/unwrap.test.ts
+++ b/tests/vanilla/utils/unwrap.test.ts
@@ -17,19 +17,19 @@ describe('unwrap', () => {
 
     expect(store.get(syncAtom)).toBe(undefined)
     resolve()
-    await new Promise((r) => setTimeout(r)) // wait for a tick
+    await store.get(asyncAtom)
     expect(store.get(syncAtom)).toBe(2)
 
     store.set(countAtom, 2)
     expect(store.get(syncAtom)).toBe(undefined)
     resolve()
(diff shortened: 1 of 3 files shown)


tests/vanilla/utils/unwrap.test.ts
```diff
@@ -114,39 +114,33 @@ describe('unwrap', () => {
...
+    await store.get(asyncAtom)
     expect(store.get(syncAtom)).toBe(3)
   })
 
   it('should unwrap to a fulfilled value of an already resolved async atom', async () => {
     const store = createStore()
     const asyncAtom = atom(Promise.resolve('concrete'))
-
-    expect(await store.get(asyncAtom)).toEqual('concrete')
-    expect(store.get(unwrap(asyncAtom))).toEqual(undefined)
```
dmaskasky, line 136: This is the change in behavior.

dai-shi:
Ohhhhh, my big apologies.
This is an intentional behavioral changed introduced in v2.10.0.
Sorry, I should have noticed when I read #2877 briefly. Thought something different with more recent changes.

This is not acceptable, but let's see if we have other solutions.

PR #2887, opened 2024-12-23
[INTERNAL] Add batch to listeners

dmaskasky:
## Summary
- Improves readability for addBatchFunc
- Passes batch to internal listeners.x
This PR does not change API or behavior.
- Adds support for [future enhancements](https://github.com/pmndrs/jotai/pull/2801)

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -198,9 +198,12 @@ const registerBatchAtom = (
 ) => {
   if (!batch.D.has(atom)) {
     batch.D.set(atom, new Set())
-    addBatchFunc(batch, 'M', () => {
-      atomState.m?.l.forEach((listener) => addBatchFunc(batch, 'M', listener))
-    })
+    const scheduleListeners = () => {
+      for (const listener of atomState.m?.l ?? []) {
+        addBatchFunc(batch, listener, 'M')
+      }
+    }
+    addBatchFunc(batch, scheduleListeners, 'M')
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/vanilla/store.ts
```diff
@@ -198,9 +200,12 @@ const registerBatchAtom = (
 ) => {
   if (!batch.D.has(atom)) {
     batch.D.set(atom, new Set())
-    addBatchFunc(batch, 'M', () => {
-      atomState.m?.l.forEach((listener) => addBatchFunc(batch, 'M', listener))
```
dai-shi, line 202: I would like to avoid this hack too.

src/vanilla/store.ts
```diff
@@ -503,7 +508,7 @@ const buildStore = (
 
     // Step 2: use the topSortedReversed atom list to recompute all affected atoms
     // Track what's changed, so that we can short circuit when possible
-    addBatchFunc(batch, 'H', () => {
+    const finishRecompute = () => {
```
dai-shi, line 511: Though increasing bundle size, this is more readable.

src/vanilla/store.ts
```diff
@@ -645,14 +651,15 @@ const buildStore = (
             isSync = false
           }
         }
-        addBatchFunc(batch, 'L', () => {
+        const processOnMount = () => {
```
dai-shi, line 654: This too.

dai-shi (changes requested):
It complicates further than I expect. I'd like to simplify. Unfortunately, this isn't acceptable *for now*.
Naming functions seems good.

PR #2888, opened 2024-12-23
alternate implementation for #2801

dai-shi:
alternate implementation for #2801

Goals:
- Do not expose atomState
- Do not expose internal listeners
- Implement onInit as minimal as possible

Remaining:
- Two `FIXME` comments in effect.test.ts
- I'm not sure if there are some missing capabilities for atomSyncEffect


src/vanilla/atom.ts
```diff
@@ -1,3 +1,5 @@
+import type { INTERNAL_PrdStore as Store } from './store'
```
dmaskasky, line 1:
```suggestion
import type { PrdOrDevStore as Store } from './store'
```

src/vanilla/store.ts
```diff
@@ -260,6 +264,7 @@ type StoreArgs = readonly [
     atom: WritableAtom<Value, Args, Result>,
     setAtom: (...args: Args) => Result,
   ) => OnUnmount | void,
+  atomOnInit: <Value>(atom: Atom<Value>, store: Store) => void,
```
dmaskasky, line 267: Should Init come before Mount in StoreArgs?

src/vanilla/store.ts
```diff
@@ -247,7 +247,11 @@ const flushBatch = (batch: Batch) => {
 
 // internal & unstable type
 type StoreArgs = readonly [
-  getAtomState: <Value>(atom: Atom<Value>) => AtomState<Value>,
+  getAtomState: <Value>(atom: Atom<Value>) => AtomState<Value> | undefined,
+  setAtomState: <Value>(
+    atom: Atom<Value>,
+    atomState: AtomState<Value>,
+  ) => AtomState<Value>,
```
dmaskasky, line 254: nice 😎

src/vanilla/atom.ts
dai-shi: That's correct for the current implementation, but as I plan to remove DevStore by default in the future, this limits more. Maybe I did wrong with `S extends Store` too.

src/vanilla/store.ts
dai-shi: I'm fine with it.

src/vanilla/store.ts
dai-shi: though, returning a (possibly modified) atomState is not very intuitive.

dmaskasky: hmm, maybe this could help, though its a little pedantic
```ts
setAtomState: <Value, AS extends AtomState<Value>>(
    atom: Atom<Value>,
    atomState: AS,
  ) => AS,
```

dai-shi: Good idea. The arg must be original one.

tests/vanilla/effect.test.ts
```diff
@@ -0,0 +1,230 @@
...
+function atomSyncEffect(effect: Effect) {
+  const refAtom = atom(
+    () => ({ deps: new Set(), inProgress: 0 }) as Ref,
+    (get) => {
+      const ref = get(refAtom)
+      return () => {
+        ref.cleanup?.()
+        ref.cleanup = null
+        ref.deps.clear()
+      }
```
dmaskasky, line 29: I'm concerned about the timing of this. unmount cleanup should run after recomputeDependents and before listeners and mount hooks.

tests/vanilla/effect.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  const internalAtom = atom((get) => {
+    const ref = get(refAtom)
+    if (!ref.get) {
+      ref.get = ((a) => {
+        ref.deps.add(a)
+        return get(a)
+      }) as Getter & { peak: Getter }
+    }
+    ref.init()
+    ref.deps.forEach(get)
```
dmaskasky, line 42: note: IIRC this hack was added in response to a breaking behavioral change this past fall dealing with promise values losing their dependencies between runs. But since the effect is now synchronous, we might be able to remove ref.deps.

tests/vanilla/effect.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  )
+  refAtom.onMount = (mount) => mount()
+  const internalAtom = atom((get) => {
+    const ref = get(refAtom)
+    if (!ref.get) {
+      ref.get = ((a) => {
+        ref.deps.add(a)
+        return get(a)
+      }) as Getter & { peak: Getter }
+    }
```
dmaskasky, line 40: if you pass `get` to init, we can move creating getter to init.

tests/vanilla/effect.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    ref.init()
+    ref.deps.forEach(get)
+    if (ref.inProgress > 0) {
+      return
+    }
+    ref.cleanup?.()
+    const cleanup = effectAtom.effect(ref.get!, ref.set!)
```
dmaskasky, line 47: We only want to fire effect synchronous to atomRead if we are recursing. Otherwise, we want to schedule the effect after all dependents have finished recomputing.

Toposort will sort atoms by their dependents - atoms who don't have any (dependents will be listed first). But since any atom can be set we can't rely on toposort. Also, atoms value can be peeked which isn't accounted for by the dependency graph.

tests/vanilla/effect.test.ts
```diff
@@ -0,0 +1,230 @@
...
+          }
+        }
+        const recurse: Setter = (a, ...args) => {
+          if (ref.fromCleanup) {
+            if (import.meta.env?.MODE !== 'production') {
+              console.warn('cannot recurse inside cleanup')
+            }
+            return undefined as any
+          }
+          return set(a, ...args)
```
dmaskasky, line 85: I think we need to wrap with isRecursing boolean. When not recursing, effect should not infinite loop when setting an watched atom.

tests/vanilla/effect.test.ts
```diff
@@ -0,0 +1,230 @@
...
+            return set(a, ...args)
+          } finally {
+            --ref.inProgress
+            ref.get(a) // FIXME why do we need this?
+          }
+        }
+        const recurse: Setter = (a, ...args) => {
+          if (ref.fromCleanup) {
+            if (import.meta.env?.MODE !== 'production') {
+              console.warn('cannot recurse inside cleanup')
```
dmaskasky, line 81: This line has been changed to `throw new Error('set.recurse is not allowed in cleanup')`

dmaskasky:
I would like to propose a few changes here so that effect runs after recomputeDependent and before listeners and mount hooks.
1. We may need a ScheduleEffect function. This function adds a listener somewhere that the store will enqueue to the batch.
1. One possible place could be to store on the internalAtom's `atomState.m.l`.
1. To access atomState, here is one way to get it.
```ts
function getAtomState(store: Store, atom: AnyAtom): AtomState {
  let atomState: AtomState
  try {
    store.unstable_derive((getAtomState) => {
      atomState = getAtomState(atom)!
      return null as any
    })
  } catch {
    // expect error
  }
  return atomState!
}
```
1. We may also need a synchronous way to determine whether an atomState has mounted. One way to do this is to add a property to atomState `ml`, which is fired synchronously after property `m` is set or deleted. There may be another way to do this, this is just one idea.
```ts
const originalMl = ref.atomState.ml
ref.atomState.ml = () => {
  if (ref.atomState.m) {
    // handle on mount
  } else {
   // handle on unmount
  }
  originalMl?.()
}
```
1. The first subscribe case is a little more tricky. Currently subscribeAtom first calls readAtomState before setting property `m`. We actually need the reverse. So we need a way to trigger a refresh to call internalAtom once more. The tricky part is we need to write in the same batch. To do this, we modify setSelf so that it calls writeAtomState as opposed to writeAtom. This is in order to share the same batch.
1. Then setSelf can be stolen from the first read of internalAtom. Calling setSelf synchronous after readAtomState inside the synchronous mount callback inside subscribeAtom will be fine. We configure internalAtom to have a write function that increments refreshAtom. Calling setSelf causes internalAtom to recompute inside recomputeDependents.
1. We also need to adjust the scheduling of atom listeners to batch. We set priority to `'H'`, and adds one more high priority scheduler. This is because the function that enqueues listeners to the batch function could run before recompute dependents in high. So if it enqueues another high function, we can guarantee it runs a the end of high. This lets is add listeners in high, medium, and low and guaranteed to be after recomputeDependents.
```ts
const scheduleListeners = () => {
  for (const [listener, priority] of atomState.l) {
    addBatchFunc(batch, priority, listener)
  }
}
// add priority listeners at the end of 'H' priority
addBatchFunc(batch, 'H', () => addBatchFunc(batch, 'H', scheduleListeners))
```
1. Internal atom's read function now needs to return something that is never equal. If read gets called, the effect should be enqueued, so epoch must never match. We can either return an always-incrementing number as its value, or we can return `NaN` which always fails `v === v`. This is to ensure `registerBatchAtom` is always called when atomEffect is read.

dai-shi:
Thanks for your review!
I think I need to understand the problem better and build my mental model.
I have two questions:
- Can I forget about recursion support in this PR?
- Can you create a failing test for "to guarantee that effect runs after recomputeDependent and before listeners and mount hooks"?

dai-shi:
We'll work on a new PR.

PR #2895, opened 2024-12-30
feat(core): add atom hooks for atom mount, unmount and improve dev store

dmaskasky:
## Summary
Adds hooks `u` and `h` to atomState that are called in-place immediately after atom mount changes or the atomState has updated.

## Check List


src/vanilla/store.ts
```diff
@@ -721,26 +724,17 @@ const deriveDevStoreRev4 = (store: Store): Store & DevStoreRev4 => {
...
-              deleteProperty(target, prop) {
-                if (prop === 'm') {
-                  debugMountedAtoms.delete(atom)
-                }
-                return Reflect.deleteProperty(target, prop)
-              },
-            })
-            proxyAtomStateMap.set(atom, proxyAtomState)
+          const atomState = getAtomState(atom)
+          let originalMounted = atomState.h
```
dai-shi, line 728:
```suggestion
          const originalMounted = atomState.h
```

dai-shi (commented):
Looks good to me. Thanks for the suggestion. This is much cleaner.

I'll merge this when we are ready to release another patch.

src/vanilla/store.ts
```diff
@@ -100,6 +100,8 @@ type AtomState<Value = AnyValue> = {
   n: number
   /** Object to store mounted state of the atom. */
   m?: Mounted // only available if the atom is mounted
+  /** Listener to notify when the atom is mounted or unmounted. */
+  h?: () => void
```
dai-shi, line 104: If you want to change this with another character, we can reconsider.

dmaskasky: How about
```ts
h?: (action: 'm' | 'u') => void
```

dai-shi: `'m' | 'u'` doesn't look too bad even though it increases the bundle size. I'm hesitant to expose batch here, and looking for an alternative.

src/vanilla/store.ts
```diff
@@ -100,6 +100,10 @@ type AtomState<Value = AnyValue> = {
   n: number
   /** Object to store mounted state of the atom. */
   m?: Mounted // only available if the atom is mounted
+  /** Listener to notify when the atom value is updated. */
+  u?: () => void
```
dai-shi, line 104: I'd like to make a patch release with only dev store improvement. So, let's remove this in this PR.
```suggestion
```

src/vanilla/store.ts
dai-shi: We may change `'h'` in the future, but as this is considered internal, it's fine.

src/vanilla/store.ts
```diff
@@ -198,6 +202,7 @@ const registerBatchAtom = (
 ) => {
   if (!batch.D.has(atom)) {
     batch.D.set(atom, new Set())
+    atomState.u?.()
```
dai-shi, line 205:
```suggestion
```

dai-shi (changes requested):

dai-shi (approved):
Much cleaner without Proxy hack.

src/vanilla/store.ts
dmaskasky: > I'm hesitant to expose batch here, and looking for an alternative.

Its pretty central to syncBatch scheduling.

dai-shi: I'm working on it, and not sure if it's successful, but I feel like reverting this PR... Will see.

dai-shi: > Will see.

It failed.

PR #2897, opened 2024-12-30
chore(eslint): migrate to flat config and simplify

sukvvon:
## Summary

* Same with https://github.com/pmndrs/zustand/pull/2912

## Check List


tests/react/dependency.test.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-unescaped-entities */
```
sukvvon, line 1: @dai-shi 

1. Maintain the current state.
2. Add `react/no-unescaped-entities: 'off'` in the rules.

This file only has a `react/no-unecaped-entities` error, so i added `eslint-disable `. Which option do you prefer?

dai-shi: Yes, this looks fine for now.

PR #2899, opened 2025-01-01
[DO_NOT_MERGE] Exploration in moving all internal functions under an object and expose to unstable_derive

dmaskasky:
## Summary
All functions are put under an object `internal`, which is then passed to unstable_derive. This allows complete modification of internals including the ability to intercept calls to certain functions. Obviously, this goes against the SOLID principle of software development, "closed for modification", but I thought it is an interesting study and might reveal new patterns or methodologies down the road.

Please treat this PR as an experiment only. This PR is _NOT FOR MERGE_.

Downsides
- method names cannot be minified
- we do not want to expose this level of control
- internals will change, so unstable_derive api would change frequently
- violates open-closed principle
- it makes unstable_derive callback not a collector. `internals` argument received should not be returned in the array, instead return `{}` to avoid overriding derived store internals.  It would be more useful to intercept the derived store internals than the parent store internals, so if we wanted to do something like this, we would probably return a store creator thunk.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -1,4 +1,5 @@
 import type { Atom, WritableAtom } from './atom.ts'
+import type { ExtractAtomValue } from './typeUtils.ts'
 
 type AnyValue = unknown
 type AnyError = unknown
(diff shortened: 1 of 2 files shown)


dai-shi:
I don't think we should do this. If we were to provide such low-level capabilities, we should expose internal functions as utility functions and let devs create their own `createStore` function.

```js
import { createBatch, ... } from 'jotai/vanilla/internals';
```

At this point, I don't feel the necessity, and these internal apis will never be stable, so depending on them won't be practical.

dmaskasky:
Yes, I agree. It was just an exploration. Please disregard.

PR #2900, opened 2025-01-02
feat(website): docs menu

sandren:
This PR introduces the new docs menu experience.


dai-shi:
There are two "Documentation" menu items in the side bar.

image

PR #2898, opened 2025-01-01
refactor(core): more readability for addBatchFunc

dmaskasky:
## Summary
~Adds iterable to batch to make flushBatch agnostic to the internals of batch.~
edit: Just refactoring for now delaying the decision

## Check List

- [x] `pnpm run prettier` for formatting code and docs

PR #2901, opened 2025-01-02
introduce atom.onInit hook

dmaskasky:
## Related PRs
https://github.com/pmndrs/jotai/pull/2888
https://github.com/pmndrs/jotai/pull/2801

## Summary
Uses atomState change and mount hooks to schedule effect and cleanup.

### Key Points
1. Introduces `atom.unstable_onInit` that fires in place when an atomState is first created in the store. `store` and `atomState` are passed as params to `atom.unstable_onInit`.
2. Uses `atomState.h` that fires in-place when an atom has mounted or unmounted, and if an atom has unmounted it schedules a cleanup to run with batch priority 0.5. Batch is a param passed to `atomState.h` function.
    - Related: https://github.com/pmndrs/jotai/pull/2895
4. Uses `atomState.u` that fires in-place when an atom has updated, and if an atom has updated it schedules an effect to run with batch priority 0.5. Batch is a param passed to `atomState.u` function.
    - Related: https://github.com/pmndrs/jotai/pull/2895
5. Makes batch and iterable to make flushBatch evaluation agnostic to its shape.
    - Related: https://github.com/pmndrs/jotai/pull/2898

## Check List


src/vanilla/store.ts
```diff
@@ -100,6 +100,10 @@ type AtomState<Value = AnyValue> = {
   n: number
   /** Object to store mounted state of the atom. */
   m?: Mounted // only available if the atom is mounted
+  /** Listener to notify when the atom value is updated. */
+  u?: (batch: Batch) => void
+  /** Listener to notify when the atom is mounted or unmounted. */
+  h?: (batch: Batch) => void
```
dmaskasky, line 106: TODO: Probably should revisit a hooks namespace
```ts
/** Atom hooks */
h?: {
  /** Listener to notify when the atom value is updated. */
  u?: (batch: Batch) => void
  /** Listener to notify when the atom is mounted or unmounted. */
  m?: (batch: Batch) => void
}
```

src/vanilla/store.ts
```diff
@@ -165,23 +169,27 @@ const addDependency = <Value>(
...
   /** High priority functions */
   H: Set<() => void>
   /** Medium priority functions */
   M: Set<() => void>
   /** Low priority functions */
   L: Set<() => void>
-}>
+  /** Atom dependents map */
+  D: Map<AnyAtom, Set<AnyAtom>>
+}
```
dmaskasky, line 186: Having an explicitly defined `H` allows us to schedule a new channel `syncEffectChannel` after `H`.

src/vanilla/store.ts
```diff
@@ -198,9 +206,11 @@ const registerBatchAtom = (
 ) => {
   if (!batch.D.has(atom)) {
     batch.D.set(atom, new Set())
-    addBatchFunc(batch, 'M', () => {
+    const scheduleListeners = () => {
+      atomState.u?.(batch)
       atomState.m?.l.forEach((listener) => addBatchFunc(batch, 'M', listener))
-    })
+    }
+    addBatchFunc(batch, 'H', scheduleListeners)
```
dmaskasky, line 213: Change priority to `H` to fire `u` as early as possible.

src/vanilla/store.ts
dai-shi: The reason we use the single character is to reduce bundle size, so using the namespace is not acceptable.

src/vanilla/atom.ts
```diff
@@ -47,6 +49,11 @@ export interface Atom<Value> {
    * @private
    */
   debugPrivate?: boolean
+  /**
+   * Fires after atom is referenced by the store for the first time
+   * For advanced use only and subject to change without notice.
+   */
+  unstable_onInit?: (store: Store, atomState: AtomState) => void
```
dai-shi, line 56: I still want to avoid atomState here. Let's use `unstable_derive`.

src/vanilla/store.ts
```diff
@@ -84,7 +84,7 @@ type Mounted = {
  * Mutable atom state,
  * tracked for both mounted and unmounted atoms in a store.
  */
-type AtomState<Value = AnyValue> = {
+export type AtomState<Value = AnyValue> = {
```
dai-shi, line 87: We keep it secret in this file.
```suggestion
type AtomState<Value = AnyValue> = {
```

src/vanilla/store.ts
```diff
@@ -165,23 +169,21 @@ const addDependency = <Value>(
...
-type Batch = Readonly<{
-  /** Atom dependents map */
-  D: Map<AnyAtom, Set<AnyAtom>>
+type Batch = Set<() => void>[] & {
   /** High priority functions */
   H: Set<() => void>
   /** Medium priority functions */
   M: Set<() => void>
   /** Low priority functions */
   L: Set<() => void>
```
dai-shi, line 178: If we were to introduce `atomState.us`, I don't think we need queue array. H/M/L would be probably more performant. If not, it still feels more readable. But, if queue array is more performant, I would change my mind.

src/vanilla/store.ts
dai-shi: Doesn't it make `.u` to run too early?

tests/vanilla/atomSyncEffect.ts
```diff
@@ -0,0 +1,186 @@
...
+  isRecursing: number
+  isRefreshing: number
+  init: (get: Getter) => void
+  refresh: () => void
+  cleanup?: Cleanup | null
+  epoch: number
+  error?: unknown
+}
+
+export function atomSyncEffect(effect: Effect): Atom<void> {
```
dai-shi, line 30: Can we keep this in `effect.test.ts`?

src/vanilla/store.ts
dmaskasky: Bundle size with queue array is smaller, and the performance is the same. A queue array is nice because it makes flush evaluation agnostic to the shape of batch.

tests/vanilla/atomSyncEffect.ts
dmaskasky: Yes, I just have test for puzzle 3 right now. This PR is not ready to merge until all jotai-effect tests are passing.

src/vanilla/store.ts
dmaskasky: actually `u` is the in-place hook so, we don't even want to fire inside scheduleListeners. I'll update the code to reflect this.

tests/vanilla/atomSyncEffect.ts
dmaskasky: I'm keeping separate for now while I have this PR in draft. I still have to remove debugLabels, labeled atomStates, and console.logs as well.

dai-shi:
Please reopen a new PR with https://github.com/pmndrs/jotai/tree/partial-sync-effect-4 branch. I will add my suggestions there. So, please avoid force pushing remote branches in the jotai repo.

PR #2903, opened 2025-01-03
feat(core): new dev (rev5) methods

dai-shi:
This introduces rev5 dev methods (prefixed `dev5_`)
The biggest change is it's no longer enabled by default, and we need to explicitly call `unstable_deriveDevStoreRev5`.

The goal is to isolate dev methods to be tree-shakable.
This also opens up using dev methods in the production build, which is sometimes requested.

Plan:
- rev5 will be introduced in v2.11.x
- rev4 will be deprecated in v2.11.y
- rev4 will be removed in v2.12.0 (or later)

@arjunvegda Does it sound okay?

### DIFF
--- a/src/vanilla.ts
+++ b/src/vanilla.ts
@@ -1,7 +1,11 @@
 export { atom } from './vanilla/atom.ts'
 export type { Atom, WritableAtom, PrimitiveAtom } from './vanilla/atom.ts'
 
-export { createStore, getDefaultStore } from './vanilla/store.ts'
+export {
+  createStore,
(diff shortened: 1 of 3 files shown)


dai-shi:
I may revert #2895 if an experiment goes well. So, meanwhile this is a draft.

dai-shi:
#2895 will stay, but I plan to do a refactor, so this is on hold for now.

dai-shi:
I will start over this after #2911.
Probably instead of providing dev methods in store.ts, we show the use case to do in tests.

PR #2904, opened 2025-01-03
chore(.github/workflows): change 'node-version' to 'lts/*' in 'actions/setup-node'

sukvvon:
## Summary

* Same with https://github.com/pmndrs/zustand/pull/2933

## Check List

### DIFF
--- a/.github/workflows/ecosystem-ci.yml
+++ b/.github/workflows/ecosystem-ci.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: latest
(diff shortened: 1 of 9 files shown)


.github/workflows/ecosystem-ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: 'latest'
```
dai-shi, line 18: I would like to change this to `lts/*` when 22 becomes 22, so let's leave it here.
```suggestion
          node-version: 22
```

dai-shi (approved):

PR #2905, opened 2025-01-04
feat(core): introduce atom.unstable_onInit hook

dmaskasky:
## Related PRs
https://github.com/pmndrs/jotai/pull/2888
https://github.com/pmndrs/jotai/pull/2801
https://github.com/pmndrs/jotai/pull/2901

## Summary
Uses atomState change and mount hooks to schedule effect and cleanup.

### Key Points
1. Introduces `atom.unstable_onInit` (`atom.unstable_onInit = (store: Store) => void`) that fires in place when an atomState is first created in the store for that atom. `store` is passed as params to `atom.unstable_onInit`.
2. Updates `store.unstable_derive` api to add a new callback param, `atomOnInit`. This is the interceptor for the `atom.unstable_onInit` api. Since getAtomState is where `atom.unstable_onInit` is called, we must add `atomOnInit` to the getAtomState params. `atomOnInit` accepts a store arg so that the current store is always passed.
3. Uses `atomState.h` that fires in-place when an atom has mounted or unmounted, and if an atom has unmounted it schedules a cleanup to run with batch priority 0.5. Batch is a param passed to `atomState.h` function.
    - Related: https://github.com/pmndrs/jotai/pull/2895
4. Uses `atomState.u` that fires in-place when an atom has updated, and if an atom has updated it schedules an effect to run with batch priority 0.5. Batch is a param passed to `atomState.u` function.
    - Related: https://github.com/pmndrs/jotai/pull/2895
5. Makes batch and iterable to make flushBatch evaluation agnostic to its shape.
    - Related: https://github.com/pmndrs/jotai/pull/2898

## Check List


src/vanilla/store.ts
```diff
@@ -100,8 +100,10 @@ type AtomState<Value = AnyValue> = {
   n: number
   /** Object to store mounted state of the atom. */
   m?: Mounted // only available if the atom is mounted
+  /** Listener to notify when the atom value is updated. */
+  u?: (batch: Batch) => void
```
dmaskasky, line 104: batch is used to schedule sync effects between priority 0 and priority 1.

src/vanilla/store.ts
```diff
@@ -100,8 +100,10 @@ type AtomState<Value = AnyValue> = {
...
   /** Listener to notify when the atom is mounted or unmounted. */
-  h?: () => void
+  h?: (batch: Batch) => void
```
dmaskasky, line 106: batch is used to schedule cleanup on atom unmount between priority 0 and priority 1.

src/vanilla/store.ts
```diff
@@ -174,28 +176,25 @@ type BatchPriority =
...
-    [BATCH_PRIORITY_HIGH]: Set<() => void>
-    [BATCH_PRIORITY_MEDIUM]: Set<() => void>
-    [BATCH_PRIORITY_LOW]: Set<() => void>
-  }
-  batch[BATCH_PRIORITY_HIGH] = new Set()
-  batch[BATCH_PRIORITY_MEDIUM] = new Set()
-  batch[BATCH_PRIORITY_LOW] = new Set()
-  return batch
+  const [H, M, L] = [new Set(), new Set(), new Set()]
+  return Object.assign([H, M, L], {
```
dmaskasky, line 192: batch is an array that is extended with properties `D`, `[BATCH_PRIORITY_HIGH]`, `[BATCH_PRIORITY_MEDIUM]`, and `[BATCH_PRIORITY_LOW]`.

This is useful so that sync effect can splice a queue between `BATCH_PRIORITY_HIGH` and `BATCH_PRIORITY_MEDIUM`. All internal functions reference the object properties instead of the array indicies so that splice does not break operation.

src/vanilla/store.ts
```diff
@@ -213,6 +212,7 @@ const registerBatchAtom = (
 ) => {
   if (!batch.D.has(atom)) {
     batch.D.set(atom, new Set())
+    atomState.u?.(batch)
```
dmaskasky, line 215: call atomState update hook synchronously in-place.

src/vanilla/store.ts
```diff
@@ -644,7 +666,7 @@ const buildStore = (
         d: new Set(atomState.d.keys()),
         t: new Set(),
       }
-      atomState.h?.()
+      atomState.h?.(batch)
```
dmaskasky, line 669: When atomState mounts, call atomState update hook `h` synchronously in-place

src/vanilla/store.ts
```diff
@@ -687,18 +709,18 @@ const buildStore = (
...
     ) {
       // unmount self
       const onUnmount = atomState.m.u
       if (onUnmount) {
-        addBatchFunc(batch, 'L', () => onUnmount(batch))
+        addBatchFunc(batch, BATCH_PRIORITY_LOW, () => onUnmount(batch))
       }
       delete atomState.m
-      atomState.h?.()
+      atomState.h?.(batch)
```
dmaskasky, line 720: When atomState unmounts, call atomState update hook `h` synchronously in-place

src/vanilla/store.ts
```diff
@@ -737,41 +759,37 @@ const deriveDevStoreRev4 = (store: Store): Store & DevStoreRev4 => {
...
+      return newAtomState
+    }
+    storeArgs[3] = (atom, getter, setter, ...args) => {
+      if (inRestoreAtom) {
+        return setter(atom, ...args)
+      }
+      return atomWrite(atom, getter, setter, ...args)
+    }
+    return storeArgs
+  })
```
dmaskasky, line 785: More concise implementation of createDevStore

src/vanilla/store.ts
```diff
@@ -687,18 +709,18 @@ const buildStore = (
...
       atomState.m &&
       !atomState.m.l.size &&
-      !Array.from(atomState.m.t).some((a) => getAtomState(a).m?.d.has(atom))
+      !Array.from(atomState.m.t).some((a) => ensureAtomState(a).m?.d.has(atom))
     ) {
       // unmount self
       const onUnmount = atomState.m.u
       if (onUnmount) {
-        addBatchFunc(batch, 'L', () => onUnmount(batch))
+        addBatchFunc(batch, BATCH_PRIORITY_LOW, () => onUnmount(batch))
```
dai-shi, line 717: Oh, it's a leftover of mine!

src/vanilla/store.ts
dai-shi: I don't like this complexity in the core. Let me use my version in `-5` branch. Or, if you don't like a sparse array, I can try something else.

dai-shi: On second thought, I don't mean I'm a big fan of the sparse array. As we assume 'M' being something anyways, let's just assume the index. It feels fragile, but it has been as well. If it causes a real issue, I'd consider something new.

src/vanilla/store.ts
dmaskasky: Its a little nicer that its self-documenting, but not critical. We can keep it if you want to.

dai-shi: I mean it's good. It minifies well too.

src/vanilla/store.ts
dmaskasky: We use the `M` property to reference the queue, not index `1`. The implementation is hardened to splice. This is a declarative approach to registering sync effects with the strong guarantee that it runs after H and before M.

dmaskasky: No, I am not a fan of the sparse array.

src/vanilla/store.ts
dmaskasky: I believe this is what we have in main currently. I just rebased my branch off of main.

src/vanilla/store.ts
dai-shi: > We use the M property to reference the queue, not index 1. 

That doesn't solve anything. I'd avoid increasing the bundle size.

src/vanilla/store.ts
dai-shi: Yeah, that's why I said it's a leftover of mine, which should have been in main.

src/vanilla/store.ts
```diff
@@ -165,45 +167,21 @@ const addDependency = <Value>(
...
 
+const createBatch = (): Batch => Object.assign([], { D: new Map() })
+
 const addBatchFunc = (
   batch: Batch,
   priority: BatchPriority,
   fn: () => void,
 ) => {
-  batch[priority].add(fn)
+  ;(batch[priority] ||= new Set()).add(fn)
```
dai-shi, line 184: This is good for a sparse array, but we only use 0,1,2. So, we should make sure that the array is dense. Otherwise, `batch.some((channel) => channel.size)` can cause a runtime error, theoretically.

src/vanilla/store.ts
```diff
@@ -302,9 +282,28 @@ type Store = {
...
+    atomOnMount,
+  ] = storeArgs
+  const ensureAtomState = <Value>(atom: Atom<Value>) => {
+    if (import.meta.env?.MODE !== 'production' && !atom) {
+      throw new Error('Atom is undefined or null')
+    }
+    let atomState = getAtomState(atom)
+    if (!atomState) {
+      atomState = { d: new Map(), p: new Set(), n: 0 }
+      atomState = setAtomState(atom, atomState)
```
dai-shi, line 301: I'm not super happy with the hack I introduced. Until we find a real use case, let me revert it.

src/vanilla/store.ts
```diff
@@ -265,7 +244,8 @@ const flushBatch = (batch: Batch) => {
 
 // internal & unstable type
 type StoreArgs = readonly [
-  getAtomState: <Value>(atom: Atom<Value>) => AtomState<Value>,
+  getAtomState: <Value>(atom: Atom<Value>) => AtomState<Value> | undefined,
+  setAtomState: <Value>(atom: Atom<Value>, atomState: AtomState<Value>) => void,
```
dai-shi, line 248: I'm not 100% sure if we need this change. I'll reconsider it in the follow-up refactor.

src/vanilla/store.ts
dmaskasky: Fyi, some skips uninitialized (empty) array indicies.

But I'm happy to refactor. Dense array code is simplier with fewer edge cases.

src/vanilla/store.ts
dmaskasky: Are you referring to setAtomState? What should we do instead?

src/vanilla/store.ts
dai-shi: > Fyi, some skips uninitialized (empty) array indicies.

Oh, yes. I actually find that out when I tried a sparse array in `-5`.
But, how about `for-of`?

src/vanilla/store.ts
dai-shi: ab1a77b did it. The reason of the hack was for proxyAtomState, which is already gone.

dai-shi (approved):
I believe `unstable_onInit`, `.h(batch)` and `.u(batch)` are all good. They are all internal things, so let's change them later if something doesn't work.

One thing I couldn't solve is the cleanup timing on unmount. Feel free to tackle it separately.

dai-shi (approved):
🎉 
Thanks tons for your contribution!

PR #2907, opened 2025-01-04
fix: derived atom is not recomputed after its dependencies changed #2906

dmaskasky:
https://github.com/pmndrs/jotai/discussions/2906

## Summary
When a writable atom sets two different atoms, and another atom indirectly depends on both, the dependent atom may fail to update correctly. Specifically, if the first atom's update results in no actual change, the dependent atom's "dirty bit" gets cleared prematurely. Then, when the second `finishRecompute` executes, the dependent atom does not recompute because its dirty bit was already cleared.

The proposed solution in this PR is to consolidate all `finishRecompute` operations into a single process, thereby handling changes to the atom graph in one go.

## Check List


src/vanilla/store.ts
```diff
@@ -174,16 +174,18 @@ type BatchPriority =
...
   | typeof BATCH_PRIORITY_LOW
 
-type Batch = Readonly<{
+type Batch = {
   /** Atom dependents map */
   D: Map<AnyAtom, Set<AnyAtom>>
+  /** Recompute dependents function */
+  R?: (batch: Batch) => void
   /** High priority functions */
   [BATCH_PRIORITY_HIGH]: Set<() => void>
```
dmaskasky, line 183: Nothing is placed in high priority anymore, but it would be a good place for _certain functions_.

dai-shi:
Thanks for working on it! I'll have a deeper look later.

dai-shi:
This does seem to fix the issue, but as you might expected it, I'm not a big fan of this change. (And, I didn't like `.D` already, so I want to find alternative way in the future.) I'll look for some other solutions. Give me some time.

dmaskasky:
> This does seem to fix the issue, but as you might expected it, I'm not a big fan of this change. (And, I didn't like .D already, so I want to find alternative way in the future.) I'll look for some other solutions. Give me some time.

Good point. I've refactored the approach so that we are no longer adding props to batch. Please have a second look. Hopefully you like this more.

dai-shi (approved):
Sorry for taking so much time.

PR #2910, opened 2025-01-06
name devStore functions

dmaskasky:
## Summary
Name dev store functions.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -116,8 +116,6 @@ type AtomState<Value = AnyValue> = {
   e?: AnyError
   /** Indicates that the atom value has been changed */
   x?: true
-  /** Debug label for the atom. */
-  _?: string | undefined
 }
 
 const isAtomStateInitialized = <Value>(atomState: AtomState<Value>) =>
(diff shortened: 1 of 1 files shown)


src/vanilla/store.ts
```diff
@@ -116,6 +116,8 @@ type AtomState<Value = AnyValue> = {
   e?: AnyError
   /** Indicates that the atom value has been changed */
   x?: true
+  /** Debug label for the atom. */
+  _?: string | undefined
```
dai-shi, line 120: As we are not using this, we don't need to define type here. Just extend it.

dai-shi: Also,`_` prefix means "unused" in the entire codebase, so it's confusing.

src/vanilla/store.ts
```diff
@@ -761,8 +763,9 @@ const deriveDevStoreRev4 = (store: Store): Store & DevStoreRev4 => {
           debugMountedAtoms.delete(atom)
         }
       }
+      atomState._ = atom.debugLabel
```
dai-shi, line 766: This feels too specific. Ordinary users don't debug atomState. You want to create your own devStore for this purpose.

dai-shi (commented):
Unfortunately, I don't feel like merging this. named functions instead of anonymous functions are okay, if that helps.

dai-shi (approved):
I generally prefer anonymous functions for minification, but this is good because it's "dev" store.

PR #2911, opened 2025-01-06
feat(core): expose internals instead of unstable_derive

dai-shi:
Following #2905, I think we could provide some internal functions because `unstable_derive` turns out to be somewhat limited.


dmaskasky:
Wow, this is incredibly powerful and flexible.

I appreciate that it doesn't allow direct mutation of an existing store's internal functions, yet provides all the necessary tools to build a new store if you know what you’re doing.

This might be overkill for jotai-scope, but it’s reassuring to have this flexibility available. My only concern is that exposing more of the API can be daunting to anyone who looks under the hood. It might be preferable to group these exports under a single `INTERNAL_`-prefixed object—or otherwise encapsulate them—to improve clarity and reduce the risk of overwhelming new users.

dai-shi:
> It might be preferable to group these exports under a single INTERNAL_-prefixed object

I don't like objects, as they are not minifyable. Let me try a namespace.

dai-shi:
I think this should come with a new minor.

dai-shi:
It's unexpected to me. I thought only jotai-effect would fail.

@arjunvegda Can you check how jotai-devtools fails?

arjunvegda:
> It's unexpected to me. I thought only jotai-effect would fail.
> 
> @arjunvegda Can you check how jotai-devtools fails?

Ah, good catch! This was a bug in devtools. This PR [fixes it](https://github.com/jotaijs/jotai-devtools/pull/174). I'll merge it by EOD Sunday to give you a chance to review.

src/vanilla/internals.ts
```diff
@@ -0,0 +1,748 @@
...
+      }
+      if (visiting.has(a)) {
+        // The algorithm calls for pushing onto the front of the list. For
+        // performance, we will simply push onto the end, and then will iterate in
+        // reverse order later.
+        if (invalidatedAtoms.get(a) === aState.n) {
+          topSortedReversed.push([a, aState, aState.n])
+        } else {
+          invalidatedAtoms.delete(a)
+        }
```
dmaskasky, line 518: I replaced this code with the code below and ran all tests. We don't have a test where invalidatedAtoms.has(a) is ever true. 
```ts
        } else if (invalidatedAtoms.has(a)) {
          throw
        }
```

dai-shi: Nice catch!

dai-shi:
My bad. I should have run it in #2967.

dmaskasky (approved):
This is a huge improvement to Jotai's versatility that will unlock amazing new capability in the jotai ecosystem. Thank you!!!

PR #2912, opened 2025-01-06
replace batch dependents map with batched changed atoms

dmaskasky:
## Summary
Batch.D was previously `Map<Atom, Set<AtomDependent>>`.
This allowed atom dependents to recompute without being mounted. There isn't an obvious reason to do so.
All tests pass after removing this functionality.

## Dropped behavior
After this change, the following test would fail.
I don't think we should be update dependents of unmounted atoms.
```ts
it('recomputes dependents of unmounted atoms', () => {
  const a = atom(0)
  const bRead = vi.fn((get) => get(a))
  const b = atom(bRead)
  const c = atom((get) => get(b))
    set(a, 1) // batch.D.set(a, new Set())
    get(c) // batch.D.get(a).add(b)
    set(a, 2) // topSortedReversed: [b, a]
    bRead.mockClear()
    // after this write function completes, flushBatch will recompute `a` and `b`
    // note: a and b are not mounted
  })
  const store = createStore()
  store.set(w)
  expect(bRead).toHaveBeenCalled()
})
```

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -180,11 +180,11 @@ type Batch = [
   priority2: Set<() => void>,
 ] & {
   /** changed Atoms */
-  D: Set<AnyAtom>
+  C: Set<AnyAtom>
 }
 
 const createBatch = (): Batch =>
-  Object.assign([new Set(), new Set(), new Set()], { D: new Set() }) as Batch
+  Object.assign([new Set(), new Set(), new Set()], { C: new Set() }) as Batch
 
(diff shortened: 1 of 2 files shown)


dai-shi:
wow, this looks cool. I will have a closer look later.

dai-shi (approved):
Oh, yes! This is something I wan't comfortable with. Thanks for fixing it. This goes towards what I have in mind. I'd like to go further.

Is this now possible because of `.x`?

Anyway, merging this.

PR #2918, opened 2025-01-07
refactor(core): use map for batch

dai-shi:
#2912 was a very good direction as I don't want to have domain knowledge in "batch".

This allows the complete isolation, and feels less hacky.


dai-shi:
closing in favor of #2925

PR #2919, opened 2025-01-07
alternate approach for #2907

dai-shi:
I want a minimal change for the fix. Refactors should be done separately with more discussions.

Changes: https://github.com/pmndrs/jotai/compare/main...fix-derived-recompute-not-changing-2


dai-shi:
Closing in favor of #2925.

PR #2920, opened 2025-01-08
fix: should update dependents with the value of the unwrapped atom when the promise resolves

dmaskasky:
https://github.com/pmndrs/jotai/discussions/2914

## Summary
Caused by https://github.com/pmndrs/jotai/commit/34b4f8e3b810e2898a1b1c6b3744ff06e4b12a42 and https://github.com/pmndrs/jotai/commit/e777f469586d685d864d9120274e83db4fbad4c8

### Fix
1. Mount and unmount of dependencies added synchronously.
2. Do not mount or unmount of dependencies added asynchronously.
3. Mount and unmount dependencies added asynchronously when the returned promise resolves.

### Implementation
atomState.m.q is a Set of all atoms added async. Atoms added async with getter in readAtomState are added to this set. This Set is cleared when a new atom value is set and after recomputeDependents has completed. Only unmount dependencies not present in this Set.

Example:
```ts
const f = atom((get) => {
  // allow mounting synchronous atoms immediately
  if (!get(a)) {
    get(b)
  } else {
    get(c)
  }
    // mount this once promise resolves and don't unmount
    if (!get(a)) {
      get(d)
    } else {
      get(e)
    }
  })
})

// mount a, b
store.sub(f, () => {})
// mount d, e
await Promise.resolve()
// unmount b, mount c
store.set(a, 1)
// unmount d, mount e
await Promise.resolve()
```

## Check List


tests/vanilla/utils/unwrap.test.ts
```diff
@@ -150,3 +150,17 @@ describe('unwrap', () => {
...
+  const target = unwrap(asyncTarget)
+  const results: string[] = []
+  const derived = atom(async (get) => {
+    await Promise.resolve()
+    results.push('effect ' + get(target))
+  })
+  store.sub(derived, () => {})
+  await new Promise((r) => setTimeout(r))
+  expect(results).toEqual(['effect undefined', 'effect value'])
+})
```
dai-shi, line 166: Looks good. So, it's a regression, right? My bad... I'll see what I can do with the impl. It will be step by step and take a bit of time. Sorry folks for waiting.

src/vanilla/store.ts
```diff
@@ -414,6 +417,11 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
       const valueOrPromise = atomRead(atom, getter, options as never)
       setAtomStateValueOrPromise(atom, atomState, valueOrPromise)
       if (isPromiseLike(valueOrPromise)) {
+        if (batch) {
+          addBatchFunc(batch, 0, () => atomState.m?.q.clear())
+        } else {
+          atomState.m?.q.clear()
+        }
```
dmaskasky, line 424: TODO: How should I do this in the refactor?

dai-shi:
Continue to #2936.

dai-shi:
I change my mind. Let me use this branch, which is easier for me to push changes.

tests/vanilla/dependency.test.tsx
```diff
@@ -94,7 +94,7 @@ it('correctly handles the same promise being returned twice from an atom getter
   await expect(store.get(derivedAtom)).resolves.toBe('Asynchronous Data')
 })
 
-it('keeps atoms mounted between recalculations', async () => {
+it('keeps sync atoms mounted between recalculations', async () => {
```
dai-shi, line 97: This is the planned breaking change in v2.12.0.

@yuneco Will you be able to check if this breaks your app?

`npm i https://pkg.csb.dev/pmndrs/jotai/commit/b4bb8a8d/jotai`

dai-shi:
Split to #2940.

PR #2925, opened 2025-01-10
refactor: eliminate batch

dai-shi:
## Background

Batch (formerly Pending) was introduced a long time ago. The major purpose was to provide an async context (Like AsyncLocalStorage) to hold state concurrently. It was really necessary at some point (before v1?), but after v2 or so, the purpose has been more for organizing store.ts code so that atoms are isolated. Like the store doesn't know about all atoms. However as we introduced #2827 and other PRs, the store now coordinates atoms as the single source of truth, and we had to tweak Batch. #2905 went further. I expected that I could change it towards the designed shape, but #2919 and #2918 didn't work.

## Moving forward

If we use the store as the single source of truth, I don't think we need batch. Interestingly, I just noticed that `batch` isn't used in readAtomState at all. We are just passing it recursively without consuming it. I don't remember when this has changed. This PR is only to eliminate batch, but I have a big picture in mind which is close to #2911, but I'll explore a bit more. The end goal is probably providing a bunch of hooks, so that the ecosystem libraries can do anything on their own risks. There should be more room to minimize the store code, but that will be tackled afterwards.


dai-shi:
Wait, I might notice some issues. I will continue working on it. Until then, this isn't ready.

src/vanilla/store.ts
```diff
@@ -458,11 +415,22 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
...
-    batch: Batch,
-    atom: Atom<Value>,
-    atomState: AtomState<Value>,
-  ) => {
+  const invalidateDependents = <Value>(atomState: AtomState<Value>) => {
+    const visited = new WeakSet<AtomState>()
+    const stack: AtomState[] = [atomState]
+    while (stack.length) {
+      const aState = stack.pop()!
+      if (!visited.has(aState)) {
```
dmaskasky, line 423:
```suggestion
      if (!visited.has(aState) && !aState.x) {
```

src/vanilla/store.ts
```diff
@@ -489,17 +457,17 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
         // The algorithm calls for pushing onto the front of the list. For
         // performance, we will simply push onto the end, and then will iterate in
         // reverse order later.
-        topSortedReversed.push([a, aState, aState.n])
+        if (aState.x) {
+          topSortedReversed.push([a, aState, aState.n])
+        }
```
dmaskasky, line 462: Good!

src/vanilla/store.ts
```diff
@@ -473,12 +441,12 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
       atomState: AtomState,
       epochNumber: number,
     ][] = []
-    const visiting = new Set<AnyAtom>()
-    const visited = new Set<AnyAtom>()
+    const visiting = new WeakSet<AnyAtom>()
+    const visited = new WeakSet<AnyAtom>()
```
dmaskasky, line 445: Good!

src/vanilla/store.ts
```diff
@@ -297,6 +233,34 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
...
+    }
+    do {
+      ;(store as any)[INTERNAL_flushStoreHook]?.()
+      changedAtoms.forEach((atomState) => atomState.m?.l.forEach(call))
+      changedAtoms.clear()
+      unmountCallbacks.forEach(call)
+      unmountCallbacks.clear()
+      mountCallbacks.forEach(call)
+      mountCallbacks.clear()
+      recomputeInvalidatedAtoms()
```
dmaskasky, line 257:
```suggestion
      recomputeInvalidatedAtoms()
      ;(store as any)[INTERNAL_flushStoreHook]?.()
      recomputeInvalidatedAtoms()
      changedAtoms.forEach((atomState) => atomState.m?.l.forEach(call))
      changedAtoms.clear()
      unmountCallbacks.forEach(call)
      unmountCallbacks.clear()
      mountCallbacks.forEach(call)
      mountCallbacks.clear()
      recomputeInvalidatedAtoms()
```

src/vanilla/store.ts
```diff
@@ -297,6 +233,34 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
...
+
+  const flushCallbacks = () => {
+    const errors: unknown[] = []
+    const call = (fn: () => void) => {
+      try {
+        fn()
+      } catch (e) {
+        errors.push(e)
+      }
+    }
```
dmaskasky, line 248:
```suggestion
    }
    if (!changedAtoms) return
```

dmaskasky:
Beautiful PR. This is the next evolution of the store code. I had a feeling the store was getting quite bloated with all the batch logic, I'm glad you figured this out.

src/vanilla/store.ts
dmaskasky, line 462: Good.

dmaskasky, line 462:
```suggestion
        topSortedReversed.push([a, aState, aState.n])
```

src/vanilla/store.ts
```diff
@@ -489,17 +457,17 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
...
-        // Mark atom dirty
-        aState.x = true
         stack.pop()
         continue
       }
       visiting.add(a)
       // Push unvisited dependents onto the stack
-      for (const [d, s] of getMountedOrBatchDependents(aState)) {
+      for (const [d, s] of getMountedOrPendingDependents(aState)) {
         if (a !== d && !visiting.has(d)) {
```
dmaskasky, line 471:
```suggestion
        if (a !== d && !visiting.has(d) && s.x) {
```

dmaskasky, line 471:
```suggestion
        if (!visiting.has(d)) {
```

src/vanilla/store.ts
```diff
@@ -508,39 +476,33 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
...
-      const changedAtoms = new Set<AnyAtom>([atom])
-      for (let i = topSortedReversed.length - 1; i >= 0; --i) {
-        const [a, aState, prevEpochNumber] = topSortedReversed[i]!
-        let hasChangedDeps = false
-        for (const dep of aState.d.keys()) {
-          if (dep !== a && changedAtoms.has(dep)) {
-            hasChangedDeps = true
-            break
-          }
+    for (let i = topSortedReversed.length - 1; i >= 0; --i) {
```
dmaskasky, line 479: nit: can we rename `topSortedReversed` to `invalidatedAtoms`?

dmaskasky:
Can you clarify the semantic difference between?
- `atomState.n === n`
- `atomState.x`
- `changedAtoms.has(atom)`

src/vanilla/store.ts
dai-shi: Hm, I prefer the current name, `invalidatedAtoms` would be good if we didn't care the order.

dai-shi:
> Could we replace `++atomState.n` with `changedAtoms.add(atom)`?

Maybe not. `n` is also used for dependency checking. `changedAtom` is for `write` only. Though, I'm not saying what it shows now is ideal. As I noted, there would be many room for improvements, and this is only the first step.

src/vanilla/store.ts
```diff
@@ -269,12 +278,11 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
             recomputeInvalidatedAtoms()
           }
           ;(store as any)[INTERNAL_flushStoreHook]?.()
-          changedAtoms.forEach((atomState) => atomState.m?.l.forEach(call))
-          changedAtoms.clear()
-          unmountCallbacks.forEach(call)
-          unmountCallbacks.clear()
-          mountCallbacks.forEach(call)
-          mountCallbacks.clear()
+          clearThenForEach(changedAtoms, (atomState) =>
```
dmaskasky, line 281:
```suggestion
          const callbacks = [
            ...changedAtoms.map((atomState) => atomState.m?.l ?? []),
            ...unmountCallbacks,
            ...mountCallbacks,
          ]          
          callbacks.forEach(call)
          changedAtoms.clear()
          unmountCallbacks.clear()
          mountCallbacks.clear()
```

src/vanilla/store.ts
```diff
@@ -269,12 +278,11 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
...
+            atomState.m?.l.forEach(call),
+          )
+          clearThenForEach(unmountCallbacks, call)
+          clearThenForEach(mountCallbacks, call)
```
dmaskasky, line 285:
```suggestion
          const callbacks = [
            ...changedAtoms.map((atomState) => atomState.m?.l ?? []),
            ...unmountCallbacks,
            ...mountCallbacks,
          ]          
          callbacks.forEach(call)
          changedAtoms.clear()
          unmountCallbacks.clear()
          mountCallbacks.clear()
```

dai-shi: That's nice.

src/vanilla/store.ts
```diff
@@ -297,6 +241,56 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
     return atomState
   }
 
+  // These are store state.
+  // As they are not garbage collectable, they shouldn't be mutated during atom read.
+  const invalidatedAtoms = new WeakMap<AnyAtom, EpochNumber>()
+  const changedAtoms = new Map<AnyAtom, AtomState>()
+  const unmountCallbacks = new Set<() => void>()
+  const mountCallbacks = new Set<() => void>()
```
dmaskasky, line 249:
```suggestion
  const unmountedAtoms = new Map<AnyAtom, AtomState>()
  const mountedAtoms = new Map<AnyAtom, AtomState>()
```

dai-shi: Just for consistency?

dmaskasky: Yeah, but I'm not sure how useful it will be.

dai-shi: Yeah, I actually thought about this too. Let me hold off.

dai-shi:
Final status:
- We are happy with this simplified implementation.
- jotai-effect v1 needs to address this change later.
- This is fairly a big rewrite, but more is coming (like #2911). We need to communicate the status with users with release notes and announcements.

PR #2936, opened 2025-01-15
fix: should update dependents with the value of the unwrapped atom when the promise resolves

dmaskasky:

Fix #2914

## Summary

Caused by 34b4f8e3b810e2898a1b1c6b3744ff06e4b12a42 and e777f469586d685d864d9120274e83db4fbad4c8

It's caused by a subtle timing issue. The promise is resolved before the derived atom is mounted because the mount process runs after the promise is resolved.

There might be a better fix, but for now in this PR, it triggers updates again if it finds atom value change after the mount process. More refactors for store.ts are coming and we may revisit this in the future.

## Check List


dai-shi:
Thanks for cleaning it up. I'll have a closer look later.

> Caused by 34b4f8e and e777f46

Is this caused by the combination of them? It doesn't fail with only one of them, does it?

dai-shi:
This is such a fundamental thing... I'll take it over.

dai-shi (approved):

PR #2937, opened 2025-01-15
docs: fix bunshi example

disjukr:
## Summary

The scope of bunshi is not divided based on the provider in the render tree but rather based on the value of the provider, so using `initialCount` as the value of the scope is not appropriate.

If multiple providers provide the same `initialCount`, they will share the same scope.

Therefore, I have modified the example code to wrap the `initialCount` once so that it gets a different reference.

## Check List

### DIFF
--- a/docs/extensions/scope.mdx
+++ b/docs/extensions/scope.mdx
@@ -114,10 +114,10 @@ npm install bunshi
 import { atom, useAtom } from 'jotai'
 import { molecule, useMolecule, createScope, ScopeProvider } from 'bunshi/react'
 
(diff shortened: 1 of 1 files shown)


dai-shi:
Can @loganvolkers review this?

dai-shi:
> - [ ] `pnpm run fix:format` for formatting code and docs

@disjukr run fix:format.

disjukr:
> @disjukr run fix:format.

Oh, I thought I had run it, but I guess I was mistaken. I ran it again.

PR #2938, opened 2025-01-16
docs: add docs for bunja

disjukr:
## Summary

Bunja is new library like Bunshi but provides more guarantee about instance lifetime.

## Check List


dai-shi (changes requested):
Please add a new library under https://github.com/pmndrs/jotai/tree/main/docs/third-party.

docs/third-party/bunja.mdx
```diff
@@ -0,0 +1,225 @@
...
+  const [count, setCount] = useAtom(countAtom)
+  // Your component logic here
+}
+```
+
+### Defining a Bunja that relies on other Bunja
+
+If you want to manage a state with a broad lifetime and another state with a narrower lifetime, you can create a (narrower) bunja that depends on a (broader) bunja.
+
+For example, you can think of a bunja that manages the WebSocket connection and disconnection, and another bunja that subscribes to a specific resource over the connected WebSocket.
```
dai-shi, line 61: Would it be possible to provide a simpler example to show the dependency case in this second section?

disjukr: Yes, replaced it with a modal screen example. https://github.com/pmndrs/jotai/pull/2938/commits/afcb0d87e601cd4ebe01eca6f62166ac2d83a626

dai-shi (approved):

PR #2940, opened 2025-01-16
[Not For Merge] BREAKING: unmount async dependencies

dai-shi:
split from #2920 

I thought it's nice to fix memory leak test, but it might not worth the breaking behavior. 

We don't plan to merge this for now. We would like to collect more feedback on it, especially the failing test case not only in this repo, but also in the ecosystem. So, one of the purposes of this PR is to run jotai-ecosystem-ci to learn the effect of this breaking behavioral change.

---

## Proposed Change In Behavior
Atom dependencies of async atoms are unmounted and remounted each time the async atom recomputes if they are accessed only after a delay. This will cause any atom.onMount and unmount functions to fire each time.

**Example**
```ts
const countAtom = atom(0)
countAtom.onMount = () => {
  console.log('countAtom mounted')
  return () => {
    console.log('countAtom unmounted')
  }
}

const asyncAtom = atom(async (get) => {
  await Promise.resolve()
  get(countAtom)
})

const store = createStore()

const unsub = store.sub(asyncAtom, () => {}) // mounts asyncAtom
await Promise.resolve()

console.log('set countAtom to 1')
store.set(countAtom, 1)
await Promise.resolve()

console.log('set countAtom to 2')
store.set(countAtom, 2)
await Promise.resolve()

unsub()
```

**Current**
```
set countAtom to 1
set countAtom to 2
```

**New**
```
countAtom mounted
set countAtom to 1
countAtom unmounted
countAtom mounted
set countAtom to 2
countAtom unmounted
countAtom mounted
countAtom unmounted
```


dai-shi:
Let's revisit this when we plan v3.

PR #2941, opened 2025-01-17
add debugStore testUtil

dmaskasky:
## Summary
`createDebugStore` is a test util for creating a store that adds `label` to the atomState from the corresponding atom's debugLabel property. This is useful when debugging store behavior.

## Check List


dai-shi:
unstable_derive will be removed in the future. So, let me consider it in the future agin. It's a little bit annoying but you can use your own util locally without committing it.

PR #2942, opened 2025-01-17
chore(tests): use babel/plugin-debug-label

dai-shi:
To make our debugging life easier and also improve babel plugin.


dai-shi:
We will see how this goes. If we find some bugs in the babel plugin, it's actually good.

PR #2946, opened 2025-01-17
refactor(store): Replace runWithTransactions to flushCallbacks

dmaskasky:
## Summary
Current implementation with `runWithTransactions` uses a single synchronous callback (`fn`) invocation followed by conditionally running store hooks. For some flows, this code reduces to simply calling `fn`. The states in which hooks should be called are well defined so refactoring to synchronously calling a finalizer (`flushCallbacks`) is introduced in this PR.

A side-effect of this PR is it resolves several incompatibilities with jotai-effect.

### Details
`flushCallbacks` is called at the end of 7 flows:
1. writeAtom
2. writeAtomState setter async
3. subscribe
4. unsubscribe
5. readAtomState getter async
6. promise complete
7. onMount setAtom async

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -433,7 +433,7 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
     return dependents
   }
 
-  const invalidateDependents = <Value>(atomState: AtomState<Value>) => {
+  const invalidateDependents = (atomState: AtomState) => {
     const visited = new WeakSet<AtomState>()
     const stack: AtomState[] = [atomState]
     while (stack.length) {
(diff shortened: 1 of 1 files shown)


src/vanilla/store.ts
```diff
@@ -440,16 +433,18 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
...
     const visited = new WeakSet<AtomState>()
     const stack: AtomState[] = [atomState]
     while (stack.length) {
       const aState = stack.pop()!
       if (!visited.has(aState)) {
         visited.add(aState)
         for (const [d, s] of getMountedOrPendingDependents(aState)) {
-          invalidatedAtoms.set(d, s.n)
-          stack.push(s)
+          if (!invalidatedAtoms.has(d)) {
```
dai-shi, line 444: Can you create a test that fails without this condition?

dmaskasky: Its a performance optimization. Transitive, dependents of dependents of A are dependents of A.

dai-shi: Hm, if it's a performance optimization, it will be difficult to write a test.

dai-shi:
This looks almost good to me, but for my preference, #2950 is a follow-up PR.

PR #2961, opened 2025-01-27
docs: update import statement for atomWithRefresh

moonheekim0118:

Update the import statement in the documentation to correctly show where atomWithRefresh is imported from.

as-is `from 'XXX'`
to-be `from 'jotai/utils'`

## Check List


dai-shi (approved):
Thanks.

PR #2965, opened 2025-01-31
fix(store): recompute dependents on eager evaluation in writeAtomState

dmaskasky:
https://github.com/pmndrs/jotai/discussions/2964

## Summary
When writeAtomState on activateAction is called, setter is called on isActiveAtom which updates the value to true. isActiveAtom is marked as dirty along with all its dependents; since isActiveAtom is currently mounted, this includes activeCountAtom. Calling getter inside writeAtomState should cause an eager readAtomState on activeCountAtom. Since activeCountAtom is dirty, this should recursively reevaluate all dependencies. In this case, countAtom is conditional and is not yet a dependency of activeCountAtom. So the issue lies with not ~~doing a full readAtomState of activeCountAtom in which its read function is called again~~ calling recomputeDependents prior to calling readAtomState on activeCountAtom. Because the readAtomState was called on activeCountAtom, that atom and its dependencies are removed from dirty. Therefore the final recomputeDependents call from flushPending does not process those atoms.

~~I wonder if this was fixed in 2.11.2. There was a [similar issue](https://github.com/pmndrs/jotai/discussions/2959#discussion-7881285) reported that was caused by early returning from recursively marking dependents as dirty when the current atom was already dirty. That change was [reverted](https://github.com/pmndrs/jotai/pull/2960/files) in 2.11.2, so the full dependents tree is marked dirty whenever an atom is changed.~~ The issue is present in 2.11.2.

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -515,7 +515,10 @@ const buildStore = (...storeArgs: StoreArgs): Store => {
     ...args: Args
   ): Result => {
     let isSync = true
-    const getter: Getter = <V>(a: Atom<V>) => returnAtomValue(readAtomState(a))
+    const getter: Getter = <V>(a: Atom<V>) => {
+      recomputeInvalidatedAtoms()
+      return returnAtomValue(readAtomState(a))
+    }
     const setter: Setter = <V, As extends unknown[], R>(
       a: WritableAtom<V, As, R>,
       ...args: As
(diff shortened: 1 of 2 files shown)


dai-shi:
That's my mental model. It feels like there's slight room for improvement, but I can try that afterwards.

dai-shi (approved):
Thanks for working on it.
I hope we don't miss other simple cases, but I'm not very confident.

PR #2966, opened 2025-02-01
refactor(core): remove unnecessary code

dai-shi:
With #2965, L483 is no longer necessary. I should have noticed when reviewing.

Now, I wonder if we can write a failing test for this. It's probably unlikely.


dai-shi:
This is included in #2911.

PR #2968, opened 2025-02-02
experimental: refactor store hooks

dmaskasky:
## Summary
The current implementation of Store Hooks has a few issues.
1. When any atom changes, mounts, or unmounts, the callback back is called. The only way to listen for changes on an atom is for every callback to test the atom on every call. On large atom graphs, this has significant overhead.
2. The only way to register callbacks is to nest the original callback inside the new one. This makes unsubscribe tricky since callbacks deeper in the recursive tree would not have knowledge of the callbacks higher up the tree.

This PR proposes the following change to StoreHooks to address these issues.
```ts
type StoreHooks = {
  /**
   * Listeners to notify when the atom value is changed.
   * This is an experimental API.
   */
  c?: WeakMap<AnyAtom, Set<() => void>>
  /**
   * Listeners to notify when the atom is mounted.
   * This is an experimental API.
   */
  m?: WeakMap<AnyAtom, Set<() => void>>
  /**
   * Listeners to notify when the atom is unmounted.
   * This is an experimental API.
   */
  u?: WeakMap<AnyAtom, Set<() => void>>
  /**
   * Listeners to notify when callbacks are being flushed.
   * This is an experimental API.
   */
  f?: Set<() => void>
}
```

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -30,6 +30,15 @@ type Mounted = {
   u?: () => void
 }
 
+type AtomHooks = {
+  /** Listeners to notify when the atom value is changed. */
+  c?: Set<() => void>
+  /** Listeners to notify when the atom is mounted. */
+  m?: Set<() => void>
+  /** Listeners to notify when the atom is unmounted. */
+  u?: Set<() => void>
+}
+
 /**
  * Mutable atom state,
  * tracked for both mounted and unmounted atoms in a store.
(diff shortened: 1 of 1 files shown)


dai-shi:
> * When any atom changes, mounts, or unmounts, the callback back is called. The only way to listen for changes on an atom is for every callback to test the atom on every call. On large atom graphs, this has significant overhead.

Please create the weak map on the user end.

> * The only way to register callbacks is to nest the original callback inside the new one. This makes unsubscribe tricky since callbacks deeper in the recursive tree would not have knowledge of the callbacks higher up the tree.

That's my concern too. But, let's leave it for now. I think it's very unlikely to unsubscribe. Because I have another concern that now it exposes both store state and hooks, which I don't like very much.

PR #2970, opened 2025-02-03
storeHooks minor improvement

dmaskasky:
## Summary
Return delete function from add in storeHooks.
- improves ergonomics
- reduces bundle size

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -648,16 +648,13 @@ const unmountAtom = <Value>(
 
 type StoreHook = {
   (): void
-  add(callback: () => void): void
-  delete(callback: () => void): void
+  add(callback: () => void): () => void
 }
 
 type StoreHookForAtoms = {
   (atom: AnyAtom): void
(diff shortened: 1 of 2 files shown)


src/vanilla/internals.ts
```diff
@@ -690,9 +687,7 @@ const createStoreHook = (): StoreHook => {
   }
   notify.add = (fn: () => void) => {
     callbacks.add(fn)
-  }
-  notify.delete = (fn: () => void) => {
-    callbacks.delete(fn)
+    return () => void callbacks.delete(fn)
```
dai-shi, line 690: Prefer block:
```suggestion
    return () => {
      callbacks.delete(fn)
    }
```

dai-shi (approved):
LGTM otherwise

src/vanilla/internals.ts
dai-shi: terser handles it anyway. so, it's just a preference.
image

PR #2972, opened 2025-02-04
refactor(internals): store state in store args

dai-shi:
The reason we had `getAtomState` is for `unstable_derive` and to avoid having state. #2911 goes with a different approach and it has store state. ~We should have atomStateMap there too. Especially, having ensureAtomState function as store state seemed weird.~ Second take is to put everything in store args.

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -57,6 +57,11 @@ type AtomState<Value = AnyValue> = {
   e?: AnyError
 }
 
+type AtomStateMap = {
+  get(atom: AnyAtom): AtomState | undefined
+  set(atom: AnyAtom, atomState: AtomState): void
+}
+
 export type INTERNAL_AtomState<Value = AnyValue> = AtomState<Value>
 
 //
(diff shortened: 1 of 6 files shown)


src/vanilla/internals.ts
```diff
@@ -780,38 +788,26 @@ export const INTERNAL_getStoreStateRev1 = (store: unknown): StoreState =>
   (store as Store)[STORE_STATE]!
 
 export const INTERNAL_buildStore = (...storeArgs: StoreArgs): Store => {
```
dmaskasky, line 790: jotai-scope originally needed unstable_derive to reuse the atomState map from the parent store. AFAIK, that may still be necessary.

How about this?
```suggestion
export const INTERNAL_buildStore = (
  storeArgs: StoreArgs,
  [
    storeHooks = {},
    mountedAtoms = new WeakMap<AnyAtom, Mounted>(),
    invalidatedAtoms = new WeakMap<AnyAtom, EpochNumber>(),
    changedAtoms = new Map<AnyAtom, INTERNAL_AtomState>(),
    mountCallbacks = new Set<() => void>(),
    unmountCallbacks = new Set<() => void>(),
    atomStateMap = new WeakMap<AnyAtom, AtomState>(),
  ]: StoreStates
): Store => {
```

dai-shi: > I'm not sure but I think that may still be necessary.

It's still possible as we don't lose any capabilities, but I like your approach. Let me try something.

dai-shi: 73a394d I intentionally avoided having defaults for now.

dmaskasky (approved):
This is so clean. Very flexible, maintainable code. Thank you.

PR #2973, opened 2025-02-04
chore(*): remove duplicate 'test-build:spec', replace with 'test:spec'

sukvvon:
## Summary

* maintain consistency with Zustand and Valtio.
* Since `test:spec` and `test-build:spec` perform the same operation, i removed the duplicate in `package.json` and replaced it with `test:spec` within the CI.”

## Check List


dai-shi (approved):
Thanks. I think it's a leftover from #2641.

PR #2974, opened 2025-02-04
expose building blocks

dmaskasky:
## Summary
Expose buildingBlocks by passing in to buildStore.

Background: We sometimes want to use a custom function in place of a building block when creating a new store. This PR is one approach that enables this functionality.

## Check List


dai-shi:
Do you have any specific use cases in your mind?
I almost like this flexibility, except that this exposes `getStore` workaround ouside internals.ts. I wonder if there's a solution to it.

dmaskasky:
> Do you have any specific use cases in your mind? I almost like this flexibility, except that this exposes `getStore` workaround ouside internals.ts. I wonder if there's a solution to it.

Previously, we had the ability to build a store with a custom getAtomState, setAtomState. jotai-scope uses this when creating scoped stores.

With the new implementation, I think jotai-scope will likely need access to modify addtional functions such as flushPending and invalidateDependents to provide scope-specific behavior.

While its more work to build a custom store, I think its since these apis are INTERNAL anyway.

dai-shi:
I'm still not sure if jotai-scope needs it, but I would like to let you worry about it. So exposing them is fine. I'll tweak a little bit as I didn't like the role of storeArgs already.

dmaskasky:
> I'm still not sure if jotai-scope needs it, but I would like to let you worry about it. So exposing them is fine. I'll tweak a little bit as I didn't like the role of storeArgs already.

I should be able to remove dependence on unstable_is soon.

dai-shi:
> I'll tweak a little bit as I didn't like the role of storeArgs already.

b24ad70

dai-shi:
Let me take care of it.

dai-shi:
I think this is coming full circle, but as we removed getAtomState, it's more like a level-up spiral. You may not like the flat building blocks array, but this is my strong opinion.

PR #2977, opened 2025-02-06
full building blocks

dmaskasky:
## Summary
BUILDING_BLOCKS value on store is equal to the sparse array of building blocks passed in to buildStore. Return the full buildingBlocks array instead.

## Check List

### DIFF
--- a/tests/vanilla/internals.test.tsx
+++ b/tests/vanilla/internals.test.tsx
@@ -1,0 +1,16 @@
+import { describe } from 'node:test'
+import { expect, it } from 'vitest'
+import { createStore } from 'jotai'
+import { INTERNAL_getBuildingBlocksRev1 as INTERNAL_getBuildingBlocks } from 'jotai/vanilla/internals'
+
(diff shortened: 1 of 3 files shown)


dai-shi:
> the sparse array of building blocks

Oh, my god. I totally missed it. Thanks for the catch. Now, I prefer my original solution. Let me modify it.

tests/vanilla/internals.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { describe, expect, it } from 'vitest'
+import { createStore } from 'jotai'
+import {
+  INTERNAL_buildStoreRev1 as INTERNAL_buildStore,
+  INTERNAL_getBuildingBlocksRev1 as INTERNAL_getBuildingBlocks,
+} from 'jotai/vanilla/internals'
+
+describe('internals', () => {
+  it('should not return a sparse building blocks array', () => {
```
dai-shi, line 9: Interesting test.

PR #2985, opened 2025-02-15
chore: use default reporter in vitest and fix TS error

kretajak:
## Summary

Vitest 3 suggest to migrate from 'basic' reporter to 'default'

```
DEPRECATED  'basic' reporter is deprecated and will be removed in Vitest v3.
Remove 'basic' from 'reporters' option. To match 'basic' reporter 100%, use configuration:
{
  "test": {
    "reporters": [
      [
        "default",
        {
          "summary": false
        }
      ]
    ]
  }
}
```

Secondly, `provider` in `coverage` is required and produces TypeScript error when not defined explicitly.

## Check List

### DIFF
--- a/vitest.config.mts
+++ b/vitest.config.mts
@@ -30,11 +30,12 @@ export default defineConfig({
     globals: true,
     environment: 'jsdom',
     dir: 'tests',
-    reporters: 'basic',
+    reporters: process.env.GITHUB_ACTIONS ? ['default', 'github-actions'] : ['default'],
     setupFiles: ['tests/setup.ts'],
     coverage: {
       reporter: ['text', 'json', 'html', 'text-summary'],
       reportsDirectory: './coverage/',
+      provider: 'v8',
       include: ['src/**'],
     },
     onConsoleLog(log) {


vitest.config.mts
```diff
@@ -30,11 +30,12 @@ export default defineConfig({
     globals: true,
     environment: 'jsdom',
     dir: 'tests',
-    reporters: 'basic',
+    reporters: [['default', { summary: false }]],
```
kretajak, line 33: Other option is to remove reporters so it will make the use of `default` reporter. For example `valtio` has this line removed.

dai-shi: Thanks for the suggestion.
We want to make this file mostly the same across zustand, jotai and valtio.
What would be your suggestion then?

kretajak: Personally, I would go with:
```
reporters: process.env.GITHUB_ACTIONS ? ['default', 'github-actions'] : ['default']
```

`Vitest` has nice [build-in Github reporter ](https://vitest.dev/guide/reporters#github-actions-reporter).

dai-shi: Sounds good. Can you open PRs to zustand and valtio too and make three files consistent as much as possible?

kretajak: PR for Zustand: https://github.com/pmndrs/zustand/pull/3015
PR for Valtio: https://github.com/pmndrs/valtio/pull/1057

dai-shi: Thanks. I'll merge all of them.

dai-shi (approved):
Thanks for your contribution!

PR #2989, opened 2025-02-17
chore: use husky and lint-staged to fix issues automaticaly

kretajak:
This PR is mend to improve developers experience. It incorporates `husky` and `lint-staged`.

* Devs will no be able to commit changes violating rules provided by `eslint` and `prettier`.
* If there are issues, but they can be automatically fixed by `eslint` and `prettier`, there will be no need to run `pnpm run fix` and amend such changes.
* It will save CI time, as there should not be any Pull request with formatting issues. 


package.json
```diff
@@ -76,9 +77,9 @@
     "postbuild": "pnpm run patch-d-ts && pnpm run copy && pnpm run patch-ts3.8 && pnpm run patch-old-ts && pnpm run patch-esm-ts && pnpm run patch-readme",
     "fix": "pnpm run \"/^fix:.*/\"",
     "fix:lint": "eslint . --fix",
-    "fix:format": "prettier \"*.{js,json,md}\" \"{src,tests,benchmarks,docs}/**/*.{ts,tsx,md,mdx}\" --write",
+    "fix:format": "prettier . --write --ignore-unknown",
```
kretajak, line 80: I believe it has more sense to ignore things via `.prettierignore` ([suggested by Prettier](https://prettier.io/docs/ignore)) rather than narrowing to these matching above patterns.

dai-shi:
Thanks but no thanks.

(A similar suggestion has been raised in some other repos, but it's our intention to not rely on commit hooks at the repo level. If you need it, setup your commit hooks locally.)

kretajak:
Right. I see your point. :+1: Would it still make sense for you to tweak Prettier config?

package.json
dai-shi: In general, I'm not a big fan of many dot files, but this is acceptable and I also ended up using `.prettierignore` file in other repos, for ex: https://github.com/jotaijs/jotai-valtio/blob/main/.prettierignore

Why do we need `--ignore-unknown`?

dai-shi:
Added a comment there: https://github.com/pmndrs/jotai/pull/2989#discussion_r1958182811

package.json
kretajak: Actually is my mistake. Prettier have problems with formatting Github workflow files, but I've have added them to `.prettierignore` so it's not longer needed.

.prettierignore
```diff
@@ -0,0 +1,5 @@
+website
+examples
+dist
+coverage
+.github
```
dai-shi, line 5: I would like to minimize the entries.
Can we only leave `dist`?
Let's format other files.
I wonder if we need to add `pnpm-lock.yaml`. I don't remember why I added it, but there was some issues back then.

kretajak: - Removing `website` produces lot of warnings, not sure why. `website` has it's own ignore, but can be removed I think. There is also another `pnpm-lock.yml`. Unless I'm mistaken there should be only once at root.  
- Removing `examples` is possible, but will require formatting of files there. 
- `coverage` can be safely removed. 
- Removing `.github` will produce error as prettier has troubles with formatting `.github/workflows/test-multiple-builds.yml`.

Summing up: I can take a look that and see if it can be simplified. My ignore config was formatting more less the same files as the current pattern in `package.json`. That was the idea.

dai-shi: > Unless I'm mistaken there should be only once at root.

We aren't using monorepo here, so having two lock files is expected.

> website has it's own ignore

Oh, I wasn't aware of it. We should have one at the root.

> Removing .github will produce error as prettier has troubles with formatting .github/workflows/test-multiple-builds.yml.

What are the troubles? Can we fix them?

> That was the idea.

Yeah, I understand it. But, here's my suggestion. Let's improve it instead of keeping it. We should format files as much as possible.

Feel free to open a new PR and let's discuss there.

PR #2996, opened 2025-02-20
fix(atomWithRefresh): non-writable refreshables throw an error if implicitly passed an argument

a10nik:
…

https://github.com/pmndrs/jotai/discussions/2995

Fixes #

## Summary
Proposed fix for the issue where passing something into the refresh function causes it to do nothing.
## Check List

### DIFF
--- a/src/vanilla/utils/atomWithRefresh.ts
+++ b/src/vanilla/utils/atomWithRefresh.ts
@@ -39,6 +39,8 @@ export function atomWithRefresh<Value, Args extends unknown[], Result>(
         set(refreshAtom, (c) => c + 1)
       } else if (write) {
         return write(get, set, ...args)
+      } else if (import.meta.env?.MODE !== 'production') {
+        throw new Error('refresh must be called without arguments')
       }
     },
   )
--- a/tests/vanilla/utils/atomWithRefresh.test.ts
+++ b/tests/vanilla/utils/atomWithRefresh.test.ts
@@ -3,7 +3,7 @@ import { createStore } from 'jotai/vanilla'
 import { atomWithRefresh } from 'jotai/vanilla/utils'
 
 describe('atomWithRefresh', () => {
-  it('throws when refresh is called with extra arguments', () => {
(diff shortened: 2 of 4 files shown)


src/vanilla/utils/atomWithRefresh.ts
```diff
@@ -35,10 +35,10 @@ export function atomWithRefresh<Value, Args extends unknown[], Result>(
       return read(get, options as never)
     },
     (get, set, ...args: Args) => {
-      if (args.length === 0) {
-        set(refreshAtom, (c) => c + 1)
```
dai-shi, line 39: I think the contract is, if no args, refresh. So, I would like to keep this as is.

tests/vanilla/utils/atomWithRefresh.test.ts
```diff
@@ -0,0 +1,12 @@
+import { describe, expect, it, vi } from 'vitest'
+import { createStore } from 'jotai/vanilla'
+import { atomWithRefresh } from 'jotai/vanilla/utils'
+
+describe('atomWithRefresh', () => {
+  it('throws when refresh is called with extra arguments', () => {
```
dai-shi, line 6:
```suggestion
  it('[DEV-ONLY] throws when refresh is called with extra arguments', () => {
```

wojtekmaj:
We have noticed that this PR introduces a code that throws in production on Next.js. I'm not sure why is that, perhaps they are not setting MODE correctly?

dai-shi:
On second thought, I think this should be either:
1. always throw
2. dev-only console.warn

dai-shi:
> I'm not sure why is that, perhaps they are not setting MODE correctly?

Yeah, if we use ESM, we need to set `import.meta.env.MODE` by ourselves, unless bundlers do it. This is tricky...

dai-shi:
> On second thought, I think this should be either:
> 
> 1. always throw
> 2. dev-only console.warn

I was going to fix it, but noticed there are many cases of dev-only throw. So, for now, let's leave it...

PR #3002, opened 2025-02-25
refactor(vanilla, react): change exported functions to 'function declarations' and non-exported ones to 'arrow functions'

sukvvon:
## Summary
* https://github.com/pmndrs/jotai/pull/3002#pullrequestreview-2640696617
  * for exported functions, use function declaration
  * for other functions, use const arrow function (and omit return statements if possible.)

## Check List

### DIFF
--- a/src/react/Provider.ts
+++ b/src/react/Provider.ts
@@ -13,14 +13,14 @@ type Options = {
   store?: Store
 }
 
-export const useStore = (options?: Options): Store => {
+export function useStore(options?: Options): Store {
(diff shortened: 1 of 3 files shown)


dai-shi (changes requested):
Thanks for the suggestion. However, the coding style is slightly different. If you want to make it consistent, please follow this rule:
- for exported functions, use function declaration
- for other functions, use const arrow function (and omit return statements if possible.)

dai-shi (approved):
LGTM. Thanks!

PR #3009, opened 2025-03-02
Refactor type set atom

nightohl:

Fixes #

## Summary
- removed unnecessary duplicated SetAtom in react

## Check List

### DIFF
--- a/src/react/useSetAtom.ts
+++ b/src/react/useSetAtom.ts
@@ -6,7 +6,8 @@ import type {
 } from '../vanilla.ts'
 import { useStore } from './Provider.ts'
 
-type SetAtom<Args extends unknown[], Result> = (...args: Args) => Result
+export type SetAtom<Args extends unknown[], Result> = (...args: Args) => Result
+
 type Options = Parameters<typeof useStore>[0]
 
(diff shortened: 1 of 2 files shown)


src/react/useSetAtom.ts
```diff
@@ -6,7 +6,8 @@ import type {
 } from '../vanilla.ts'
 import { useStore } from './Provider.ts'
 
-type SetAtom<Args extends unknown[], Result> = (...args: Args) => Result
+export type SetAtom<Args extends unknown[], Result> = (...args: Args) => Result
```
dai-shi, line 9: It's intentional to avoid exporting this for misusing the internal types. We may reconsider ts bundling in the future, but let's leave it for now.

dai-shi:
Thanks anyway.

src/react/useSetAtom.ts
nightohl: @dai-shi 
Ah! if it's intentional, that's okay :)
at first I tried to use internal types and remove types in react directory, but I thought it's intentional of two things.
- when it is used in vanilla then it should support subtype of Args, 
- when it used in react it should be follow strict types of Args,
so I only handled this.

Thanks for replying!

PR #3010, opened 2025-03-02
refactor(unmountAtom): use atomStateMap.get instead of ensureAtomState

nightohl:

Fixes #

## Summary

When unmounting an atom, we don't need to create atom state since we only care about already mounted atoms.
but ensureAtomState make new atom whenever there isn't atom.
I thought this process is kinda unnecessary

- Prevents unnecessary atom state creation
- Makes the unmounting logic clearer
- Returns early if atom state doesn't exist

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -838,7 +838,11 @@ const buildStore = (
   const unmountAtom =
     buildingBlockFunctions[8] ||
     ((atom) => {
-      const atomState = ensureAtomState(atom)
+      const atomState = atomStateMap.get(atom)
+      if (!atomState) {
+        return
+      }
+
       let mounted = mountedMap.get(atom)
       if (
         mounted &&


src/vanilla/internals.ts
```diff
@@ -838,7 +838,11 @@ const buildStore = (
   const unmountAtom =
     buildingBlockFunctions[8] ||
     ((atom) => {
-      const atomState = ensureAtomState(atom)
```
dai-shi, line 841: I think we want to keep this, as this is the single point of accessing the atomStateMap. There's possibility to extend it, and I don't feel like adding more lines for this.

nightohl: @dai-shi 
Got it! I thought it could be intentional too.

Maybe it's a small overhead, but I thought it was causing unnecessary work right now, and as a beginner, I couldn't determine if atomOnInit hooks might cause any unintended problems or not.

That seems pretty reasonable if it's just for the purpose of ensuring and adhering to the converntions.

Thank you for your reply!

dai-shi: Yeah, for the normal use case, atom state should have been created in advance.

PR #3011, opened 2025-03-02
feat(flushCallbacks): improve error handling with multiple errors

nightohl:

Fixes #

## Summary

Previously, only the first error was thrown and subsequent errors were ignored. Now all errors are collected and:
- If there's only one error, throw it directly
- If there are multiple errors, throw them as AggregateError

This change improves debugging by preserving all error information during callback execution.

## Check List


dai-shi:
@dmaskasky What do you think?

src/vanilla/internals.ts
```diff
@@ -490,8 +486,14 @@ const buildStore = (
...
       )
-      if (hasError) {
-        throw error
+
+      if (errors.length === 1) {
+        throw errors[0]
+      } else if (errors.length > 1) {
+        throw new AggregateError(
+          errors,
+          'Multiple errors occurred during flushCallbacks',
```
dai-shi, line 495: I would like to remove the message at least.
```suggestion
```

src/vanilla/internals.ts
```diff
@@ -490,8 +486,14 @@ const buildStore = (
...
+        )
```
dai-shi, line 496: I would even go with this. Not sure about the drawback, but it's predictable.
```suggestion
      if (errors.length) {
        throw new AggregateError(errors)
```

nightohl: @dai-shi I think it's good way to go!
I'd change commit with it and force-push, thanks!

dmaskasky:
> @dmaskasky What do you think?

My main concern is that AggregateError will not pass an instance check.

```
try {
  store.set(someAtom, ...values)
} catch(e) {
  if (e extends FooError) {
    // handle error
  }
}
```

Are there examples where AggregateError is used in other open source libraries?

dmaskasky:
Reading more on AggregateError, I think this makes sense.

Could we do
```ts
if (errors.length === 1) {
  throw errors[0]
} else if (errors.length) {
  throw new AggregateError(errors) 
}
```

dai-shi:
> My main concern is that AggregateError will not pass an instance check.

I don't think it's solved anyways, because you don't know if other subscribers throw an error.

src/vanilla/internals.ts
```diff
@@ -490,8 +486,9 @@ const buildStore = (
         unmountCallbacks.size ||
         mountCallbacks.size
       )
-      if (hasError) {
-        throw error
+
```
dai-shi, line 489:
```suggestion
```

dai-shi (approved):
I like this all-or-nothing approach. It's more predictable.

dmaskasky:
> I don't think it's solved anyways, because you don't know if other subscribers throw an error.

Should we wrap all errors in AggregateError? Is this something we should teach?

dai-shi:
> Should we wrap all errors in AggregateError? Is this something we should teach?

Yes, that's exactly my suggestion. It's harder to tell when it's wrapped otherwise.

PR #3012, opened 2025-03-05
Docs: Remove 'published' metadata

dngur9801:
## Summary
Missing documents were discovered in each section of https://jotai.org/docs

### Detail
image
In the Extensions section, documents for valtio, zustand, relay, and redux are not visible.
However, access to the detailed pages is possible.

The missing documents commonly had a ```'published: false'``` metadata, and once this was removed, the missing documents began to be exposed.

It was confirmed that the role of 'published' was solely to indicate document visibility, so related code was also removed.

If this is an intended behavior, please close these Pull Requests. 🙏

## Check List

### DIFF
--- a/docs/extensions/redux.mdx
+++ b/docs/extensions/redux.mdx
@@ -3,7 +3,6 @@ title: Redux
 description: This doc describes Redux extension.
 nav: 4.98
 keywords: redux
-published: false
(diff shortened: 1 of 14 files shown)


dai-shi:
Yes, it's intentional.

PR #3013, opened 2025-03-06
fix: micro-suspenses

tien:
Fixes #2848

I honestly don't understand why or how, so would need help with reviewing this. Worth noting this issue only happen with React 19.

### DIFF
--- a/src/react/useAtomValue.ts
+++ b/src/react/useAtomValue.ts
@@ -138,11 +138,16 @@ export function useAtomValue<Value>(atom: Atom<Value>, options?: Options) {
   const delay = options?.delay
   useEffect(() => {
     const unsub = store.sub(atom, () => {
-      if (typeof delay === 'number') {
+      try {
         const value = store.get(atom)
         if (isPromiseLike(value)) {
           attachPromiseMeta(createContinuablePromise(value))
         }
+      } catch {
+        /* empty */
+      }
+
+      if (typeof delay === 'number') {
         // delay rerendering to wait a promise possibly to resolve
         setTimeout(rerender, delay)
         return
--- a/src/react/useAtomValue.ts
+++ b/src/react/useAtomValue.ts
@@ -138,16 +138,11 @@ export function useAtomValue<Value>(atom: Atom<Value>, options?: Options) {
   const delay = options?.delay
   useEffect(() => {
     const unsub = store.sub(atom, () => {
-      try {
+      if (typeof delay === 'number') {
         const value = store.get(atom)
         if (isPromiseLike(value)) {
           attachPromiseMeta(createContinuablePromise(value))
         }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -168,6 +168,7 @@
     "ts-expect": "^1.3.0",
     "typescript": "^5.7.3",
     "typescript-eslint": "^8.24.0",
+    "vite": "^6.2.0",
```
tien, line 171: This is so the Vitest VSCode extension can work.

src/react/useAtomValue.ts
```diff
@@ -138,11 +138,12 @@ export function useAtomValue<Value>(atom: Atom<Value>, options?: Options) {
   const delay = options?.delay
   useEffect(() => {
     const unsub = store.sub(atom, () => {
+      const value = store.get(atom)
+      if (isPromiseLike(value)) {
+        attachPromiseMeta(createContinuablePromise(value))
+      }
+
```
dmaskasky, line 145: So it seems to work when we add a microtask delay.

Can we avoid having to add a delay?

tien: As per this issue: https://github.com/pmndrs/jotai/discussions/2848

Adding a delay is actually a workaround for a React 19 specific bug.

Though when I test around with it, seems like this is the actual part that fix it `attachPromiseMeta(createContinuablePromise(value))`

package.json
dai-shi: Please don't add it only for that purpose.

dai-shi (commented):
My gut feeling is that the fix isn't correct. But, I know `atomWithObservable` is of hacks.
The test should be helpful. I'll check #2848 first.

dai-shi:
Maybe, I should create a separate PR and revert my change here.

dai-shi (changes requested):
To get convinced with src/react/useAtomValue.ts change, we need a test without atomWithObservable. However, even if we have a test, the change might not be appropriate. So, there's no guarantee that this PR will be merged.

For the atomWithObservable limitation, I opened #3016.

dai-shi:
> Though when I test around with it, seems like this is the actual part that fix it `attachPromiseMeta(createContinuablePromise(value))`

It feels like it works by chance, which means it may break in the future versions of React.

Just curious why you put try-catch, which also looks bad.

tien:
@dai-shi 

> It feels like it works by chance

Yeah, I was literally trying random stuffs until things stop misbehaving :))

The `try-catch` block is needed since `store.get(atom)` could throws a synchronous error.

dai-shi:
As #3026 does, we shouldn't depend on the hack in React 19.

I'm still interested in the problem without atomWithObservable. Let's continue the discussion in #2848.

PR #3026, opened 2025-03-12
fix(react): experimental option to make attaching promise status controllable

dai-shi:
And rename to `React`.

### DIFF
--- a/src/react/useAtomValue.ts
+++ b/src/react/useAtomValue.ts
@@ -171,9 +171,6 @@ export function useAtomValue<Value>(atom: Atom<Value>, options?: Options) {
   // `instanceof Promise` actually works fine in this case.
   if (isPromiseLike(value)) {
     const promise = createContinuablePromise(value, () => store.get(atom))
-    if (promiseStatus) {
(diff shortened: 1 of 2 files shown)


dai-shi:
https://bsky.app/profile/sebmarkbage.calyptus.eu/post/3lku7b7xjmk2w

Or, we should always attach promise meta if the value is a promise? (It's actually what I did previously.)

dai-shi:
Actually, it doesn't help as it's already triggering the update.

dai-shi:
> Actually, it doesn't help as it's already triggering the update.

Maybe it helps for subsequent updates, as #3013 suggested by @tien.

https://github.com/pmndrs/jotai/discussions/2848#discussioncomment-12621556

tien:
@dai-shi are there any downside of attaching promise status? Also, I assume this is used internally by React's `use` and is not documented?

dai-shi:
What they suggest is slightly different from what we do here.
Anyway, this is behind the experimental options, and you can experiment with it. If it works well, let's consider finalizing it in the future version.

PR #3045, opened 2025-04-12
fix: expose an internal function for devtools

dai-shi:

#3023

## Summary

This allows overriding createStore. It's not very recommended, but can't be helped to avoid breaking changes. We may revisit this in the next major version. cc @arjunvegda 

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -84,12 +84,22 @@ type PrdOrDevStore =
   | INTERNAL_PrdStore
   | (INTERNAL_PrdStore & INTERNAL_DevStoreRev4)
 
+let overiddenCreateStore: typeof createStore | undefined
+
+export function INTERNAL_overrideCreateStore(
+  fn: (prev: typeof createStore | undefined) => typeof createStore,
+): void {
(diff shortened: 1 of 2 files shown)


arjunvegda (approved):
LGTM! I'll test this build tomorrow too

src/vanilla/store.ts
```diff
@@ -84,12 +84,22 @@ type PrdOrDevStore =
...
   | (INTERNAL_PrdStore & INTERNAL_DevStoreRev4)
 
+let overiddenCreateStore: typeof createStore | undefined
+
+export function INTERNAL_overrideCreateStore(
+  fn: (prev: typeof createStore | undefined) => typeof createStore,
+): void {
+  overiddenCreateStore = fn(overiddenCreateStore)
+}
+
```
arjunvegda, line 94: Should we export it out of `jotai/vanilla` too?

dai-shi: Yeah, I forgot there was a follow-up commit. Thanks.

arjunvegda (approved):
Nice! Works as expected! 🎉

PR #3051, opened 2025-04-17
docs: fix atom-with-listeners usage example

CanRau:
I might be completely off but on first glance I think it should be `useAtom(countAtom)` instead of `useState(0)`, shouldn't it? 🤔

### DIFF
--- a/docs/recipes/atom-with-listeners.mdx
+++ b/docs/recipes/atom-with-listeners.mdx
@@ -15,7 +15,7 @@ those state changes.
 
 ```ts
 import { useEffect } from 'react'
-import { atom, useSetAtom, Getter, Setter, SetStateAction } from 'jotai'
+import { atom, useAtom, useSetAtom, Getter, Setter, SetStateAction } from 'jotai'
 
 type Callback<Value> = (
   get: Getter,
@@ -59,7 +59,7 @@ In a component:
 const [countAtom, useCountListener] = atomWithListeners(0)
 
 function EvenCounter() {
-  const [evenCount, setEvenCount] = useState(0)
+  const [evenCount, setEvenCount] = useAtom(countAtom)
 
   useCountListener(
     useCallback(
--- a/docs/recipes/atom-with-listeners.mdx
+++ b/docs/recipes/atom-with-listeners.mdx
@@ -15,7 +15,14 @@ those state changes.
 
 ```ts
 import { useEffect } from 'react'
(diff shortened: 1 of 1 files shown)


dai-shi (commented):
Nice catch.

Please check the CI errors.

dmaskasky:
Yes, I agree. But I also wonder if we really need this recipe at all? The example is pretty complicated to be a recipe. There are easier and better ways to listen for atom changes. 

For instance, [jotai-history](https://github.com/jotaijs/jotai-history) exports an `withHistory`. The code would look something like this.

```ts
import { useRef } from 'react'
import { useCallbackOne as useStableCallback } from 'use-memo-one'
import { withHistory } from 'jotai-history'
import { atom } from 'jotai'
import { useAtomCallback } from 'jotai/utils'

const countWithHistory = withHistory(atom(0), 2) // 2 for [curr, prev]

function useCountListener(cb) {
  const cbRef = useRef(cb)
  cbRef.current = cb
  useAtomCallback(
    useStableCallback((get) => {
      const [curr, prev] = get(countWithHistory)
      cbRef.current?.(curr, prev)
    }, [])
  )
}

function Component() {
  useCountListener((curr, prev) => {
    console.log(curr, prev)
  })
  // ...
}
```

A few more points:
1. As with useMemo, you shouldn't rely on React's useCallback as it doesn't provide a strong guarantee that the callback will always be memoized correctly to the dependencies passed in. Prefer to use useMemoOne, useCallbackOne from 'use-memo-one' instead.

2. We shouldn't need to require callback memoization at every level. Instead we can use a useRef to always call the latest callback passed in.

3. jotai-utils already exposes useAtomCallback. I believe this is an appropriate use of that api.

dai-shi:
> The example is pretty complicated to be a recipe.

I think it's still good. Consider it a learning material.

> A few more notes:

Good point. PRs are welcome.

docs/recipes/atom-with-listeners.mdx
```diff
@@ -15,7 +15,7 @@ those state changes.
 
 ```ts
 import { useEffect } from 'react'
-import { atom, useSetAtom, Getter, Setter, SetStateAction } from 'jotai'
+import { atom, useAtom, useSetAtom, Getter, Setter, SetStateAction } from 'jotai'
```
dai-shi, line 18:
```suggestion
import {
  atom,
  useAtom,
  useSetAtom,
  Getter,
  Setter,
  SetStateAction,
} from 'jotai'
```

PR #3055, opened 2025-04-24
fix(utils): allow undefined for atomWithStorage subscriber return

caffeinewriter:

Fixes #3054 

## Summary

This changes the return type of the `subscribe` function to allow undefined. Alternatively, the type of `Unsubscribe` could be changed to `() => void | undefined`, which may be preferable. This doesn't change the functionality of Jotai, and is a type-only issue, so no tests needed to be added or changed, and this simply encompasses the intended existing functionality.

## Check List

### DIFF
--- a/src/vanilla/utils/atomWithStorage.ts
+++ b/src/vanilla/utils/atomWithStorage.ts
@@ -11,12 +11,12 @@ type Subscribe<Value> = (
   key: string,
   callback: (value: Value) => void,
   initialValue: Value,
-) => Unsubscribe
+) => Unsubscribe | undefined
 
 type StringSubscribe = (
   key: string,
   callback: (value: string | null) => void,
-) => Unsubscribe
+) => Unsubscribe | undefined
 
 type SetStateActionWithReset<Value> =
   | Value


src/vanilla/utils/atomWithStorage.ts
```diff
@@ -11,12 +11,12 @@ type Subscribe<Value> = (
   key: string,
   callback: (value: Value) => void,
   initialValue: Value,
-) => Unsubscribe
+) => Unsubscribe | undefined
```
dai-shi, line 14: `void` should be more appropriate?

caffeinewriter: Ah, that actually causes some issues since `void` is considered the absence of a return value, rather than undefined. It's subtle, but Typescript won't accept that. The workaround here would be to type [the `unsub` variable](https://github.com/caffeinewriter/jotai/blob/bfe474f588472136002b0783bbb947b116a0d4ec/src/vanilla/utils/atomWithStorage.ts#L239) as `Unsubscribe | void | undefined`, but considering the intent, I think `Unsubscribe | undefined` as the return type might make more sense.

dai-shi: Okay, makes sense.

dai-shi (approved):
Thanks. I'll merge it when I prepare next release.

PR #3061, opened 2025-05-06

dai-shi:
for v2.12.4


src/react/Provider.ts
```diff
@@ -18,8 +18,6 @@ export function useStore(options?: Options): Store {
   return options?.store || store || getDefaultStore()
 }
 
-/* eslint-disable react-compiler/react-compiler */
-// TODO should we consider using useState instead of useRef?
```
dai-shi, line 22: Cool. react-compiler understands our pattern.
#2771 is no longer necessary.

PR #3075, opened 2025-05-23
test(react/useSetAtom): add test for throwing error when called with read-only atom

sukvvon:

Fixes #

### AS-IS

image

### TO-BE

image

## Check List


tests/react/useSetAtom.test.tsx
```diff
@@ -103,3 +103,32 @@ it('useSetAtom with write without an argument', async () => {
   fireEvent.click(screen.getByText('increment'))
   expect(screen.getByText('count: 1')).toBeInTheDocument()
 })
+
+it('useSetAtom throws when called with a read-only atom', () => {
+  const originalEnv = import.meta.env
+  Object.defineProperty(import.meta, 'env', {
```
Wendystraite: Tip : you can use `vi.stubEnv` instead https://vitest.dev/api/vi.html#vi-stubenv

My usual single-test-setup is to pair a cleanup with `onTestFinished` https://vitest.dev/api/#ontestfinished

sukvvon: Thanks for your feedback!. BTW we have to handle old TS versions.

tests/react/useSetAtom.test.tsx
```diff
@@ -103,3 +103,21 @@ it('useSetAtom with write without an argument', async () => {
...
+
+it('useSetAtom throws when called with a read-only atom', () => {
+  const countAtom = atom(0)
+  const readOnlyAtom = atom((get) => get(countAtom))
+
+  let setAtomFn: ((v: number) => void) | undefined
+
+  function TestComponent() {
+    // eslint-disable-next-line react-hooks/globals
+    setAtomFn = useSetAtom(readOnlyAtom as any)
```
dai-shi: Instead of using the out scope variable, can we use `useEffect` instead?

tests/react/useSetAtom.test.tsx
```diff
@@ -103,3 +103,21 @@ it('useSetAtom with write without an argument', async () => {
...
+    return null
+  }
+
+  render(<TestComponent />)
+
+  expect(setAtomFn).toBeDefined()
+  expect(() => setAtomFn!(1)).toThrow()
```
dai-shi: Otherwise, I'm not sure what's throwing.

tests/react/useSetAtom.test.tsx
```diff
@@ -103,3 +103,20 @@ it('useSetAtom with write without an argument', async () => {
   fireEvent.click(screen.getByText('increment'))
   expect(screen.getByText('count: 1')).toBeInTheDocument()
 })
+
+it('[DEV-ONLY] useSetAtom throws when called with a read-only atom', () => {
```
dai-shi, line 107: Does this work?
```suggestion
it('[DEV-ONLY] useSetAtom throws when called with a read-only atom', () => {
  expect.assertions(1)
```

PR #3077, opened 2025-05-25
test(react/vanilla-utils/selectAtom): add test for creating fresh cache path when deps differ (memo3)

sukvvon:

Fixes #

### before

image

### after

image

## Check List

### DIFF
--- a/tests/react/vanilla-utils/selectAtom.test.tsx
+++ b/tests/react/vanilla-utils/selectAtom.test.tsx
@@ -1,7 +1,7 @@
 import { StrictMode, useEffect, useRef } from 'react'
 import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
-import { it } from 'vitest'
+import { expect, it } from 'vitest'
 import { useAtomValue, useSetAtom } from 'jotai/react'
 import { atom } from 'jotai/vanilla'
 import { selectAtom } from 'jotai/vanilla/utils'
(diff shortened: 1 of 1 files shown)


tests/react/vanilla-utils/selectAtom.test.tsx
```diff
@@ -130,3 +130,26 @@ it('do not update unless equality function says value has changed', async () =>
...
+    const value = useAtomValue(derivedAtom2)
+    return <div>{JSON.stringify(value)}</div>
+  }
+
+  const { unmount } = render(<Component1 />)
+  await screen.findByText('{"a":0,"b":1}')
+  unmount()
+
+  render(<Component2 />)
+  await screen.findByText('{"a":0,"b":1}')
```
dai-shi, line 154: How does this test "creates fresh cache path when deps differ"?

sukvvon: @dai-shi 

This test ensures that a new derived atom is created when the dependencies to `memo3` differ — specifically, when the `selector` function is a different reference, even if semantically identical.

We call `selectAtom(baseAtom, (value) => value)` twice with new inline functions. Since each inline function creates a new reference, the second call results in a new memoization path. This triggers `memo3` to generate a fresh WeakMap chain and cache the result separately.

To verify this behavior not just at the code level but also at runtime, we render and unmount `Component1` (which uses the first derived atom), then render `Component2` (which uses the second derived atom). Both render correctly with the expected value, demonstrating that the two derived atoms are treated as independent instances with separate cache paths.

Although this rendering/unmounting logic might seem unnecessary at first glance, it plays a crucial role in verifying the correct isolation and behavior of the newly memoized atom instance. It confirms that the cache differentiation isn’t just theoretical — it actually results in functionally distinct atoms that operate independently, even after the first one is unmounted.

In short, this test verifies that `memo3` correctly isolates cache paths based on all three dependencies (`anAtom`, `selector`, and `equalityFn`) and that the resulting atom instances behave as expected in practice.

dai-shi: I think this test passes even if we don't have `memo3` or if `memo3` doesn't cache things correctly.

tests/react/vanilla-utils/selectAtom.test.tsx
```diff
@@ -130,3 +130,18 @@ it('do not update unless equality function says value has changed', async () =>
...
+  const derivedAtom2 = selectAtom(baseAtom, (v) => v)
+
+  expect(derivedAtom1).not.toBe(derivedAtom2)
+
+  const selector = (v: { a: number; b: number }) => v.a
+  const derivedAtom3 = selectAtom(baseAtom, selector)
+  const derivedAtom4 = selectAtom(baseAtom, selector)
+
+  expect(derivedAtom3).toBe(derivedAtom4)
+})
```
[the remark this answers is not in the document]
dai-shi: This looks nice! Thanks!!

dai-shi (approved):

PR #3085, opened 2025-06-10
only throw AggregateError if more than one error was thrown in flushPending

dmaskasky:
## Summary
Changes the error handling behavior of flushPending. Only throw AggregateError if more than one error was thrown during flushPending. 

- No errors: proceed as normal
- A single error: rethrow the error thrown
- Multiple errors: throw `AggregateError`

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -473,7 +473,7 @@ const buildStore = (
         mountCallbacks.size
       )
       if (errors.length) {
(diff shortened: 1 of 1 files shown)


dai-shi:
As far as I remember, we've already discussed about this before and this was intentional. With this change, the behavior gets unpredictable. I think we should always wrap an error or some errors with AggregateError.

dmaskasky:
Makes sense. I forgot.

PR #3088, opened 2025-06-17
docs(third-party): add '@suspensive/jotai' library

sukvvon:

Fixes #

## Check List


dai-shi:
I'm a bit hesitant to push this pattern. Our official recommendation should be using hooks.
(This means we can't accept everything even for `third-party`.)

Closing this, but feel free to continue discussions here.

PR #3095, opened 2025-07-03
breaking: drop deprecated devstore

dai-shi:
Drop deprecated internal types and functions.


dai-shi:
@arjunvegda Are we ready for this, or should we wait a little longer?

PR #3099, opened 2025-07-13
jotai-scope docs update

dmaskasky:
## Summary
Updates jotai-scope website docs

## Check List


dai-shi (approved):
LGTM

I wonder if we should prefer function declaration or const arrow function for components throughout the doc, but that's relatively minor issue.

dmaskasky:
> I wonder if we should prefer function declaration or const arrow function for components throughout the doc, but that's relatively minor issue.

Personally, I like function declaration syntax. But its just a personal preference, I could be  persuaded either way.

dai-shi:
Apart from library development, one difference I care in app development is the minified size:

In https://try.terser.org/
```js
export function Foo () {
  return createElement('h1', null, 'Hello');
}
```
is 62 bytes, whereas

```js
export const Foo = () => createElement('h1', null, 'Hello');
```
is 54 bytes.

That's a subtle thing though. In terms of coding style, I used to prefer const function that can't be overwritten, but nowadays I prefer function declaration too (as we don't do such mistakes in modern JS/TS.) So, I don't care much in general.

dmaskasky:
> Apart from library development, one difference I care in app development is the minified size:
> 
> In https://try.terser.org/
> 
> ```js
> export function Foo () {
>   return createElement('h1', null, 'Hello');
> }
> ```
> 
> is 62 bytes, whereas
> 
> ```js
> export const Foo = () => createElement('h1', null, 'Hello');
> ```
> 
> is 54 bytes.
> 
> That's a subtle thing though. In terms of coding style, I used to prefer const function that can't be overwritten, but nowadays I prefer function declaration too (as we don't do such mistakes in modern JS/TS.) So, I don't care much in general.

The one thing that bothers me with function declarations are that you need to fall back to const functions whenever you want to apply a higher order function. Oh well.

PR #3104, opened 2025-07-23
[Not For Merge] wip: internals rev2

dai-shi:
To expand our idea.

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -375,16 +375,12 @@ type StoreState = readonly [
   storeHooks: StoreHooks,
 ]
 
-export type INTERNAL_StoreStateRev2 = StoreState
-
 type AtomInterceptors = readonly [
   atomRead: AtomRead,
   atomWrite: AtomWrite,
   atomOnInit: AtomOnInit,
   atomOnMount: AtomOnMount,
 ]
-
-export type INTERNAL_AtomInterceptorsRev2 = AtomInterceptors
 
 type BuildingBlocks = readonly [
   ensureAtomState: EnsureAtomState,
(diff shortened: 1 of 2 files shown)


dai-shi:
image

Hmm, I'm not super happy with it.

src/vanilla/internals.ts
```diff
@@ -391,474 +394,439 @@ type BuildingBlocks = readonly [
...
+    storeHooks,
+  ] = storeState
+
+  const [atomRead, atomWrite, atomOnInit, atomOnMount] = atomInterceptors
+
+  const ensureAtomState: EnsureAtomState = (atom) => {
+    if (import.meta.env?.MODE !== 'production' && !atom) {
+      throw new Error('Atom is undefined or null')
+    }
+    let atomState = atomStateMap.get(atom)
```
dmaskasky, line 443: This might be the same issue. `ensureAtomState` closes over the storeState destructure which causes it to reuse those captured values. If we want swappable pieces, my guess is that we would have to destructure within `ensureAtomState` and other building block functions.

dmaskasky: I think a good test would be to do,
```ts
const asm0 = { get: vi.fn(() => ({ d: new Set() })), has: () => true }
const buildingBlocks0 = asm0
const store0 = INTERNAL_buildStore(...buildingBlocks0)
const [...buildingBlocks1] = INTERNAL_getBuildingBlocks(store0)
const asm1 = { get: vi.fn(() => ({ d: new Set() })), has: () => true }
buildingBlocks1[0] = asm1
const store1 = INTERNAL_buildStore(...buildingBlocks1)
const a = atom(0)
store1.get(a)
expect(asm1.get).toBeCalled()
expect(asm0.get).not.toBeCalled()
```

dai-shi: With this rev2, my expectation is to call createBuildingBlocks again for the second store. You could argue that building blocks of the first store has changed, but we can't assure they don't use closure state anyways. We need some _assumptions_.

dai-shi:
Closing in favor of #3105.

PR #3108, opened 2025-07-24
fix(tests): fix type testing for useHydrateAtoms for old typescript versions

dai-shi:
Problem: tests are broken for old typescript versions
Solution: granular SKIP- and ONLY- control for specific ts versions

### DIFF
--- a/tests/react/utils/types.test.tsx
+++ b/tests/react/utils/types.test.tsx
@@ -6,23 +6,21 @@ it('useHydrateAtoms should not allow invalid atom types when array is passed', (
   function Component() {
     const countAtom = atom(0)
     const activeAtom = atom(true)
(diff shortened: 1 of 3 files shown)


tests/react/utils/types.test.tsx
```diff
@@ -6,30 +6,28 @@ it('useHydrateAtoms should not allow invalid atom types when array is passed', (
   function Component() {
     const countAtom = atom(0)
     const activeAtom = atom(true)
-    // Adding @ts-ignore for typescript 3.8 support
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
```
dai-shi, line 11: I didn't notice, but this actually skips checking types even besides 3.8.

PR #3111, opened 2025-07-29
feat: add `unstable_defaultUse`

himself65:
From v3 discussion https://github.com/pmndrs/jotai/discussions/2889

```
Do not use use in useAtomValue (migration path: add { use: true } option which is true by default in v2, and warns it)
```

### DIFF
--- a/src/react/useAtomValue.ts
+++ b/src/react/useAtomValue.ts
@@ -102,22 +102,25 @@ const createContinuablePromise = <T>(
 
 type Options = Parameters<typeof useStore>[0] & {
   delay?: number
+  unstable_defaultUse?: boolean
   unstable_promiseStatus?: boolean
 }
 
-export function useAtomValue<Value>(
+export function useAtomValue<Value, O extends Options = Options>(
   atom: Atom<Value>,
-  options?: Options,
-): Awaited<Value>
-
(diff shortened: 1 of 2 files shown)


dai-shi:
I would like to wait for https://github.com/pmndrs/jotai/discussions/2889#discussioncomment-12935507 a little more.

Once we decided our plan, we don't need `unstable_` prefix. btw, what's the implication behind `defaultUse` instead of `use`?

src/react/useAtomValue.ts
```diff
@@ -102,22 +102,25 @@ const createContinuablePromise = <T>(
...
+  options?: O,
+): O extends { unstable_defaultUse: true } ? Awaited<Value> : Value
 
 export function useAtomValue<Value>(atom: Atom<Value>, options?: Options) {
-  const { delay, unstable_promiseStatus: promiseStatus = !React.use } =
-    options || {}
+  const {
+    delay,
+    unstable_promiseStatus: promiseStatus = !React.use,
+    unstable_defaultUse = true,
```
dai-shi, line 122: We should warn if it's undefined for the migration. So, it's too early to merge this PR.

dai-shi:
I changed the plan and we will be adding new apis in v3.

PR #3117, opened 2025-08-05
fix: useHydrateAtoms tuple types support looser definitions

dmaskasky:

Fixes #
https://github.com/pmndrs/jotai/discussions/3116

## Summary
The workaround to support older versions of Typescript caused issues when using loose types such as 
```ts
useHydrateAtoms([[atom, value]] as (readonly [AnyWritableAtom, unknown])[])
```

This PR drops support for older TS versions.

## Check List


dai-shi:
See https://github.com/pmndrs/jotai/pull/3108/commits/bcc0fcd294e7acdd14ef4edaeab33be0a64fb09b commit how to use SKIP and ONLY.

.github/workflows/test-old-typescript.yml
```diff
@@ -41,8 +41,10 @@ jobs:
         with:
           node-version: 'lts/*'
           cache: 'pnpm'
-      - run: pnpm install
-      - run: pnpm run build
+      - name: Install dependencies
+        run: pnpm install
+      - name: Build project
+        run: pnpm run build
```
dai-shi: The original one is very intentional. Was it hard to understand without names?

dmaskasky:
> See [bcc0fcd](https://github.com/pmndrs/jotai/commit/bcc0fcd294e7acdd14ef4edaeab33be0a64fb09b) commit how to use SKIP and ONLY.

This works with `@ts-expect-error`, but I would need to provide the inverse for many cases.

.github/workflows/test-old-typescript.yml
dmaskasky: I'll revert.

dai-shi:
> I would need to provide the inverse for many cases.

The inverse is `@ts-ignore`, no?

Anyway, in #3108, I originally took your first fix https://github.com/pmndrs/jotai/commit/c2080e03282d133bab5bffbc5ee8d21f82e4567a . Isn't it different from this PR?

dmaskasky:
> The inverse is `@ts-ignore`, no?

It would be something like `// @ts-expect-pass [SKIP-TS-4.0.5]`

> Anyway, in #3108, I originally took your first fix [c2080e0](https://github.com/pmndrs/jotai/commit/c2080e03282d133bab5bffbc5ee8d21f82e4567a) . Isn't it different from this PR?

This PR drops the compatibility hack that my earlier PR had to accomodate TS versions that did not properly support variadic tuple types, so more old TS versions will not be supported.

dai-shi:
I thought https://github.com/pmndrs/jotai/commit/c2080e03282d133bab5bffbc5ee8d21f82e4567a was the commit before you added the hack for the compatibility.

dmaskasky:
Yes, same overall approach. Slightly improved with `Rest extends Args` check.

dai-shi (approved):

PR #3121, opened 2025-08-06
fix(types): Symbol.observable missing

a10nik:
https://github.com/pmndrs/jotai/discussions/3120

Fixes #

## Check List

- [ ] `pnpm run fix` for formatting and linting code and docs

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "build:vanilla:internals": "rollup -c --config-vanilla_internals",
     "build:react": "rollup -c --config-react --client-only",
     "build:react:utils": "rollup -c --config-react_utils --client-only",
-    "postbuild": "pnpm run patch-d-ts && pnpm run copy && pnpm run patch-ts3.8 && pnpm run patch-old-ts && pnpm run patch-esm-ts && pnpm run patch-readme",
+    "postbuild": "pnpm run patch-d-ts && pnpm run copy && pnpm run patch-ts3.8 && pnpm run patch-old-ts && pnpm run patch-observable && pnpm run patch-esm-ts && pnpm run patch-readme",
     "fix": "pnpm run fix:lint && pnpm run fix:format",
     "fix:format": "prettier . --write",
     "fix:lint": "eslint . --fix",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -94,6 +94,7 @@
     "copy": "shx cp -r dist/src/* dist/esm && shx cp -r dist/src/* dist && shx rm -rf dist/src && shx rm -rf dist/{src,tests} && downlevel-dts dist dist/ts3.8 --to=3.8 && shx cp package.json README.md LICENSE dist && json -I -f dist/package.json -e \"this.private=false; this.devDependencies=undefined; this.optionalDependencies=undefined; this.scripts=undefined; this.prettier=undefined;\"",
     "patch-ts3.8": "node -e \"require('shelljs').find('dist/ts3.8/**/*.d.ts').forEach(f=>require('fs').appendFileSync(f,'declare type Awaited<T> = T extends Promise<infer V> ? V : T;'))\"",
     "patch-old-ts": "shx touch dist/ts_version_3.8_and_above_is_required.d.ts",
+    "patch-observable": "node -e \"require('shelljs').find('dist/esm/**/*.d.ts').forEach(f=>{require('shelljs').sed('-i','Symbol.observable','// @ts-ignore: Symbol.observable is optional\\nSymbol.observable',f);})\"",
```
a10nik, line 97: However, I realized that ts-ignore directives don't get automatically transferred from the source code to d.ts files. So the only way I came up with was to add yet another patching script in package.json. Too much of a hack, IMO, not sure if it should be accepted. Might be helpful to look into some other options

Separate entry point aside these might also work but have their trade-offs:

- Polyfill the type ourselves
  - **trade-off** That will pollute the global scope and lead the users to believe they've got Symbol.observable only to cause runtime errors or confusion later. Besides, saw you remove that type declaration a year ago, apparently, for performance reasons.
- Get rid of that optional `[Symbol.observable]?: ...` type requirement, at least until the proposal is accepted -- if it will be at all. We could drop the `[Symbol.observable]?: ...` type part but keep checking for the symbol in runtime.
  - **trade-off**: That would allow the users to put something incompatible into Symbol.observable
  - IMO it makes little sense to require both `subscribe()` method and an optional `Symbol.observable` field at the same time like it does now. Shouldn't it accept either of the two instead?

dai-shi: This seems too complicated.

dai-shi: Does #3122 work for you?

a10nik: Yeah, that should work. Seems pretty much what I've tried to describe here:

> Get rid of that optional [Symbol.observable]?: ... type requirement, at least until the proposal is accepted -- if it will be at all. We could drop the [Symbol.observable]?: ... type part but keep checking for the symbol in runtime.

PR #3122, opened 2025-08-07
fix(utils/atomWithObservable): avoid external symbol type

dai-shi:
alternate take for #3121 and #3120

### DIFF
--- a/src/vanilla/utils/atomWithObservable.ts
+++ b/src/vanilla/utils/atomWithObservable.ts
@@ -17,9 +17,7 @@ type Observer<T> = {
   complete: () => void
 }
 
-type ObservableLike<T> = {
-  [Symbol.observable]?: () => ObservableLike<T> | undefined
-} & (
+type ObservableLike<T> =
   | {
       subscribe(observer: Observer<T>): Subscription
     }
@@ -31,7 +29,6 @@ type ObservableLike<T> = {
       // Overload function to make typing happy
       subscribe(next: (value: T) => void): Subscription
     }
-)
 
 type SubjectLike<T> = ObservableLike<T> & Observer<T>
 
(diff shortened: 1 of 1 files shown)


src/vanilla/utils/atomWithObservable.ts
```diff
@@ -79,7 +76,9 @@ export function atomWithObservable<Data>(
 
   const observableResultAtom = atom((get) => {
     let observable = getObservable(get)
-    const itself = observable[Symbol.observable]?.()
+    const itself = (
+      observable as Partial<Record<symbol, () => typeof observable>>
+    )[Symbol.observable]?.()
     if (itself) {
```
a10nik, line 82: Now that the type checking system doesn't check for the `observable.[Symbol.observable]()`'s type itself, it might be safer to check some of it in runtime. At least its being a function.

Suggested fix:
```
const observableField = (
      observable as Partial<Record<symbol, () => typeof observable>>
)[Symbol.observable];
if (typeof observableField === 'function') {
  observable = observableField();
}
```

On the other hand, it feels even weirder now that we are ignoring a perfectly suitable instance of Observable with a type-checked subscribe function that is already passed to us and trying to extract something from the inside of it instead.

Suggested fix: Just remove the `Symbol.observable` introspection and the reassignment altogether

dai-shi: Changing the JS behavior is out of the scope of this PR, or the change for this time.

I'm not sure if I understand the use of the symbol from the start.
Some contributor said it's the standard.
I'm fine to drop it if it's not relevant.
Maybe, instead of discussing it here, can you open a new discussion?

a10nik: Ok, will do! Btw can confirm that the main typing issue is fixed, tested in stackblitz.

dai-shi: Cool! Thanks for the confirmation.

PR #3132, opened 2025-08-20
docs: atomWithStorage getOnInit clarification for React Native

peterpme:

Clarifies the confusing behavior of `getOnInit` that I felt wasn't properly clarified in the docs.

## Check List

### DIFF
--- a/docs/utilities/storage.mdx
+++ b/docs/utilities/storage.mdx
@@ -45,7 +45,6 @@ The `atomWithStorage` function creates an atom with a value persisted in `localS
 
 - **getOnInit** (optional, by default **false**): A boolean value indicating whether to get item from storage on initialization. Note that in an SPA with `getOnInit` either not set or `false` you will always get the initial value instead of the stored value on initialization. If the stored value is preferred set `getOnInit` to `true`.
 
-
 If not specified, the default storage implementation uses `localStorage` for storage/retrieval, `JSON.stringify()`/`JSON.parse()` for serialization/deserialization, and subscribes to `storage` events for cross-tab synchronization.
 
 <Stackblitz id="vitejs-vite-wewrmi" file="src%2FApp.tsx" />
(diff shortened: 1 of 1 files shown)


docs/utilities/storage.mdx
```diff
@@ -140,6 +179,8 @@ const content = {} // anything JSON serializable
 const storedAtom = atomWithStorage('stored-key', content, storage)
 ```
 
+**Note** you probably want to set `getOnInit` to true if you're expecting the storage value to be available immediately instead of the default value
```
dai-shi: This is RN section, so instead of adding another note here, how about changing the example above?

docs/utilities/storage.mdx
```diff
@@ -43,7 +43,46 @@ The `atomWithStorage` function creates an atom with a value persisted in `localS
...
+
+```jsx
+// Assume localStorage already contains: { "symbol": "BTC_USDC" }
+
+// Without getOnInit (default behavior)
+const symbolAtom = atomWithStorage('symbol', 'SOL_USDC')
+
+function App() {
+  const symbol = useAtomValue(symbolAtom)
+  console.log('symbol', symbol)
```
dai-shi: We don't usually recommend console.log in the render function. Can we put it in useEffect?
```suggestion
  useEffect(() => {
    console.log('symbol', symbol)
  })
```

dai-shi (commented):
Thanks for the additional notes for the clarification.
It feels a bit verbose, but I know where it's coming from.
Maybe we can simply recommend getOnInit=true in the RN section. Then, can we remove or hide console.log behavior difference section with `<details>`?

peterpme:
Hi @dai-shi thank you for the prompt review and I appreciate the comments and feedback. Those are all very good suggestions I should have considered myself!

Please let me know if this works better for you.

dai-shi (approved):
Looks clean and straightforward.
Thanks for your contribution!

PR #3133, opened 2025-08-21
feat: add allowTearing option in useAtom and useAtomValue to avoid double render on initial mount. (fix #1137)

### LINKED ISSUE #1137: The component is re-rendered the first time it is rendered.
``` javascript

// Create your atoms and derivatives
const textAtom = atom('hello')
const uppercaseAtom = atom(
  (get) => get(textAtom).toUpperCase()
)

// Use them anywhere in your app
const Input = () => {
  const [text, setText] = useAtom(textAtom)
  const handleChange = (e) => setText(e.target.value)
  return (
    <input value={text} onChange={handleChange} />
  )
}

const Uppercase = () => {
  const [uppercase] = useAtom(uppercaseAtom)
  return (
    <div>Uppercase: {uppercase}</div>
  )
}

const App = () => {
    //Print twice
    console.log("rerender....")
  return (
    <>
      <Input />
      <Uppercase />
    </>
  )
}
```
The component is re-rendered the first time it is rendered. That is my console.log will be printed twice. I don't encounter this kind of problem when I use recoil and useState. Is there any way to solve it, thank you.

pilaoda:

Fixes #1137

## Summary

I guess rerender here is to handle atom value changing between initial render and subscribe in useEffect, also avoid tearing in concurrent mode.

However some app just care performance more than tearing. 
For me, I use some [craft.js](https://craft.js.org/) like component to support UI edit. All component do `useAtomValue(enableUIEditorAtom)` to switch on editing, so all components double render will cause a big impact.

## Check List

### DIFF
--- a/src/react/useAtomValue.ts
+++ b/src/react/useAtomValue.ts
@@ -1,6 +1,6 @@
 /// <reference types="react/experimental" />
 
-import React, { useDebugValue, useEffect, useReducer } from 'react'
+import React, { useDebugValue, useEffect, useReducer, useRef } from 'react'
 import { INTERNAL_registerAbortHandler as registerAbortHandler } from '../vanilla/internals.ts'
 import type { Atom, ExtractAtomValue } from '../vanilla.ts'
 import { useStore } from './Provider.ts'
@@ -103,6 +103,7 @@ const createContinuablePromise = <T>(
 type Options = Parameters<typeof useStore>[0] & {
   delay?: number
   unstable_promiseStatus?: boolean
+  allowTearing?: boolean
 }
 
 export function useAtomValue<Value>(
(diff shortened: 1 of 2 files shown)


dai-shi:
Thanks for raising this up.
Our preference is to avoid options, and prefer different hooks.
As the current version doesn't fully solve tearing anyway, I think we should provide two hooks separately.
This is also a good time to consider about avoiding `use`.
Let me work on it, taking some time. Please be patient.

---

<details>
<summary>(memo to myself)</summary>

```js
import { useAtomValue } from 'jotai/react/sync' // uSES (no use)
import { useAtomValue } from 'jotai/react/async' // useEffect (no use)
import { useAtomValue } from 'jotai/react' // useEffect + use
```

OR

```js
import { useAtomVal } from 'jotai/react' // useEffect
import { useAtomValue } from 'jotai/react' // uSES + use
```

OR

```js
import { useAtomVal } from 'jotai/react' // useEffect
import { useAtomSyncVal } from 'jotai/react' // useSES
import { useAtomValue } from 'jotai/react' // useAtomVal + use
```

uSES requires React 18, so this can only be done with a major bump.
</details>

PR #3135, opened 2025-08-26
fix(core/types): Wrong variance on the WritableAtom Result type parameter

atomanyih:
## Reproduction:
```ts
export const createWritableAtom = (): WritableAtom<
  null,
  [],
  number | null
> =>
  atom(
    () => null,
    (get, set, someArg: string) => {
      return null;
    },
  );
```
Error:
```
1. Type 'WritableAtom<null, [someArg: string], null>' is not assignable to type 'WritableAtom<null, [], number | null>'.
2.   Types of property 'read' are incompatible.
3.     Type 'Read<null, SetAtom<[someArg: string], null>>' is not assignable to type 'Read<null, SetAtom<[], number | null>>'.
4.      Type 'SetAtom<[], number | null>' is not assignable to type 'SetAtom<[someArg: string], null>'.
5.        Type 'number | null' is not assignable to type 'null'.
6.          Type 'number' is not assignable to type 'null'. ts(2322)
```

This code should be valid: a function that returns `null` is assignable to a function that returns `number | null`, and I would expect writable atom to work the same way. 

## Explanation
From some digging, it seems like the issue is stemming from `Result` being propagated down to the `SetSelf` parameter of `Read` in `atom.ts`

```ts
type Read<Value, SetSelf = never> = (get: Getter, options: {
    readonly signal: AbortSignal;
    readonly setSelf: SetSelf;
}) => Value;
```

I think what is happening is that, because `SetSelf` is an argument `Read`, typescript infers that it should be contravariant. That's why the assignability gets inverted from lines 3 to 4 of the error. 

## Fix
I haven't fully examined the internal usage of `setSelf`, so forgive me if this is totally off the mark. But it looks like the return value is not used in any of the call sites. So I think we could modify the typing of `WritableAtom` to remove `Result` from read.

```ts
export interface WritableAtom<Value, Args extends unknown[], Result>
  extends Atom<Value> {
  read: Read<Value, SetAtom<Args, void>>
  write: Write<Args, Result>
  onMount?: OnMount<Args, Result>
}
```

This type change is saying, in effect, that we don't care about the return value of `SetAtom` in the `Read` function.
I've also added a type test to ensure the assignability of `WritableAtom` remains correct.

Let me know if I've gotten anything wrong!

## Check List

### DIFF
--- a/src/react/useAtomValue.ts
+++ b/src/react/useAtomValue.ts
@@ -1,5 +1,3 @@
-/// <reference types="react/experimental" />
-
 import React, { useDebugValue, useEffect, useReducer } from 'react'
 import { INTERNAL_registerAbortHandler as registerAbortHandler } from '../vanilla/internals.ts'
 import type { Atom, ExtractAtomValue } from '../vanilla.ts'
--- a/tests/react/transition.test.tsx
+++ b/tests/react/transition.test.tsx
@@ -1,4 +1,3 @@
-/// <reference types="react/experimental" />
 import ReactExports, { StrictMode, Suspense, useEffect } from 'react'
 import { act, fireEvent, render, screen } from '@testing-library/react'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
@@ -14,66 +13,62 @@ afterEach(() => {
   vi.useRealTimers()
 })
 
-const { use, useTransition } = ReactExports
+const { useTransition } = ReactExports
 
 describe.skipIf(typeof useTransition !== 'function')('useTransition', () => {
-  // FIXME some tests are failing with react@experimental
-  it.skipIf(typeof use === 'function')(
-    'no extra commit with useTransition (#1125)',
-    async () => {
-      const countAtom = atom(0)
-      const delayedAtom = atom(async (get) => {
-        await sleep(100)
-        return get(countAtom)
+  it('no extra commit with useTransition (#1125)', async () => {
+    const countAtom = atom(0)
+    const delayedAtom = atom(async (get) => {
+      await sleep(100)
+      return get(countAtom)
+    })
+
+    const committed: { pending: boolean; delayed: number }[] = []
+
+    const Counter = () => {
+      const setCount = useSetAtom(countAtom)
+      const delayed = useAtomValue(delayedAtom)
+      const [pending, startTransition] = useTransition()
+      useEffect(() => {
+        committed.push({ pending, delayed })
       })
-
-      const committed: { pending: boolean; delayed: number }[] = []
-
-      const Counter = () => {
-        const setCount = useSetAtom(countAtom)
-        const delayed = useAtomValue(delayedAtom)
-        const [pending, startTransition] = useTransition()
-        useEffect(() => {
-          committed.push({ pending, delayed })
-        })
-        return (
-          <>
-            <div>delayed: {delayed}</div>
-            <button
-              onClick={() => startTransition(() => setCount((c) => c + 1))}
-            >
-              button
-            </button>
-          </>
-        )
-      }
-
-      await act(() =>
-        render(
-          <>
-            <Suspense fallback="loading">
-              <Counter />
-            </Suspense>
-          </>,
-        ),
+      return (
+        <>
+          <div>delayed: {delayed}</div>
+          <button onClick={() => startTransition(() => setCount((c) => c + 1))}>
+            button
+          </button>
+        </>
       )
+    }
 
-      expect(screen.getByText('loading')).toBeInTheDocument()
-      await act(() => vi.advanceTimersByTimeAsync(100))
-      expect(screen.getByText('delayed: 0')).toBeInTheDocument()
+    await act(() =>
+      render(
+        <>
+          <Suspense fallback="loading">
+            <Counter />
+          </Suspense>
+        </>,
+      ),
+    )
 
+    expect(screen.getByText('loading')).toBeInTheDocument()
+    await act(() => vi.advanceTimersByTimeAsync(100))
+    expect(screen.getByText('delayed: 0')).toBeInTheDocument()
+
+    await act(async () => {
       fireEvent.click(screen.getByText('button'))
-      expect(screen.getByText('delayed: 0')).toBeInTheDocument()
-      await act(() => vi.advanceTimersByTimeAsync(100))
-      expect(screen.getByText('delayed: 1')).toBeInTheDocument()
+    })
+    expect(screen.getByText('delayed: 0')).toBeInTheDocument()
+    await act(() => vi.advanceTimersByTimeAsync(100))
+    expect(screen.getByText('delayed: 1')).toBeInTheDocument()
 
-      expect(committed).toEqual([
-        { pending: false, delayed: 0 },
-        { pending: true, delayed: 0 },
-        { pending: false, delayed: 1 },
-      ])
-    },
-  )
+    expect(committed).toEqual([
+      { pending: false, delayed: 0 },
+      { pending: true, delayed: 0 },
+      { pending: false, delayed: 1 },
+    ])
+  })
 
   it('can update normal atom with useTransition (#1151)', async () => {
     const countAtom = atom(0)


src/vanilla/atom.ts
```diff
@@ -58,7 +58,7 @@ export interface Atom<Value> {
 
 export interface WritableAtom<Value, Args extends unknown[], Result>
   extends Atom<Value> {
-  read: Read<Value, SetAtom<Args, Result>>
+  read: Read<Value, SetAtom<Args, void>>
```
dai-shi:
```suggestion
  read: Read<Value, SetAtom<Args, unknown>>
```

src/vanilla/atom.ts
```diff
@@ -75,7 +75,7 @@ let keyCount = 0 // global key count for all atoms
 
 // writable derived atom
 export function atom<Value, Args extends unknown[], Result>(
-  read: Read<Value, SetAtom<Args, Result>>,
+  read: Read<Value, SetAtom<Args, void>>,
```
dai-shi:
```suggestion
  read: Read<Value, SetAtom<Args, unknown>>,
```

src/vanilla/atom.ts
```diff
@@ -98,7 +98,7 @@ export function atom<Value>(
 ): PrimitiveAtom<Value> & WithInitialValue<Value>
 
 export function atom<Value, Args extends unknown[], Result>(
-  read?: Value | Read<Value, SetAtom<Args, Result>>,
+  read?: Value | Read<Value, SetAtom<Args, void>>,
```
dai-shi:
```suggestion
  read?: Value | Read<Value, SetAtom<Args, unknown>>,
```

src/vanilla/atom.ts
```diff
@@ -110,7 +110,7 @@ export function atom<Value, Args extends unknown[], Result>(
     },
   } as WritableAtom<Value, Args, Result> & { init?: Value | undefined }
   if (typeof read === 'function') {
-    config.read = read as Read<Value, SetAtom<Args, Result>>
+    config.read = read as Read<Value, SetAtom<Args, void>>
```
dai-shi:
```suggestion
    config.read = read as Read<Value, SetAtom<Args, unknown>>
```

dai-shi (changes requested):
Thanks for digging into it. Yeah, SetSelf is tricky, and I'm not super happy with it.
I think using `unknown` is more appropriate. (If not, `never`? But if `unknown` works, it feels safer.)

atomanyih:
@dai-shi Yeah, I flip flopped on this a couple times. My thinking was:

`() => void` means a function whose return value cannot/will not be used
whereas `() => unknown` means a function whose return value is not known, but can be used if it is cast or narrowed to a know type.
I think `() => never` would mean a function that _never_ returns, which wouldn't be correct.

But I think either `void` or `unknown` seems good and I'm happy to change it to `unknown`

dai-shi:
@atomanyih Sorry for the delay. On second thought, I think the original idea is to support this case.

```ts
it('shoud type setSelf return value', () => {
  function Component() {
    const numberAtom = atom(0)
    const writableAtom: WritableAtom<string, [], number> = atom(
      (_get, { setSelf }) => {
        const doubleNum = setSelf()
        return String(doubleNum + 1)
      },
      (get, _set) => {
        return get(numberAtom) * 2
      },
    )

    expectType<WritableAtom<string, [], number>>(writableAtom)
  }
  expect(Component).toBeDefined()
})
```

Can you reconsider the solution to support both? We may end up dropping either if we don't find any solutions.

tests/vanilla/types.test.tsx
```diff
@@ -67,3 +67,12 @@ it('type utils should work', () => {
   }
   expect(Component).toBeDefined()
 })
+
+it('WritableAtom Result type should be covariant', () => {
+  function Component() {
+    const writableAtom = atom(null, () => null)
```
dai-shi: This test doesn't represent your original reproduction. Can you do something similar to your repro in the PR description?

atomanyih: Which part of the original reproduction do you mean? `() => null` or `[someArg: number]`?

The `someArg` was actually a mistake. That wouldn't be assignable anyway (the arguments should be contravariant). I fixed the example in my local reproduction and updated the PR description

atomanyih:
> @atomanyih Sorry for the delay. On second thought, I think the original idea is to support this case.
> 
> ```ts
> it('shoud type setSelf return value', () => {
>   function Component() {
>     const numberAtom = atom(0)
>     const writableAtom: WritableAtom<string, [], number> = atom(
>       (_get, { setSelf }) => {
>         const doubleNum = setSelf()
>         return String(doubleNum + 1)
>       },
>       (get, _set) => {
>         return get(numberAtom) * 2
>       },
>     )
> 
>     expectType<WritableAtom<string, [], number>>(writableAtom)
>   }
>   expect(Component).toBeDefined()
> })
> ```
> 
> Can you reconsider the solution to support both? We may end up dropping either if we don't find any solutions.

Can you explain more about this use case? 
It looks like you have a derived atom that, when read, calls the its own set method?

tests/vanilla/types.test.tsx
dai-shi: I mean in this case, we could do:
```suggestion
    const writableAtom = atom(null, () => null as null | number)
```

I think we need the actual case in the test:
```ts
export const createWritableAtom = (): WritableAtom<
  null,
  [],
  number | null
> =>
  atom(
    () => null,
    (get, set, someArg: string) => {
      return null;
    },
  );
```

dai-shi:
That is a very exceptional use case, we hack reading other atoms in our "read" function. It's probably used in some of third-party libraries. So, if we change the type from `Return` to `unknown`, those library will be broken in types. Which is kind of acceptable, if this PR case is more important.

dai-shi:
btw, I want to drop `setSelf` in the future version... #2889

tests/vanilla/types.test.tsx
dai-shi: ☝️ Does it make sense?

dai-shi (changes requested):
I'm considering to merge this for the next patch.
@atomanyih Please check comments.

dai-shi:
@atomanyih Are you still around?

atomanyih:
Hi @dai-shi , sorry for the long delay. I haven't had a chance to work on this part of the project in a while. Let me take a look

tests/vanilla/types.test.tsx
atomanyih: I don't think I understand.

In this bug I was surfacing
```ts
export const createWritableAtom = (): WritableAtom<
  null,
  [],
  number | null
> =>
  atom(
    () => null,
    (get, set) => {
      return null;
    },
  );
```
We are saying that `WritableAtom<null, [], null>` should be assignable to `WritableAtom<null, [], number | null`.

And the test that I think reflects that
```ts
expectType<WritableAtom<null, [], number | null>>(
  atom(null, () => null)
)
```

To assert a similar test for functions: That `() => null` is assignable to `() => number | null`. You would write this, I think
```ts
expectType<() => number | null>(() => null)
```

Sorry for not understanding, can you explain what isn't correct?

dai-shi: I think I meant that the test should look like the bug your were having.
At least, it felt to me that it didn't.
So, the request was to make the test more like the actual use case.

> And the test that I think reflects that
> 
> ```ts
> expectType<WritableAtom<null, [], number | null>>(
>   atom(null, () => null)
> )
> ```

Ah, now I understand what the test means.
Thanks for your explanation.

dai-shi:
@atomanyih Could you resolve conflicts?

atomanyih:
@dai-shi No worries, it's my fault for taking so long to address it.

I merged and resolved the conflicts from the recent changes. I see that the type testing interface has changed. It seems like a clear translation, but can you double check that I am using the new utilities correctly? Thanks

dai-shi (changes requested):
Ah, we need to deal with old TypeScript versions. See CI errors.

tests/vanilla/types.test.tsx
```diff
@@ -71,3 +71,9 @@ it('type utils should work', () => {
     Promise<void>
   >()
 })
+
+it('WritableAtom Result type should be covariant', () => {
+  const writableAtom = atom(null, () => null)
+
+  expectTypeOf(writableAtom).toExtend<WritableAtom<null, [], number | null>>()
```
dai-shi, line 80:
```suggestion
  // NOTE: expectTypeOf is not available in TypeScript 4.0.5 and below, toExtend is not available in TypeScript 4.6.4 and below
  // [ONLY-TS-4.6.4] [ONLY-TS-4.5.5] [ONLY-TS-4.4.4] [ONLY-TS-4.3.5] [ONLY-TS-4.2.3] [ONLY-TS-4.1.5] [ONLY-TS-4.0.5] [ONLY-TS-3.9.7] [ONLY-TS-3.8.3] @ts-ignore
  expectTypeOf(writableAtom).toExtend<WritableAtom<null, [], number | null>>()
```

atomanyih:
> Ah, we need to deal with old TypeScript versions. See CI errors.

Thanks for looking at that

PR #3137, opened 2025-08-28
allow atom with default update function to return reset and add tests

scamden:

Fixes # https://github.com/pmndrs/jotai/discussions/3136

## Summary

Makes atomWithDefault behave the same way as atomWithReset in the sense that the update when passed as a function is allowed to return RESET. ~~It even now shares the same exact type~~. Also added some basic tests for atomWithDefault and this new functionality.

## Check List

- [ x] `pnpm run fix` for formatting and linting code and docs

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -6,9 +6,9 @@ jobs:
   compressed_size:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: pnpm/action-setup@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           node-version: 'lts/*'
           cache: 'pnpm'
--- a/.github/workflows/ecosystem-ci.yml
+++ b/.github/workflows/ecosystem-ci.yml
@@ -9,11 +9,11 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/ecosystem-ci run') }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           repository: 'jotaijs/jotai-ecosystem-ci'
       - uses: pnpm/action-setup@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           node-version: 22
           cache: 'pnpm'
--- a/.github/workflows/livecodes-preview.yml
+++ b/.github/workflows/livecodes-preview.yml
@@ -6,9 +6,9 @@ jobs:
   build_and_prepare:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: pnpm/action-setup@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
           node-version: 'lts/*'
           cache: 'pnpm'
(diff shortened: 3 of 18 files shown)


src/vanilla/utils/atomWithReset.ts
```diff
@@ -2,7 +2,7 @@ import { atom } from '../../vanilla.ts'
 import type { WritableAtom } from '../../vanilla.ts'
 import { RESET } from './constants.ts'
 
-type SetStateActionWithReset<Value> =
+export type SetStateActionWithReset<Value> =
```
dai-shi: It's our coding style not to export internal types. It's a bit tedious, but please define the same type in atomWithDefault.ts.

src/vanilla/utils/atomWithDefault.ts
```diff
@@ -9,37 +10,31 @@ type Read<Value, Args extends unknown[], Result> = WritableAtom<
...
+          return overwritten
+        }
+        return getDefault(get, options)
+      },
+      (get, set, update) => {
+        const newValue =
+          typeof update === 'function'
+            ? (update as (prev: Value) => Value)(get(anAtom))
+            : update
+        set(overwrittenAtom, newValue === RESET ? EMPTY : newValue)
```
dai-shi, line 40: Nice. Looks clean!

src/vanilla/utils/atomWithReset.ts
scamden: Oh ok happy to! I just had recreated this type in our code base (at 2 different companies now haha) and always wanted it to be exported. May I ask the reason we wouldn't want to?

scamden: ok went ahead and defined it, but still curious about your answer! 

Also added a couple more unit tests to clarify and enforce the behavior of the dynamic default getter.

src/vanilla/utils/atomWithDefault.ts
```diff
@@ -1,5 +1,5 @@
 import { atom } from '../../vanilla.ts'
-import type { SetStateAction, WritableAtom } from '../../vanilla.ts'
```
scamden, line 2: Not sure if it's useful but in our codebase we defined a second generic type parameter to SetStateAction that allows a different return type from update functions but defaults to the first type parameter to let us express things like the SetStateWithReset type:

`export type SetStateAction<Value, Update = Value> = Update | ((prevState: Value) => Update);`

so you can say

`type SetStateActionWithReset<Value> = SetStateAction<Value, Value | typeof RESET>`

Maybe that's over optimizing if reset is the only use case for a differing return type but just thought I'd mention it :)

src/vanilla/utils/atomWithReset.ts
dai-shi: In the past, some people depended on internal types, and when we changed the name of the internal type, they got confused. (It felt like a breaking change.) So, it's for us to freely change the internal types. And, this is due to our build process, which doesn't bundle type files unlike js files.

src/vanilla/utils/atomWithDefault.ts
dai-shi: Yeah, usually SetStateAction has the same type for the arg value and the return value.

tests/vanilla/utils/atomWithDefault.test.ts
dai-shi, line 1: Very nice.

dai-shi (approved):
Looks good. Thanks for adding the test.

I'll merge this when I prepare a new release.

dai-shi:
Oops, please check CI errors.

scamden:
> Oops, please check CI errors.

Whoops I'll format and resubmit

PR #3142, opened 2025-09-09
Update "Preview in LiveCodes" to use pkg.pr.new

hatemhosny:
Hi,
I noticed you are using pkg.pr.new github action.

I updated the template for "Preview in LiveCodes" to use it (through esm.sh)
e.g. https://esm.sh/pr/jotai@de789bf

This is simpler and lighter-weight than using data URLs for all the build files.

By the way, I'm not sure if you have had a look on [LiveCodes](https://livecodes.io/) recently. Lots of things have changed!
Docs: https://livecodes.io/docs

Cheers

### DIFF
--- a/.livecodes/react.json
+++ b/.livecodes/react.json
@@ -16,12 +16,8 @@
   "customSettings": {
     "jotai commit sha": "{{LC::SHORT_SHA}}",
     "imports": {
-      "jotai": "{{LC::TO_DATA_URL(./dist/esm/index.mjs)}}",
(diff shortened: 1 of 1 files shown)


dai-shi:
Oh, I like the standalone solution. If it depends on other services, it's not very useful because we can't use it as a fallback solution for the case when pkg.pr.new is down.
On the side note, users don't often use LiveCodes if pkg.pr.new exists, so we may consider dropping it and codesandbox as well from the workflow in the future. I think the good news is that users who like livecodes (like myself) can continue using it with pkg.pr.new as this PR does.

PR #3145, opened 2025-09-10
feat: only remove atoms from atomFamily when fully unmounted

dmaskasky:
https://github.com/pmndrs/jotai/discussions/3143#discussioncomment-14365117

## Summary
New behavior - `atomFamily.remove` and `atomFamily.setShouldRemove` marks the atom(s) for removal. Atoms marked for removal are only removed after they are unmounted from all stores they have previously been mounted in.

This change fixes the issue where an atom can be removed from an atomFamily while a React component is actively using that atom from the atomFamily. In this case, when the React component rerenders, it throws the TypeError "Invalid value used as weak map key" because `undefined` is now being passed to useAtom / useAtomValue.

## Check List

### DIFF
--- a/src/vanilla/utils/atomFamily.ts
+++ b/src/vanilla/utils/atomFamily.ts
@@ -1,4 +1,9 @@
 import { type Atom } from '../../vanilla.ts'
+import {
+  INTERNAL_getBuildingBlocksRev2 as INTERNAL_getBuildingBlocks,
+  INTERNAL_initializeStoreHooksRev2 as INTERNAL_initializeStoreHooks,
+  type INTERNAL_Store as Store,
+} from '../internals.ts'
 
 /**
  * in milliseconds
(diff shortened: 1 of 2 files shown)


src/vanilla/utils/atomFamily.ts
```diff
@@ -87,28 +173,27 @@ export function atomFamily<Param, AtomType extends Atom<unknown>>(
...
       for (const [key, [atom]] of atoms) {
         if (areEqual(key, param)) {
-          atoms.delete(key)
-          notifyListeners('REMOVE', key, atom)
+          markForRemoval(atom)
           break
         }
       }
     }
+    flushFullyUnmounted()
```
taylorcode, line 185: I'm likely missing something there, but shouldn't `flushFullyUnmounted` only be called when the subscribing component unmounts? My read of this logic is that `flushFullyUnmounted` is performed when `.remove()` is called.

dmaskasky: The call to flush here is to handle the case that the atom is already fully unmounted or has never been mounted. If the atom is still mounted, this is a no-op.

taylorcode (approved):

dai-shi:
wow, another use case of internal building blocks.
I'm not sure if I like the complexity of the implementation, but as I keep saying, I'm happy to see it extends for the community expectation. (If I were to implement it, I'd probably go with `useAtomFamily(aFamily, param)` which would avoid relying on the building blocks.)
As we discussed previously somewhere, this is the timing to move to [`jotai-family`](https://github.com/jotaijs/jotai-family).
Please publish the first version of `jotai-family` with the current implementation of `atomFamily`, and then apply this PR change and publish a new version.
I'll make a PR to deprecate `atomFamily` in `jotai/utils`.

PR #3149, opened 2025-09-20
feat(internals): External building blocks surface

dmaskasky:
https://github.com/jotaijs/jotai-scope/pull/89

## Summary
ScopeProvider must intercept all store api to provide resolveAtom functionality. atomEffect uses building blocks to react to changes to watched atoms. For atomEffects to run in a ScopeProvider, scopeProvider must provide resolveAtom functionality for all buildingBlocks.

Before jotai@2.14.0, we technically had two sets of buildingBlocks; the buildStore-scoped ones and the buildingBlocks stored behind the BUILDING_BLOCKS symbol on the store. With jotai@2.14.0, we collapsed this to a single set of buildingBlocks.

This change adds back the ability to treat buildingBlocks as an external api surface and allows for the transformation of those buildingBlocks.
1. We continue to store buildingBlocks on a WeakMap (instead of behind a symbol)
2. We continue to freeze both internal and external buildingBlocks.

```ts
// internals.ts
function getExternalBuildingBlocks(store: Store): Readonly<BuildingBlocks> {
  const buildingBlocks = getInternalBuildingBlocks(store)
  const enhanceBuildingBlocks = buildingBlocks[24]
  if (enhanceBuildingBlocks) {
    return enhanceBuildingBlocks(buildingBlocks)
  }
  return buildingBlocks
}

export {
  getExternalBuildingBlocks as INTERNAL_getBuildingBlocksRev2,
}

// Usage
const buildingBlocks = []
buildingBlocks[24] = function enhanceBuildingBlocks(internalBuildingBlocks) {
  const externalBuildingBlocks = [...internalBuildingBlocks]
  externalBuildingBlocks[21] = customStoreGet
  return externalBuildingBlocks
}
// store internally uses the default storeGet implementation
const store = buildStore(...buildingBlocks)
// enhanceBuildingBlocks allows us to shim the external interface of buildingBlocks
const storeBB = INTERNAL_getBuildingBlocksRev2(store)
storeBB[21].name // customStoreGet
```

## Check List

### DIFF
--- a/.github/workflows/ecosystem-ci.yml
+++ b/.github/workflows/ecosystem-ci.yml
@@ -30,10 +30,11 @@ jobs:
         id: run_command
         run: |
           echo "x<<EOF" >> $GITHUB_OUTPUT
-          pnpm run ecosystem-ci | grep -A999 -- '---- Jotai Ecosystem CI Results ----' >> $GITHUB_OUTPUT
+          pnpm run ecosystem-ci | tee >(grep -A999 -- '---- Jotai Ecosystem CI Results ----' >> $GITHUB_OUTPUT)
           echo "EOF" >> $GITHUB_OUTPUT
         env:
           JOTAI_PKG: https://pkg.csb.dev/pmndrs/jotai/commit/${{ steps.short_sha.outputs.x }}/jotai
+          VERBOSE: 1
       - uses: peter-evans/create-or-update-comment@v4
         with:
           issue-number: ${{ github.event.issue.number }}
--- a/docs/utilities/storage.mdx
+++ b/docs/utilities/storage.mdx
@@ -47,7 +47,7 @@ The `atomWithStorage` function creates an atom with a value persisted in `localS
 
 If not specified, the default storage implementation uses `localStorage` for storage/retrieval, `JSON.stringify()`/`JSON.parse()` for serialization/deserialization, and subscribes to `storage` events for cross-tab synchronization.
 
-<Stackblitz id="vitejs-vite-wewrmi" file="src%2FApp.tsx" />
+<Stackblitz id="vitejs-vite-xvlgfxrk" file="src%2FApp.tsx" />
 
 ### `createJSONStorage` util
 


dmaskasky:
> ## Ecosystem CI Output
> ```
> ---- Jotai Ecosystem CI Results ----
> {
>   "bunshi": "PASS",
>   "jotai-devtools": "FAIL",
>   "jotai-effect": "PASS",
>   "jotai-scope": "PASS",
>   "waku-jotai": "PASS"
> }
> ```

Something is not right with Jotai Ecosystem CI Results. I expect jotai-scope to fail until I refactor jotai-scope to use this PR.

dai-shi:
> Something is not right with Jotai Ecosystem CI Results. I expect jotai-scope to fail until I refactor jotai-scope to use this PR.

Does jotai-scope fail with the main branch tests? I thought the failing tests are only in some branch.

dmaskasky:
Oh nevermind, I just realized I disabled the atomEffect scope tests.

src/vanilla/internals.ts
```diff
@@ -155,33 +158,34 @@ type Store = {
...
+  unmountAtom: UnmountAtom, //                                 19
+  setAtomStateValueOrPromise: SetAtomStateValueOrPromise, //   20
   // store api
-  storeGet: StoreGet, //                                     21
-  storeSet: StoreSet, //                                     22
-  storeSub: StoreSub, //                                     23
+  storeGet: StoreGet, //                                       21
+  storeSet: StoreSet, //                                       22
+  storeSub: StoreSub, //                                       23
+  enhanceBuildingBlocks: EnhanceBuildingBlocks | undefined, // 24
```
dai-shi, line 188: I wonder if this can be something like this:
```suggestion
  alternateBuildingBlocks: Readonly<BuildingBlocks> | undefined, // 24
```

(I'm not sure if "alternate" is the right word. "real"? "overriding"?)

src/vanilla/internals.ts
```diff
@@ -930,14 +934,23 @@ const storeSub: StoreSub = (store, atom, listener) => {
...
   if (import.meta.env?.MODE !== 'production' && !buildingBlocks) {
     throw new Error(
       'Store must be created by buildStore to read its building blocks',
     )
   }
-  return buildingBlocks!
+  return buildingBlocks
+}
+
+function getExternalBuildingBlocks(store: Store): Readonly<BuildingBlocks> {
```
dai-shi: This function isn't used here. I guess we can move this function outside of this repo, to `jotai-scope`.

src/vanilla/internals.ts
dmaskasky:
```ts
const buildingBlocks = [new WeakMap()]
buildingBlocks[24] = [
  // ScopeProvider needs access to
  // all the internal building blocks
  // here so that it can shim them
  // but they are created internal
  // to buildStore and not exposed
]
const store = buildStore(...buildingBlocks)
```

src/vanilla/internals.ts
dmaskasky: Its exported as INTERNAL_getBuildingBlocksRev2. Libraries like jotai-effect would import this and call it.

Wendystraite:
Could also be used by jotai-logger since it also patch the store get/set/sub and state.set (see the current hack in [bind-atoms-logger-to-store.ts#L38-L56](https://github.com/Wendystraite/jotai-logger/blob/ca900853fae440119ab8c54965c651fd5ac0ded5/src/bind-atoms-logger-to-store.ts#L38-L56)).

src/vanilla/internals.ts
dai-shi: That makes totally sense. I missed that point. Can you add a tiny test that requires this case?

src/vanilla/internals.ts
dai-shi: But this is basically a pure function. It just depends on the original `getBuildingBlocks`. Why can't it be defined in jotai-effect?

dmaskasky: Because jotai-scope internal building blocks would be exposed. jotai-effect should not have to care about this implementation detail. There are no TypeScript or runtime checks to help with this either.

dai-shi: So are you saying to have this as a contract here? Do you agree that it's technically possible to define this function in jotai-effect?

dmaskasky: I think its an important encapsulation. While Jotai Effect could technically have this, it would break this encapsulation.

I just don't want to have issues such as "ScopeProvider doesn't work with library-X". I think other libraries such as devTools could benefit from this too.

src/vanilla/internals.ts
```diff
@@ -930,14 +934,23 @@ const storeSub: StoreSub = (store, atom, listener) => {
 
 const buildingBlockMap = new WeakMap<Store, Readonly<BuildingBlocks>>()
 
-function getBuildingBlocks(store: Store): Readonly<BuildingBlocks> {
-  const buildingBlocks = buildingBlockMap.get(store)
+const getBuildingBlocks = (store: Store): Readonly<BuildingBlocks> => {
```
dai-shi: Can we name this like `getBaseBuildingBlocks`? Or, `getInternalBuildingBlocks`?

dmaskasky:
> Could also be used by jotai-logger since it also patch the store get/set/sub and state.set (see the current hack in [bind-atoms-logger-to-store.ts#L38-L56](https://github.com/Wendystraite/jotai-logger/blob/ca900853fae440119ab8c54965c651fd5ac0ded5/src/bind-atoms-logger-to-store.ts#L38-L56)).

I think the best way to add in custom storeGet, storeSet, and storeSub functions would be to create a new store derived from the buildingBlocks of the target store rather than mutating buildingBlocks. How do you feel about prohibiting mutating buildingBlocks with Object.freeze?

src/vanilla/internals.ts
dai-shi: Okay, I understand it. It's the same as `initializeStoreHooks`, which isn't used in the core, but to provide a contract or an encapsulation.

dai-shi: Okay, I understand it. It's the same as `initializeStoreHooks`, which isn't used in the core, but to provide a contract or an encapsulation.

src/vanilla/internals.ts
```diff
@@ -996,7 +1010,7 @@ export {
   // Export internal functions
   //
   buildStore as INTERNAL_buildStoreRev2,
-  getBuildingBlocks as INTERNAL_getBuildingBlocksRev2,
+  getExternalBuildingBlocks as INTERNAL_getBuildingBlocksRev2,
```
dai-shi: Let's simply call the new one `getBuildingBlocks` as the user of this function doesn't care if it's internal or enhanced. It's also consistent for readability.
```suggestion
  getBuildingBlocks as INTERNAL_getBuildingBlocksRev2,
```

src/vanilla/internals.ts
```diff
@@ -930,28 +934,37 @@ const storeSub: StoreSub = (store, atom, listener) => {
...
   }
-  return buildingBlocks!
+  return buildingBlocks
+}
+
+function getExternalBuildingBlocks(store: Store): Readonly<BuildingBlocks> {
+  const buildingBlocks = getInternalBuildingBlocks(store)
+  const enhanceBuildingBlocks = buildingBlocks[24]
+  if (enhanceBuildingBlocks) {
+    return enhanceBuildingBlocks(buildingBlocks)
```
dai-shi, line 951: I wonder if we should recursively enhance building blocks util it hits `undefined`. Or, is it _correct_ to apply enhanceBuildingBlocks just once?

dai-shi (commented):
Thanks for your work and explaining.
I may have some more requests and questions, but here're some so far.

src/vanilla/internals.ts
dmaskasky: I had similar questions regarding ensuring buildingBlocks
1. is not sparse with:
```ts
const enhanced = enhanceBuildingBlocks(buildingBlocks)
Object.assign(enhanced, buildingBlocks, enhanced)
```
2. is frozen with:
```ts
Object.freeze(enhanceBuildingBlocks(buildingBlocks))
```

But I think I would leave it up to user-land to handle this.

---

My thinking is that store configuration is defined just once when calling `buildStore`, and store config is not mutable.

I was also thinking of proposing Object.freeze on the store itself, but that seems more of an unfriendly breaking change.

dai-shi: > But I think I would leave it up to user-land to handle this.

That's related to my original thinking of that we don't even need this to be exported here.
But, you should have deeper understanding of the problem and the requirement, so I'll just count on you.

For now, I assume the user-land code should do the recursion if necessary.

dai-shi (approved):
LGTM

src/vanilla/internals.ts
dmaskasky: Just to clarify this, there is the store creator userland aka jotai-scope, then there is store consumer userland aka jotai-effect. We will leave implementation of the external building blocks up to the store creator, and we will require the store consumer to use the getBuildingBlocks api to read the buildingBlocks.

dai-shi: Yeah, I understand it. My question was probably for the case when there are two or more store creator libraries. Never mind for now.

dmaskasky: Its a hard problem. My hope is that nested storeCreators would derive from the parent and mixin their capability. But I'm pretty sure this would not be the case.

Wendystraite:
> > Could also be used by jotai-logger since it also patch the store get/set/sub and state.set (see the current hack in [bind-atoms-logger-to-store.ts#L38-L56](https://github.com/Wendystraite/jotai-logger/blob/ca900853fae440119ab8c54965c651fd5ac0ded5/src/bind-atoms-logger-to-store.ts#L38-L56)).
> 
> I think the best way to add in custom storeGet, storeSet, and storeSub functions would be to create a new store derived from the buildingBlocks of the target store rather than mutating buildingBlocks. How do you feel about prohibiting mutating buildingBlocks with Object.freeze?

You are right about deriving the store. I think I got it wrong when I wanted the jotai-logger API to be like jotai-devtools but didn’t use `INTERNAL_overrideCreateStore`. I’ll try to change that.

`Object.freeze` is a good tool for APIs that want to ensure they are used as intended.
For example, it could have prevented me from patching the store instead of deriving it.

Using `Object.freeze` for buildingBlocks and for the store is a good idea.. if the internal APIs are able to expose all the necessary hooks for valid extensions or overrides without needing mutation.
The same thing apply for `Object.seal` but that may be overkill 😄

PR #3154, opened 2025-09-21
idea: addOnInit function to replace atom.unstable_onInit

dmaskasky:
## Summary
The issue with atom.unstable_onInit is that there can only be one per atom. This is incompatible with composable patterns such as mixins that incrementally add new functionality. atom.unstable_onInit is also mutable.

This PR explores a simple function for adding onInit functions.

## Check List


dai-shi:
> The issue with atom.unstable_onInit is that there can only be one per atom.

This is the designed intended limitation. We shouldn't overuse `unstable_onInit`. I guess most use cases should be covered by deriving atoms.

dai-shi:
On second thought, I don't completely disagree with this idea.
`INTERNAL_addOnInit(anAtom, fn)` is more explicit about that it's only for lib authors than `atom.unstable_onInit`. It's considerable, especially if we never plan to remove `unstable_` prefix from it.
(Though, having addOnInit in internals.ts feels uncomfortable as it's not a store hook. API-wise, `atom.unstable_onInit` feels cleaner. `INTERNAL_storeAtom` might be simpler implementation-wise, but not sure if it's cool.)

dai-shi:
👉 #3159

dmaskasky:
I think INTERNAL_onInit would be better than INTERNAL_storeAtom. It provides onInit hook timing and encourages not mixing store with atom read.

PR #3163, opened 2025-10-06
fix(unwrap): recover from errors

darrenclark:
When `unwrap` encountered an error, it would get stuck in the error state.  This commit fixes that.

Discussion https://github.com/pmndrs/jotai/discussions/3162

## Summary

Discussion goes into more detail above, but basically an `unwrap` atom gets stuck in an error state via:

1. underlying `Promise` fails
2. the internal `promiseAndValueAtom` saves an error
3. then in the future, when `promiseAndValueAtom` reads it's previous value, the original error gets thrown
4. goes back to 2. above

To break this loop, added a try/catch around reading the previous value.

## Check List

### DIFF
--- a/src/vanilla/utils/unwrap.ts
+++ b/src/vanilla/utils/unwrap.ts
@@ -58,7 +58,12 @@ export function unwrap<Value, Args extends unknown[], Result, PendingValue>(
       } = atom(
         (get, { setSelf }) => {
           get(refreshAtom)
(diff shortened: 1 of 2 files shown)


darrenclark:
Happy to help, thanks @dai-shi !

meow2149:
> Happy to help, thanks @dai-shi !

Thanks too

PR #3165, opened 2025-10-08
investigate side-effect regression

dmaskasky:
https://github.com/pmndrs/jotai/discussions/3164

Fixes #

## Summary
- add failing test
- revert https://github.com/pmndrs/jotai/commit/f5d843c87842df5a9c034ef54bf2a77b367d0c36

I don't think we should merge this PR. I don't think Jotai should support side-effects in atom read.

The use-case which spawned this investigation was with
```ts
function atomShallowEqual<T>(fn: (get: Getter) => T): Atom<T> {
  const previousValueAtom = atom<T | void>();
  return atom((get) => {
    const jotaiStore = getDefaultStore();
    const oldValue = jotaiStore.get(previousValueAtom);
    const newValue = fn(get);
    jotaiStore.set(previousValueAtom, newValue);
    if (!oldValue) {
      return newValue;
    }
    return shallowEquals(newValue, oldValue) ? oldValue : newValue;
  });
}
```

I think there could be a better pattern for above. I haven't found a pattern that supports side-effects in read.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -457,6 +457,7 @@ const recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (store) => {
   const mountedMap = buildingBlocks[1]
   const invalidatedAtoms = buildingBlocks[2]
   const changedAtoms = buildingBlocks[3]
+  const storeHooks = buildingBlocks[6]
   const ensureAtomState = buildingBlocks[11]
   const readAtomState = buildingBlocks[14]
   const mountDependencies = buildingBlocks[17]
@@ -465,7 +466,11 @@ const recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (store) => {
   // This is a topological sort via depth-first search, slightly modified from
   // what's described here for simplicity and performance reasons:
   // https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search
-  const topSortedReversed: [atom: AnyAtom, atomState: AtomState][] = []
+  const topSortedReversed: [
+    atom: AnyAtom,
+    atomState: AtomState,
+    epochNumber: EpochNumber,
+  ][] = []
   const visiting = new WeakSet<AnyAtom>()
   const visited = new WeakSet<AnyAtom>()
   // Visit the root atom. This is the only atom in the dependency graph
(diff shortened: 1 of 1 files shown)


dmaskasky:
> ` "jotai-effect": "FAIL",`

I expected this.

dai-shi:
(I'll have a look later.)
(Sorry for the delay, not on it yet.)

tests/vanilla/store.test.tsx
```diff
@@ -1270,3 +1270,19 @@ it('updates dependents when it eagerly recomputes dirty atoms', () => {
...
   expect(store.get(activeCountAtom)).toBe(1)
 })
+
+it('triggers the sub when setting inside an atom read', () => {
+  const store = createStore()
+  const basicAtom = atom(0)
+  const previousValueAtom = atom()
+  const testAtom = atom((get) => {
+    const newValue = get(basicAtom)
+    store.set(previousValueAtom, newValue)
```
dai-shi, line 1280: We don't need to support this.
If we can't support this, we should probably warn it (dev only) if we can detect it.

dai-shi:
> The use-case which spawned this investigation was with
> 
> ```ts
> function atomShallowEqual<T>(fn: (get: Getter) => T): Atom<T> {
>   const previousValueAtom = atom<T | void>();
>   return atom((get) => {
>     const jotaiStore = getDefaultStore();
>     const oldValue = jotaiStore.get(previousValueAtom);
>     const newValue = fn(get);
>     jotaiStore.set(previousValueAtom, newValue);
>     if (!oldValue) {
>       return newValue;
>     }
>     return shallowEquals(newValue, oldValue) ? oldValue : newValue;
>   });
> }
> ```

My suggestion is something like this:
```js
function atomShallowEqual(initialValue) {
  const valueAtom = atom(initialValue);
  return atom(
    (get) => get(valueAtom),
    (get, set, newValue) => {
      const oldValue = get(valueAtom);
      if (!shallowEqual(oldValue, newValue)) {
        set(valueAtom, newValue);
      }
    }
  );
}
```

So, the challenge is how we can do with `fn: (get: Getter) => T`.
If the requirement is only about `shallowEqual`, and if we only need to support one jotai store, we could simply use a scope variable.

```js
function atomShallowEqual(fn) {
  let lastValue;
  return atom((get) => {
    const value = fn(get);
    if (!shallowEqual(lastValue, value)) {
      lastValue = value;
    }
    return lastValue;
  });
}
```

It still violates purity, but it feels like a better hack.

PR #3173, opened 2025-10-31
chore: update dev dependencies

dai-shi:
for v2.15.1.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -1,0 +1,21 @@
+## Getting Started
+
+Install the dependencies:
+
+```bash
+pnpm install --ignore-workspace
+```
+
+Make a cache directory:
+
+```bash
+mkdir .cache
+```
+
+Run the development server:
+
+```bash
+pnpm run dev
+```
+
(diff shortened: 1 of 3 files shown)


src/react/Provider.ts
```diff
@@ -26,16 +26,21 @@ export function Provider({
...
 }): ReactElement<
   { value: Store | undefined },
-  FunctionComponent<{ value: Store | undefined }>
+  FunctionComponent<{ value: Store }>
 > {
-  const storeRef = useRef<Store>(undefined)
-  if (!store && !storeRef.current) {
+  const storeRef = useRef<Store>(null)
+  if (store) {
+    return createElement(StoreContext.Provider, { value: store }, children)
```
dai-shi, line 33: Technically, this is a change in the library code.

dai-shi:
92bdc7e
@Aslemammad We can't upgrade to vitest 4. Would you help?

dai-shi:
Hmmm, the last commit is equivalent with e1f6fe0. Why does it work now...

PR #3179, opened 2025-11-17
fix(internals): add dev-only warning for unsupported side effect

dai-shi:
close #3165

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -6,7 +6,7 @@ jobs:
   compressed_size:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v6
         with:
--- a/.github/workflows/ecosystem-ci.yml
+++ b/.github/workflows/ecosystem-ci.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/ecosystem-ci run') }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           repository: 'jotaijs/jotai-ecosystem-ci'
       - uses: pnpm/action-setup@v4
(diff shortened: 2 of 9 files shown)


src/vanilla/internals.ts
```diff
@@ -523,6 +523,9 @@ const recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (store) => {
   }
 }
 
+// Dev only
+const storeMutationMap: WeakMap<Store, number> = new WeakMap()
```
dai-shi: I wonder if there's a more straightforward way than introducing this.

src/vanilla/internals.ts
```diff
@@ -700,6 +714,9 @@ const writeAtomState: WriteAtomState = (store, atom, ...args) => {
           // NOTE technically possible but restricted as it may cause bugs
           throw new Error('atom not writable')
         }
+        if (import.meta.env?.MODE !== 'production') {
+          storeMutationMap.set(store, (storeMutationMap.get(store) || 0) + 1)
```
dmaskasky: Can we move this to storeSet instead?

dai-shi: Yeah, I thought about it too. We could, but there can be "pure" write atom. Also, I thought there would be a way to bypass `store.set`.

I'm fine either way. Do you have pros and cons?

Also, I wonder if you have any other idea about the implementation.

dmaskasky: I wonder if atomEffect is affected by this. If you change it to throw, does CI still pass for jotai-effect?

I can also disable this warning if necessary.

dmaskasky: I would have implemented something similar.

dai-shi: > If you change it to throw, does CI still pass for jotai-effect?

I'll try it later (in a few days).

dai-shi:
Oh, it's because we use csb...

dai-shi:
> If you change it to throw, does CI still pass for jotai-effect?

Yes, it looks good.

src/vanilla/internals.ts
dai-shi: https://github.com/pmndrs/jotai/pull/3179#issuecomment-3555399999

PR #3181, opened 2025-11-19
fix(internals): add storeHooks.i to be called when atomState is created

dmaskasky:
## Summary
Add a `storeHooks.i` to be called when atomState is created.

## Check List


dmaskasky:
would `storeHooks.a` be better?

PR #3186, opened 2025-12-04
fix(internals): Minor tweaks to store internals

dmaskasky:
## Summary
Minor tweaks to store internals.

## Changes
1. Unify weak map types and set types
  - adds `has` and `delete` to AtomStateMap type
  - adds `delete` to ChangedAtoms type
2. Add `callbacks` prop to StoreHook and StoreHookForAtoms
3. Relocates `storeHooks.u?.(atom)` to below `aMounted?.t.delete(atom)`

## Check List

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -12,6 +12,6 @@ jobs:
         with:
           node-version: 'lts/*'
           cache: 'pnpm'
-      - uses: preactjs/compressed-size-action@v2
+      - uses: preactjs/compressed-size-action@v3
         with:
           pattern: './dist/**/*.{js,mjs}'
--- a/src/react/Provider.ts
+++ b/src/react/Provider.ts
@@ -38,7 +38,7 @@ export function Provider({
   return createElement(
     StoreContext.Provider,
     {
-      // TODO should we use useState instead?
+      // TODO: If this is not a false positive, consider using useState instead of useRef like https://github.com/pmndrs/jotai/pull/2771
       // eslint-disable-next-line react-hooks/refs
       value: storeRef.current,
     },


src/vanilla/internals.ts
```diff
@@ -320,17 +312,22 @@ function getMountedOrPendingDependents(
 type StoreHook = {
   (): void
   add(callback: () => void): () => void
+  callbacks: SetLike<() => void>
```
dai-shi: To me, this feels kind of overwrapping.
If we aren't be afraid of breaking changes (which is the purpose of "internals" and "Rev"), I hope this to be:
```suggestion
  hooks: SetLike<() => void>
```

src/vanilla/internals.ts
```diff
@@ -320,17 +312,22 @@ function getMountedOrPendingDependents(
...
 }
 
 type StoreHookForAtoms = {
   (atom: AnyAtom): void
   add(atom: AnyAtom, callback: () => void): () => void
   add(atom: undefined, callback: (atom: AnyAtom) => void): () => void
+  callbacks: WeakMapLike<
+    AnyAtom | Record<any, never>,
+    SetLike<(atom?: AnyAtom) => void>
+  >
```
dai-shi: This is tricker, but still:
```suggestion
  hooksMap: WeakMapLike<
    AnyAtom | Record<any, never>,
    SetLike<(atom?: AnyAtom) => void>
  >
```

dai-shi: It feels like there should be a better way though.

dai-shi (commented):
I'm not exactly sure what are reasons for the changes, but I don't mind much. Though, I do care the minimalist API.

src/vanilla/internals.ts
dmaskasky: This will have to be a new "Rev", but I agree.

src/vanilla/internals.ts
dmaskasky: Actually, I like what we already have more. I added callbacks because I wanted to hook in with my test debug tools, not for any production code. But I think there are better ways to do this. I'll remove these changes from the PR.

dai-shi:
This is more breaking than I expected. Do we really need all methods in WeakMap and Set?

dmaskasky:
`delete` is needed for O(1) removal by jotai-effect and jotai-scope. Happy to discuss, and this can wait untill next minor if you prefer.

CI Fixes:
- jotai-effect@2.1.6 is out now (fixed).
- Here is the PR for waku-jotai: https://github.com/wakujs/waku-jotai/pull/3

dai-shi:
> `delete` is needed for O(1) removal by jotai-effect and jotai-scope.

I just wondered if this is to simplify types or new methods are required. Good to know that.

Adding more methods means larger contract. So, I prefer keeping them minimal.

src/vanilla/internals.ts
```diff
@@ -57,37 +57,29 @@ type Mounted = {
...
+type SetLike<T> = {
   readonly size: number
-  add(fn: () => void): void
+  add(value: T): void
+  has(value: T): boolean
+  delete(value: T): boolean
   clear(): void
-  forEach(callback: (fn: () => void) => void): void
+  forEach(callback: (value: T) => void): void
+  [Symbol.iterator](): IterableIterator<T>
```
dai-shi, line 74: In that sense, I wonder if this is ever used. Can remove otherwise?

dmaskasky: `[Symbol.iterator]()` isn't used but was already there for ChangedAtoms type. Shall I remove it?

dai-shi: Yeah, let's remove it for this time.

dai-shi: Ohhh, we actually use it at L471.

dai-shi: So, I think we somewhat intentionally changed ChangedAtoms and Callbacks types.

dai-shi: Well, I can give it up then. This is unstable internal APIs anyway.
Revert 9c65140 and good to go. We can change it in the future.

dai-shi (approved):

PR #3187, opened 2025-12-05
feat(utils): deprecate atomFamily in `jotai/utils` (migrate to `jotai-family`)

dmaskasky:
## Summary
- adds depreciate notice for atomFamily (atomFamily will be removed in V3).
- fixes issues found in the atomFamily website docs.

## Check List

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -12,6 +12,6 @@ jobs:
         with:
           node-version: 'lts/*'
           cache: 'pnpm'
-      - uses: preactjs/compressed-size-action@v2
+      - uses: preactjs/compressed-size-action@v3
         with:
           pattern: './dist/**/*.{js,mjs}'
--- a/src/react/Provider.ts
+++ b/src/react/Provider.ts
@@ -38,7 +38,7 @@ export function Provider({
   return createElement(
     StoreContext.Provider,
     {
-      // TODO should we use useState instead?
+      // TODO: If this is not a false positive, consider using useState instead of useRef like https://github.com/pmndrs/jotai/pull/2771
       // eslint-disable-next-line react-hooks/refs
       value: storeRef.current,
     },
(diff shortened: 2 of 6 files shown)


dai-shi (commented):
Could you add runtime dev-only one-time deprecation warning?

Use `didWarnDeprecation` flag like https://github.com/pmndrs/zustand/blob/adce5e6f2adff35994b8b9c9d04aa6a6dbee5e77/src/middleware/devtools.ts#L277

docs/utilities/family.mdx
```diff
@@ -6,6 +6,30 @@ keywords: family
...
+`atomFamily` is deprecated and will be removed in Jotai v3.
+
+Please migrate to the [`jotai-family`](https://github.com/jotaijs/jotai-family) package, which provides the same API with additional features like `atomTree`.
+
+**Migration:**
+
+```bash
+npm install jotai-family
+# or
+pnpm add jotai-family
```
dai-shi: I would stick with `npm install` only.
(Otherwise, people start adding other tools.)

src/vanilla/utils/atomFamily.ts
```diff
@@ -24,6 +39,23 @@ export interface AtomFamily<Param, AtomType> {
   unstable_listen(callback: Callback<Param, AtomType>): Cleanup
 }
 
+let didWarnDeprecation = false
+
+/**
+ * @deprecated atomFamily is deprecated and will be removed in v3.
+ * Please use the `jotai-family` package instead: https://github.com/jotaijs/jotai-family
+ *
+ * Install: `npm install jotai-family` or `pnpm add jotai-family`
```
dai-shi: here too, and there too.

dmaskasky: Oops, looks like I missed those.

dai-shi (approved):
LGTM!

songkeys:
may i ask the reason behind this? is it just for the pkg size?

dai-shi:
Yes, I want to keep the package small, not for the size, but for maintaining.
atomFamily started as a simple util, but there are many requirements which can't be handled here (mostly due to the project scope as well as my capacity), so separating it will speed up the development allowing adding more features.

PR #3188, opened 2025-12-05
test: extract 'sleep' and 'useCommitCount' to 'test-utils'

sukvvon:

Fixes #

## Check List


dai-shi:
I didn't notice this because GitHub doesn't notify when making PR ready from draft .

I don't feel very necessary for this change particularly for `sleep`, but it's harmless anyway. I may change something in the future.

PR #3189, opened 2025-12-05
fix(internals): only mounted dependents can be added to mounted.t

dmaskasky:
## Summary
Ensure only mounted dependents can be added to `mounted.t`. Previously, unmounted atoms were being added by mistake. This was not caught due to a guard in `getMountedOrPendingDependents`. The new approach is more efficient and may prevent a memory leak.

## Check List

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -12,6 +12,6 @@ jobs:
         with:
           node-version: 'lts/*'
           cache: 'pnpm'
-      - uses: preactjs/compressed-size-action@v2
+      - uses: preactjs/compressed-size-action@v3
         with:
           pattern: './dist/**/*.{js,mjs}'
--- a/docs/utilities/family.mdx
+++ b/docs/utilities/family.mdx
@@ -6,6 +6,28 @@ keywords: family
 
 ## atomFamily
 
+:::caution Deprecated
+`atomFamily` is deprecated and will be removed in Jotai v3.
+
+Please migrate to the [`jotai-family`](https://github.com/jotaijs/jotai-family) package, which provides the same API with additional features like `atomTree`.
+
+**Migration:**
+
+```bash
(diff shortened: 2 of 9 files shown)


dai-shi:
pls check failing ci

dmaskasky:
> Is the change from function style to for-break style for performance?

`Array.from(atomState.d).every` and `Array.from(atomState.d).some` allocates an extra array into memory. Its a slight performance bump to iterate over it directly.

---

The main purpose of this PR is actually [L599](https://github.com/pmndrs/jotai/pull/3189/files#diff-e93c16aa25ffe514c1401b80b686250aaad3f79f98db6116c01a0d980de82011R599). We were accidentally adding unmounted atoms to `.t` prop. This was only not an issue because [L306](https://github.com/pmndrs/jotai/pull/3189/files#diff-e93c16aa25ffe514c1401b80b686250aaad3f79f98db6116c01a0d980de82011L306) was filtering out unmounted atoms. But the comment for `t` says "Set of mounted atoms that depends on the atom.". Its cleaner and more efficient to filter out unmounted atoms before adding them to `t`.

dai-shi:
> The main purpose of this PR is actually [L599](https://github.com/pmndrs/jotai/pull/3189/files#diff-e93c16aa25ffe514c1401b80b686250aaad3f79f98db6116c01a0d980de82011R599). We were accidentally adding unmounted atoms to `.t` prop. This was only not an issue because [L306](https://github.com/pmndrs/jotai/pull/3189/files#diff-e93c16aa25ffe514c1401b80b686250aaad3f79f98db6116c01a0d980de82011L306) was filtering out unmounted atoms. But the comment for `t` says "Set of mounted atoms that depends on the atom.". Its cleaner and more efficient to filter out unmounted atoms before adding them to `t`.

Nice.

dai-shi (approved):

PR #3191, opened 2025-12-07
docs(react/Provider): clarify TODO comment related with useRef usage

sukvvon:

N/A

## Summary

- Replace TODO comment with NOTE comment in `Provider` component
- Clarify why `useRef` is used instead of `useState` for store initialization

## Why `useRef` instead of `useState`?

The original TODO comment asked: "should we use useState instead?"

The answer is **no** - `useRef` is the correct choice here for performance reasons.

### The Problem with `useState`

```typescript
// If we used useState:
const [storeState] = useState(() => createStore()) // Always runs on first render

if (store) {
  return createElement(StoreContext.Provider, { value: store }, children)
}
return createElement(StoreContext.Provider, { value: storeState }, children)
```

With `useState`, the initializer function `() => createStore()` **always runs on the first render**, even when the `store` prop is provided. This means `createStore()` would be called unnecessarily.

### The Solution with `useRef`

```typescript
// Current implementation:
const storeRef = useRef<Store>(null)
if (store) {
  // Early return - createStore() is never called
  return createElement(StoreContext.Provider, { value: store }, children)
}
if (storeRef.current === null) {
  storeRef.current = createStore() // Only called when store prop is not provided
}
```

With `useRef`, we can use an early return pattern to **completely skip** the `createStore()` call when the `store` prop is provided.

## Check List


src/react/Provider.ts
```diff
@@ -38,7 +38,9 @@ export function Provider({
   return createElement(
     StoreContext.Provider,
     {
-      // TODO should we use useState instead?
+      // NOTE: Using useRef instead of useState to avoid calling createStore()
+      // when the store prop is provided. useState's initializer always runs on
+      // the first render, but useRef with early return allows us to skip it.
```
dai-shi:
```suggestion
      // TODO: If this is not a false positive, consider using useState instead of useRef like https://github.com/pmndrs/jotai/pull/2771
```

sukvvon: @dai-shi Do you want to change it to what you suggested?

dai-shi: Yes, if it solves your original concern.

dai-shi (approved):

PR #3197, opened 2025-12-12
fix(internals): buildingblocks should not invoke buildingblock definitions from top-scope

dmaskasky:
## Summary
Building blocks should be derived from the store arg, not referenced from top-scope. This change fixes this implementation error.

## Check List


dai-shi:
>   "jotai-effect": "FAIL",

@dmaskasky Is it expected?

PR #3198, opened 2025-12-12
[Experimental][WIP] New buildingBlock createGetter

dmaskasky:
## Summary
This new buildingBlock makes it easier for library maintainers to shim the getter from readAtomState. This change is just an experiment at this point.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -141,7 +141,7 @@ type EnhanceBuildingBlocks = (
 type CreateGetter = <S extends Store, A extends AnyAtom>(
   store: S,
   atom: A,
-  isSync: boolean,
+  getIsSync: () => boolean,
 ) => Getter
 
 type Store = {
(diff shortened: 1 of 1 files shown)


src/vanilla/internals.ts
```diff
@@ -580,36 +634,8 @@ const readAtomState: ReadAtomState = (store, atom) => {
...
-      }
-      if (mountedMap.has(atom)) {
-        mountedMap.get(a)?.t.add(atom)
-      }
-      if (!isSync) {
-        mountDependenciesIfAsync()
-      }
-    }
-  }
+  let getter = createGetter(store, atom, true)
```
dai-shi: Not a big fan of `let` here. Could we do like this?
```js
const getter = createGetter(store, atom, () => isSync)
```

dmaskasky:
I'm not super confident on this api yet. Should we hold off?

dai-shi:
I'm not confident either. It feels a bit over-complicated, particularly leaking `sync` flag.

PR #3199, opened 2025-12-13
[Experimental] recomputeDependents respects scoped atom native store

dmaskasky:
## Summary
When working with scopes, if the base store makes a change that causes scoped atom dependents to be invalidated, recomputeDependents can call `readAtomState(baseStore, scopedAtom)`. This can be confusing and causes issues that are unfixable by jotai-scope.

This PR changes the implementation of recomputeDependents to respect the scoped atom's native store when calling readAtomState.

## Core Change
`mounted.d` and `mounted.t` are a map of the dependencies / dependents to their respective stores. An atom + store is treated as a unique identity.
```diff
type Mounted = {
  /** Set of listeners to notify when the atom value changes. */
  readonly l: Set<() => void>
-  /** Set of mounted atoms that the atom depends on. */
-  readonly d: Set<AnyAtom>
-  /** Set of mounted atoms that depends on the atom. */
-  readonly t: Set<AnyAtom>
+  /** Map of mounted atoms that the atom depends on to their stores. */
+  readonly d: Map<AnyAtom, Set<Store>>
+  /** Map of mounted atoms that depend on this atom to their stores. */
+  readonly t: Map<AnyAtom, Set<Store>>
  /** Function to run when the atom is unmounted. */
  u?: () => void
}
```

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -26,7 +26,7 @@ type AtomState<Value = AnyValue> = {
    */
   readonly d: Map<AnyAtom, EpochNumber>
   /**
-   * Map of atoms with pending promise that depend on this atom to their stores.
+   * Set of atoms with pending promise that depend on the atom.
    *
    * This may cause memory leaks, but it's for the capability to continue promises
    * TODO(daishi): revisit how to handle this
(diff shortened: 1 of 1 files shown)


dai-shi:
As you might guess, this complexity isn't acceptable in jotai internals. It's not about the implementation but the concept. I wonder if some new store hooks help for the jotai-scope case.

PR #3200, opened 2025-12-14
store hook for when the atom graph recomputes

dmaskasky:
## Summary
Fires after recomputeInvalidatedAtoms is called.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -330,6 +330,8 @@ type StoreHooks = {
   readonly m?: StoreHookForAtoms
   /** Listener to notify when the atom is unmounted. */
   readonly u?: StoreHookForAtoms
+  /** Listener to notify when the atom graph is recomputed. */
+  readonly g?: StoreHook
   /** Listener to notify when callbacks are being flushed. */
   readonly f?: StoreHook
 }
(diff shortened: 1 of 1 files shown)


dmaskasky:
Is there a better letter name than `g`?

src/vanilla/internals.ts
```diff
@@ -530,6 +534,7 @@ const recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (store) => {
     }
     invalidatedAtoms.delete(a)
   }
+  storeHooks.g?.()
```
dai-shi, line 537: Is this enough for your case? I was thinking about adding hook in the above if loop for each recomputed atom.

This looks pretty similar to `storeHooks.f`. Doesn't it work for your case? If not, I would like to understand what is the difference.

dmaskasky: I was thinking I would use `storeHooks.c` for inside the loop and `storeHooks.g` once the loop has finished.

I agree `storeHooks.g` is too similar to `storeHooks.f`. My concern with using `storeHook.f` was more about semantics. I really don't want to add more hooks if it can be avoided, but I worry that it might just be coincidence that `flushCallbacks` always runs after `recomputeInvalidatedAtoms`.

Also, it might race things that run in `storeHooks.f`, such as atomEffect, but I have a workaround in mind for atomEffect of ot comes to it.

What do you suggest?

dai-shi: My thought is yes, `.f` and `.g` are too similar, and not good for the future.
If you care about the semantics, my suggestion is to drop both the current `.f` and this new `.g` and create a new hook that works both for -scope and -effect. I'd expect the new one to have better semantics, and probably it will be moved out of flushCallback internal loop.

dmaskasky: Ok, I think I'll close the PR for now. I'll try to make it work with `storeHooks.f`. Keeping `storeHooks.f` inside the flushCallbacks loop is intentional to process the atoms set by effect.

dai-shi: > Keeping `storeHooks.f` inside the flushCallbacks loop is intentional to process the atoms set by effect.

I see.

PR #3213, opened 2026-01-08
fix(utils): unwrap should not violate the store mutation rule

dai-shi:

Fixes #3208

It was a regression with #3172.

## Summary

- `unwrap` should not violate the store mutation rule (avoids dev-only warning)
- same for `loadable`

## Check List


dhenneke:
I applied this PR to our repo and all tests are now green and don't emit the warning anymore ✅. Thank you!

PR #3214, opened 2026-01-11
test(react/vanilla-utils/atomWithStorage): add test for 'createJSONStorage' with async string storage

sukvvon:

N/A

## Summary

Add test case for `createJSONStorage` with async string storage (e.g., React Native `AsyncStorage`).

This test covers line 139 of `atomWithStorage.ts` where async storage returns a string value that needs to be parsed.

### before

image

### after

image

## Check List


dai-shi:
@sukvvon I didn't notice this is ready. GitHub doesn't notify when changing draft to ready, so would you mind posing a comment when you mark a PR ready, for the next time?

PR #3217, opened 2026-01-18
deprecate loadable util

dai-shi:
`loadable` is deprecated infavor of `unwrap`.

Userland implementation of loadable:
```js
function loadable(anAtom) {
  const LOADING = { state: 'loading' }
  const unwrappedAtom = unwrap(anAtom, () => LOADING)
  return atom((get) => {
    try {
      const data = get(unwrappedAtom)
      if (data === LOADING) {
        return LOADING
      }
      return { state: 'hasData', data }
    } catch (error) {
      return { state: 'hasError', error }
    }
  })
}
```

### DIFF
--- a/src/vanilla/utils/loadable.ts
+++ b/src/vanilla/utils/loadable.ts
@@ -1,12 +1,12 @@
 import { atom } from '../../vanilla.ts'
 import type { Atom } from '../../vanilla.ts'
+// XXX We don't usually depend on another util,
+// but our future plan is to deprecate loadable in favor of unwrap.
+import { unwrap } from './unwrap.ts'
 
 const cache1 = new WeakMap()
 const memo1 = <T>(create: () => T, dep1: object): T =>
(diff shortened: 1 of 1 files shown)


src/vanilla/utils/loadable.ts
```diff
@@ -17,64 +17,18 @@ const LOADING: Loadable<unknown> = { state: 'loading' }
 
 export function loadable<Value>(anAtom: Atom<Value>): Atom<Loadable<Value>> {
```
dmaskasky, line 38: Should we deprecate loadable now?

dai-shi: I was thinking of doing it with another step, but yeah we can do that now. Let me try.

wojtekmaj:
Note for folks that will land here because of a deprecation message: If you can't be bothered with figuring out unwrap() now, I published [jotai-loadable](https://github.com/wojtekmaj/jotai-loadable) you can use.

PR #3223, opened 2026-01-27
test(vanilla/store): add test for 'getDefaultStore' warning on multiple Jotai instances

sukvvon:

N/A

## Summary

Add test case for `getDefaultStore` to verify it warns when multiple Jotai instances are detected via `globalThis.__JOTAI_DEFAULT_STORE__`.

## Check List

### DIFF
--- a/tests/react/vanilla-utils/atomWithStorage.test.tsx
+++ b/tests/react/vanilla-utils/atomWithStorage.test.tsx
@@ -694,10 +694,11 @@ describe('atomWithStorage (with non-browser storage)', () => {
 })
 
 describe('withStorageValidator', () => {
+  const isNumber = (v: unknown): v is number => typeof v === 'number'
+
   it('should use withStorageValidator with isNumber', () => {
     const store = createStore()
     const storage = createJSONStorage()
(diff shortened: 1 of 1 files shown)


tests/vanilla/store.test.tsx
```diff
@@ -1331,3 +1331,13 @@ it('[DEV-ONLY] should warn store mutation during read', () => {
     'Detected store mutation during atom read. This is not supported.',
   )
 })
+
+it('[DEV-ONLY] should warn when multiple Jotai instances are detected', () => {
+  const prev = (globalThis as any).__JOTAI_DEFAULT_STORE__
+  ;(globalThis as any).__JOTAI_DEFAULT_STORE__ = {}
```
dai-shi, line 1337: Do you think this is a valid test? It's assuming an implementation detail.
Improving test coverage is nice, but we should test the behavior, not the implementation.

sukvvon: @dai-shi You're right. This test relies on `__JOTAI_DEFAULT_STORE__` which is an implementation detail, not a public behavior. I'll close this PR.

PR #3227, opened 2026-01-27
feat: allow projects to override `getDefaultStore`

mxmul:

#3226 

## Summary

- allows projects to override the behavior of`getDefaultStore`
- this allows one to opt-out of provider-less mode in an app by throwing an exception
- or for a server-side app to replace the default store with one that is request-scoped

## Check List

### DIFF
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -11,26 +11,29 @@ export function INTERNAL_overrideCreateStore(
   overriddenCreateStore = fn(overriddenCreateStore)
 }
 
-export function createStore(): Store {
+export function createStore(unstable_isDefaultStore?: boolean): Store {
   if (overriddenCreateStore) {
-    return overriddenCreateStore()
+    return overriddenCreateStore(unstable_isDefaultStore)
   }
-  return INTERNAL_buildStore()
+  if (!unstable_isDefaultStore) {
+    return INTERNAL_buildStore()
+  }
+
+  defaultStore ||= INTERNAL_buildStore()
+  if (import.meta.env?.MODE !== 'production') {
+    ;(globalThis as any).__JOTAI_DEFAULT_STORE__ ||= defaultStore
+    if ((globalThis as any).__JOTAI_DEFAULT_STORE__ !== defaultStore) {
+      console.warn(
+        'Detected multiple Jotai instances. It may cause unexpected behavior with the default store. https://github.com/pmndrs/jotai/discussions/2044',
+      )
+    }
+  }
+
+  return defaultStore
 }
 
 let defaultStore: Store | undefined
 
 export function getDefaultStore(): Store {
-  if (!defaultStore) {
-    defaultStore = createStore()
-    if (import.meta.env?.MODE !== 'production') {
-      ;(globalThis as any).__JOTAI_DEFAULT_STORE__ ||= defaultStore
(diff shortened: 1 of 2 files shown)


dai-shi:
Thanks for the suggestion, while I understand the motivation, I'm not a big fan of having two methods for overriding. I'm also hesitant about making this API public.

How about this?
```diff
diff --git a/src/vanilla/store.ts b/src/vanilla/store.ts
index 8255db7..e9e5356 100644
--- a/src/vanilla/store.ts
+++ b/src/vanilla/store.ts
@@ -11,18 +11,16 @@ export function INTERNAL_overrideCreateStore(
   overriddenCreateStore = fn(overriddenCreateStore)
 }
 
-export function createStore(): Store {
+let defaultStore: Store | undefined
+
+export function createStore(unstable_isDefaultStore?: boolean): Store {
   if (overriddenCreateStore) {
-    return overriddenCreateStore()
+    return overriddenCreateStore(unstable_isDefaultStore)
   }
+  if (!unstable_isDefaultStore) {
     return INTERNAL_buildStore()
   }
-
-let defaultStore: Store | undefined
-
-export function getDefaultStore(): Store {
-  if (!defaultStore) {
-    defaultStore = createStore()
+  defaultStore ||= INTERNAL_buildStore()
   if (import.meta.env?.MODE !== 'production') {
     ;(globalThis as any).__JOTAI_DEFAULT_STORE__ ||= defaultStore
     if ((globalThis as any).__JOTAI_DEFAULT_STORE__ !== defaultStore) {
@@ -31,6 +29,9 @@ export function getDefaultStore(): Store {
       )
     }
   }
-  }
   return defaultStore
 }
+
+export function getDefaultStore(): Store {
+  return createStore(true)
+}
```

---

> or for a server-side app to replace the default store with one that is request-scoped

That's actually interesting. We might consider a dedicated api based on the internals if there were any use cases. Though, it's safe if we don't write to atoms on the server anyway. I'm not sure if there are good use cases.

See also my take on this area: https://github.com/wakujs/waku-jotai

mxmul:
@dai-shi thanks for reviewing! 

Yes, your suggested API would work as well. My only concern is that 'INTERNAL_overrideCreateStore' may go away (I thought I saw an issue suggesting it be deprecated?) -- if I can convince you to keep it around for this purpose, then that's great.

dai-shi:
Yes, that's the point. I mean `INTERNAL_` is kind of `unstable_` but we'll keep it. We may change it though. It's basically for ecosystem library, so a wrapper library would be nice. But, advanced users may use it.

> I thought I saw an issue suggesting it be deprecated?

Good eyes. That's my original idea ("if possible" which isn't likely). But now with this change, there's no immediate plan to deprecate it.

---

Would you update this PR, or do you want me to do it?

mxmul:
I'll update with your suggestions.

dmaskasky:
I think we should deprecate overrideCreateStore. IMO, there are better ways of accomplishing this without adding another API.

dai-shi:
Okay, let's hold. I'll consider @dmaskasky idea seriously. Let's take time before landing this PR. (I'm interested in "if possible" case.)

Note that overrideCreateStore is used in jotai-devtools, which needs to be resolved too.

dai-shi:
Let's keep discussing in https://github.com/pmndrs/jotai/discussions/3226.
Marking this as draft for now.

dai-shi:
https://github.com/pmndrs/jotai/discussions/3226#discussioncomment-15655943

PR #3129, opened 2025-08-18
Unmounted pending atoms do not recompute

dmaskasky:
## Related Bug Reports or Discussions
https://github.com/pmndrs/jotai/discussions/3124#discussioncomment-14100673

## Summary
Adds failing test: "pending asyncAtom -> syncAtom++"

### Test incompatibility
I believe this behavior conflicts with:
- (tests/vanilla/dependency.test.tsx) "refreshes deps for each async read"

## Check List

- [x] `pnpm run fix` for formatting and linting code and docs

PR #3232, opened 2026-01-31
refactor(internals): simplify promise handing

dai-shi:
- Played based on #3129 and concluded that it's not possible.
- Based on the play, added some improvements.

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -6,12 +6,12 @@ jobs:
   compressed_size:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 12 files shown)


dai-shi:
https://github.com/pmndrs/jotai/actions/runs/21563292297 

> . test:spec:  ❯ tests/effect.test.tsx (2 tests | 2 failed) 87ms
. test:spec:    × atomEffect > should work with atomEffect 72ms
. test:spec:      → Cannot read properties of undefined (reading 'has')
. test:spec:    × atomEffect > should work with atomEffect in a scope 13ms
. test:spec:      → Cannot read properties of undefined (reading 'has')

PR #3235, opened 2026-02-07
ci(.github/workflows): pin GitHub Actions to commit hashes

sukvvon:

N/A

## Summary

- Pin all GitHub Actions `uses` to commit hashes for supply chain security
  - Ensures reproducibility by always using the exact same commit
  - Protects against tag mutation (tags can be moved to different commits)

## Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 visit [jotai.org](https://jotai.org) or `npm i jotai`
 
-[![Build Status](https://img.shields.io/github/actions/workflow/status/pmndrs/jotai/test.yml?branch=main&style=flat&colorA=000000&colorB=000000)](https://github.com/pmndrs/jotai/actions?query=workflow%3ALint)
+[![Build Status](https://img.shields.io/github/actions/workflow/status/pmndrs/jotai/test.yml?branch=main&style=flat&colorA=000000&colorB=000000)](https://github.com/pmndrs/jotai/actions/workflows/test.yml?query=branch%3Amain)
 [![Build Size](https://img.shields.io/bundlephobia/minzip/jotai?label=bundle%20size&style=flat&colorA=000000&colorB=000000)](https://bundlephobia.com/result?p=jotai)
 [![Version](https://img.shields.io/npm/v/jotai?style=flat&colorA=000000&colorB=000000)](https://www.npmjs.com/package/jotai)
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+    labels: []
```
dai-shi: What is labels?

Can we sync with zustand config. 👇 

```suggestion
    ignore:
      - dependency-name: '*'
        update-types:
          - 'version-update:semver-patch'
          - 'version-update:semver-minor'
```

dai-shi (changes requested):
https://docs.github.com/en/actions/reference/security/secure-use 👀 

I'm not sure how this is maintainable for the future with multiple repos. Maybe we would create our own action so that we don't need to change all of them.

.github/dependabot.yml
dai-shi: Actually, we don't need this file as we notice it with zustand repo.

dai-shi:
> with multiple repos

Perhaps, we could pin only in the three z/j/v repos.

.github/dependabot.yml
sukvvon: @dai-shi 6bfb6cd Added the `ignore` config. `labels: []` is set to match the existing  convention — current PRs in this repo have no labels.

sukvvon: @dai-shi c6b2ce2 Synced with zustand config —  added `npm` ecosystem (`daily`, ignore patch/minor). Kept `labels: []` to match the existing convention in this repo (PRs have no labels).

.github/dependabot.yml
dai-shi, line 1: I don't think we need this file because we maintain z/j/v all at once and zustand dependabot notifies us. Then we decide.

sukvvon: @dai-shi 84f92af Removed `dependabot.yml`. Would you like to keep the hash pinning as is?

sukvvon:
> > Would you like to keep the hash pinning as is?
> 
> Yes. I don't love it, but unless GitHub provides other solutions (like lock file), it seems unavoidable.
> 
> Please open the same PR in Zustand and Valtio. We'll use commit hash only in these three repos.

@dai-shi 👌

PR #3243, opened 2026-02-19
feat: add `subscribed` option to `useAtomValue`

peterpme:
## Summary

Adds a `subscribed` option (default `true`) to `useAtomValue`. When `subscribed: false`, the hook skips the `store.sub()` call inside `useEffect`, so the component won't re-render when the atom's value changes.

This is useful for react-navigation screens that **freeze but don't unmount**. Without this, off-screen components remain subscribed to atom changes, causing unnecessary work on screens the user isn't looking at.

When `subscribed` transitions from `false` back to `true`, the effect re-runs, re-subscribes, and calls `rerender()` to pick up any changes that occurred while unsubscribed.

### Usage

```ts
const isFocused = useIsFocused();
const value = useAtomValue(myAtom, { subscribed: isFocused });
```

### Changes

- Added `subscribed?: boolean` to the `Options` type
- Destructure `subscribed` (default `true`) from options
- Early return in `useEffect` when `subscribed` is `false`
- Added `subscribed` to the `useEffect` dependency array

### DIFF
--- a/src/react/useAtomValue.ts
+++ b/src/react/useAtomValue.ts
@@ -104,6 +104,7 @@ const createContinuablePromise = <T>(
 type Options = Parameters<typeof useStore>[0] & {
   delay?: number
   unstable_promiseStatus?: boolean
+  subscribed?: boolean
 }
 
 export function useAtomValue<Value>(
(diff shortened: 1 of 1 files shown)


dai-shi:
I don't think this is the scope of the official hook. We generally prefer primitive approach and in v3, we plan to remove some options.

That said, this should be useful for some users and should be covered by ecosystem libs. Would you like to create one?

While it's not related with your use case, there's another ecosystem lib which might be of your interest just as an example. Check this out: https://github.com/jotaijs/jotai-game

PR #3245, opened 2026-02-19
fix(vanilla): subscriber not notified when derived read calls store.set

dmaskasky:
## Summary
A bug was introduced in v2.12.1 (commit f5d843c, PR #2982 — "fix: remove extra onChange store hook in recomputeInvalidatedAtoms").
- Adds a failing  test

## Bug
When a derived atom's **read** calls `store.set` (e.g. via a write-only atom), that atom's **subscribers are not notified** when a dependency changes:
- `store.get(derivedAtom)` returns the correct updated value.
- The subscription callback passed to `store.sub(derivedAtom, listener)` never runs for that update.

## References
- https://github.com/koutaro-masaki/jotai-atom-in-component-with-usesetatom
- Commit: f5d843c87842df5a9c034ef54bf2a77b367d0c36

### DIFF
--- a/tests/react/vanilla-utils/atomWithStorage.test.tsx
+++ b/tests/react/vanilla-utils/atomWithStorage.test.tsx
@@ -119,6 +119,15 @@ describe('atomWithStorage (sync)', () => {
 
     expect(screen.getByText('count: 9')).toBeInTheDocument()
   })
+
+  it('should get stored value on init with getOnInit option', () => {
+    const store = createStore()
+    const countAtom = atomWithStorage('count', 0, dummyStorage, {
+      getOnInit: true,
+    })
+
+    expect(store.get(countAtom)).toBe(10)
+  })
 })
 
 describe('with sync string storage', () => {


tests/vanilla/store.test.tsx
```diff
@@ -1363,3 +1363,35 @@ it('should keep reactivity when a derived atom returns a function that calls get
...
+// https://github.com/koutaro-masaki/jotai-atom-in-component-with-usesetatom
+it('notifies derived-atom subscriber when read calls store.set', () => {
+  const store = createStore()
+  const counterAtom = atom(0)
+  const queryAtom = atom(null, (get, _, v: number) => v + get(counterAtom))
+  const dataAtom = atom((get) => {
+    const v = get(counterAtom)
+    const result = store.set(queryAtom, v)
+    return result
+  })
```
dmaskasky:
```suggestion
  const queryAtom = atom(null, (_get, _set, v: number) => v)
  const dataAtom = atom((get) => {
    const v = get(counterAtom)
    const result = store.set(queryAtom, v * 2)
    return result
  })
```

dai-shi: Would you suggest to change our test?

dmaskasky: Yes, it was a TODO reminder for myself to do so. I'll take care of it tomorrow.

src/vanilla/internals.ts
```diff
@@ -491,6 +491,7 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
       }
     }
     if (hasChangedDeps) {
+      invalidatedAtoms.set(a, aState.n)
```
dmaskasky, line 494: I don't understand how adding this line fixes "it notifies subscriber through chain of derived atoms when root calls store.set". Wouldn't the invalidatedAtoms.set represent atoms needing to have readAtomState called? But then we immediately call `readAtomState`, so to me, this semantically doesn't make sense. What am I missing?

dmaskasky: Nevermind, I understand now.

Line 494's purpose is to re-invalidate atoms whose invalidation markers were prematurely cleared by a nested recompute cycle. The nested recompute happens because `store.set` was called inside a read function, which triggers `recomputeInvalidatedAtoms` before the outer recompute has finished propagating changes through the graph. The nested recompute processes (and clears) atoms like derivedAtom before baseAtom's new value has been committed, so it effectively no-ops for those atoms but still removes their `invalidatedAtoms` entries. Line 494 restores the entry right before `readAtomState` to prevent the cache shortcut from swallowing the update.

tests/vanilla/store.test.tsx
```diff
@@ -1363,3 +1363,236 @@ it('should keep reactivity when a derived atom returns a function that calls get
...
+it('does not recompute derived atom redundantly when store.set in read modifies a shared dependency', () => {
+  const store = createStore()
+  const counterAtom = atom(0)
+  const sideEffectAtom: PrimitiveAtom<number> = atom(0)
+
+  let baseReadCount = 0
+  const baseAtom = atom((get) => {
+    baseReadCount++
+    const v = get(counterAtom)
+    store.set(sideEffectAtom, v * 10)
```
dai-shi: This should cause "Detected store mutation during atom read. This is not supported." warning. So, it's unsupported usage. Should we throw rather than warning?

dmaskasky: I think it should throw. There is no guarantee of correctness otherwise. If this functionality is required, atomEffect is a better option.

dmaskasky: I modified the test to make it valid. It demonstrates that `derivedAtom` would have an extra recompute for each atom dependency that calls `store.set`, so the following `derivedAtom` would recompute 3 times when `counterAtom` is incremented.
```ts
const counterAtom = atom(0)
const queryAtom = atom(null, (_get, _set, v: number) => v)
const baseAtom = atom((get) => store.set(queryAtom, get(counterAtom)))
const baseAtom2 = Object.assign({}, baseAtom)
const derivedAtom = atom((get) => [get(counterAtom), get(baseAtom), get(baseAtom2)])
```

PR #3252, opened 2026-03-04
test(react/vanilla-utils/atomWithStorage): add test for 'getOnInit' option reading stored value on init

sukvvon:

N/A

## Summary

- Add test for `atomWithStorage` with `getOnInit: true` option
  - Verifies that stored value (`10`) is used as initial value instead of `initialValue` (`0`) on first render

## Check List

### DIFF
--- a/tests/react/vanilla-utils/atomWithStorage.test.tsx
+++ b/tests/react/vanilla-utils/atomWithStorage.test.tsx
@@ -121,22 +121,12 @@ describe('atomWithStorage (sync)', () => {
   })
 
   it('should get stored value on init with getOnInit option', () => {
+    const store = createStore()
     const countAtom = atomWithStorage('count', 0, dummyStorage, {
       getOnInit: true,
     })
 
-    const Counter = () => {
-      const [count] = useAtom(countAtom)
-      return <div>count: {count}</div>
-    }
-
-    render(
-      <StrictMode>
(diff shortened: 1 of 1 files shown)


tests/react/vanilla-utils/atomWithStorage.test.tsx
```diff
@@ -119,6 +119,25 @@ describe('atomWithStorage (sync)', () => {
 
     expect(screen.getByText('count: 9')).toBeInTheDocument()
   })
+
+  it('should get stored value on init with getOnInit option', () => {
+    const countAtom = atomWithStorage('count', 0, dummyStorage, {
+      getOnInit: true,
```
dai-shi, line 126: This test passes even if we make `getOnInit: false`.

dai-shi (changes requested):
Not a valid test.

tests/react/vanilla-utils/atomWithStorage.test.tsx
sukvvon: @dai-shi 707894c Changed to use `store.get` to verify the value before mount. The test now fails with `getOnInit: false` (`expected +0 to be 10`), confirming it correctly validates the `getOnInit` behavior.

dai-shi (approved):

PR #3253, opened 2026-03-04
fix(vanilla/utils/atomWithObservable): add 'Symbol.observable' type support to 'ObservableLike'

sukvvon:

N/A

## Summary

- Add `SubscribableObservable<T>` and `SymbolObservable<T>` type aliases
- Extend `ObservableLike<T>` to accept `SymbolObservable<T>`
- Improve `Symbol.observable` resolution logic using `SymbolObservable` type alias
- Add test for `Symbol.observable` interop protocol
- Remove redundant `subscribe` from `Symbol.observable` test

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -491,6 +491,7 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
       }
     }
     if (hasChangedDeps) {
+      invalidatedAtoms.set(a, aState.n)
       readAtomState(store, a)
       mountDependencies(store, a)
     }
@@ -632,6 +633,7 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
     },
   }
   const prevEpochNumber = atomState.n
+  const prevInvalidated = invalidatedAtoms.get(atom) === prevEpochNumber
   try {
     if (import.meta.env?.MODE !== 'production') {
       storeMutationSet.delete(store)
@@ -662,10 +664,7 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
     return atomState
   } finally {
     isSync = false
-    if (
-      prevEpochNumber !== atomState.n &&
-      invalidatedAtoms.get(atom) === prevEpochNumber
-    ) {
+    if (atomState.n !== prevEpochNumber && prevInvalidated) {
       invalidatedAtoms.set(atom, atomState.n)
       changedAtoms.add(atom)
       storeHooks.c?.(atom)
@@ -930,14 +929,18 @@ const BUILDING_BLOCK_storeGet: StoreGet = (store, atom) => {
 
 const BUILDING_BLOCK_storeSet: StoreSet = (store, atom, ...args) => {
   const buildingBlocks = getInternalBuildingBlocks(store)
+  const changedAtoms = buildingBlocks[3]
   const flushCallbacks = buildingBlocks[12]
   const recomputeInvalidatedAtoms = buildingBlocks[13]
   const writeAtomState = buildingBlocks[16]
+  const prevChangedAtomsSize = changedAtoms.size
   try {
     return writeAtomState(store, atom, ...args) as any
   } finally {
-    recomputeInvalidatedAtoms(store)
-    flushCallbacks(store)
+    if (changedAtoms.size !== prevChangedAtomsSize) {
+      recomputeInvalidatedAtoms(store)
+      flushCallbacks(store)
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


tests/react/vanilla-utils/atomWithObservable.test.tsx
```diff
@@ -87,6 +87,30 @@ it('count state', () => {
   expect(screen.getByText('count: 1')).toBeInTheDocument()
 })
 
+it('count state with Symbol.observable', () => {
+  const subject = new BehaviorSubject(1)
+  const observable = {
+    subscribe: subject.subscribe.bind(subject),
```
dai-shi: Shouldn't we remove this to test the behavior?

btw, I'm not sure if we should preserve supporting Symbol.observable in the future.

sukvvon: @dai-shi Removed the redundant `subscribe` as you suggested (230091f).

After removing it, the test required `as any` since `ObservableLike` didn't accept objects with only `[Symbol.observable]`. So I extracted `SubscribableObservable<T>` and extended `ObservableLike<T>` to also accept `{ [Symbol.observable]: () => SubscribableObservable<T> }` (a5d16e9).

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -75,13 +81,11 @@ export function atomWithObservable<Data>(
   }
 
   const observableResultAtom = atom((get) => {
-    let observable = getObservable(get)
+    const source = getObservable(get)
```
dai-shi: Why do you name it `source` here?

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -156,7 +160,7 @@ export function atomWithObservable<Data>(
         }
       }
     }
-    return [resultAtom, observable, makePending, start, isNotMounted] as const
+    return [resultAtom, source, makePending, start, isNotMounted] as const
```
dai-shi: This seems to change the behavior from the previous one. Shouldn't we return `observable`?

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -75,13 +81,11 @@ export function atomWithObservable<Data>(
   }
 
   const observableResultAtom = atom((get) => {
-    let observable = getObservable(get)
```
dai-shi, line 78: Avoiding `let` is nice.

src/vanilla/utils/atomWithObservable.ts
sukvvon: @dai-shi Renamed back to `observable` and used `subscribable` for the resolved result (1aa0313).

src/vanilla/utils/atomWithObservable.ts
sukvvon: @dai-shi You're right, restored to return `observable` (1aa0313).

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -75,13 +81,11 @@ export function atomWithObservable<Data>(
...
+    const observable = getObservable(get)
     const itself = (
-      observable as Partial<Record<symbol, () => typeof observable>>
+      observable as Partial<Record<symbol, () => SubscribableObservable<Data>>>
     )[Symbol.observable]?.()
-    if (itself) {
-      observable = itself
-    }
+    const subscribable = itself || (observable as SubscribableObservable<Data>)
```
dai-shi: Can we inline `itself`?

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -30,6 +30,12 @@ type ObservableLike<T> =
       subscribe(next: (value: T) => void): Subscription
     }
 
+type ObservableLike<T> =
+  | SubscribableObservable<T>
+  | {
+      [Symbol.observable]: () => SubscribableObservable<T>
+    }
```
dai-shi: Let's define this as a separate type alias.

sukvvon: @dai-shi Extracted `SymbolObservable<T>` as a separate type alias (bcb221f).

src/vanilla/utils/atomWithObservable.ts
sukvvon: @dai-shi Inlined `itself` into `subscribable` (bcb221f).

src/vanilla/utils/atomWithObservable.ts
```diff
@@ -75,13 +81,13 @@ export function atomWithObservable<Data>(
...
-    )[Symbol.observable]?.()
-    if (itself) {
-      observable = itself
-    }
+    const observable = getObservable(get)
+    const subscribable =
+      (
+        observable as Partial<
+          Record<symbol, () => SubscribableObservable<Data>>
+        >
```
dai-shi: Can we use the defined type alias?

sukvvon: @dai-shi Used `SymbolObservable<Data>` type alias in the assertion (4ce892a).

dai-shi (approved):

PR #3259, opened 2026-03-07
perf: optimize atom store for unmounted atom reads and DAG traversal

edkimmel:

Fixes discussion # https://github.com/pmndrs/jotai/discussions/2334

## Summary

The discussion linked is two years old now, time flies.
My old suggestions and approaches no longer work with the big store rewrite, and I do not believe they ever worked with async atoms.  My goals are the same, albeit I did use AI to figure out how exactly it could be re-implemented in this new jotai world.

1) React SSR and React Hydration are *slow* due to repeated tree traversal reading unmounted atom values. During the synchronous react render, there aren't any mutations to the store to warrant re-traversal, so a caching strategy was desired.

2) There has been some slight performance regressions over time that have added up in the dependency updates flow since the initial conversion to topological sort a few years ago.

---

Two key performance optimizations:

1. Store-level epoch cache for unmounted atoms: Repeated store.get() calls on unmounted derived atoms no longer recursively walk the entire dependency chain when no mutations have occurred. A per-store epoch counter (incremented on each store.set()) allows storeGet to skip the expensive dep walk entirely when the epoch is unchanged.

2. Inline getMountedOrPendingDependents: The hot-path functions invalidateDependents and recomputeInvalidatedAtoms now iterate mounted.t and atomState.p directly, avoiding a Set allocation on every call during DAG traversal.

## Check List

- [ ] `pnpm run fix` for formatting and linting code and docs

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -473,7 +473,16 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
     }
     visiting.add(a)
     // Push unvisited dependents onto the stack
-    for (const d of getMountedOrPendingDependents(a, aState, mountedMap)) {
+    // Inline iteration to avoid Set allocation from getMountedOrPendingDependents
+    const aMounted = mountedMap.get(a)
+    if (aMounted) {
+      for (const d of aMounted.t) {
+        if (!visiting.has(d)) {
+          stack.push(d)
+        }
+      }
+    }
+    for (const d of aState.p) {
       if (!visiting.has(d)) {
         stack.push(d)
       }
@@ -496,6 +505,33 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
     }
     invalidatedAtoms.delete(a)
   }
+}
+
+// Store-level epoch counter for unmounted atom cache optimization.
+// Incremented on each store.set() call so that repeated store.get() calls
+// on unmounted atoms can skip redundant recursive dependency walks when
+// no mutations have occurred between reads.
+// The verified map is keyed per-store (not on AtomState) because
+// AtomState objects can be shared across derived stores.
+// Verified entries store [storeEpoch, atomEpoch] - we check both
+// because another store may mutate the shared AtomState.
+type VerifiedEntry = [storeEpoch: number, atomEpoch: EpochNumber]
+type StoreEpochState = {
+  epoch: number
+  verified: WeakMap<AnyAtom, VerifiedEntry>
+}
+const storeEpochMap = new WeakMap<Store, StoreEpochState>()
+const getStoreEpochState = (store: Store): StoreEpochState => {
+  let state = storeEpochMap.get(store)
+  if (!state) {
+    state = { epoch: 0, verified: new WeakMap() }
+    storeEpochMap.set(store, state)
+  }
+  return state
+}
+const incrementStoreEpoch = (store: Store): void => {
+  const state = getStoreEpochState(store)
+  ++state.epoch
 }
 
 // Dev only
@@ -685,7 +721,19 @@ const BUILDING_BLOCK_invalidateDependents: InvalidateDependents = (
   while (stack.length) {
     const a = stack.pop()!
     const aState = ensureAtomState(store, a)
-    for (const d of getMountedOrPendingDependents(a, aState, mountedMap)) {
+    // Inline iteration over mounted dependents and pending-promise dependents
+    // to avoid Set allocation from getMountedOrPendingDependents
+    const mounted = mountedMap.get(a)
+    if (mounted) {
+      for (const d of mounted.t) {
+        const dState = ensureAtomState(store, d)
+        if (invalidatedAtoms.get(d) !== dState.n) {
+          invalidatedAtoms.set(d, dState.n)
+          stack.push(d)
+        }
+      }
+    }
+    for (const d of aState.p) {
       const dState = ensureAtomState(store, d)
       if (invalidatedAtoms.get(d) !== dState.n) {
         invalidatedAtoms.set(d, dState.n)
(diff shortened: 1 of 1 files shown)


dai-shi:
I'm almost forgetting the past discussion, but if I remember correctly, we don't plan to merge this as this is only beneficial for some use cases. Thanks to new building blocks, I think you can create an ecosystem library to implement this capability (if not, we should fix the building blocks). I can invite you to https://github.com/jotaijs, let me know if you are interested.

edkimmel:
@dai-shi Ah, that's a fun development!  I'll have to take a look.

edkimmel:
@dai-shi 
I will take a look at whether it's more realistic for us to maintain an ecosystem library or a fork for ourselves like we have been.  Given where the changes are, we may have to re-implement more as an ecosystem than maintaining a fork.

The current benchmarks do not test wide and deep dependency trees.  Here's an example addition
https://github.com/edkimmel/jotai/blob/performance-optimizations/benchmarks/derived-read.ts

I would still encourage looking at the problem of redundant tree traversal during that first render before effects mount in core jotai, independent of my proposals.  Every read scales linearly with depth/width of the dependency tree, but could be cut down to O(1) with a guarantee that data has not changed since the last read.
 
Screenshot 2026-03-07 at 5 45 26 PM

dai-shi:
@edkimmel 

Thanks for the follow-up.

> The current benchmarks do not test wide and deep dependency trees. Here's an example addition https://github.com/edkimmel/jotai/blob/performance-optimizations/benchmarks/derived-read.ts

Ah, I think I misunderstood something. This should be the general improvement for deep/wide use cases (at the cost of complexity).

Please send another PR to add the benchmark file.

> I would still encourage looking at the problem of redundant tree traversal during that first render before effects mount in core jotai, independent of my proposals.

Thanks. Sure, I'll take another deeper look soon. Meanwhile, do you think you can reproduce with a failing test (without relying on "time")?

> Every read scales linearly with depth/width of the dependency tree, but could be cut down to O(1) with a guarantee that data has not changed since the last read.

Does the guarantee hold? Okay, I think I get it. I see, it's store epoch to guarantee. I think it's valid. Clever. But, I'm not sure if it covers 100%. We can update store async.
Using the WeakMap solves my concern about GC. https://github.com/pmndrs/jotai/discussions/2334#discussioncomment-8132025
Did you solve it two years ago? I didn't notice it.

The implementation needs to be re-designed. I'll take care of it as I can't tell my preference easily (and it may end up with something like yours). If you can help the work, the failing test should be really helpful.

dmaskasky:
Can you refactor this to use Building Blocks? I would start by moving the store epoch to buildingBlocks and verified atomEpoch to atomState. We want both to be accessible by stores that compose these building blocks.

We don't rely on derived stores anymore, since stores are now composed by building blocks, so I would drop that requirement and leave it up to the developer to integrate with store cache - but we have to ensure the default behavior is always correctness so if they choose to do nothing, it will just result in extra readAtomState traversals.

src/vanilla/internals.ts
```diff
@@ -473,7 +473,16 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
     }
     visiting.add(a)
     // Push unvisited dependents onto the stack
-    for (const d of getMountedOrPendingDependents(a, aState, mountedMap)) {
```
dmaskasky, line 476: Nice. Can we remove getMountedOrPendingDependents completely or is it still used somewhere?

src/vanilla/internals.ts
```diff
@@ -924,7 +972,32 @@ const BUILDING_BLOCK_setAtomStateValueOrPromise: SetAtomStateValueOrPromise = (
...
+    const epochState = getStoreEpochState(store)
+    const atomStateMap = buildingBlocks[0]
+    const cachedState = atomStateMap.get(atom)
+    if (cachedState && ('v' in cachedState || 'e' in cachedState)) {
+      const entry = epochState.verified.get(atom)
+      if (
+        entry &&
+        entry[0] === epochState.epoch &&
+        entry[1] === cachedState.n
+      ) {
```
dmaskasky, line 992:
```suggestion
      const [storeEpoch, atomEpoch] = epochState.verified.get(atom) || []
      if (
        storeEpoch === epochState.epoch &&
        atomEpoch === cachedState.n
      ) {
```

dmaskasky:
Can you revisit the benchmarks? I would like to understand the performance of various atom trees:
- wide (one atom having 10,000 dependencies)
- tall (one atom having one dependency recursive 10,000 times)
- balanced (the same width as height)

src/vanilla/internals.ts
```diff
@@ -924,7 +972,32 @@ const BUILDING_BLOCK_setAtomStateValueOrPromise: SetAtomStateValueOrPromise = (
...
+  const readAtomState = buildingBlocks[14]
+  // For unmounted atoms, check if we can skip the read entirely.
+  // If the store epoch hasn't changed since the last verification, and the
+  // atom's own epoch hasn't changed (possible with shared AtomState across
+  // derived stores), we can return the cached value.
+  if (!mountedMap.has(atom)) {
+    const epochState = getStoreEpochState(store)
+    const atomStateMap = buildingBlocks[0]
+    const cachedState = atomStateMap.get(atom)
+    if (cachedState && ('v' in cachedState || 'e' in cachedState)) {
```
dmaskasky, line 986:
```suggestion
    const ensureAtomState = buildingBlocks[11]
    const atomState = ensureAtomState(atom)
    if (isAtomStateInitialized(atomState)) {
```

dai-shi:
Note that I'll revisit this and #3265 when I get time (maybe next months). I hope to find a minimal solution that should be comfortable. Meanwhile, feel free to continue experiments, but current plan is not merge this PR.

> Can you refactor this to use Building Blocks?

Sounds good as an experiment. And, if it can be done with a third-party library, it is very nice as an experiment too.

dai-shi:
Superseded by #3265.

PR #3266, opened 2026-03-12
test(react/vanilla-utils/atomWithObservable): add test for write error on non-subject observable

sukvvon:

N/A

## Summary

Add tests for `'observable is not subject'` error when writing to a read-only observable (without `next`) in both React and vanilla contexts.

## Check List


sukvvon:
@dai-shi 👌

PR #3274, opened 2026-03-25
fix(vanilla/utils/atomWithObservable): use symbol index signature to avoid 'Symbol.observable' type reference

sukvvon:

Fixes #3120

## Summary

Replace `[Symbol.observable]` with `[key: symbol]` index signature in `SymbolObservable<T>` type to avoid direct reference to `Symbol.observable`, which caused a type error for users with `skipLibCheck: false` and without `ES2015.Symbol.WellKnown` in their `lib` config.

## Check List


sukvvon:
@dai-shi 👌

PR #3278, opened 2026-03-27
refactor(internals): replace nextDeps with prevDeps

dmaskasky:


src/vanilla/internals.ts
```diff
@@ -593,7 +592,7 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
     try {
       return returnAtomValue(aState)
     } finally {
-      nextDeps.set(a, aState.n)
+      prevDeps.delete(a)
```
dai-shi, line 593: Oh, I see. I didn't realize this would work. Nice.

src/vanilla/internals.ts
```diff
@@ -555,13 +555,12 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
   }
   // Compute a new state for this atom.
   let isSync = true
+  // Track only the previous deps. As we encounter live deps during this read,
+  // remove them from prevDeps. Whatever remains gets pruned.
```
dai-shi: Maybe we don't need this comment.

PR #3279, opened 2026-03-28
fix: improve perf of internals

dmaskasky:

image
Shown above, this pr is version "building-blocks-best"

## Summary

1. all: lazy building block assignment
2. `readAtomState`: drop unnecessary nextDeps
3. `mountAtom`: replace recursive with iterative approach
4. `mountAtom`: remove unnecessary ensureAtomState
5. `unmountAtom`: replace recursive with iterative approach
6. `flushCallbacks`: wrap forEach calls in conditionals
7. `recomputeInvalidatedAtoms`: early return
8. `recomputeInvalidatedAtoms`: remove extra ensureAtomState call
9. `invalidateDependents`: remove extra ensureAtomState call
10. `getInternalBuildingBlocks`: switch from WeakMap to Symbol store property
11. `ensureAtomState`: pass atomStateMap as param
12. `getMountedOrPendingDependents`: avoid unnecessary Set instantiation in most cases
13. `invalidateDependents`: change api to accept list of atoms

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -102,6 +102,7 @@ type AtomOnMount = <Value, Args extends unknown[], Result>(
 type EnsureAtomState = <Value>(
   store: Store,
   atom: Atom<Value>,
+  atomStateMap: AtomStateMap,
 ) => AtomState<Value>
 type FlushCallbacks = (store: Store) => void
 type RecomputeInvalidatedAtoms = (store: Store) => void
@@ -109,7 +110,7 @@ type ReadAtomState = <Value>(
   store: Store,
   atom: Atom<Value>,
 ) => AtomState<Value>
-type InvalidateDependents = (store: Store, atom: AnyAtom) => void
+type InvalidateDependents = (store: Store, atoms: readonly AnyAtom[]) => void
 type WriteAtomState = <Value, Args extends unknown[], Result>(
   store: Store,
   atom: WritableAtom<Value, Args, Result>,
@@ -156,6 +157,7 @@ type Store = {
     ...args: Args
   ) => Result
   sub: (atom: AnyAtom, listener: () => void) => () => void
+  [BUILDING_BLOCKS_KEY]: Readonly<BuildingBlocks>
 }
 
 type BuildingBlocks = [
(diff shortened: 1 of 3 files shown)


src/vanilla/internals.ts
```diff
@@ -102,14 +102,15 @@ type AtomOnMount = <Value, Args extends unknown[], Result>(
 type EnsureAtomState = <Value>(
   store: Store,
   atom: Atom<Value>,
+  atomStateMap: AtomStateMap,
```
dai-shi, line 105: This feels too arbitrary. If we care the performance, why not passing all necessary building blocks as params?

src/vanilla/internals.ts
```diff
@@ -156,6 +157,7 @@ type Store = {
     ...args: Args
   ) => Result
   sub: (atom: AnyAtom, listener: () => void) => () => void
+  [BUILDING_BLOCKS_KEY]: Readonly<BuildingBlocks>
```
dai-shi, line 160: It was intentional not to have this here. This is not related to performance right?

src/vanilla/internals.ts
```diff
@@ -266,17 +268,26 @@ function addPendingPromiseToDependency(
   }
 }
 
+const EMPTY_SET: ReadonlySet<AnyAtom> = new Set()
```
dai-shi, line 271: Better to use `if` check for readability? Not sure though.

src/vanilla/internals.ts
```diff
@@ -414,14 +436,22 @@ const BUILDING_BLOCK_flushCallbacks: FlushCallbacks = (store) => {
...
     const callbacks = new Set<() => void>()
     const add = callbacks.add.bind(callbacks)
-    changedAtoms.forEach((atom) => mountedMap.get(atom)?.l.forEach(add))
-    changedAtoms.clear()
-    unmountCallbacks.forEach(add)
-    unmountCallbacks.clear()
-    mountCallbacks.forEach(add)
-    mountCallbacks.clear()
+    if (changedAtoms.size) {
+      mountedMap ||= buildingBlocks[1]
```
dai-shi, line 443: hardly readable. I'd go for passing all necessary building blocks.

src/vanilla/internals.ts
dmaskasky: Because the act lf destructiring them is a little expensive and adds up with this hot function.

src/vanilla/internals.ts
dmaskasky: Weakmap access is a little slower than symbol property access. I can quantifybthisbif you like. Its a tradeoff I'm not happy with either.

src/vanilla/internals.ts
dmaskasky: Yeah, readibility was one of the first casualities. I can quantify how much of a difference this makes.

PR #3280, opened 2026-03-30
breaking(internals): pass buildingBlockContext directly to internals

dmaskasky:
## Summary
Since 2.14.0, the building block architecture relies on passing the store as the first parameter to internal functions. This store was used as a WeakMap key to access building blocks. WeakMap access is not super performant so in certain heavy applications, performance was impacted.

This PR reworks the building block architecture to directly pass buildingBlocksContext `readonly [...buildingBlocks, store]` instead of `store`. This eliminates the WeakMap access per call site.

The breaking change is that BUILDING_BLOCK functions now accept buildingBlocksContext instead of store as the first parameter.

## What was done
- switch internal function signatures and call sites from store-first to `buildingBlocksContext`-first.
- update `useAtomValue` and internals tests to pass buildingBlocksContext instead of store as the first parameter.

## Next Steps
- We should drop TS3.8.3 and TS3.9.7. These versions of TS do not support spreading tuples.

## Other Changes
- WriteAtomState now uses `args` instead of `...args`
- Reworked types for `WithOnInit` and `WithOnMount`
- added `hasOnInit` for conditionally calling `AtomOnInit`
- dropped `getInternalBuildingBlocks`

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -90,60 +90,63 @@ type ChangedAtoms = SetLike<AnyAtom>
 type Callbacks = SetLike<() => void>
 
 type AtomRead = <Value>(
-  ctx: Readonly<Context>,
+  ctx: BuildingBlockContext,
   atom: Atom<Value>,
   ...params: Parameters<Atom<Value>['read']>
 ) => Value
 type AtomWrite = <Value, Args extends unknown[], Result>(
-  ctx: Readonly<Context>,
+  ctx: BuildingBlockContext,
   atom: WritableAtom<Value, Args, Result>,
   ...params: Parameters<WritableAtom<Value, Args, Result>['write']>
 ) => Result
 type AtomOnInit = <Value>(
-  ctx: Readonly<Context>,
+  ctx: BuildingBlockContext,
   atom: Atom<Value> & WithOnInit,
 ) => void
 type AtomOnMount = <Value, Args extends unknown[], Result>(
-  ctx: Readonly<Context>,
+  ctx: BuildingBlockContext,
   atom: WritableAtom<Value, Args, Result> & WithOnMount<Args, Result>,
   setAtom: (...args: Args) => Result,
 ) => OnUnmount | void
 
 type EnsureAtomState = <Value>(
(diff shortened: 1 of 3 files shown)


src/vanilla/internals.ts
```diff
@@ -184,16 +190,17 @@ type BuildingBlocks = [
   unmountAtom: UnmountAtom, //                                 19
   setAtomStateValueOrPromise: SetAtomStateValueOrPromise, //   20
   // store api
-  storeGet: StoreGet, //                                       21
-  storeSet: StoreSet, //                                       22
-  storeSub: StoreSub, //                                       23
+  UNUSED_storeGet: undefined, //                               21
+  UNUSED_storeSet: undefined, //                               22
+  UNUSED_storeSub: undefined, //                               23
```
dmaskasky: Should we rework the ordering of buildingBlocks? I was thinking a nice pattern would be to create zone ranges in buildingBlocks. Each  zone range would leave a few empty slots at the end for future capability. The current strategy of putting new buildingBlocks at the end gets a bit messy overtime. If we like this idea, I think this should go in a separate PR.

src/vanilla/internals.ts
```diff
@@ -995,89 +976,100 @@ const BUILDING_BLOCK_registerAbortHandler: RegisterAbortHandler = (
...
-}
-
-function getBuildingBlocks(store: Store): Readonly<BuildingBlocks> {
-  const buildingBlocks = getInternalBuildingBlocks(store)
   const enhanceBuildingBlocks = buildingBlocks[24]
   if (enhanceBuildingBlocks) {
     return enhanceBuildingBlocks(buildingBlocks)
   }
-  return buildingBlocks
+  return [...buildingBlocks]
```
dai-shi: I don't think copying here is good. Users should copy it on their end.

dmaskasky: Ok, I'll revert this. It should be in its own pr anyways.

dai-shi:
https://github.com/pmndrs/jotai/pull/3293#issuecomment-4286938006

Thanks so much for your experiments and insights!

PR #3281, opened 2026-03-30
refactor(internals): iterative mountAtom (scale)

dmaskasky:
## Summary
Refactors mountAtom to use an iterative solution.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -814,35 +814,53 @@ const BUILDING_BLOCK_mountAtom: MountAtom = (store, atom) => {
   const mountCallbacks = buildingBlocks[4]
   const storeHooks = buildingBlocks[6]
   const atomOnMount = buildingBlocks[10]
-  const ensureAtomState = buildingBlocks[11]
   const flushCallbacks = buildingBlocks[12]
   const recomputeInvalidatedAtoms = buildingBlocks[13]
   const readAtomState = buildingBlocks[14]
   const writeAtomState = buildingBlocks[16]
-  const mountAtom = buildingBlocks[18]
-  const atomState = ensureAtomState(store, atom)
-  let mounted = mountedMap.get(atom)
-  if (!mounted) {
-    // recompute atom state
-    readAtomState(store, atom)
-    // mount dependencies first
-    for (const a of atomState.d.keys()) {
-      const aMounted = mountAtom(store, a)
-      aMounted.t.add(atom)
+  const mounted = mountedMap.get(atom)
+  if (mounted) {
+    return mounted
+  }
+  const atomStack: AnyAtom[] = [atom]
+  const stateStack: (AtomState | undefined)[] = [undefined]
+  while (atomStack.length) {
+    const top = atomStack.length - 1
+    const a = atomStack[top]!
+    const existingMounted = mountedMap.get(a)
+    if (existingMounted) {
+      atomStack.pop()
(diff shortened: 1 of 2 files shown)


dai-shi (commented):
Do you think you can create a failing test case with recursive mountAtom?

src/vanilla/internals.ts
```diff
@@ -814,35 +814,53 @@ const BUILDING_BLOCK_mountAtom: MountAtom = (store, atom) => {
...
-    // mount dependencies first
-    for (const a of atomState.d.keys()) {
-      const aMounted = mountAtom(store, a)
-      aMounted.t.add(atom)
+  const mounted = mountedMap.get(atom)
+  if (mounted) {
+    return mounted
+  }
+  const atomStack: AnyAtom[] = [atom]
+  const stateStack: (AtomState | undefined)[] = [undefined]
```
dai-shi, line 826: Have you tried `const stack: [AnyAtom, AtomState][]` and compare the performance?

dmaskasky: Yeah, its worse because of the added overhead of creating new arrays on every iteration.

dai-shi: Oh, really... That means push/pop are more lightweight, correct?

dmaskasky: Parallel arrays are faster because they avoid allocating a tiny tuple array for every entry.

With tuple arrays, each push does all of this:

* allocate a new `[a, aState]` array
* write its two elements
* push a reference to that tuple into the outer stack

Then each pop does this:

* pop the tuple reference
* dereference that tuple object
* read element `0` and element `1`
* later let GC reclaim that tuple

With parallel arrays, each entry is just:

* push `a` into `aStack`
* push `aState` into `stateStack`

and later:

* pop from `aStack`
* pop from `stateStack`

So parallel arrays usually win because they have:

* no per-entry object allocation
* less garbage collection
* less pointer indirection
* better memory locality

The extra push/pop calls are usually cheaper than creating and later collecting a fresh tuple object for every item.

dai-shi:
@copilot resolve the merge conflicts in this pull request

dai-shi:
@dmaskasky could you resolve conflicts as copilot fails?

dmaskasky:
@copilot resolve the merge conflicts in this pull request

dai-shi:
@dmaskasky Should we try to merge #3281, #3282, #3283, or drop them? I'll start working on v3 sometime soon-ish.

dmaskasky:
What is your preference? The code is harder to read, performance is a little slower, bundle size is larger — all to handle an extreme edge case of a super deep atom graph where depth > max stack size.

dai-shi:
My preference is to avoid complexity. I'm fine to drop these as we know that readAtomState has to be recursive anyway.

dmaskasky:
Let's just do a careful review of each and make a decision of whether therebisbanything we want to keep.

dai-shi:
I spent some time and reconsidered this (and other two PRs). My original hope was that the iterative approaches not only unlimit the graph depth, but also improve performance.
On another thought, as `readAtomState` is recursive anyway, it's too rare that we want deeper graph for `mountAtom` without reading the atom state.
I might be something wrong about it, but let me drop three PRs for now and wait for users to report actual use cases in the future.
I appreciate for your work anyway. Without it, I wouldn't have noticed this.

PR #3283, opened 2026-03-30
breaking(internals): iterative batched invalidateDependents (scale)

dmaskasky:
## Summary
invalidateDependents is called inside a for loop in mountDependencies. InvalidateDependents involves deeply traversing dependents. To help improve performance, this PR moved invalidateDependents out of the for loop and changes the invalidateDependents api to accept a list of atoms.

The breaking change is changing invalidateDependents to accept atom list.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -109,7 +109,7 @@ type ReadAtomState = <Value>(
   store: Store,
   atom: Atom<Value>,
 ) => AtomState<Value>
-type InvalidateDependents = (store: Store, atom: AnyAtom) => void
+type InvalidateDependents = (store: Store, atoms: Iterable<AnyAtom>) => void
 type WriteAtomState = <Value, Args extends unknown[], Result>(
   store: Store,
   atom: WritableAtom<Value, Args, Result>,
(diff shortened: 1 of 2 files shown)


src/vanilla/internals.ts
```diff
@@ -785,19 +791,27 @@ const BUILDING_BLOCK_mountDependencies: MountDependencies = (store, atom) => {
   const atomState = ensureAtomState(store, atom)
   const mounted = mountedMap.get(atom)
   if (mounted) {
+    const staleDeps: AnyAtom[] = []
```
dai-shi: Could we use Set instead of array? I'm not sure which is better in terms of performance.

src/vanilla/internals.ts
```diff
@@ -785,19 +791,27 @@ const BUILDING_BLOCK_mountDependencies: MountDependencies = (store, atom) => {
...
     for (const [a, n] of atomState.d) {
       if (!mounted.d.has(a)) {
         const aState = ensureAtomState(store, a)
         const aMounted = mountAtom(store, a)
         aMounted.t.add(atom)
         mounted.d.add(a)
         if (n !== aState.n) {
-          changedAtoms.add(a)
```
dai-shi, line 795: Is it bad to keep it here?

src/vanilla/internals.ts
```diff
@@ -109,7 +109,7 @@ type ReadAtomState = <Value>(
   store: Store,
   atom: Atom<Value>,
 ) => AtomState<Value>
-type InvalidateDependents = (store: Store, atom: AnyAtom) => void
+type InvalidateDependents = (store: Store, atoms: readonly AnyAtom[]) => void
```
dai-shi: So, I wonder if this can be `Iterable<AnyAtom>`.

src/vanilla/internals.ts
dmaskasky: Array is better. Set add and delete have higher overhead.

src/vanilla/internals.ts
dmaskasky: Good call.

src/vanilla/internals.ts
dai-shi: I didn't know that.

dmaskasky: But this function isn't very hot, and it might actually reduce workload by having fewer noop iterations in invalidateDependents. To use Set, I just need to check correctness since change hooks would now fure only after the full atom forest has been mounted.

dmaskasky: but `a` is the atomState.d map key so it is guaranteed to be unique. Switching to Set would be more of a semantic change than actually affecting runtime.

dai-shi:
I'll review this later, but resolving conflicts helps in the meantime.

dmaskasky:
@copilot resolve the merge conflicts in this pull request

dai-shi:
https://github.com/pmndrs/jotai/pull/3281#issuecomment-4644837043

dai-shi:
A review from AI:

Benchmarked vs merge-base:

1. `store.set` perf is within noise on all graph shapes; small graphs regress +3–4%.
2. The batched path never fires in practice — every `invalidateDependents` call comes from `writeAtomState` with 1 atom.
3. Deep chains still overflow on both builds (`readAtomState` recursion), so no depth gain.

PR #3284, opened 2026-03-30
refactor(internals): reduce recomputeInvalidatedAtoms overhead (performance)

dmaskasky:
## Summary
- short circuit when changedAtoms.size is zero.
- adds a parallel stack for atomState, this avoids calling ensureAtomState on every iteration.
- getMountedOrPendingDependents early returned to avoid creating a Set when either pendingDependents or mountedDependents size is zero.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -578,12 +578,9 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
   // Compute a new state for this atom.
   let isSync = true
   const prevDeps = new Set<AnyAtom>(atomState.d.keys())
-  const nextDeps = new Map<AnyAtom, EpochNumber>()
   const pruneDependencies = () => {
     for (const a of prevDeps) {
-      if (!nextDeps.has(a)) {
-        atomState.d.delete(a)
-      }
+      atomState.d.delete(a)
     }
   }
   const mountDependenciesIfAsync = () => {
@@ -615,7 +612,7 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
     try {
       return returnAtomValue(aState)
     } finally {
-      nextDeps.set(a, aState.n)
+      prevDeps.delete(a)
       atomState.d.set(a, aState.n)
       if (isPromiseLike(atomState.v)) {
         addPendingPromiseToDependency(atom, atomState.v, aState)
--- a/docs/utilities/storage.mdx
+++ b/docs/utilities/storage.mdx
@@ -36,7 +36,7 @@ The `atomWithStorage` function creates an atom with a value persisted in `localS
 
 **storage** (optional): an object with the following methods:
 
-- **getItem(key, initialValue)** (required): Reads an item from storage, or falls back to the `intialValue`
+- **getItem(key, initialValue)** (required): Reads an item from storage, or falls back to the `initialValue`
 - **setItem(key, value)** (required): Saves an item to storage
 - **removeItem(key)** (required): Deletes the item from storage
 - **subscribe(key, callback, initialValue)** (optional): A method which subscribes to external storage updates.
--- a/src/vanilla/utils/atomWithObservable.ts
+++ b/src/vanilla/utils/atomWithObservable.ts
@@ -31,7 +31,7 @@ type SubscribableObservable<T> =
     }
 
 type SymbolObservable<T> = {
-  [Symbol.observable]: () => SubscribableObservable<T>
+  [key: symbol]: () => SubscribableObservable<T>
 }
 
 type ObservableLike<T> = SubscribableObservable<T> | SymbolObservable<T>


src/vanilla/internals.ts
```diff
@@ -434,9 +441,12 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
   store,
 ) => {
   const buildingBlocks = getInternalBuildingBlocks(store)
+  const changedAtoms = buildingBlocks[3]
```
dai-shi: I think this is micro optimization, but not confirmed. just feels so.

src/vanilla/internals.ts
```diff
@@ -445,26 +455,35 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
   // This is a topological sort via depth-first search, slightly modified from
   // what's described here for simplicity and performance reasons:
   // https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search
-  const topSortedReversed: [atom: AnyAtom, atomState: AtomState][] = []
+  const sortedReversedAtoms: AnyAtom[] = []
+  const sortedReversedStates: AtomState[] = []
```
dai-shi, line 459: Looks less readable, but I understand this is _not_ micro optimization.

src/vanilla/internals.ts
```diff
@@ -271,12 +271,19 @@ function getMountedOrPendingDependents(
...
+  const pendingDependents = atomState.p
+  if (!mountedDependents || mountedDependents.size === 0) {
+    return pendingDependents
+  }
+  if (pendingDependents.size === 0) {
+    return mountedDependents
   }
-  for (const atomWithPendingPromise of atomState.p) {
-    dependents.add(atomWithPendingPromise)
+  // only pay the union cost when both sides are non-empty
```
dai-shi, line 283: Sounds reasonable.
I thought it's safe to return the copy, but it's readonly anyway, right?

src/vanilla/internals.ts
```diff
@@ -271,12 +271,19 @@ function getMountedOrPendingDependents(
   atomState: AtomState,
   mountedMap: MountedMap,
 ): Iterable<AnyAtom> {
-  const dependents = new Set<AnyAtom>()
-  for (const a of mountedMap.get(atom)?.t || []) {
-    dependents.add(a)
+  const mounted = mountedMap.get(atom)
+  const mountedDependents = mounted?.t
+  const pendingDependents = atomState.p
+  if (!mountedDependents || mountedDependents.size === 0) {
```
dai-shi: Can it be this?
```suggestion
  if (!mountedDependents?.size) {
```

src/vanilla/internals.ts
```diff
@@ -271,12 +271,19 @@ function getMountedOrPendingDependents(
...
+    return pendingDependents
+  }
+  if (pendingDependents.size === 0) {
```
dai-shi:
```suggestion
  if (!pendingDependents.size) {
```

src/vanilla/internals.ts
dmaskasky: Yes

src/vanilla/internals.ts
dmaskasky: yes, I was worried about readability.

src/vanilla/internals.ts
dmaskasky: Yes, I think the type informs the api very well here.

src/vanilla/internals.ts
dmaskasky: The early return is measurable but recomputeInvalidatedAtoms is not hot.

src/vanilla/internals.ts
dai-shi: For me personally, I can read it pretty well like "if there's no size".

src/vanilla/internals.ts
```diff
@@ -434,9 +441,12 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
   store,
 ) => {
   const buildingBlocks = getInternalBuildingBlocks(store)
+  const changedAtoms = buildingBlocks[3]
+  if (changedAtoms.size === 0) {
+    return
+  }
```
dai-shi: Let's do early return after extracting building blocks. It's probably subtle even if it's measurable, but it's the coding pattern.

dmaskasky: Ok, I think I agree. It feels like we are declaring usage at the top of the function, I can see how that's much more readable. I think the only place where its measurable is in ensureAtomState.

src/vanilla/internals.ts
```diff
@@ -434,37 +441,49 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
   store,
 ) => {
   const buildingBlocks = getInternalBuildingBlocks(store)
+  const changedAtoms = buildingBlocks[3]
```
dai-shi:
```suggestion
```

src/vanilla/internals.ts
```diff
@@ -434,37 +441,49 @@ const BUILDING_BLOCK_recomputeInvalidatedAtoms: RecomputeInvalidatedAtoms = (
...
   const mountedMap = buildingBlocks[1]
   const invalidatedAtoms = buildingBlocks[2]
```
dai-shi, line 445:
```suggestion
  const invalidatedAtoms = buildingBlocks[2]
  const changedAtoms = buildingBlocks[3]
```

PR #3285, opened 2026-03-30
refactor(internals): reduce flushPending overhead (performance)

dmaskasky:
## Summary
- Adds an early return when there is no work to do
- Replaces forEach with for..of loops - slightly better performance?

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -271,12 +271,19 @@ function getMountedOrPendingDependents(
   atomState: AtomState,
   mountedMap: MountedMap,
 ): Iterable<AnyAtom> {
-  const dependents = new Set<AnyAtom>()
-  for (const a of mountedMap.get(atom)?.t || []) {
-    dependents.add(a)
+  const mounted = mountedMap.get(atom)
+  const mountedDependents = mounted?.t
+  const pendingDependents = atomState.p
+  if (!mountedDependents?.size) {
+    return pendingDependents
   }
-  for (const atomWithPendingPromise of atomState.p) {
-    dependents.add(atomWithPendingPromise)
+  if (!pendingDependents.size) {
+    return mountedDependents
+  }
+  // only pay the union cost when both sides are non-empty
+  const dependents = new Set<AnyAtom>(mountedDependents)
+  for (const a of pendingDependents) {
+    dependents.add(a)
   }
   return dependents
 }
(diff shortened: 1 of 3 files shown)


dmaskasky:
FlushPending isn't all that hot. We can close this PR if we don't like the readability.

src/vanilla/internals.ts
```diff
@@ -394,11 +394,19 @@ const BUILDING_BLOCK_ensureAtomState: EnsureAtomState = (store, atom) => {
...
   const storeHooks = buildingBlocks[6]
+  if (
+    changedAtoms.size === 0 &&
+    mountCallbacks.size === 0 &&
+    unmountCallbacks.size === 0 &&
+    !storeHooks.f
+  ) {
+    return
+  }
+  const mountedMap = buildingBlocks[1]
```
dai-shi: This should be a micro optimization. Let's avoid it.

src/vanilla/internals.ts
```diff
@@ -394,11 +394,19 @@ const BUILDING_BLOCK_ensureAtomState: EnsureAtomState = (store, atom) => {
...
   const unmountCallbacks = buildingBlocks[5]
   const storeHooks = buildingBlocks[6]
+  if (
+    changedAtoms.size === 0 &&
+    mountCallbacks.size === 0 &&
+    unmountCallbacks.size === 0 &&
+    !storeHooks.f
+  ) {
+    return
+  }
```
dai-shi, line 417: I think changing do-while loop to while loop should be more readable and improve the performance for most cases. Only the concern is the store hook, right?

src/vanilla/internals.ts
```diff
@@ -413,14 +421,26 @@ const BUILDING_BLOCK_flushCallbacks: FlushCallbacks = (store) => {
       call(storeHooks.f)
     }
     const callbacks = new Set<() => void>()
-    const add = callbacks.add.bind(callbacks)
-    changedAtoms.forEach((atom) => mountedMap.get(atom)?.l.forEach(add))
+    for (const atom of changedAtoms) {
```
dai-shi, line 431: This looks good.

src/vanilla/internals.ts
dmaskasky: Yeah. Also, btw flushCallbacks is not a hot path. Let me see if I can switch to a while.

dmaskasky: its ugly. I think I prefer the code as-written.
```ts
...
if (storeHooks.f) {
  call(storeHooks.f)
}
let iterations = 0
while (changedAtoms.size || unmountCallbacks.size || mountCallbacks.size) {
  if (storeHooks.f && iterations++ > 0) {
    call(storeHooks.f)
  }
  ...
}
```

src/vanilla/internals.ts
```diff
@@ -400,6 +400,12 @@ const BUILDING_BLOCK_flushCallbacks: FlushCallbacks = (store) => {
   const unmountCallbacks = buildingBlocks[5]
   const storeHooks = buildingBlocks[6]
   const recomputeInvalidatedAtoms = buildingBlocks[13]
+  if (
+    !storeHooks.f &&
+    !(changedAtoms.size || mountCallbacks.size || unmountCallbacks.size)
```
dai-shi:
```suggestion 
    !changedAtoms.size && !mountCallbacks.size && !unmountCallbacks.size
```

dai-shi (approved):

PR #3286, opened 2026-03-30
writeAtomState setter early return

dmaskasky:
## Summary
Adds an early return when setting a value that is already set. This avoids calling `setAtomValueOrPromise`, `mountDependencies`, (async) `recomputeInvalidatedAtoms`, (async) `flushCallbacks`.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -747,6 +747,9 @@ const BUILDING_BLOCK_writeAtomState: WriteAtomState = (
         }
         const prevEpochNumber = aState.n
         const v = args[0] as V
+        if ('v' in aState && Object.is(aState.v, v)) {
+          return undefined as R
+        }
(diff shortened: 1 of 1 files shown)


src/vanilla/internals.ts
```diff
@@ -747,6 +747,9 @@ const BUILDING_BLOCK_writeAtomState: WriteAtomState = (
         }
         const prevEpochNumber = aState.n
         const v = args[0] as V
+        if ('v' in aState && Object.is(aState.v, v)) {
+          return undefined as R
+        }
```
dai-shi, line 752: I don't like having this and the same check in BUILDING_BLOCK_setAtomStateValueOrPromise. I would like to have `Object.is` only one place. If this is not micro optimization and improves much, I would like to consider bigger refactor.

dmaskasky: Closing for now.

PR #3287, opened 2026-03-30
breaking: ensureAtomState internal function

dmaskasky:
## Summary
- move ensureAtomState out of buildingBlocks (slot is void).
- export ensureAtomsState as INTERNAL_ensureAtomState.
- pass atomStateMap to ensureAtomState and lazily access storeHooks and atomOnInit behind atom init check.

The breaking change is setting buildingBlocks[11] type to void and adding atomStateMap to the parameters of ensureAtomState.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -102,6 +102,7 @@ type AtomOnMount = <Value, Args extends unknown[], Result>(
 type EnsureAtomState = <Value>(
   store: Store,
   atom: Atom<Value>,
+  atomStateMap: AtomStateMap,
 ) => AtomState<Value>
 type FlushCallbacks = (store: Store) => void
 type RecomputeInvalidatedAtoms = (store: Store) => void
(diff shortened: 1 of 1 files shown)


src/vanilla/internals.ts
```diff
@@ -281,6 +281,23 @@ function getMountedOrPendingDependents(
   return dependents
 }
 
+const ensureAtomState: EnsureAtomState = (store, atom, atomStateMap) => {
+  if (import.meta.env?.MODE !== 'production' && !atom) {
+    throw new Error('Atom is undefined or null')
+  }
+  let atomState = atomStateMap.get(atom)
+  if (!atomState) {
+    const buildingBlocks = getInternalBuildingBlocks(store)
```
dai-shi, line 290: This is not acceptable. If `ensureAtomState` become an internal function instead of a building block, it shouldn't rely on bulding blocks.

dmaskasky: I originally had paired this with passing buildingBlocks as the first argument instead of store. But your point stands. Perhaps we should close this PR.

dai-shi: yeah

PR #3290, opened 2026-04-01
fix(internals): check if atom has dependencies before doing mountDependencies work (performance)

dmaskasky:
## Summary
Check if atom as dependencies before doing mountDependencies work.

## Check List


dai-shi:
> I would like to merge this after

I changed my mind. With ca01cdf, let me merge this for the patch.

PR #3291, opened 2026-04-01
fix(internals): check if atom has onMount property before queueing processOnMount callback (performance)

dmaskasky:
## Summary
Check if atom has onMount property before queueing processOnMount callback.

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -7,11 +7,10 @@ type AnyValue = unknown
 type AnyError = unknown
 type AnyAtom = Atom<AnyValue>
 type AnyWritableAtom = WritableAtom<AnyValue, unknown[], unknown>
-type WritableAtomWithOnMount<
-  Value,
-  Args extends unknown[],
-  Result,
-> = WritableAtom<Value, Args, Result> & {
+type WritableAtomWithOnMount<Value, Args extends unknown[], Result> = Omit<
+  WritableAtom<Value, Args, Result>,
+  'onMount'
+> & {
   onMount: NonNullable<WritableAtom<Value, Args, Result>['onMount']>
 }
 type OnUnmount = () => void


src/vanilla/internals.ts
```diff
@@ -837,7 +837,7 @@ const BUILDING_BLOCK_mountAtom: MountAtom = (store, atom) => {
       t: new Set(),
     }
     mountedMap.set(atom, mounted)
-    if (isActuallyWritableAtom(atom)) {
+    if (isActuallyWritableAtom(atom) && atom.onMount) {
```
dai-shi: This is not semantically correct as `atomOnMount` can be overridden.

dmaskasky: This was my previous thought, but we pay a performance penalty for queuing up a noop function every time an atom mounts, while most atoms do not have an onMount method.

My thinking is that a reasonable assumption is that atomOnMount requires onMount to exist. Expanding on this, I think we should add onMount as a required property to atomOnMount atom argument.

`atomWrite`, `atomRead`, and `atomOnMount` functions are intercepters to provide additional capability or override default behavior. For other capability, I think storeHooks would be a better solution.

dai-shi: Sounds fair.

src/vanilla/internals.ts
```diff
@@ -95,10 +98,9 @@ type AtomWrite = <Value, Args extends unknown[], Result>(
 type AtomOnInit = <Value>(store: Store, atom: Atom<Value>) => void
 type AtomOnMount = <Value, Args extends unknown[], Result>(
   store: Store,
-  atom: WritableAtom<Value, Args, Result>,
+  atom: WithRequired<WritableAtom<Value, Args, Result>, 'onMount'>,
   setAtom: (...args: Args) => Result,
 ) => OnUnmount | void
-
 type EnsureAtomState = <Value>(
```
dai-shi, line 108:
```suggestion

type EnsureAtomState = <Value>(
```

src/vanilla/internals.ts
```diff
@@ -11,6 +11,9 @@ type OnUnmount = () => void
 type Getter = Parameters<AnyAtom['read']>[0]
 type Setter = Parameters<AnyWritableAtom['write']>[1]
 type EpochNumber = number
+type WithRequired<T, K extends keyof T> = Omit<T, K> & {
+  [P in K]-?: NonNullable<T[P]>
+}
```
dai-shi: What do you think about defining `WithOnMount` specifically? Maybe I prefer that.

dmaskasky: Yes, I like it

src/vanilla/internals.ts
```diff
@@ -95,7 +101,7 @@ type AtomWrite = <Value, Args extends unknown[], Result>(
 type AtomOnInit = <Value>(store: Store, atom: Atom<Value>) => void
 type AtomOnMount = <Value, Args extends unknown[], Result>(
   store: Store,
-  atom: WritableAtom<Value, Args, Result>,
+  atom: WritableAtomWithOnMount<Value, Args, Result>,
```
dmaskasky, line 104: Could we import OnMount<Args, Result> to use here instead?

dai-shi: No, we don't export internal types to avoid misusing.

PR #3293, opened 2026-04-01
breaking(internals): avoid getInternalBuildingBlock function

dai-shi:
alternative to #3280. Though, it became pretty similar.


dmaskasky:
My initial thought is that rest/spread adds overhead to these internal functions.
Many of the functions only require passing `store` since atomOnInit needs it. But this pollutes the api quite a bit. I'm inclined to shove store into buildingBlocks. Semantically, buildingBlocks are not buildingBlocks for the store, they are the buildingBlocks used by the store internals.
The `BB` prefix is nice. I find code width a bigger problem now that I also have a chat side panel open. We can also use `bb` instead of `buildingBlocks` for the identifier.

dai-shi:
This gets closer to #3280. One of meaningful changes is that I dislike putting store in buildingBlocks.

src/vanilla/internals.ts
```diff
@@ -711,11 +736,10 @@ const BUILDING_BLOCK_invalidateDependents: InvalidateDependents = (
 }
 
 const BUILDING_BLOCK_writeAtomState: WriteAtomState = (
-  store,
+  buildingBlocks,
   atom,
-  ...args
+  args,
```
dmaskasky, line 838: did you intend on keeping this change?

src/vanilla/internals.ts
```diff
@@ -1065,19 +1056,28 @@ function buildStore(...buildArgs: Partial<BuildingBlocks>): Store {
...
       // abortable promise support
       new WeakMap(), // abortHandlersMap
       BUILDING_BLOCK_registerAbortHandler,
       BUILDING_BLOCK_abortPromise,
       // store epoch
       [0],
+      // store api
+      store,
     ] satisfies BuildingBlocks
   ).map((fn, i) => buildArgs[i] || fn) as BuildingBlocks
```
dmaskasky: Passing store in buildArgs is technically possible, but should not affect anything.

This is a common pattern. The intent is not to overwrite the store (since that is what we are building).
```ts
const buildingBlocks = getBuildingBlocks(store)
buildingBlocks[0] = new AtomState()
const store2 = buildStore(...buildingBlocks)
```

I recommend putting `buildingBlocks[29] = store` after this line to overwrite buildArgs[29].

src/vanilla/internals.ts
```diff
@@ -730,86 +754,87 @@ const BUILDING_BLOCK_writeAtomState: WriteAtomState = (
...
-        recomputeInvalidatedAtoms(store)
-        flushCallbacks(store)
+        recomputeInvalidatedAtoms(buildingBlocks)
+        flushCallbacks(buildingBlocks)
       }
     }
   }
   try {
-    return atomWrite(store, atom, getter, setter, ...args)
+    return atomWrite(buildingBlocks, atom, getter, setter, ...args)
```
dmaskasky: Or should we also make atomWrite non-variadic?

src/vanilla/internals.ts
```diff
@@ -995,49 +979,56 @@ const BUILDING_BLOCK_registerAbortHandler: RegisterAbortHandler = (
...
+const BUILDING_BLOCK_abortPromise: AbortPromise = (buildingBlocks, promise) => {
   const abortHandlersMap = buildingBlocks[25]
   const abortHandlers = abortHandlersMap.get(promise)
   abortHandlers?.forEach((fn) => fn())
 }
 
 const buildingBlockMap = new WeakMap<Store, Readonly<BuildingBlocks>>()
 
-const getInternalBuildingBlocks = (store: Store): Readonly<BuildingBlocks> => {
+function getBuildingBlocks(store: Store): Readonly<BuildingBlocks> {
```
dmaskasky, line 1144: I think the return value should be a mutable copy of buildingBlocks.

Mutating a single buildingBlock to form a derived store is a common usage pattern for getBuildingBlocks.

src/vanilla/internals.ts
```diff
@@ -184,16 +191,18 @@ type BuildingBlocks = [
   unmountAtom: UnmountAtom, //                                 19
   setAtomStateValueOrPromise: SetAtomStateValueOrPromise, //   20
   // store api
-  storeGet: StoreGet, //                                       21
-  storeSet: StoreSet, //                                       22
-  storeSub: StoreSub, //                                       23
+  UNUSED_storeGet: undefined, //                               21
+  UNUSED_storeSet: undefined, //                               22
+  UNUSED_storeSub: undefined, //                               23
```
dmaskasky: Should we rework the ordering of buildingBlocks? I was thinking a nice pattern would be to create zone ranges in buildingBlocks. Each zone range would leave a few empty slots at the end for future capability. The current strategy of putting new buildingBlocks at the end gets a bit messy overtime. If we like this idea, I think this should go in a separate PR.

src/vanilla/internals.ts
dai-shi: kind of. both work, right? do you see any pros and cons?

src/vanilla/internals.ts
dai-shi: I don't like it then. If we specialize store like that, I wouldn't put it into the building blocks.

src/vanilla/internals.ts
dai-shi: I think it's my mistake.

src/vanilla/internals.ts
dai-shi: I'm thinking of re-arranging it in v3. I don't think sparse array would be nice. so, no empty slots.

src/vanilla/internals.ts
dmaskasky: `args`
- Slightly better performance, fewer array instantiations / reads.
- Better semantics, args for atom.write. In theory if writeAtomState / atomWrite were to call two variadic functions, you would have vf1Args and vf2Args.
- Consistency between writeAtomState and atomWrite.
- Future non-breaking extensibility.

`...args`
- Doesn't change the api, fewer api changes.

src/vanilla/internals.ts
dmaskasky: Then let's `void` it from the buildArgs.

src/vanilla/internals.ts
dai-shi: this is an internal api. let's do `args`.

src/vanilla/internals.ts
dai-shi: I mean I don't like adding `store` in buildingBlocks.

src/vanilla/internals.ts
dai-shi: I don't see any use cases, but feels consistent for all atom interceptors.

dai-shi:
@dmaskasky How do you like/dislike this? https://github.com/pmndrs/jotai/pull/3293/commits/d6e7bd31ea23108383472acdfacdf0ffa1ab0525

dmaskasky:
> @dmaskasky How do you like/dislike this? https://github.com/pmndrs/jotai/pull/3293/commits/d6e7bd31ea23108383472acdfacdf0ffa1ab0525

It's confusing to me why store is not equal to store1.
```ts
const store1 = createStore()
const anAtom = atom(0)
atom.INTERNAL_onInit = (store) => {
  expect(store).toBe(store1)
}
```

So that would also mean we couldn't get buildingBlocks in atom.INTERNAL_onInit, right?

dmaskasky:
> @dmaskasky How do you like/dislike this? https://github.com/pmndrs/jotai/pull/3293/commits/d6e7bd31ea23108383472acdfacdf0ffa1ab0525

On a side note, I think we should change onInit type signature to have the store param be typed as `Readonly<Store>`.

dai-shi:
Actually, I'm a bit skeptical about `INTERNAL_onInit`. It's too tricky if it depends on store instance.

dmaskasky:
How do you like voiding out the slot for buildArgs?

```ts
type BuildArgs = Partial<BuildingBlocks> & { [29]: void }
```

We allow any value to be passed here, but we are saying we will throw it away. We only allow this value so that buildingBlocks can be reused to make a derived store.

So here we have,

```ts
const buildingBlocks = getInternalStore(store1)
buildingBlocks[someSlotId] = override
// buildingBlocks[29] has store but we don't buildStore doesn't care
const store2 = buildStore(...buildingBlocks)
```

---

If we are truely not comfortable with this, my second proposal is for only internal buildingBlocks to include store, but getInternalBuildingBlocks will only return buildingBlocks.

```ts
type WithStore = { [29]: Store }

// internal functions accept buildingBlocks with a store
const AtomOnInit = (
  buildingBlocks: WithStore<BuildingBlocks>,
  atom: WithOnInit<Atom<unknown>>
) => {
  const store = buildingBlocks[29]
  atom.INTERNAL_onInit?.(store)
}
```

dai-shi:
> How do you like voiding out the slot for buildArgs?

Not a fan of it.

> We allow any value to be passed here, but we are saying we will throw it away.

Then, I mean it's not a building block.

dai-shi:
What's wrong with passing both `buildingBlocks` and `store`? If both are required, that's expected. I don't think "hiding" is good. If it feels weird by design, I guess the design of `INTERNAL_onInit` is not good.

src/vanilla/internals.ts
dai-shi: Actually, this is fine. enhanceBuildingBlocks returns readonly one. I don't think we ever allowed mutating building blocks.

dmaskasky: Oh enhance should also return mutable buildingBlocks. We don't allow mutating the original, but the common case for derive is to overwrite some of the slots.

dmaskasky:
> Then, I mean it's not a building block.

It is a buildingBlock if an internal function relies on it. Otherwise, we are back to having impure internal functions.

BuildArgs are not the same as buildingBlocks, but it shares a compatible shape for ergonomics. `store` not a buildArg.

```ts
type BuildArgs = BuildingBlocks & { [29]: void }
```

dmaskasky:
> What's wrong with passing both `buildingBlocks` and `store`? If both are required, that's expected. I don't think "hiding" is good. If it feels weird by design, I guess the design of `INTERNAL_onInit` is not good.


Then, every internal function should have store arg, whether or not it uses it. Otherwise, it would not be possible to call ensureAtomState from any function.

dai-shi:
> Then, every internal function should have store arg, whether or not it uses it. Otherwise, it would not be possible to call ensureAtomState from any function.

Yes, I will go with it until we find a better API for `INTERNAL_onInit`.

src/vanilla/internals.ts
dai-shi: Are you sure? I thought we always copy building blocks if necessary.

Would you like to open a new PR if that's the intention? It should fix types and add some tests.

dmaskasky: Yes, let's make this change in its own pr.

dai-shi: In case we don't use enhanceBuildingBlocks, our buildingBlocks is frozen. So, I'm not sure how "the common case for derive is to overwrite some of the slots" works.

dmaskasky: It just returns a copy.

dai-shi: There are use cases of read-only building blocks. So, we shouldn't copy unnecessarily.

dmaskasky: Yes, I'm convinced. It also informs that the api will not allow for mutating the original buildingBlocks.

src/vanilla/internals.ts
```diff
@@ -701,17 +739,22 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
   const prevInvalidated = invalidatedAtoms.get(atom) === prevEpochNumber
   try {
     if (import.meta.env?.MODE !== 'production') {
-      storeMutationSet.delete(store)
+      storeMutationSet.delete(buildingBlocks)
```
dmaskasky: This should remain store.

src/vanilla/internals.ts
```diff
@@ -781,86 +824,88 @@ const BUILDING_BLOCK_writeAtomState: WriteAtomState = (
...
-        recomputeInvalidatedAtoms(store)
-        flushCallbacks(store)
+        recomputeInvalidatedAtoms(buildingBlocks, store)
+        flushCallbacks(buildingBlocks, store)
       }
     }
   }
   try {
-    return atomWrite(store, atom, getter, setter, ...args)
+    return atomWrite(atom, getter, setter, ...args)
```
dmaskasky: Should we use `args`?

src/vanilla/internals.ts
```diff
@@ -384,17 +414,20 @@ function initializeStoreHooks(storeHooks: StoreHooks): Required<StoreHooks> {
 // Main functions
 //
 
-const BUILDING_BLOCK_atomRead: AtomRead = (_store, atom, ...params) =>
+const BUILDING_BLOCK_atomRead: AtomRead = (atom, ...params) =>
```
dmaskasky: We should include buildingBlocks here, right?

src/vanilla/internals.ts
dai-shi: Whichever works, but using store should be more readable unless we name buildingBlocksMutationSet which isn't very intuitive.

src/vanilla/internals.ts
dai-shi: I have thought about it. It feels natural to keep `atom.write` signature.

src/vanilla/internals.ts
dai-shi: This is intentional to reduce store dependency.

src/vanilla/internals.ts
dmaskasky: Don't you want to keep store as the identifier? So,

```ts
storeMutationSet.delete(store)
```

dai-shi: I mean, yes. But, when we were able to reduce the store dependency, we might use buildingBlocks as the identifier in the future.

src/vanilla/internals.ts
```diff
@@ -89,70 +89,100 @@ type ChangedAtoms = SetLike<AnyAtom>
 type Callbacks = SetLike<() => void>
 
 type AtomRead = <Value>(
-  store: Store,
```
dmaskasky, line 92: I think we should continue to pass the store to `atomRead`. Otherwise these functions have to be impure.

dai-shi: Are you using `store` in atomRead in one of ecosystem libraries? My intention is to drop supporting `store` in atomRead.

dmaskasky: jotai-scope uses it in a PR right now. The issue is if the function is impure, then it makes deriving stores not possible.

What is your reasoning for removing it?

dai-shi: Oh, I wasn't aware it's being used. The reason I wanted to drop was to reduce the dependency of `store` in the future. I'm not intending to making them impure. Let me revert it until we find a better solution in the future.

dai-shi:
I'll try `const store = storeMap.get(buildingBlocks)` idea and see how I feel it.

(result: I don't feel it's good.)

src/vanilla/internals.ts
```diff
@@ -179,7 +207,7 @@ type BuildingBlocks = [
   atomOnInit: AtomOnInit, //                                   9
   atomOnMount: AtomOnMount, //                                 10
   // building-block functions
-  ensureAtomState: EnsureAtomState, //                         11
+  obsolete_ensureAtomState: undefined, //                      11
```
dmaskasky: Should probably be `void` instead?

dai-shi: I don't think it matters much, but I think `undefined` is better here.

dmaskasky: actually, I think we should keep ensureAtomState as a buildingBlock.

dai-shi: I'm fine with it. Is it because we want the extensibility? Or, is it better in performance?

dmaskasky: There is not really any benefit moving `ensureAtomState` out of buildingBlocks.

dai-shi:
For some subtle reasons, I don't prefer #3308 nor #3280.
I will go with this for this time. (Let's revisit it in the future. It's "internal" thing.)

dai-shi:
Now, time for ecosystem.

dai-shi:
@arjunvegda @dmaskasky 
For v2.20.0, I would like to introduce this breaking change. Please prepare for the change. Ideally, I would like to merge this after all "PASS".

src/vanilla/internals.ts
dmaskasky, line 93: atom interceptors should also have buildingBlocks.

src/vanilla/internals.ts
```diff
@@ -106,53 +106,86 @@ type AtomOnMount = <Value, Args extends unknown[], Result>(
...
   atom: AnyAtom,
   listener: () => void,
 ) => () => void
 type EnhanceBuildingBlocks = (
   buildingBlocks: Readonly<BuildingBlocks>,
+  store: Store,
 ) => Readonly<BuildingBlocks>
 type AbortHandlersMap = WeakMapLike<PromiseLike<unknown>, Set<() => void>>
 type RegisterAbortHandler = <T>(
-  store: Store,
```
dmaskasky, line 151: we should still pass the store as the second param.

src/vanilla/internals.ts
```diff
@@ -106,53 +106,86 @@ type AtomOnMount = <Value, Args extends unknown[], Result>(
...
+  buildingBlocks: Readonly<BuildingBlocks>,
   promise: PromiseLike<T>,
   abortHandler: () => void,
 ) => void
-type AbortPromise = <T>(store: Store, promise: PromiseLike<T>) => void
+type AbortPromise = <T>(
+  buildingBlocks: Readonly<BuildingBlocks>,
+  promise: PromiseLike<T>,
```
dmaskasky, line 196: We should pass store here as well.

dai-shi:
I'll work on them.

dmaskasky:
> @arjunvegda @dmaskasky 
> 
> For v2.20.0, I would like to introduce this breaking change. Please prepare for the change. Ideally, I would like to merge this after all "PASS".

Ok, and we should remember to set the minimum version of Jotai peer dep to this version.

dai-shi:
@dmaskasky reported this breaking change makes jotai-scope unmaintainable. Let's start over when we get time.

@arjunvegda ~sorry about the confusion.~ sorry about the double confusion. I'll wait for your reply.

dmaskasky:
> @dmaskasky reported this breaking change makes jotai-scope unmaintainable. Let's start over when we get time.
> 
> @arjunvegda sorry about the confusion.

I was able to reach a solution for jotai-scope and jotai-effect.
- https://github.com/jotaijs/jotai-scope/pull/99
- https://github.com/jotaijs/jotai-effect/pull/83

dai-shi:
wow, that's great to hear.

PR #3294, opened 2026-04-02
bug: strange mounting behavior near the stack limit

dmaskasky:
## Summary
I encountered a strange behavior when mounting a long chain of atoms. I am able to reliably reproduce this locally and in CI.

When mounting the leaf atom of a long chain of atoms (>996), only 996 atoms are mounted. Testing locally I got this result:
```
 FAIL   jotai  tests/vanilla/internals.test.tsx > internals > every dependency (deep) should be mounted
AssertionError: expected 996 to be 7548 // Object.is equality

- Expected
+ Received

- 7548
+ 996

 ❯ tests/vanilla/internals.test.tsx:33:29
     31|     const leaf = chain[chain.length - 1]!
     32|     store.sub(leaf, () => {})
     33|     expect(mountedMap.size).toBe(depth)
       |                             ^
     34|   })
     35| 
```

Why are only 996 atoms mounted?
Why are there no error throw / reported?

## Check List

### DIFF
--- a/src/vanilla/internals.ts
+++ b/src/vanilla/internals.ts
@@ -200,6 +200,7 @@ type BuildingBlocks = [
   abortPromise: AbortPromise, //                               27
   // store epoch
   storeEpochHolder: StoreEpochHolder, //                       28
+  thrownErrors: Map<AnyAtom, unknown>, //                        29
 ]
 
 export type {
@@ -419,6 +420,7 @@ const BUILDING_BLOCK_flushCallbacks: FlushCallbacks = (store) => {
   const unmountCallbacks = buildingBlocks[5]
   const storeHooks = buildingBlocks[6]
   const recomputeInvalidatedAtoms = buildingBlocks[13]
+  const thrownErrors = buildingBlocks[29]
   if (
     !storeHooks.f &&
     !changedAtoms.size &&
@@ -436,6 +438,8 @@ const BUILDING_BLOCK_flushCallbacks: FlushCallbacks = (store) => {
     }
   }
   do {
+    errors.push(...Array.from(thrownErrors.values()))
+    thrownErrors.clear()
     if (storeHooks.f) {
       call(storeHooks.f)
     }
@@ -577,6 +581,7 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
   const setAtomStateValueOrPromise = buildingBlocks[20]
   const registerAbortHandler = buildingBlocks[26]
   const storeEpochHolder = buildingBlocks[28]
+  const thrownErrors = buildingBlocks[29]
   const atomState = ensureAtomState(store, atom)
   const storeEpochNumber = storeEpochHolder[0]
   // See if we can skip recomputing this atom.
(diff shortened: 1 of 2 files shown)


dai-shi:
> Why are only 996 atoms mounted?

It's different in CI:

```
- Expected
+ Received

- 9465
+ 1320
```

> Why are there no error throw / reported?

I have no idea...

dai-shi:
> Why are only 996 atoms mounted?

Isn't is simply that there are more recursions somewhere? 

> Why are there no error throw / reported?

They are caught by try-catch in readAtomState. `RangeError: Maximum call stack size exceeded`

dmaskasky:
But I don't see the error being rethrown in `returnAtomValue`.

tests/vanilla/internals.test.tsx
```diff
@@ -15,6 +16,25 @@ import {
...
+    const store = INTERNAL_buildStore(...bb)
+    const base = atom(0)
+    const chain: Atom<unknown>[] = [base]
+    for (let i = 0; i < depth; i++) {
+      const parent = chain[chain.length - 1]!
+      chain.push(atom((get) => get(parent)))
+    }
+    const leaf = chain[chain.length - 1]!
+    store.sub(leaf, () => {})
+    store.get(leaf)
```
dmaskasky: This shouldn't be necessary to see the error.

dai-shi: If you are suggesting to capture the system error with distinguishing from the user error and propagate differently, I'm afraid it's going to be very complicated.

dmaskasky: When I run this test locally, there is no error reported from `store.sub`. `store.sub` is incomplete but still the pointer is allowed past that line. This silent error is an issue because it affects correctness. The system should never swallow errors and not surface them to the user.

dai-shi: I know it's desired. Hope we find a reasonable solution.

dmaskasky: I'm still trying to understand where exactly the stack overflow error is getting swallowed. Is it by `readAtomState`? But then `returnAtomValue` is never called? Yes, I think I see.

Ok, then I think we need to have a thrownErrors list in buildingBlocks and throw an AggregateError in `store.sub` if thrownErrors.length.

src/vanilla/internals.ts
```diff
@@ -200,6 +200,7 @@ type BuildingBlocks = [
   abortPromise: AbortPromise, //                               27
   // store epoch
   storeEpochHolder: StoreEpochHolder, //                       28
+  thrownErrors: Map<AnyAtom, unknown>, //                        29
```
dai-shi, line 203: Are you sure this needs to be a store-wide thing? It would be nice if it were a local variable outside the loop. I thought we should catch the error with try-catch.

src/vanilla/internals.ts
```diff
@@ -726,6 +731,7 @@ const BUILDING_BLOCK_readAtomState: ReadAtomState = (store, atom) => {
   } catch (error) {
     delete atomState.v
     atomState.e = error
+    thrownErrors.set(atom, error)
```
dai-shi, line 734: This looks wrong. We are duplicating the error. So, a user could see the error twice.

dmaskasky:
This PR is just a concept right now. What should the behavior be? If we throw on store.sub, this will have a meaningful impact on the jotai behavior. I might be best to not handle stack overflow errors at all since there is no way to distinguish them for other types of errors.

Actually, my vite os to close this PR with the caveat that we acknowledge there is no guarantee of correctness at or near the stack limit for some store operations. This is inherit to the api as iterative readAtomState is not possible given the current api.

PR #3308, opened 2026-04-16
wip: building block context

dai-shi:


src/react/useAtomValue.ts
```diff
@@ -63,6 +66,8 @@ const createContinuablePromise = <T>(
   getValue: () => PromiseLike<T> | T,
 ) => {
   const buildingBlocks = INTERNAL_getBuildingBlocks(store)
```
dmaskasky, line 68: Neat idea with the `s` prop.
Should we do `INTERNAL_getBuildingBlocksContext` instead?

dai-shi:
https://github.com/pmndrs/jotai/pull/3293#issuecomment-4286938006

PR #3313, opened 2026-04-21
refactor(internals): lazy hooks in ensureAtomState for new atom state

dmaskasky:
## Summary
- lazy buildingBlocks assignment of storeHooks and atomOnInit in ensureAtomState.
- removes dev warning. I could be wrong, but I think this condition has a performance cost of checking the environment in production - unless vite is doing something different.

## Justification
`ensureAtomState` is hot, we should do everything we can to optimize this function.


src/vanilla/internals.ts
```diff
@@ -467,13 +467,10 @@ const BUILDING_BLOCK_ensureAtomState: EnsureAtomState = (
...
   const atomStateMap = buildingBlocks[0]
-  const storeHooks = buildingBlocks[6]
-  const atomOnInit = buildingBlocks[9]
-  if (import.meta.env?.MODE !== 'production' && !atom) {
-    throw new Error('Atom is undefined or null')
-  }
   let atomState = atomStateMap.get(atom)
   if (!atomState) {
+    const storeHooks = buildingBlocks[6]
+    const atomOnInit = buildingBlocks[9]
```
dai-shi, line 475:
```suggestion
    // Because ensureAtomState is a hot spot, we intentionally delay the evaluation even though this pattern is exceptional.
    // Ref: https://github.com/pmndrs/jotai/pull/3313
    const storeHooks = buildingBlocks[6]
    const atomOnInit = buildingBlocks[9]
```

PR #3326, opened 2026-06-07
fix(vanilla/utils/atomWithStorage): apply reviver when parsing subscription updates

chatman-media:
## Bug

When a custom `reviver` is passed to `createJSONStorage`, it is correctly applied in `getItem` (via `JSON.parse(str, options?.reviver)`), but **not** in the internal `createHandleSubscribe` handler that fires when the underlying string storage emits a change (e.g. via a cross-tab `storage` event or a custom `subscribe` callback).

This means that if an atom value is a class instance or otherwise requires revival from the raw JSON object graph, any subscription-driven update receives a plain object instead of the revived value — causing runtime errors for callers that invoke methods on the result.

Reported in discussion #3101.

## Root cause

`createHandleSubscribe` called `JSON.parse(v || '')` without forwarding `options?.reviver`:

```ts
// before
newValue = JSON.parse(v || '')

// after
newValue = JSON.parse(v || '', options?.reviver)
```

## Fix

One-line change in `src/vanilla/utils/atomWithStorage.ts`: pass `options?.reviver` as the second argument to `JSON.parse` inside `createHandleSubscribe`.

## Test

A new test `createJSONStorage reviver is applied on subscription updates (#3101)` is added to `tests/react/vanilla-utils/atomWithStorage.test.tsx`. It:
1. Creates a `createJSONStorage` with a reviver that turns plain `{ n: number }` objects into `Item` class instances.
2. Renders a component that calls `value.doubled()` — a method that only exists on `Item`.
3. Simulates an external storage update via a `CustomEvent`.
4. Without the fix the subscription callback returns a plain object, `value instanceof Item` is `false`, and the component renders `NOT_ITEM`. With the fix it renders the correct doubled value.

Closes #3101

### DIFF
--- a/src/vanilla/utils/atomWithStorage.ts
+++ b/src/vanilla/utils/atomWithStorage.ts
@@ -154,7 +154,7 @@ export function createJSONStorage<Value>(
       subscriber(key, (v) => {
         let newValue: Value
         try {
-          newValue = JSON.parse(v || '')
+          newValue = JSON.parse(v || '', options?.reviver)
         } catch {
           newValue = initialValue
         }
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Looks great! Thanks for your contribution.
I'll merge this when I prepare a patch release.

PR #3337, opened 2026-06-11
v3

dai-shi:
- #3336
- #3338
- #3339
- #3340
- #3341
- #3342
- #3344
- #3345
- #3346
- #3347
- #3348 
- #3349 
- #3350
- #3351
- #3358
- #3359


src/react/useAtomValueRaw.ts
```diff
@@ -0,0 +1,54 @@
...
+      },
+      undefined,
+      () => [store.get(atom), store, atom],
+    )
+  let value = valueFromReducer
+  if (storeFromReducer !== store || atomFromReducer !== atom) {
+    rerender()
+    value = store.get(atom)
+  }
+  useEffect(() => store.sub(atom, rerender), [store, atom])
```
Wendystraite, line 48: There is a slight behavioral difference in `useAtomValue` between v2 and v3. I'm not sure if this is intentional or could cause an issue.

```ts
// before (in useAtomValue)
useEffect(() => {
  const unsub = store.sub(atom, () => {
    // ...
    rerender()
  })
  rerender() // <-- rerender was called on mount
  return unsub
}, [store, atom, delay, promiseStatus])

// after (in useAtomValueRaw)
useEffect(() => store.sub(atom, rerender), [store, atom])
```

In v2, `rerender()` was explicitly called on mount, whereas in v3 it seems to only be triggered by the subscription.

AjaxSolutions: This is intentional and documented.

dai-shi: Yes, this is intentional and one of the biggest reasons for the major bump. Suggestions are welcome to improve the docs.

dai-shi:
@arjunvegda Could you work on migrating jotai-devtools?

src/react/useAtomValueRaw.ts
Wendystraite: I can't open a PR myself (locked to collaborators), so here's an attempt to better explain the differences between the three hooks: https://github.com/pmndrs/jotai/compare/v3...Wendystraite:jotai:v3_docs?expand=1

dai-shi: https://github.com/pmndrs/jotai/pull/3360 is open.

PR #3351, opened 2026-06-12
[v3] useAtomValueRaw & useAtomValueRawSync

dai-shi:
for #3337

Motivation:
- We want a hook that doesn't use `use`.
- We want a hook that is more performant with sacrificing tearing issue.
- We don't want to break v2 app.

Solution:
- Introduces a new hook `useAtomValueRaw` without `use` and with tearing.
- Introduces another hook `useAtomValueRawSync` to solve tearing completely.
- `useAtomValue` works as before (with one subtle difference.)

### DIFF
--- a/docs/core/use-atom.mdx
+++ b/docs/core/use-atom.mdx
@@ -119,6 +119,36 @@ const Counter = () => {
 
 Similar to the `useSetAtom` hook, `useAtomValue` allows you to access a read-only atom. Nonetheless, it can also be used to access read-write atom's values.
 
+## useAtomVal
+
+⚠️ This hook is only available since Jotai v3.
+
+```jsx
+const countAtom = atom(0)
+
+const Counter = () => {
+  const count = useAtomVal(countAtom)
+  // ...
+}
+```
+
+`useAtomVal` is like `useAtomValue`, but it doesn't resolve promises: the value of an async atom is returned as a promise, and the component never suspends. It subscribes to the store in an effect, so it doesn't cause an extra re-render on initial mount. As a trade-off, a value change that happens between the initial render and the subscription is not picked up until the next change.
+
+## useAtomSyncVal
+
+⚠️ This hook is only available since Jotai v3.
+
+```jsx
+const countAtom = atom(0)
+
+const Counter = () => {
+  const count = useAtomSyncVal(countAtom)
+  // ...
+}
+```
+
+`useAtomSyncVal` has the same signature as `useAtomVal` and doesn't resolve promises either, but it's implemented with `useSyncExternalStore`. It avoids tearing and picks up values written during mount. As a trade-off, updates are always rendered synchronously without taking advantage of concurrent rendering features.
+
 ## useSetAtom
 
 ```jsx
--- a/docs/core/use-atom.mdx
+++ b/docs/core/use-atom.mdx
@@ -119,35 +119,35 @@ const Counter = () => {
 
 Similar to the `useSetAtom` hook, `useAtomValue` allows you to access a read-only atom. Nonetheless, it can also be used to access read-write atom's values.
 
-## useAtomVal
+## useAtomValueRaw
 
-⚠️ This hook is only available since Jotai v3.
-
-```jsx
-const countAtom = atom(0)
-
-const Counter = () => {
-  const count = useAtomVal(countAtom)
-  // ...
-}
-```
-
-`useAtomVal` is like `useAtomValue`, but it doesn't resolve promises: the value of an async atom is returned as a promise, and the component never suspends. It subscribes to the store in an effect, so it doesn't cause an extra re-render on initial mount. As a trade-off, a value change that happens between the initial render and the subscription is not picked up until the next change.
-
-## useAtomSyncVal
-
-⚠️ This hook is only available since Jotai v3.
+⚠️ This hook is only available since Jotai v3, and it's intended for advanced users.
 
 ```jsx
 const countAtom = atom(0)
 
 const Counter = () => {
-  const count = useAtomSyncVal(countAtom)
+  const count = useAtomValueRaw(countAtom)
   // ...
 }
 ```
 
-`useAtomSyncVal` has the same signature as `useAtomVal` and doesn't resolve promises either, but it's implemented with `useSyncExternalStore`. It avoids tearing and picks up values written during mount. As a trade-off, updates are always rendered synchronously without taking advantage of concurrent rendering features.
+`useAtomValueRaw` is like `useAtomValue`, but it doesn't resolve promises: the value of an async atom is returned as a promise, and the component never suspends. It subscribes to the store in an effect, so it doesn't cause an extra re-render on initial mount. As a trade-off, a value change that happens between the initial render and the subscription is not picked up until the next change.
+
+## useAtomValueRawSync
+
+⚠️ This hook is only available since Jotai v3, and it's intended for advanced users.
+
+```jsx
+const countAtom = atom(0)
+
+const Counter = () => {
+  const count = useAtomValueRawSync(countAtom)
+  // ...
+}
+```
+
+`useAtomValueRawSync` has the same signature as `useAtomValueRaw` and doesn't resolve promises either, but it's implemented with `useSyncExternalStore`. It avoids tearing and picks up values written during mount. As a trade-off, updates are always rendered synchronously without taking advantage of concurrent rendering features.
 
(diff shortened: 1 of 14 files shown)


Wendystraite:
Do you see these two new hooks as long term replacement for `useAtomValue` ?

Would a `useAtomSyncValue` hook make sense to provide the same implicit `React.use` call ?

AjaxSolutions:
Names feel too abbreviated for an API that will be public, and they don’t convey behavioral differences.

dai-shi:
Thanks for the feedback. @Wendystraite @AjaxSolutions 

- For v3, I don't think introducing a breaking change is nice, so we keep `useAtomValue` with `use` (still the behavior changes a little bit.)
- So, the basic idea is to deprecate `useAtomValue` in v3, and drop it in v4. But, it may change at all if `use(store)` becomes real.
- I don't believe people would use `useAtomSyncVal` with `use`, but if they want, they should just use `use`, so we don't need to provide `useAtomSyncValue`.
- Any suggestion for the name? If this feels too confusing, my second take is not to include `useAtomSyncVal` and provide it as a library or a recipe.

AjaxSolutions:
Claude suggests:

"My top pick: useAtomValueSync — it's the most self-documenting. Anyone familiar with useAtomValue immediately understands the relationship, and Sync clearly signals it won't suspend. It also parallels React's own useSyncExternalStore naming convention."

dai-shi:
What would be the one for `useAtomValue` minus `use`?

To avoid abbreviation, my idea is `useAtomRawValue` and `useAtomRawValueSync`?

AjaxSolutions:
Sorry to repost from Claude, but getting names right is not easy.

"The `Raw` framing is a bit ambiguous — it could imply "unprocessed data" rather than "without React.use."

Some alternatives that more clearly signal "bypasses React.use":

- `useAtomPlainValue` / `useAtomPlainValueSync`
- `useAtomDirectValue` / `useAtomDirectValueSync`
- `useAtomEagerValue` / `useAtomEagerValueSync`

But honestly, the most precise framing might be **`useAtomValueWithoutSuspense`** / **`useAtomSyncValue`** — since what you're really saying is "don't suspend, give me what you have."

If the goal is to stay close to `useAtomValue` with minimal surprise, **`useAtomSyncValue`** (which Wendystraite also suggested) seems like the cleanest pair:

- `useAtomValue` — may suspend via `use`
- `useAtomSyncValue` — never suspends, always returns synchronously

It's the same length pattern, no abbreviation, and `Sync` is already a recognized React concept (`useSyncExternalStore`). The pair reads naturally together."

Wendystraite:
What about not using "Value" as a suffix at all, especially if `useAtomValue` is intended to be deprecated?

Something more aligned with `useSetAtom` could be :

- `useAtom`
- `useSetAtom`
- `useGetAtom` (or `useReadAtom`)
- `useGetAtomSync` (or `useReadAtomSync`)

This gives a more symmetric API surface around read/write operations, instead of centering the naming on “Value”.

One concern with `GetAtom` though is that it may read like it returns a getter function rather than the current value, which could be slightly misleading.

dai-shi:
Let me give you more constraints.

In v3:
- `useAtom` shouldn't change the basic behavior from v2.
- `useAtomValue` shouldn't change the basic behavior from v2.
- a new hook (`useAtomValue` minus `use`) should be the **recommended** hook with React 19.
- an alternative new hook (with `uSES` without `use`) is an escape hatch, and it's **not** recommended for typical use case.

That is tough, right?

---

> One concern with `GetAtom` though is that it may read like it returns a getter function rather than the current value, which could be slightly misleading.

While it may sound good, yeah, I feel it's too misleading. btw, `getAtom` is kind of used in https://github.com/jotaijs/jotai-game.

lqhuang:
Hey Daishi,

Coming from your X post:

> Looking for feedback on the Jotai v3 API. I'd like to hear your thoughts.

First of all, thanks for your persistent efforts to make Jotai better! I'm a big fan of both Jotai and Waku 🥰

Let's dive into main topics:

Last year I proposed and improved the main interface for `jotai-ai`

- https://github.com/jotaijs/jotai-ai/pull/15
  - https://github.com/jotaijs/jotai-ai/pull/16
- https://github.com/jotaijs/jotai-ai/pull/20
  - the main context I'd like to discuss, long texts
  - Would you mind taking a quick look when you have a moment
  - Or I also wrote a blog [post](https://lqhuang.io/post/redesign-jotai-ai-api-in-jotai-way) with a linear timeline about it 

However, the latest one (https://github.com/jotaijs/jotai-ai/pull/20) has stalled and hasn't seen much progress.

Asides from Alex's interests shifting, I believe another imporant reason is that jotai itself currently lacks a built-in mechanism to sync external data source or store efficiently. This often results in one more redundant re-renders.

This basically what `useSyncExternalStore` solves. But we know that Jotai doesn't use `useSyncExternalStore` internally, since it doesn't support time slicing. So, solving re-render issue within my PR felt out of scope.

I originally considered seeking help with upstream or submitting a temporary workaround inspired `atomWithObserver` and [recoil's `syncEffect`](https://recoiljs.org/docs/recoil-sync/api/syncEffect/), but I couldn't find the time to finalize it. Despite this, I've been still closely following the news and progress of your all projects:). Like recently, [Naman](https://github.com/nmn) created a PoC project [gh:nmn/react-concurrent-jotai](https://github.com/nmn/react-concurrent-jotai) to explore Jotai style atoms in React 19.

I'm excited about the coming v3 upgrade. So I'd love to discuss the possibility of addressing these sync problems in Jotai v3. (I haven't fully reviewed the general design for v3 yet, will dive into it soon). And I'm interested if the new features introduced in React 19 will help to resolve the time slicing issue when syncing with external sources. 

Please let me know if I've misunderstood anything. I'm more than happy to help with early prototyping, validation, or testing. If there's anything else I can contribute to, just let me know, I'm always here to help!

Regards,
Lanqing

lqhuang:
I thought `useAtomSyncVal` may work perfectly for my issues 🤔

dai-shi:
@lqhuang Thanks for your feedback, though your question is about
> jotai itself currently lacks a built-in mechanism to sync external data source or store efficiently.

is not the scope of this PR, or even Jotai v3 which isn't trying to introduce new features. New feature should come with v2.x or v3.x.

Please open a new discussion https://github.com/pmndrs/jotai/discussions to continue discussion for a new possible feature, independent from Jotai v3. I believe your requirement should be implemented as an ecosystem library, not Jotai core.

lqhuang:
@dai-shi I've checked the PR content, I think the newly introduced `useAtomSyncVal` is enough to address the unnecessary re-renders for `jotai-ai`. Not perfect, but as I mentioned, downgraded version might be acceptable currently. Let's just make it work first.

I've also opened a new discussion https://github.com/pmndrs/jotai/discussions/3352 for full concurrent mode for syncing with external sources

dai-shi:
> `useAtomSyncVal` is enough to address the unnecessary re-renders

Oh, I see what you mean now. Yes, it's the other part of the motivation for this change. Did you try `useAtomVal`?

lqhuang:
> Did you try `useAtomVal`?

Oh, I haven't pulled the current branch into `jotai-ai` as an alpha version to test it yet, only gone through the newly added code and docs so far. But I indeed plan to do so. Will let you know once I have some results in the near future.

dai-shi:
Another idea is to change the import.

```js
import { useAtom, useAtomValue } from 'jotai/react'; // same as v2
import {
  useAtomValue, // no `use`
  useAtomValueSync, // no `use` too
} from 'jotai/react18'; // or jotai/react/...
```

But, this is bad for searchability.

dai-shi:
Let me go with this:
```ts
useAtomValueRaw<T>(atom: Atom<T>): T
useAtomValueRawSync<T>(atom: Atom<T>): T
```

Docs should clearly mention that these new hooks are for advanced users.

lqhuang:
@dai-shi, After digging into it, I believed I made mistakes that my PR for jotai-ai https://github.com/jotaijs/jotai-ai/pull/20 is actually out scope of current PR. You can ignore the re-render issue as I fixed that a while back. I'll hide the unrelated context to avoid any confusion.

But now I finally understand your original intent for the v3 upgrade.

Regarding API compatibility, I guess changing the import from `jotai` to `jotai/react` would break a lot of automated CI runs, LoL.

I prefer the naming style of new `useAtomValueRaw` and `useAtomValueRawSync`.

PR #3358, opened 2026-07-16
[v3] redesign buildingBlocks

dai-shi:
for #3337

- buildingBlocks is a record with one-char keys.


dmaskasky:
Should we use Symbols instead of string keys?
```ts
const flushCallbacksKey = Symbol('flushCallbacks')
...
const flushCallbacks = buildingBlocks[flushCallbacksKey]
```

dai-shi:
No, because one of the goals is to reduce the bundle size.

dai-shi:
The last commit is my answer to @dmaskasky 's DX concern.

